PR #7, opened 2017-03-07
Animations, layout/material design improvements

Gargron:
* Slide left/right animation when navigating
* Pressing back button on main activity will switch tabs when there is a history of navigating tabs
* Nicer login page
* Emoji shortcodes in statuses converted to unicode for native display
* Material design for compose form
* More image views to use Picasso for loading
* Removed list footers but improved how loading indicator looks

screenshot from 2017-03-07 15-06-59
screenshot from 2017-03-07 15-07-29
screenshot from 2017-03-07 15-07-57
screenshot from 2017-03-07 15-08-32


Vavassor:
This looks really good but also introduces a few bugs. I'm just going to merge it into master anyway and then we can sort those out.

So far I've noticed:
-Crash due to Picasso being asked to load empty avatars and that throwing an exception.
-Can't remove attachments in the composer anymore, I think because the edittext steals input focus from behind the little preview.
-If you scroll up and down in the notifications timeline a little while the text from notifications starts appearing in other notifications. I'm guessing a bunch of view holders aren't getting reset to their initial state when they're bound. Also the text "Example status here" pops up in fav/boost notifications.
-Ellipsizing long usernames in statuses is still a little messed up. I think it literally is a one-line fix of assigning layout_toLeftOf to the timestamp.

I'm mostly writing this down so I don't forget because I'm leaving for a couple of hours.

PR #10, opened 2017-03-08
Retrofit for API requests

Gargron:
* One main definition of how the API actually behaves
* Centralized instantiation of API class (removes duplication of domain/access token logic)
* Entities returned from the API now have a more high-fidelity structure, are organized separately


Gargron:
TODO: Find any and all remaining API calls that don't use retrofit, upgrade them

Gargron:
- Retrofit now used everywhere to make API calls
- When opening the app, notifications are automatically dismissed
- Share toot URL action
- Muting/unmuting accounts
- Back button on favourites/blocks
- Drawer on main activity instead of dropdown menu

PR #13, opened 2017-03-10
Improve drawer, integrate floating search view

Gargron:
Although search isn't actually hooked up to anything at the moment


Gargron:
Compose activity now looks like this: https://mastodon.social/users/Gargron/updates/1335760

PR #16, opened 2017-03-12
Push notifications

Gargron:
* Requires tusky-api server, currently running at tusky.zeonfederated.com (setting is under `R.string.tusky_api_url`)
* Replaces pull notifications service as they were doing the same thing


Gargron:
Server code: https://github.com/Gargron/tusky-api

PR #36, opened 2017-03-19
Added German localization

koyuawsmbrtn:
Just thought to drop it here :)


Vavassor:
Ah, I hadn't even asked for localization contributions! But thank you so much, I'm sure folks will really appreciate this.

PR #37, opened 2017-03-21
Translated array.xml

koyuawsmbrtn:
Found a new file, so I thought to translate it too :)


Vavassor:
Actually, that file was for the preferences for the old notifications but since we changed to push notifications this file isn't read anymore. I just need to remove it from the repo, sorry.

PR #41, opened 2017-04-01
New default/error avatar images

nevillepark:
Added a default avatar image similar to Mastodon web's, added a similar, less garish "error" version.


Vavassor:
Yeah, that red error X is still basically programmer art from the first week or so of development. While it does make loading errors easy to spot for me, it def should be out of here, so thanks!

Gargron:
Awwww that unhappy face <3

PR #50, opened 2017-04-03
modified manifest to allow for a quick settings tile

theZacAttacks:
Added support for a custom quick settings tile that accesses the Compose Activity


Vavassor:
I hadn't actually heard of this feature, seems like a nice little thing!

PR #53, opened 2017-04-04
Update German translations

iGEL:
* The plural of Status is Status (pronounced differently, tho), not Statuse, see https://de.wiktionary.org/wiki/Status - Not just for that reason, I suggest to remove the word *Status* completely and use *Toot* or *Toots* instead
* "Dieser Benutzer war nie *x*" translates to "This user was never *x*". The new translations are similar to the English ones, but I'd prefer something like "Couldn't follow that user".
* "%s ist dir gefolgt" -> "%s was following you"
* With the visibility I took more freedom and translated them "Public visible", "Public visible, but not in the public timeline" and "Only visible to followers and mentioned"


Vavassor:
Okay, thank you!

PR #55, opened 2017-04-05
Fixed small typo

koyuawsmbrtn:


Vavassor:
Thanks. I actually didn't realise that was visible because I didn't know android studio's refactor had updated the names from the "no account" text. The english was rewritten a bit.

Issue #60, opened 2017-04-05
Can translate app


Dryusdan:
Hi
i search on code (i'm not very good on java so maybe i don't see if possible) but i don't found translate file. 
It's cool if app Can be translate (i Can translate on french) :)

connyduck:
The strings to translate are in the file Tusky/app/src/main/res/values/strings.xml and the translated french strings would be placed in Tusky/app/src/main/res/values-fr/strings.xml

Vavassor:
Yes, that's basically how it works. All the english is in that .xml file. You can just copy the whole file to a new values-fr directory. Then, in the file, replace the text between each of these tags ```<string></string>``` with the translated text.

You can take a look if you'd like. Do you know how to make a pull request?

Vavassor:
Also, there's a few odd things. Apostrophe is supposed to be written ```\'```, newline is written ```\n```, and some have a code like ```%s``` that will later be replaced with a name or some other text.

I can always fix these symbols when I review, but that's what they mean if anyone is curious.

Dryusdan:
Yes, I know make a pull request ;) 
Is not more simple to write "an error occured" rather write an entire error ? For "basic" people ?

Vavassor:
Yeah, to be honest there's a lot of redundant error messages. Like the "failed to fetch timeline", "failed to follow", etc, could just be "an error occurred". The instructional ones, like "attachment must be less than 4MB" are more important though.

I know a literal translation can create awkward phrases, but it's fine for now. It's always possible to revise parts that don't sound good later.

daycode:
Now it's translate, is it possible to close this issue ? 🤔

Dryusdan:
Yes, I close it :)

PR #61, opened 2017-04-05
Make french translation

Dryusdan:
Reference of #60 issues


Vavassor:
Thank you so much for doing this! I'm sure folks will really appreciate it.

PR #65, opened 2017-04-06
Turkish translation

alerque:
I've started the process of localizing all the strings to Turkish. If anybody wants to help please let me know. Also if anybody has ideas for localizing Mastodon specific terminology or links to good discussions on other languages that would be great. For example

* What languages are importing "toot" as a noun or verb? What languages have picked a substitute for this? The literal verb would be "ötmek" in Turkish but that doesn't seem quite right. It doesn't have as light hearted a connotation an the English.

* A direct translation of "instance" as used in programming such as an instance of a class would be "oluşum", however this can also mean version or composition or any number of other things. The English is confusing enough here, I don't want the  confusion to be even worse in translation. For now I've been going with the equivalent of "server" just because it's a lot clearer what kind of value is requested, but I'm open to ideas.

* For "boost" I went with "yükseltmek" which seems to fit the meaning quite closely, but as this is a very  Mastodon specific usage it's possible importing the original English term would be less confusing in the long run.

If anybody sees activity or discussion related to localizing Mastodon itself or other clients into Turkish please cross link it here.

### DIFF
--- a/app/src/main/res/values-tr/strings.xml
+++ b/app/src/main/res/values-tr/strings.xml
@@ -8,7 +8,7 @@
     <string name="error_authorization_denied">Kimlik doğrulama reddedildi.</string>
     <string name="error_retrieving_oauth_token">Giriş jetonu alınamadı.</string>
     <string name="error_compose_character_limit">İleti fazlasıyla uzun!</string>
-    <string name="error_media_upload_size">Dosya 4MB'ten küçük olmalı.</string>
+    <string name="error_media_upload_size">Dosya 4MB\'ten küçük olmalı.</string>
     <string name="error_media_upload_type">O biçim dosya yüklenmez.</string>
     <string name="error_media_upload_opening">O dosya açılamadı.</string>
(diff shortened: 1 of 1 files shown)


carlosjs23:
Hello, I'm working on translating the Mastodon server into Spanish.

I preferred to use the translation of "normal" terms like Publish instead of toot, share instead of boost, server rather than instance.

The reason is what you mention, English can contribute to confusion in other languages.

So, until there are clear alternatives in the community to those words, I suggest that the translations be maintained in the way I mentioned. (Publish, share, server).

alerque:
I finished off the Turkish translation (including new strings in the master branch since I originally branched) and hence have dropped the WIP from this PR. It should be ready to merge. I'm sure there will be things to fix and tweak, but it should be enough to get people started.

Vavassor:
Thanks so much for the translation!

umitkabuli:
I translate to Turkish v20.0 untranslated parts in menu Preferences submenu and approved. The app can update.

PR #71, opened 2017-04-06
Better fix for the follow account FAB button

theZacAttacks:
changed AccountActivity so that it shows FAB that allows a user to un/follow a user from it instead of the menu

added an xml svg for unfollowing someone
removed the useless line from build.gradle that I added in the last pull


Vavassor:
Ha, I don't think I've seen a minus-person button before but it's sensible. Anyhow, this is nice.

theZacAttacks:
Haha yeah I had to make it myself. It just made sense.

PR #80, opened 2017-04-06
complete german translation

ntngrtltr:
heyho,
I added missing german translations, it should be complete now :3


Vavassor:
Thank you for these translation fixes!

PR #83, opened 2017-04-07
Chrome Custom Tab Support

sturmen:
Just a small change to enable Chrome Custom Tabs.

Also upgrades some libraries.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -28,10 +28,11 @@ dependencies {
     compile('com.mikepenz:materialdrawer:5.8.2@aar') {
         transitive = true
     }
-    compile 'com.android.support:appcompat-v7:25.2.0'
-    compile 'com.android.support:recyclerview-v7:25.2.0'
-    compile 'com.android.support:support-v13:25.2.0'
-    compile 'com.android.support:design:25.2.0'
+    compile 'com.android.support:appcompat-v7:25.3.1'
+    compile 'com.android.support:customtabs:25.3.1'
+    compile 'com.android.support:recyclerview-v7:25.3.1'
+    compile 'com.android.support:support-v13:25.3.1'
+    compile 'com.android.support:design:25.3.1'
     compile 'com.squareup.picasso:picasso:2.5.2'
     compile 'com.pkmmte.view:circularimageview:1.1'
     compile 'com.github.peter9870:sparkbutton:master'
--- a/app/src/main/java/com/keylesspalace/tusky/CustomTabURLSpan.java
+++ b/app/src/main/java/com/keylesspalace/tusky/CustomTabURLSpan.java
@@ -1,0 +1,60 @@
+package com.keylesspalace.tusky;
+
+import android.content.ActivityNotFoundException;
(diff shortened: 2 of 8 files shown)


alerque:
I was going to request Chrome tabs for links, not only did you beat me to the request you implemented it! I really like using browser tabs esp since Brave's adblocking works in them quite well.

Could I suggest you rebase this on master to get rid of the extraneous back-merge commit?

sturmen:
Rebase complete! It actually doesn't work with Brave on my GS7e right now, but I think I have some time tonight if we want to make sure that's in there.

alerque:
Great! I'll just mention that [Trittnuker](https://github.com/vanita5/twittnuker)'s implementation of this does work with Brave if you need something to compare too.

sturmen:
Okay I verified Brave is working but it seems like this needs a few changes before it's production-ready.
Known issues when the custom tabs are enabled:
- if you have no compatible browsers, links just won't open at all. Needs a fallback path.
- When you click usernames or hashtags it seems to exit Tusky and open your default browser

@Vavassor I'll try to fix these 2 things tonight, but I'd still appreciate feedback.

SkyzohKey:
Please make this optional, I have **no** trust in chrome/webkit browser and dont want my links to open in it. I'm probably not the only one.

SkyzohKey:
Ok checked the commit and it looks this is already implemented, nice ! :3

Vavassor:
In an unrelated issue #116, I was considering using WebView but I'll have to keep in mind this is also an option.

alerque:
@SkyzohKey If you don't want to use it and  want your links to  open in an external browser you can either ① turn off the Tusky preference included as part of this PR or ② just don't install a browser that provides remote tab support. This feature typically falls back to an external link handler if now browser provides tap support to apps.

That being said you should be aware that Chrome is not the only option for such tab support.

wictory:
Hi!

I'm not sure if I should open an issue for this or write it here.

I recently updated to Tusky 1.1.4, I'm running Android 7.1.1.

My problem is that links now always opens in chrome, which I don't want to use and don't use. My "Use Chrome Custom Tabs" is set to false. When I later inactivated chrome links don't open at all.

I don't have a default browser set.

Thanks!

alerque:
@wictory If you don't have a default browser set and you've disabled Chrome, what exactly do you expect to have happen? When you click a link where do you expect it to open up?

wictory:
@alerque I expect that the menu which lets me decide which application to open the link with, will appear. 

However, when I try it now, it seems that it works for me now, must have been some other problem. I don't know why it didn't work last time, I remember that I even rebooted the phone to see if it would go away.

It might be that I have wasted your time.

Vavassor:
@wictory It's actually true that the "Use Chrome Custom Tabs" option wasn't doing anything until a commit I just made 9381f21e160002c3fe37712ad3aa18a3bf456985, that hasn't been released yet. Apparently, I renamed the preference key at some point and didn't update it in all places, so it's been broken a while.

PR #97, opened 2017-04-07
inherit content warnings in replies

ntngrtltr:
hi,
Mastodon-Web automatically adds the content-warnings from the toot one is replying to.
I added a few lines that check if a new toot is a reply and if so, it inherits the content notes from its predecessor.


Vavassor:
Thank you, this is convenient so I'm glad to have it!

PR #98, opened 2017-04-07
removed splash screen

connyduck:
Removes the splash screen. It really is unnecessary to steal the user 2 seconds everytime the app starts.
https://github.com/Vavassor/Tusky/issues/81


theZacAttacks:
Could you turn it into an option in the preferences?

connyduck:
Of course. But please help me understand why anyone would want to see the splash screen if it doesn't really load anything? Because it looks pretty?

theZacAttacks:
I mean, that's kinda why I want it as a preference haha. I actually do like seeing it

alerque:
It is pretty, but now I've seen it and it's a waste of seconds every time I open the app. Perhaps it can be moved to the first-run only location before prompting a user to connect to an instance.

koyuawsmbrtn:
why? i love that splash...

Vavassor:
I'd really like to replace the splash screen with another one that's either a static drawable or some kind of animation that doesn't require loading a layout file. So that there's still something other than a black screen but it doesn't waste 2 seconds.

Anyhow, that's why I haven't merged this for the moment.

connyduck:
I see we now have a better solution for this, thx @Vavassor

PR #100, opened 2017-04-07
added proguard config

connyduck:
Added proguard config for release build. Reduces apk size 30%. There is really no reason to not use this.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -14,8 +14,9 @@ android {
     }
     buildTypes {
         release {
-            minifyEnabled false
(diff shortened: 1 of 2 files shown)


Vavassor:
Huh, I didn't know about this. I did a quick release build and the .apk was 2.7MB, compared to 3.5MB for version 1.1.0. So that's nice indeed!

connyduck:
But you have to be careful now - if you add anything that requires variable/method/class names to be preserved in the final apk or uses reflection, the proguard config probably also needs to change.
You can hit me up anytime if you have any questions about it!

PR #107, opened 2017-04-09
Adjust light theme tab bar area

ubiquill:
Adds a shadow below the TabLayout. This makes it easier to distinguish UI elements as before both the tabs and list below had identical background colors

screen shot 2017-04-09 at 1 59 47 am
screen shot 2017-04-09 at 2 00 17 am

### DIFF
--- a/app/src/main/res/layout/activity_account.xml
+++ b/app/src/main/res/layout/activity_account.xml
@@ -166,6 +166,15 @@
         android:layout_width="match_parent"
         android:layout_height="match_parent" />
 
+    <View
+        android:id="@+id/tab_bottom_shadow"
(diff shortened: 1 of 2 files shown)


wavebeem:
Does this work for the dark theme also? It's missing a shadow on that theme too.

ubiquill:
@wavebeem yes, it works on both, but as it is slightly darker it's a little subtler. However in the dark theme two colors are utilized for the tab background and the list background so the shadow doesn't need to be as visible to distinguish UI elements.

Vavassor:
I messed around with a bunch of off-white shades before but didn't like them. The shadow is nicer, though, thank you!

PR #121, opened 2017-04-10
Update the compose activity while posting

theZacAttacks:


Vavassor:
Nice, this appears to work just fine. No overlap if there's a "media uploading" box, and light/dark themes check out.

PR #143, opened 2017-04-12
Share menu now allows for sharing of link/actual post

theZacAttacks:
changed the share button menu to allow for sharing either the link, or the actual content of the post
added values into @string file to accommodate for the change
added the new share submenu into the status_more layout xml


alerque:
This partially addresses #147! Great.

Issue #119, opened 2017-04-10
Replying: parent toot review on swiping down


eidolies:
In the reply view, when composing a reply to a toot, swiping down the compose area and holding could show the content of the toot you're replying to by sliding it from above. Releasing would make it disappear so you can continue writing. 

Useful when you need to read the toot again to clarify your point or avoid neglecting details, especially useful with long toots.

alerque:
I like the idea, but the UI sounds complex and unpredictable. Just putting the toot at the top of the screen and then the compose area would be enough. Scrolling up would get you a look at what your were replying to.

theZacAttacks:
Would something (like this)[https://drive.google.com/open?id=1Ayo3DiMF9O-JJAGsJ1MbPHheHuQfjrlSRQ] be okay?

eidolies:
I was more thinking of a way to see both at the same time, I like @alerque's scrolling up idea.

PR #144, opened 2017-04-12
Show reply text

theZacAttacks:
adds a button in the compose activity that caches your composed toot and displays an uneditable copy of the status you're replying to
the button only appears when you are replying to a message


theZacAttacks:
I'll tweak this a bit more to try to get it more like specified in #119

app/src/main/java/com/keylesspalace/tusky/ComposeActivity.java
```diff
@@ -356,6 +360,16 @@ public void onCreate(Bundle savedInstanceState) {
         floatingBtn.setOnClickListener(new View.OnClickListener() {
             @Override
             public void onClick(View v) {
+<<<<<<< HEAD
```
alerque: This looks like it's left over from a merge or rebase gone bad. These markers should never have ended up in a commit. Please check what happened and amend the commit so these aren't in here.

theZacAttacks: Yeah, I didn't see them before I pushed (after I thought I had successfully merged the files). I was in a rush and haven't updated it yet.

PR #157, opened 2017-04-14
german translation

connyduck:
added missing german translation


daycode:
I think we should synchronize with @Vavassor before a release to translate all of the app and missing string from new features instead of had translation in the process of development 🤔

connyduck:
Well, the app is now live in the Play Store with missing strings, so yes it would be better to wait with a release until everything is translated.

alerque:
If people want to translate as development happens that's great — it just means that much less work to do to roll a release. Most open source projects are not able to coordinate translation teams to have all languages in sync at release points. I mean it's something that can be tried with an announcement and pinging language contributors, but it just doesn't work out that way. People aren't always able to contribute on demand.

It might waste a little effort with strings being in flux, but we're already doing pretty good at keeping that to a minimum. I suggest that translating whenever people can get to it is the best place to start. Then a call can go out to fix anything missing nearing a release.

Vavassor:
Yeah, I'll try to give notice the next update I do. I know those working on the french translation were also annoyed with the state of their translation when I chose to last update.

At any rate, thanks for the improvements!

PR #162, opened 2017-04-15
Adding arabic localization for Tusky

BoFFire:
Hi,
Work made via Transifex via :
https://www.transifex.com/opendz/tusky-in-arabic/
12 missing strings only that I will complete when I'll have enough time :)
Salutations.


Vavassor:
Ah, thank you so much for doing this! I think this is the first translation that's for a right-to-left script so let me know if there's any strangeness related to that.

BoFFire:
Okay :)

Issue #62, opened 2017-04-06
Notification tab spins indefinitely 


exadeci:
It seems that if you never received a notification the tab loading will spin indefinitely, not a big issue but it'll happen to all new users.

raphaelm:
This does not only apply to the notifications, but to all tabs. I'm working on a simple solution.

daycode:
now the commit is merged, could you close this issue ? 😉

connyduck:
This still happens on the "Blocked Users" and "Muted Users" page.

Vavassor:
I fixed it to not occur on account lists like blocked and muted users back in commit 275cd51a6db2a1f119ff01e3091603fda32860db. It should be the same in all lists and timelines, now.

PR #165, opened 2017-04-15
Timeline footer improvements

raphaelm:
The current implementation of endless scrolling has a few user interface issues, for example that the loading animation is shown even if nothing is loaded in the background (see eg. #62).

This pull request is not perfect in all situations, but improves the overall situation, especially for new users.

Before:
 

After:


Vavassor:
I actually had a similar footer setup during the alpha version, but @Gargron removed it and I didn't ask why. The no-notifications case is especially prominent for new users, though, as well as going on accounts that don't have many posts and reaching the end with an infinite loading sign.

PR #159, opened 2017-04-15
Allow pull-to-refresh in thread view (closes #73)

raphaelm:
Hi :)

Thanks for Tusky! Hope this is useful!

PR #166, opened 2017-04-15
Update thread view after sent reply (closes #74)

### LINKED ISSUE #74: Update toot view once response is sent.

raphaelm:
Fixes #74 

**IMPORTANT: This PR is built on top of #159 (which introduces refreshing in the first place) and therefore includes two commits. Please merge #159 first and then review the second commit here separately. If necessary, I can rebase this after #159 is merged :)**


raphaelm:
I also added this to TimelineFragment so it refreshes your home timeline after you posted a new toot, which feels intuitive to me. This might lead to some unnecessary refreshes in some cases, what do you think?

Vavassor:
Thanks for these changes! Since it uses onRefresh, it checks if it only needs to load the top bit when there's already statuses in the list. So it loads in the new one and doesn't lose position if you're scrolled far down the timeline, which is nice.

PR #167, opened 2017-04-15
Allow to filter boosts and replies from home (closes #133)

### LINKED ISSUE #133: Do not display boosts and replies in the Home panel
On the web interface of Mastodon, you can activate or deactivate "Show Boosts" or "Show Replies" in order not to read the toots boosted/replied by your followers. That's a great feature (because sometime you like toots from a guy and not the ones from his followers) and it would be great to have it in Tusky.

mastodon-interface

raphaelm:
This is a simple approach to #133. It currently has one problem: If you change the value of the settings, as the current timeline is still cached in the fragment/adapter, you will only see the results after restarting the app or doing similar things. Does anyone have a good idea on how to solve this?


Vavassor:
Hmm, since there wouldn't be any way to restore removed statuses when you uncheck either filter, I think the solution would be to fully refresh the timeline.

So, adding this TimelineFragment's onActivityCreated should work?
```
preferences.registerOnSharedPreferenceChangeListener(new SharedPreferences.OnSharedPreferenceChangeListener {
    boolean onSharedPreferenceChanged(SharedPreferences sharedPreferences, String key) {
        if (key.equals("tabFilterHomeBoosts") || key.equals("tabFilterHomeReplies"))
            sendFetchTimelineRequest();
        }
    }
});
```
Note that this would reset the position to the top of the timeline. Which is probably acceptable, just keeping in mind. Currently, the timeline only actually loads statuses when scrolling down and never unloads them. So there's not yet any way to start at an arbitrary point on the timeline and be able to scroll up.

raphaelm:
This was my first idea as well, but apparently ``onSharedPreferenceChanged`` doesn't run as the fragment is not active while the preference changes.

Vavassor:
Oh, I guess you could keep copies of the settings and then compare them to preferences in onResume?

raphaelm:
Yep, that was my other thought but it feels so *ugly* :D

raphaelm:
I pushed an implementation that should be sufficient :)

raphaelm:
Hi @Vavassor I just resolved the merge conflicts with master :)

Vavassor:
Sorry I let this PR languish for a while, it's in now.

PR #168, opened 2017-04-15
Margins in account tab

raphaelm:
Small spacing correction.

Before:

After:


alerque:
Much better. This has bugged me before and it's exacerbated by the number fields being blank until the profile loads making the labels way off center.

Issue #64, opened 2017-04-06
Missing expected Android UI elements


alerque:
There are a few basic UI elements one expects to find that are missing.

Most notably headers for the preferences panes need to be added that tell what screen you're even looking at and have back buttons. Individual toot screens have this as expected but the preference screens don't.

When scrolling way down in a timeline there doesn't seem to be an easy way back to the top. One is forced to flick enough to scroll all the way back.

An "About" entry in the main nav with links to the project page and where to file issues would also be nice.

Vavassor:
I hadn't given the preferences thing much thought, since it only recently expanded to a second screen. But, especially as more preferences are added, the app bar would be good.

You should just be able to tap the tab for the timeline you're currently on to jump to the top. But with things like the favourites timeline and block list, that don't have tabs, there's no way to jump upward yet.

And yeah, an about page is nice although talking with translators has made me wary of adding unnecessary text, haha

alerque:
Should I split this up into separate issues for tracking? I was writing this from mobile and got lazy.

Yes I agree on keeping the strings to a minimum, but a very basic one that shows the version number and links to keep places like the issue tracker shouldn't need more than a couple words.

sbetelmal:
I am happy to add a simple about page that shows the version number.

connyduck:
I added ActionBar to Preferences https://github.com/Vavassor/Tusky/commit/93fcd33c48b34a1d84c17122c756845873cea9ea

charlag:
@connyduck as you've added Toolbar to Preferences, there's a link for bugs in About and one can scroll up by clicking on the tab icon on the top I think this can be closed now?

PR #169, opened 2017-04-15
Adding an about activity to the App

sbetelmal:


alerque:
Great, but to meaningfully address what I brought up in #64 this needs more than the version number. The main reason I suggested this was as a place to hang useful links such as to the project page and the issue tracker.

sbetelmal:
Happy to add that, could you prepare the links you see fit and I will add them to the activity.

alerque:
> **Project website:**
> https://tusky.keylesspalace.com
>
> **Bug reports & feature requests:**
> https://github.com/Vavassor/Tusky/issues

Keep in mind that any strings here need to be localized so the less text the better, but pointing out these two key URLs should be worth the effort.

sbetelmal:
@Vavassor Are you happy with @alerque suggestions? or do you want to take the about page to a different direction?

Vavassor:
@sbetelmal Yeah, those sound like good contents. It might also be good to link to the Tusky@mastodon.social account, but it's not as vital.

sbetelmal:
Hey guys,
let me know how you feel about this. I have tested it with Nexus 6P.
If you could let me know how to looks like on other devices.

@alerque @Vavassor

PR #170, opened 2017-04-15
Useful title for AccountActivity

raphaelm:
Before:

After:

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/AccountActivity.java
+++ b/app/src/main/java/com/keylesspalace/tusky/AccountActivity.java
@@ -107,8 +107,18 @@ public class AccountActivity extends BaseActivity {
                 @AttrRes int attribute;
                 if (collapsingToolbar.getHeight() + verticalOffset
                         < 2 * ViewCompat.getMinimumHeight(collapsingToolbar)) {
+                    if (getSupportActionBar() != null && loadedAccount != null) {
+                        getSupportActionBar().setTitle(loadedAccount.getDisplayName());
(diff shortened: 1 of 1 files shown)


Vavassor:
This is nice, but it needs to be tinted for its theme, because it doesn't show up under the light theme (white letters on white background). Also, it'd be good to animated it so it fades in with the bar background.

I might merge this as is and make these changes myself, but that's why I didn't merge it just yet.

raphaelm:
I'm not sure what I'm doing wrong, but adding ``app:collapsedTitleTextAppearance="?attr/android:textColorPrimary"`` to the ``CollapsingToolbarLayout`` does not change anything for me regarding the light theme. If you have a good idea, go ahead!

SkyzohKey:
Would it be possible to display the instance address too ? (ie. @ramy_test@instance.tld)

raphaelm:
@SkyzohKey Sure, good idea. I'm currently not moving this forward as I'm not experienced with the theming issue, so I'd be glad about some help there.

Vavassor:
I merged this and fixed the theming issue. It still animates the background differently than the text and icons, which I'd like to change, but I'm okay with it for the moment.

Also, @SkyzohKey it shows their instance name if they're on another instance, just not for the same instance as you.

SkyzohKey:
Nice :)
Is a release planned soon ?

Vavassor:
@SkyzohKey Yeah, I'm planning for Wednesday. So, day after tomorrow.

PR #184, opened 2017-04-17
Improves Japanese translation

FlameArche:
Improved and added new translation.

In string name="dialog_whats_an_instance", fixed link of List-of-Mastodon-instances.md from [old URL](https://github.com/tootsuite/mastodon/blob/master/docs/Using-Mastodon/List-of-Mastodon-instances.md) to [new URL](https://github.com/tootsuite/documentation/blob/master/Using-Mastodon/List-of-Mastodon-instances.md) that has moved.

### DIFF
--- a/app/src/main/res/values-ja/strings.xml
+++ b/app/src/main/res/values-ja/strings.xml
@@ -1,20 +1,21 @@
 <resources>
 
     <string name="error_generic">エラーが発生しました。</string>
-    <string name="error_invalid_domain">不正なドメイン名が入力されました</string>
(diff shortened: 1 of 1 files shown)


alerque:
@indication Could you review these strings and comment on this PR? Since most of us don't speak Japanese it's hard to review such translation changes. Since saying a translation is "improved" is rather a subjective matter it would be nice to hear your feedback as the original translator.

indication:
@alerque All of @FlameArche 's translation is correct.
I cannot say any more! Good translations, and thanks.

Issue #155, opened 2017-04-14
Add Direct to visibility options panel


daycode:
Add the option "Direct", like in the web UI, to send a direct message to people mentionned and use the same description and icon as web UI.

![direct-message](https://cloud.githubusercontent.com/assets/13915512/25042751/acd14dd8-211a-11e7-9d60-5730dfab346f.png)

Vavassor:
Yes, I'd like to reformat the options to be more like the main site, at the same time this is added. Especially because associating the icon with the visibility level for that drop-down is a lot more understandable than the button indicator (TOOT!, TOOT, or TOOT🔒).

daycode:
@Vavassor I'm doing that right now, the functionnality is done and I put icons instead of radio basic style

Work in progress in mind 😄 
![visibility-panel](https://cloud.githubusercontent.com/assets/13915512/25057945/f141e572-2174-11e7-96cd-4bc964dbf184.png)

daycode:
Can I put the content warning button like in web UI next to the visibility button instead of in the visibility panel ? 🤔

Vavassor:
@daycode Oh nice! Yes, I was going to move the content warning to its own button like the site.

PR #194, opened 2017-04-18
Add Direct Message to visibility panel

daycode:
PR for #155 

direct-message-visibility-panel


Vavassor:
So, apparently setting vector icons to radio buttons like ```button="@drawable/ic_whatever"``` is not supported on Android versions Lollipop and earlier. I was testing on emulator with API level 17 and it crashes when it tries to load the layout. There isn't a alternative support library attribute like there is for images (app:srcCompat), so assigning at runtime is apparently the only way. Commit bd687fb45d4c2fdda0aa94064a7d27ecfce01e84 is my fix, but it doesn't animate the same way, so it's kind of a bummer.

I might switch the icons to non-vector, I just thought this was important for future reference.

PR #202, opened 2017-04-19
change mastodon social illustration at login

daycode:
mastodon-login


alerque:
Should we really be using the "mastodon't" image for this usage?

connyduck:
No please don't. This image is used when an error occurs on the web client, this is very confusing.
Also see https://github.com/Vavassor/Tusky/issues/99

PR #207, opened 2017-04-20
Add a custom User-Agent

connyduck:
Add a custom User-Agent that contains Tusky & Android Version to all requests
https://github.com/Vavassor/Tusky/issues/206


Vavassor:
Ah, I didn't even realise this was how it was showing up to servers. Thanks for fixing it!

PR #210, opened 2017-04-21
Traduction

boced66:
Translate to french

Traduction vers le français


alerque:
Thanks for taking the time to contribute translations!

This branch is a little bit messed up and including all the back-merges will make it very difficult down the road to use tools like `git bisect` or `git blame` to find where changes got introduced or who contributed what. Could we request that you rebase this branch on the current master so that it ends up being just the commits you added with the French translation?

Thanks!

alerque:
You don't have to close this. You can rebase it and push the fixed commits  back up to the same branch.

alerque:
I just cleaned this up for you and pulled your updates to French into #212 — so you shouldn't have to do anything at this point. Thanks for the translations!

boced66:
Thanks, 
Sorry, i'm not yet comfortable with github
But I try and learn

alerque:
I understand if you aren't used to what `git` can do it can be a bit confusing. Also this is not a good way to manage translations anyway (see #213). Just for your learning, what I did was add your fork as a remote to mine and checked out your branch. Then I rebased your branch onto the one that I'd been working on, then I edited the French to fix the file order. Then I squashed your two commits and my order edit into one commit. I then changed back to my branch and cherry picked that new commit into my branch and pushed it.

What you _could_ have done (ignoring my other  work, just what normally would have happened) was pull down changes from the upstream master branch to your master branch. Then checking out your branch, you could have used rebase to update your branch to be based on top of the updated master branch. That would have brought this down to two commits. Then you could have pushed this branch again (using "force" to over-ride this one) and this PR would have automatically updated to reflect the new state of the branch.

PR #212, opened 2017-04-21
Normalize strings

alerque:
Since most of the translations were initially started new features have been added requiring new strings. Also the strings were re-arranged a bit.

I went through all the existing translations and made sure the keys match the original. For ones that were missing, I added the current English strings. Where applicable I also organized them to be in the same order for easy comparison.

Note this highlights what translators need to catch up on for new strings, but it doesn't show off where strings have been edited. In particular they should review the `visibility_*` strings and `dialog_whats_an_instance` as those seem to be the most significant  changes.

I brought the TR strings up to date, but after this merges we should ping other translators to review their languages.

### DIFF
--- a/app/src/main/res/values-fr/strings.xml
+++ b/app/src/main/res/values-fr/strings.xml
@@ -1,7 +1,7 @@
 <resources>
 
     <string name="error_generic">Une erreur s’est produite.</string>
-    <string name="error_empty">This cannot be empty.</string>
+    <string name="error_empty">Ce champs ne peut pas être vide</string>
     <string name="error_invalid_domain">Le domaine est invalide</string>
     <string name="error_failed_app_registration">L’application n’a pu s’authentifier avec l’instance.</string>
     <string name="error_no_web_browser_found">Impossible de trouver un navigateur web.</string>
@@ -13,7 +13,7 @@
     <string name="error_media_upload_type">Ce type de fichier n’est pas accepté.</string>
     <string name="error_media_upload_opening">Le fichier ne peut être ouvert.</string>
     <string name="error_media_upload_permission">Une permission pour lire ce média est requise pour le mettre en ligne.</string>
-    <string name="error_media_download_permission">Permission to store media is required.</string>
+    <string name="error_media_download_permission">Permission d\'enregistrer le fichier requise.</string>
 
     <string name="error_media_upload_image_or_video">Impossible de mettre une vidéo et une image sur le même pouet.</string>
     <string name="error_media_upload_sending">Ce média ne peut être mis en ligne.</string>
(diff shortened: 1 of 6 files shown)


alerque:
This now includes @boced66's work translating the missing French strings. I rebased his branch and edited the commit so the file would stay in the same order as the English.

app/src/main/res/values-fr/strings.xml
```diff
@@ -139,4 +153,6 @@
     <string name="notification_title_summary">%d nouvelles interactions</string>
 
     <string name="description_account_locked">Compte bloqué</string>
+    <string name="status_share_content">Partager le conteunu du pouet</string>
```
SkyzohKey: contenu *

FlameArche:
Hmm... I feel disinclined to add English strings into "ones that were missing" of other languages.
We need to have common recognition for "Which strings have not been translated."

By default, Translations Editor in Android Studio shows empty and highlighted that strings has not added to localization files.
translation_editor

If we add English strings to other languages as default, Translations Editor no longer shows highlight. 
translation_editor_2
It make a little difficult to finding strings has not translated. Even translation to Japanese, it may be same words as English such as "NSFW."

alerque:
Fair enough @FlameArche. I haven't been using an IDE for this and was unfamiliar with how it handles the files. However the fact that the files were massively out of sync with each-other makes me think other's aren't either.

Can you tell me how Android Studio handles builds where ① the key is not present in the strings file at all vs. ② the key is present but the value is empty?

FlameArche:
Ok, I'll tell you how Android Studio builds resources.

When values/strings.xml has following strings.
`<string name="example_key">Hello world!</string>`
For example, if Japanese translation files (values-ja/strings.xml) have following cases, Android platform processes...

1. There are localization strings.
`<string name="example_key">Hello Japan!</string>`
=> The device displays "Hello Japan!" (uses localization file.)

2. There are not localization strings.
`(no name="example_key" line)`
=> The device displays "Hello world!" (uses default file.)

3. There are localization strings but it's empty.
`<string name="example_key"></string>`
=> The device displays "" (uses localization file but empty!)

More info can be found at [Android Developers document](https://developer.android.com/guide/topics/resources/localization.html).

alerque:
Thanks @FlameArche, that's helpful. And yes, given that compile time behavior what I did with the strings is wrong. I'll submit a tweak it a bit.

app/src/main/res/values-nl/strings.xml
```diff
@@ -87,6 +89,9 @@
     <string name="action_compose_options">Opties</string>
     <string name="action_open_drawer">Menu openen</string>
     <string name="action_clear">Leegmaken</string>
+    <!-- <string name="action_clear">Clear</string> -->
```
FlameArche: Already exists.

alerque:
I've edited the commits in this PR to address what @FlameArche in regard to the way Android Studio handles translation strings. For the convenience of anybody that wants to contribute translations directly from the Github editor or via their own text editor I've still copied the missing strings, but I commented out the keys so that and IDEs are not deluded into thinking the translations are done. This should effectively…

* keep actual builds falling back to the EN strings if not translated.
* alert anybody using AS to what translations are missing.
* make it easy for anybody with a plain editor to find and translate new strings.

Long term an external translation platform is still preferable, this mostly just cleans up my previous mistake and adds the updated  TR & FR translations.

Note I also corrected the mistake noted by @SkyzohKey in code review.

app/src/main/res/values-fr/strings.xml
alerque: Fixed, thanks!

app/src/main/res/values-nl/strings.xml
alerque: Oops, I must have been going diff crosseyed. Corrected.

PR #216, opened 2017-04-22
fix bugs with fab showing/hiding on scroll

connyduck:
- fixes bug where fab hiding on scroll would not work in NotificationsFragment
- fixes bug where fab would be hidden completely until app restart when the option was toggled while fab was invisible
- also SharedPreferences.getBoolean is not called in the ScrollListener anymore as it is quite an expensive operation and a ScrollListener is called very often

@theZacAttacks


theZacAttacks:
Awesome! Sorry if my code isn't optimal, I'm not very fluent in Android's best practices.

PR #217, opened 2017-04-22
use material design icon for local timeline tab

connyduck:
fits much better with the other material design icons

before:
screen-1

after:
screen-2


Vavassor:
Also, no longer reminiscent of the ionizing radiation hazard symbol ☢.

PR #232, opened 2017-04-24
(untested) Fixed grammatical errors in the french translation

Elarcis:
I don't have the necessary SDK for Android apps, however I was able to fix some awful grammatical errors in the french locales and made the whole thing more coherent.

Could anyone please confirm the app works with my commit? (first ever contribution to an open source project, I might have missed some obvious community guideline)

### DIFF
--- a/app/src/main/res/values-fr/strings.xml
+++ b/app/src/main/res/values-fr/strings.xml
@@ -1,9 +1,9 @@
 <resources>
 
     <string name="error_generic">Une erreur s’est produite.</string>
-    <string name="error_empty">Ce champs ne peut pas être vide</string>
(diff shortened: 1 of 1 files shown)


daycode:
you should use \' instead of ' in that file, plus it's 'POUET!' less whitespace between the text and "!" 😉

Elarcis:
In french there is a (non-breaking) whitespace before any exclamation point, as well as for `?`, `;` and `:`.

Regarding the character `’`, it's typographically valid, and although it was used inconsistently in the file before, I can switch it to an escaped `'`.

Vavassor:
Everything checked out fine, thanks for the changes!

PR #246, opened 2017-04-29
Improve French translation

Poulpette:


yannicka:
Line 35: "%s a boosté" instead of "%s boosté"
Line 151: "%1$s, %2$s, et %3$s</string>" instead of "%1$s, %2$s et %3$s</string>"

Poulpette:
Agreed.
About Line 35, I thought that it might mean “Pouet boosté”. I may have missed the appropriate layout in Android Studio, so I left it that way.
Also one thing that might be discussed is the use of imperative mood/infinitive form. Both are correct, of course, but I wonder whether it’s a good idea to use both. For example, Line 135: since we mostly use the imperative form everywhere else, I used “Indiquer”, but “Indiquez” may be better…

app/src/main/res/values-fr/strings.xml
```diff
@@ -98,61 +98,60 @@
...
-    <string name="link_whats_an_instance">Qu’est ce qu’une instance ?</string>
+    <string name="link_whats_an_instance">Qu’est-ce qu’une instance ?</string>
 
-    <string name="dialog_whats_an_instance">L’adresse ou le domaine d’une instance peut être entré
-         ici, comme mastodon.social, icosahedron.website, social.tchncs.de,
-        <a href="https://github.com/tootsuite/mastodon/blob/master/docs/Using-Mastodon/List-of-Mastodon-instances.md">ou autre</a> (en anglais) !
-        \n\nSi vous n’avez pas de compte, vous pouvez entrer le nom de l’instance que vous voulez rejoindre et créer un compte ici.\n\nUne instance est l’endroit où votre compte est
-        stocké, mais vous pouvez facilement communiquer et suivre d’autres personnes sur d’autres instances bien que vous soyez sur le même site
-        \n\nPlus d’info <a href="https://mastodon.social/about">mastodon.social</a> (anglais).
+    <string name="dialog_whats_an_instance">Indiquer ici l’adresse ou le domaine d’une instance, comme mastodon.social, icosahedron.website, social.tchncs.de,
```
nicosomb: Puisque dans le reste du texte, on dit `si vous ne disposez`, `vous pouvez renseigner`, etc., je verrais plus ici `Indiquez`.

PR #254, opened 2017-05-01
Update French translation

turb:
Space before colon, dash...
Better translation on some strings


turb:
Sorry, did not see the more complete one. Feel free to pick some elements.

Issue #77, opened 2017-04-06
Allow profile edition within the app.


SkyzohKey:

Vavassor:
There's isn't any way to edit any of the profile contents in Mastodon's API currently https://github.com/tootsuite/mastodon/blob/master/docs/Using-the-API/API.md. Until it does, I don't have any way to make this feature.

SkyzohKey:
See this: https://mamot.fr/users/kinux/updates/7570

Vavassor:
Yee, that pull request was closed so this is now possible!

SkyzohKey:
@daycode Here's a quick mockup made with [Gravit Designer](https://designer.io) for the edition profile Activity. Hope you like it, I can provide the .gvdesign/.svg file(s) if you want to. :)

![Edit profile mockup](https://0x0.st/3ij.png)

Vavassor:
This is good! I hadn't thought to put the existing header & avatar like that, it seems nicer. I don't think you can make the account private/locked in the API, though, so I'd have to omit that bit.

SkyzohKey:
I guess we could ask @Gargron to add this to the API. Everything accessible via the web UI should also be available via the API, in order for a third-party app/service to provide the best UX possible. :stuck_out_tongue:

daycode:
I have that right now, and I think it will be finish tomorrow 😉 
but I have a question: I think the preview image should change the background of its button to not break the layout.
 what do you think about that ? 🤔 

![edit_profile](https://cloud.githubusercontent.com/assets/13915512/25565110/95464084-2dc1-11e7-9ae7-300e0ad66c20.png)

SkyzohKey:
@Daycode instead of using a button for the cover, you may use an Image and then place place a button as an overlay with reduced opacity (I used black at .7 on the mockup) 😎

daycode:
I know, I downloaded the software to do the integration 😎

PR #256, opened 2017-05-01
Allow profile edition within the app

daycode:
see #77 

edit_profile_final


Vavassor:
This looks great!

Just to let you know, the header and avatar buttons were reading as "no text available" in TalkBack with the labels invisible, so I just moved the text to the buttons themselves and deleted the labels. Also, I put it in a scroll view because not all the fields were reachable in landscape orientation (I forgot to do this in my previous layout).

Issue #190, opened 2017-04-17
Changing screen orientation after scrolling causes timeline to reload, losing scroll position


mdiehr:
Tusky v 1.1.2
Android 7.1.2 (Google Pixel)

Steps to reproduce:

- Start in Portrait orientation
- Navigate to "Home" timeline
- Scroll down several toots so that the top toot is not visible
- Rotate phone to Landscape orientation

Results:

- The Home timeline reloads from scratch
- The scroll position is reset to the top

Expected results:

- The timeline probably shouldn't reload from scratch
- The timeline should stay at the same scroll position when changing orientations

Notes:

This also happens for the Notifications, Local, and Federated timelines.

theZacAttacks:
It also loses the place when you refresh, which can be frustrating.

Vavassor:
Commit dc971ec36d7e75d2f4c109986fc6405df7ef8b8c fixes this.

Issue #259, opened 2017-05-03
When changing orientation of app, 'Search accounts...' text is lost


PhotonQyv:
Just noticed that if while running the app you rotate the screen of your device either from portrait to landscape or vice versa the 'Search accounts...' default text disappears. Also if you search and then rotate the search criteria as entered is lost though the search results remain across rotation.

Vavassor:
Commit dc971ec36d7e75d2f4c109986fc6405df7ef8b8c fixes this (obvs).

PR #261, opened 2017-05-03
Fixed issues #190 & #259 screen orientation in MainActivity

PhotonQyv:
I added an onConfigurationChanged() listener to MainActivity class so that when you change the orientation of the screen the Activity doesn't get destroyed. This fixes issues 190 & 259 that both relate to changing screen orientation while running Tusky.

The diffs look worse than they should due I think to the difference in line end characters, between my version of the files and the original version.

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -37,7 +37,9 @@
                     android:scheme="@string/oauth_scheme" />
             </intent-filter>
         </activity>
-        <activity android:name=".MainActivity" />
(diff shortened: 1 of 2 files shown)


Vavassor:
Hey, I think your editor defaults line endings to Windows CRLF when you save files, I'd guess? It made the diff a little hard to review because git treats it as **every** line changed. I'm keeping this code using LF "\n", so please use that.

PhotonQyv:
I'll get it right next time, Glad to help (:*

PR #271, opened 2017-05-04
i18n - Updating Arabic strings translation

BoFFire:
i18n - Updating Arabic strings translation and added missing ones from "en" version to "ar".


BoFFire:
State of translation to arabic on this link (updated) : https://www.transifex.com/opendz/tusky-in-arabic/

Issue #235, opened 2017-04-25
[Feature request]: Confirm dialogue on logout


betsythefc:
I accidentally logged myself out of the app (big fingers, small screen). I think a confirmation dialog box would be useful.

PR #273, opened 2017-05-05
Logout Confirmation Dialog

tryton-vanmeer:
Pressing logout in the drawer now shows a dialog to the user asking if they wish to continue.

For #235


Vavassor:
Thanks for this addition!

PR #278, opened 2017-05-08
Add loading on login activity while fetching tokens

Pangoraw:
When coming from the chrome tab, the app was still showing the domain input for ~1 second before switching to the main activity. Now, the app shows a progress bar during this period.

I am not really aware of the naming convention, so don't hesitate to request some changes.

French Version

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/LoginActivity.java
+++ b/app/src/main/java/com/keylesspalace/tusky/LoginActivity.java
@@ -32,6 +32,7 @@ import android.text.method.LinkMovementMethod;
 import android.view.View;
 import android.widget.Button;
 import android.widget.EditText;
+import android.widget.LinearLayout;
 import android.widget.TextView;
 
 import com.keylesspalace.tusky.entity.AccessToken;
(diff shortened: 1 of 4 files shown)


connyduck:
oh very nice, that always irritated me
thank you :3

Two remarks:
- I would prefer "Connecting..." instead of "Connection..." as text 
- You probably should make the instance input visible again in all error cases so the users does not see the ProgressBar forever and can try again

Pangoraw:
I am going to implement the following function :
```java
private void setLoading(boolean loadingState) {
   if (loadingState) {
          loading.setVisibility(View.VISIBLE);
          input.setVisibility(View.GONE);
   } else {
          loading.setVisibility(View.GONE);
          input.setVisibility(View.VISIBLE);
   }
}
```
and call `setLoading(false);` whenever an error is detected.
Does it work for you ?

Also, you're right about "Connecting..." I am going to change that.

connyduck:
Sounds good!

Pangoraw:
Updated !

Issue #228, opened 2017-04-23
Download image button or menu option


Vavassor:
As a design principle, I really don't like actions that have no visible indication that they're there. Right now images can be downloaded, but only by long-pressing the image in the viewer. The primary way to do it should be a button or something visible on-screen and long-press could be kept as a shortcut.

torrentcome:
Hi,

Can i take care of it ?
What do you thing about a toolbar with the button download at the right of the screen ?
May ask you if you have you'r own "Code style" for automatic format ?

R.

daycode:
it will be better to have a top bar for image action (back arrow, download button) like that

![imageview](https://cloud.githubusercontent.com/assets/13915512/25896317/0602a762-3584-11e7-969c-38f0a516445a.png)

PR #281, opened 2017-05-10
Download image button or menu option #228

torrentcome:
+ toolbar on media
+ navigation button 
+ download button when image is loaded


Vavassor:
Thanks for making this, it works well!

PR #282, opened 2017-05-11
Allow broadcasting event across activities, closes #280

### LINKED ISSUE #280: Allow activities to be synced
Right now when an action is taken on a user profile for example, the action is not synced with the main timelines.

Here are the actions that should be synced :

- When following a user, add his statuses to the home timeline
- When unfollowing/muting/blocking a user, remove his statuses from the home timeline
- When fav or boosting, show the status as fav/boosted in the home timeline

I don't know how such thing can be done. Do you have any clue ?

Pangoraw:
Implements removing statuses of a user from home timeline when unfollowing, and from all timelines when muting/blocking.

There is still work to be done in order to sync favourite and reblog.

I don't know if it should have the same behavior when muting and blocking. I did not get the difference.
Also, I don't if it is relevent to put `TimelineReceiver` in the `util` package, maybe it should be in `adapter`.

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/AccountActivity.java
+++ b/app/src/main/java/com/keylesspalace/tusky/AccountActivity.java
@@ -32,6 +32,7 @@ import android.support.design.widget.FloatingActionButton;
 import android.support.design.widget.Snackbar;
 import android.support.design.widget.TabLayout;
 import android.support.v4.app.Fragment;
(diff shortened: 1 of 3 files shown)


Vavassor:
Blocking is like muting, but it also forces you and target to unfollow each other. Also mutes I think you can still receive notifications from the muted user but blocked users you don't get mentions/replies.

PR #284, opened 2017-05-11
Add the ability to keep your device screen on.

PhotonQyv:
I've added a preference so that user can keep their screen on, it defaults to false, and is turned on at the BaseActivity level.

No idea if anyone else wants this ability, I just thought it might be a 'nice to have'.


connyduck:
I have no idea why this would be useful except when a video is playing? I think we can do without this change or have it limited to the ViewVideoActivity without a setting. 

Also your pull request should probably contain only 1 commit and it is better to not translate a string than to paste in crappy translations from google.

PhotonQyv:
Fair enough (:*

alerque:
Agree with @connyduck on all counts here. I think this setting would just be clutter. If you want to change your phone screen's settings I would expect to do this in the phone settings, not per app. The only exception should be exception events: notably when actively playing video media. This exception is something I would expect to be setup by default and shouldn't need a setting.

Also, please do re-base this branch on the current upstream master to eliminate all the duplicate merges.

PR #294, opened 2017-05-23
remove unnecessary Log utility class

connyduck:
...replace Exception.printStackTrace with logging

-> production apk does not log anything anyway if build with proguard, see proguard rules
-> less classes is always awesome
-> Exception.printStackTrace prints to stdout, better log it so proguard can remove it too


Vavassor:
I only introduced that utility because logs were still showing up in release builds so I wanted to stop that. I didn't even know proguard rules could also be a solution, so this is good to be rid of. Also, thanks for making output and logging nicer!

PR #297, opened 2017-05-26
updated Polish translation

m4sk1n:
Also, what about entry with info who translated Tusky (and Mastodon profile to report bugs) in About section?


Vavassor:
TBH it'd be a long list of translators, but I'd be glad to acknowledge in-app if you wanted? Or I could make a web page and link it there.

Also, right now @Tusky@mastodon.social is sort of doubling as publicity and support account so it's probably all that's needed at the moment.

m4sk1n:
@Vavassor, I meant just mentioning author of the currently used translation, it would be nice to have ability to leave some info like 'If you have found any mistake in translation, or have any suggestion to make it better, contact me on @m4sk1n@mastodon.xyz'.

PR #299, opened 2017-05-27
Adds Japanese translations.

FlameArche:
New translations added for Japanese.

I did not translate strings used in "About" page because I'm worried that users may expect that we can understand Japanese, especially with bug reports & feature requests.
Showing "About" page in English would be better than discouraging them when they see the project site wrote in English.


Vavassor:
That's fair. It could also be structured like "Link description (English)" next to the links, which I've seen when linking to sites in a particular language. But it's probably fair enough to leave it in english, but if so I should probably move it to donottranslate.xml so there's no confusion among the translators.

PR #321, opened 2017-06-28
Add boost click handling

charlag:
Since tapping on a status opens original (not boosted) status, it is impossible to jump to the user who booster the status. So, now clicking on a boost line opens account.
I also renamed "id" in AccountViewHolder because it confused me at first. Hope you don't mind that I left it in this commit.


charlag:
I'm not quite sure why there's a conflict. It wasn't there when I opened it...

charlag:
I updates it with the master

Issue #142, opened 2017-04-12
Save toots as drafts


theZacAttacks:
Maybe add an option to the menu bar on the side that shows your locally saved drafts

ebrulato:
Good Idea, as we have 500c, we may need some times to write some toots :)

raphaelm:
Currently, when adding a photo to a toot, it is instantly uploaded. In this case, would we want to store the photo locally as well or just keep the URL?

Vavassor:
I forgot to close this when #330 was merged! I added media to drafts later in commit a1e007eb2adf23e8ae19e7810bf3acbd6ec02c21 and chose to just save copies of any attachments on external storage (SD card) and upload them when the draft is opened.

Ideally, I'd like to do it the other way, but I wanted to avoid the situation where someone loses connection, can't finish uploading, so it won't let them save the draft. You could save a local copy in that case and build extra logic to distinguish already-uploaded URIs from local copy URIs, but it was enough trouble without that, for the moment.

PR #330, opened 2017-07-06
#142/save toots

torrentcome:
So, with the "Room Persistence Library"

+ Feature : Saved toot
+ With a long click, save the toot, go to new section "Draft", select the toot for go to composition with.
+ Drafted Toot can be deleted too

Available and tested. For sure it has to be improved (:


Vavassor:
This is so cool, I didn't know you were working on this! It seems like it's working well, although I notice it doesn't save content warnings or media yet.

A few things:
- If you save a draft, load it into the compose page, and then save that, it adds a new draft. It possibly would make more sense to save over the previous draft. It's not super important to change, but I think I'd like it better.
- Also, if you load a draft into the composer, save, and then back out, it doesn't show the new save on the drafts page unless you back out and re-enter.
- I didn't understand it was long-press to save a draft. Is that a convention other apps use? I was thinking it'd be a stand-alone save button, but maybe it's just something I'm not familiar with.

Should I merge it as is or do you want to improve it more?

torrentcome:
Hi Vavassor,

I will see what can i do for improve that, i let you know about (;

Ahm , about the long press, what one of the solution for no "corrupt/change" the initial design,

I assume that is not so familiar but :
The long press will be improve with the time with a Floatting Menu. (see here quick search https://stackoverflow.com/questions/30699302/android-design-support-library-fab-menu, i think will be really cool to have some open feature link to that)

torrentcome:
should be fine now ;)

PR #332, opened 2017-07-12
Preserve status states on updates. UI layer refactoring.

charlag:
Some things were pulled out of adapters to fragments. New classes were introduced - StatusViewData and NotificationViewData. They not only have view state in them but also help decoupling.
Because introducing parallel model list requires a lot of synchronization PairedList was added.
Also synchronization between fragments and adapters is quiet tedious and error-prone and should be replaces with better solution.
Oh, I also couldn’t resist and fixed bug with buttons animation in the same commit.

I checked it and seems like everything works but there may be some corner cases so I think it should be double-checked at least.
I understand that this may be quite a lot for one PR but it's quiet hard to fix separately. Thanks for your time.

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/TuskyApplication.java
+++ b/app/src/main/java/com/keylesspalace/tusky/TuskyApplication.java
@@ -59,10 +59,6 @@ public class TuskyApplication extends Application {
             throw new RuntimeException(e);
         }
 
-        if (BuildConfig.DEBUG) {
-            Picasso.with(this).setLoggingEnabled(true);
-        }
-
         /* Install the new provider or, if there's a pre-existing older version, replace the
          * existing version of it. */
         final String providerName = "BC";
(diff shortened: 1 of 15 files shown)


charlag:
Oh, I've found a bug. It may open wrong thread from the timeline sometimes.
Will investigate and fix, stay tuned!

charlag:
@Vavassor I've fixed it, event added some assertions, now it should work like a charm!

charlag:
I'm not sure but something may be wrong about notifications order. I would be glad if someone could confirm that.

charlag:
There was a bug with fav/boost in threads, fixed now

Vavassor:
@charlag The notification order is a bug I introduced. The server sends ```limit``` notifications every fetch no matter what, so at the end it sends however many it has, plus copies of the last one for padding. So I had a couple places where I just lazily removed duplicates with new ArrayList<>(new HashSet<>(newNotifications)). BUT that construction doesn't preserve order, so that's the bug.

PR #335, opened 2017-07-14
Add small avatars to boosted toots and to notifications.

charlag:
screenshot_20170714-073837
screenshot_20170714-080732


Vavassor:
Oh, yeah, this is a nice feature to bring over! I didn't even have it on my radar. (I should've)

PR #336, opened 2017-07-14
avoid crash when new notification list is empty

hgourvest:


Vavassor:
woop, thanks for this one

charlag:
@hgourvest Thanks for the fix, it probably was me who introduced this bug, such a tricky thing to get it right.

hgourvest:
@Vavassor @charlag you're welcome.

PR #337, opened 2017-07-14
Update Dutch strings for 1.4

jeroenpraat:


Vavassor:
Thanks so much for these updates!

PR #341, opened 2017-07-15
Improve media browser and video viewer

charlag:
* show/hide status bar by tapping a photo
* dim and color status bar in video/media viewers
* show/hide status bar in video viewer
* use shared element transition when opening a photo is possible
* center video in VideoView


Vavassor:
This is really nice! I only saw one small thing: if you swipe to the side in a gallery and back out, the shared element transition plays on the element you entered on rather than the last one viewed. It could be a pain to change, though, I'm not sure.

charlag:
One way to fix that is to have a listener for viewPager and if we're not in the initial page then use regular finish() instead of transition one. We can put a listener and reject/change bindings.

PR #353, opened 2017-07-18
fix permission problem when downloading image

connyduck:
fixes https://github.com/Vavassor/Tusky/issues/347

tested on my Android 5 & Android 7 devices

**ATTENTION this introduces a new permission in the manifest**

I just wonder... why did nobody check if the download code actually works? You can't request a permission that does not exist in the manifest...


Vavassor:
Ah! I changed the manifest while working on drafts. I thought we were only using WRITE_EXTERNAL_STORAGE for writing to application-specific directories, which doesn't require the permission on API level 19 and above. So I broke this after the fact.

PR #360, opened 2017-07-28
(fragment) quick changed

torrentcome:
The duplicated interface will no appeared anymore. The problem is with the "stack memory" who do weird thing with his fragments.

no more duplicated stuff after that ;)


Vavassor:
!!! I just assumed that issue was related to ```RecyclerView``` "scrapped" views, but I hadn't found any mistakes there. So I didn't have good leads on how to fix it. I hadn't looked at the possibility the whole fragment might be duplicated!

Thanks for finding this!

charlag:
Oh, thanks, I _knew_ it was it but never remembered to fix it.

Issue #85, opened 2017-04-07
Option to use Emojione instead of system emoji


dpogue:
Particularly on older phones that don't get the latest OS updates, the newer emoji characters render as sad little boxes.  The art style and visual meaning of the emoji can also be lost in translation between icon sets sometimes.

The Mastodon web client uses [EmojiOne](http://emojione.com/)

daycode:
I wanted to use mastodon's emoji from Tusky too. I'm not and android developer but I'll see how to implement it and perhaps I'll do a pull request.

pythe:
At work we just tore EmojiOne out of our Android app. It may be worth considering Noto: https://www.google.com/get/noto/

Lionirdeadman:
Mastodon now uses Twemoji as of 2.0 so that'd be better.

charlag:
We could use Google's downloadable fonts for emoji, there's a PR but it's very old.

connyduck:
The google emoji support library only works with the standard Android emojis.
We could use this library https://github.com/vanniktech/Emoji
I already use it in other projects, it is really easy to plug in, but it will increase our apk size significally. Personally, I would be willing to make that tradeoff, but i am not sure everybody agrees with that.

charlag:
@connyduck that's not true, you can configure any font and font provider you want.

connyduck:
Oh I see, cool. Lets try it out!

dpogue:
I was poking around at updating [my EmojiCompat PR](https://github.com/Vavassor/Tusky/pull/364) last night, and realized that the only downloadable font provider right now is Google Play Services.

There is a way to have EmojiCompat bundle the fonts with the application, but as you noted that would result in a larger apk size. However, that might also allow us to generate our own "TwemojiCompat" font to bundle so that the emoji would be consistent with the web version of Mastodon rather than Android's Noto Emoji.

charlag:
I'm still not sure what's the best way to deal with this issue. The library mentioned by @connyduck doesn't provide you with choice as far as I understand. i'd go with EmojiCompat + bundled font (noto or emoji one).

charlag:
I've tried using EmojiCompat to use EmojiOne bundled font but it won't work just like that, it needs some metadata in the font.
We *could* try to generate such metadata but it seems crazy.
[...]

PR #364, opened 2017-08-04
Experiment bringing in the EmojiCompat support library

dpogue:
This was a fun experiment to try out the EmojiCompat stuff.

I'm probably missing a bunch of TextViews that might possibly contain emoji, but it does work for the most common ones (toots in the lists, expanded toots, and account pages). Mainly, it might be a starting point for anyone who wants to try to complete this and get it into a mergable state.

If you do want to merge this, it should be rebased after #365 is merged, since that provides a better way of doing a bunch of the support library updates.

The font_certs.xml I had to copy from Google's [example project](https://github.com/googlesamples/android-EmojiCompat/). It seems really weird to me that this isn't handled automatically by the EmojiCompat library, but apparently the only other option is to bundle the font with the app (and bloat the binary size).


connyduck:
see the discussions #85 #516

PR #367, opened 2017-08-05
Account activity improvements

connyduck:
-> improve Tabs: stretch to whole width in landscape mode, highlight text, bigger font
-> hide fab on scroll if set in settings
-> change fab to compose, new follow button in header, addresses https://github.com/Vavassor/Tusky/issues/359
-> show if an account follows you back

I am open to discussions how to improve this even further. Maybe the "Follows you" TextView and the Follow Button could have a fancier layout (e.g. look more like badges?), but I actually like it in the raw material design they have now.
Of the two possibilities
1) leaving username next to avatar, new button below avatar
2) username below avatar, new button next to avatar
I liked 2) better because it leaves more space for the username, on small displays instancenames would always get cut off

Before
screenshot_20170805-101007

After
screenshot_20170805-100820


connyduck:
Maybe there should be a warning dialog before unfollowing too?

connyduck:
Also, maybe show "blocked" and "muted" states more prominently?

charlag:
@connyduck I love how it looks! I am also for the warning dialog before unfollowing.

Vavassor:
I tried it on a fairly narrow screen and I think having the layout with the button next to the avatar and names below works well in that scenario.

I'm going to add the warning dialog for unfollowing, also.

Thanks for working on this @connyduck!

PR #368, opened 2017-08-05
improve buttons in follow request/muted users/saved toots/blocked users

connyduck:
The buttons where really hard to touch, much better now :)
Also, I changed from android:text to tools:text -> hardcoded strings will not be in final apk, but still in layout preview of android studio

Before:
screenshot_20170805-105514_1
After:
screenshot_20170805-105514

### DIFF
--- a/app/src/main/res/layout/item_blocked_user.xml
+++ b/app/src/main/res/layout/item_blocked_user.xml
@@ -1,65 +1,69 @@
 <?xml version="1.0" encoding="utf-8"?>
 <RelativeLayout xmlns:android="http://schemas.android.com/apk/res/android"
     xmlns:app="http://schemas.android.com/apk/res-auto"
+    xmlns:tools="http://schemas.android.com/tools"
     android:layout_width="match_parent"
     android:layout_height="72dp"
+    android:gravity="center_vertical"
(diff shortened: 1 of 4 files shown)


app/src/main/res/layout/item_blocked_user.xml
```diff
@@ -1,65 +1,69 @@
...
+        android:layout_height="32dp"
         android:layout_alignParentEnd="true"
         android:layout_alignParentRight="true"
-        android:layout_marginStart="16dp"
-        android:layout_marginLeft="16dp"
         android:layout_centerVertical="true"
-        android:contentDescription="@string/action_unblock" />
+        android:layout_marginLeft="12dp"
+        android:layout_marginStart="12dp"
+        android:background="?attr/selectableItemBackgroundBorderless"
```
charlag: Isn't `selectableItemBackgroundBorderless` is API 21+?

connyduck:
`?android:attr/selectableItemBackground` is API 21+, `?attr/selectableItemBackgroundBorderless` is from the support library :)

charlag:
@connyduck ah, thanks, for some reason I thought it only works with `selectableItemBackground`.

PR #370, opened 2017-08-10
l10n: Update Japanese translations

lindwurm:
* tested on my device
* support new strings (ex. drafts, unfollow dialog, search)


Vavassor:
Thanks for updating these!

PR #371, opened 2017-08-11
optimize build files

connyduck:
In https://github.com/Vavassor/Tusky/pull/365 I upgraded libraries but did not check if the production build still works, sorry.

I have no idea, why the production logging was enabled in this commit https://github.com/Vavassor/Tusky/commit/bb0ea876fa9844d620cdf25bf45582f3d5d1ee58 so I disabled it again. (Please explain if it wasnt a mistake)

Also, I had a look at the files contained in the apk and noticed some unnecessary metadata files bundled.

Yes, I do like my apks really small xD

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -26,6 +26,11 @@ android {
     lintOptions {
         disable 'MissingTranslation'
     }
+    packagingOptions { //remove metadata files of jsoup library from final build
+        exclude 'META-INF/CHANGES'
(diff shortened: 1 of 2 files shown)


Vavassor:
@connyduck Oops, yeah, production logging being enabled was a mistake. I was debugging why ```BouncyCastleProvider``` was behaving so differently in release builds and forgot to add it back.

PR #389, opened 2017-09-24
Add feedback when downloading image

### LINKED ISSUE #388: Provide visual feedback when saving an image
I ended up with 10+ copy of the same image because I had no clue that the download actually went well. We could show a toast indicating that the download started and an other one when it ends. Also the icon could have a ripple providing instant feedback when pressing it.

flyingrub:
fixes #388


charlag:
@flyingrub sorry for no response, I guess everyone's is super busy.
Honestly, we should rewrite most of the post submitting but this LGTM. @Vavassor ?

PR #395, opened 2017-10-17
Update Japanese Translations

lindwurm:
add japanese strings for #386 #389 #391 .


connyduck:
Thank you very much :)

PR #397, opened 2017-10-18
Improvements for adding link info while composing toots

charlag:
- Now link title and picture are not automatically inserted white posting a link. User is given an option to add them. I had to make something like a Snackbar because it only appears at the bottom of the screen and it was too easy to miss and tap 'toot' instead.
- If user adds link info, progress is shown until everything is loaded or until some error occurs
- I changed progress bar to horizontal one. I think in the future we will post toots in the background so this progress bar will only be used for the link info loading.

There is also some automatic cleanup.

screenshot_20171018-201604

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/ComposeActivity.java
+++ b/app/src/main/java/com/keylesspalace/tusky/ComposeActivity.java
@@ -16,6 +16,8 @@
 package com.keylesspalace.tusky;
 
 import android.Manifest;
+import android.animation.Animator;
+import android.animation.AnimatorListenerAdapter;
 import android.app.ProgressDialog;
 import android.content.ContentResolver;
 import android.content.Context;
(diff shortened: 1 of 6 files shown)


charlag:
I've checked it and seems that there are some problems with the layout, I've probably forgot something. If someone won't fix this before me, I can fix it tomorrow.

charlag:
I wasn't able to reproduce it, lgtm

screenshot_20171019-113715

connyduck:
This looks good and works fine - but I am really not sure about it. It still feels really quirky and does only work with pasted links - sometimes I type links out or edit a pasted one.

I personally would prefer removing all link preview stuff from ComposeActivity and instead implement the Mastodon card endpoint on detail statuses https://github.com/tootsuite/documentation/blob/master/Using-the-API/API.md#getting-a-card-associated-with-a-status 
This way writing a toot gets less confusing and we get link previews for toots not created with Tusky.

charlag:
@connyduck wow, I didn't know about such API! yes, let's just strip it out of good then. I wonder how many useless pictures were uploaded because of this...

connyduck:
So I think I implemented most of the api. Have a look at this: https://github.com/Vavassor/Tusky/commit/70c1385104844f89d64919f9fbee4fd1725def46  (code needs some cleanup, but interface and functionality look good to me)
Do you want to do the part of removing link previews from ComposeActivity, @charlag?

charlag:
I will look at it. Yeah, I can I'd you want

connyduck:
yes please do & don't forget to remove the jsoup library

PR #400, opened 2017-10-19
Load custom emoji in statuses

charlag:
- This PR loads custom emoji in timelines (I hope in all of them)
- I've also added HttpLoggingInterceptor for debug builds because it make life much easier. I hope it's okay (if not I will remove it)

I've tried to not introduce any leaks, but span should not outlive the text content anyway.


charlag:
Wow, it's 400 already!

connyduck:
I just tried this out, looks really good, thank you!
I think we should remove the EmojiOne class now? https://github.com/Vavassor/Tusky/blob/master/app/src/main/java/com/emojione/Emojione.java

charlag:
@connyduck thanks! 
Wow, I didn't know about it (still not sure what it does). If it just replaced shortcodes with emojis - then yeah, probably we can

connyduck:
Yes it replaces shortcodes with emojis and is wired in via the json parser. But it does not make any sense because the emojis wont show up on any other app or on the web.
I will remove it and then merge this pull request.

connyduck:
I also like the HttpLoggingInterceptor, but next time please make two pull request or commits so our commit history really shows all changes

charlag:
@connyduck you're right, sorry for smashing everything together, I was in hurry but it should be done separately.

PR #414, opened 2017-10-24
Fix extra autocompletion requests

charlag:
The problem was that Tusky kept sending requests for autocompletion while writing toots even when the user wasn't typing a username anymore. As it happened very often we would exceed the API request limit and user wouldn't be able to send the toot.
    
This happened because Tokenizer is not used as expected. In fact, during testing, findTokenEnd() and terminateToken() were never called. I've tried setting a Validator but it wasn't used either. I'm not sure what is the reason.
    
I am afraid it still may work incorrectly for the full nicknames (ones with the instance name, like@name@isntance) because the search may happen for the instance name but it's not as critical.


charlag:
Oh crap, looks like I've used wrong parent commit, let me fix this

Vavassor:
Oh, weird. I was originally intending to replace all detection of mentions with the regex in SpanUtils.java, but I missed this class. It's not important to change, since it's not causing issues now, though.

charlag:
@vavassor well, it won't properly search for full nicknames as I said but I'm not sure how to use regexp there.

PR #415, opened 2017-10-25
i18n: PT-BR translation added

contraexemplo:
- European Portuguese and Brazilian Portuguese have fundamental differences between them, so it's reasonable to distinguish them. That's why the folder is named values-pt-rBR rather than values-pt.
- I tried to keep it similar to translations on Mastodon and other apps.


connyduck:
Thank you so much for the translation, your work is really appreciated <3

PR #418, opened 2017-10-26
Stop adding link info when composing toot

charlag:
Well, it's pretty self-explanatory. As discussed.


connyduck:
I think you can remove ParserUtils completely. Also there is more jsoup stuff in build.gradle and in proguard-rules.pro

charlag:
Yeah, probably.
Oh, crap, sorry, yeah, you're right, will do tomorrow, kinda tired so I missed it.

charlag:
I think that's it for jsoup now

PR #420, opened 2017-10-27
Use Google's maven repository

lindwurm:
`com.android.tools.build:gradle:3.0.0` is not available on jcenter.

now Google's Maven repository is officially recommended.

ref: https://developer.android.com/studio/build/dependencies.html#google-maven

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -2,6 +2,7 @@
 
 buildscript {
     repositories {
+        google()
         jcenter()
     }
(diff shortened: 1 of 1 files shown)


connyduck:
We already have that in another pull request #418, so merging this would only produce merge conflicts. But thanks nonetheless!

lindwurm:
oh I overlooked that.

PR #425, opened 2017-10-28
Fix issues with media uploads restoring. Fixes #419. Fixes #308.

### LINKED ISSUE #308: "Hide follow button while scrolling" text (under Preferences > Appearance) gets cut off
**Environment**
- HTC Desire 610
- Android 4.4.2
- API 19
- Display: 4.7 inches, 540 x 960 pixels
- Tusky version: 1.1.4-beta.3

**Reproduction steps** 
1. Open Tusky application
2. Open _Preferences_ from options menu
3. Under the _Appearance_ section, one will see the string _Hide follow button while scrolling_ gets displayed but only shows the word _scrolling_ partially (i.e. _Hide follow button while sc_).

### LINKED ISSUE #419: android.os.TransactionTooLargeException
There is an android.os.TransactionTooLargeException showing up in the PlayStore crash reports.
I suspect the problem is the onSaveInstanceState method of ComposeActivity. It writes whole bitmaps to parcels, very possible that the max transaction size gets exceded.

charlag:
Previews are not saved anymore but rather re-generated on restore.
Also it won't upload finished media anymore.

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/ComposeActivity.java
+++ b/app/src/main/java/com/keylesspalace/tusky/ComposeActivity.java
@@ -27,7 +27,6 @@ import android.content.res.AssetFileDescriptor;
 import android.content.res.Resources;
 import android.graphics.Bitmap;
 import android.graphics.BitmapFactory;
-import android.graphics.drawable.BitmapDrawable;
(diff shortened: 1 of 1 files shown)


connyduck:
makes sense, looks good, works, thank you!

Issue #174, opened 2017-04-16
Impossible to upload pictures as attachment


manuel-uberti:
Hi,

first of all thanks for Tusky.

I am using version `1.1.2` of Tusky on Android `6.0 MRA58K`, on a Xiaomi Redmi Note 4.

Whenever I try to attach a picture (both from the gallery and from the file manager), Tusky gives this message:

`That file could not be opened`.

If you need more details, please feel free to ask and I see what I can do.

tchncs:
Maybe tusky is lacking permission?

manuel-uberti:
Every permission available on my phone for Tusky is already enabled:

- Send MMS messages
- Access and change files on internal storage
- Show on lock screen
- Start in background

Vavassor:
It requests permission before opening the image picker, so the fact that it got past that and let you pick a photo seems to indicate that part is fine.

This error can appear in a couple of places. First it tries to look at the database entry to find out the file size (OpenableColumns.SIZE) and this failing would cause the error. Or after that, it opens the stream and makes a thumbnail to show as the preview image before it can be put in the upload queue proper. So the opening & thumbnail creation process failing would also lead to that error.

manuel-uberti:
Thanks for the explanation.

On the other hand, on my other phone (Fairphone, running Fairphone OS 1.13.0) latest Tusky works just fine. I was able to take a picture with my phone and toot it correcty (see: https://mastodon.social/@manuel_uberti/2623754).

manuel-uberti:
Since it seems to be related to the Redmi, is it ok if I close this one?

rototom:
Redmi 4 prime with same problem here.

charlag:
I think we will need to add (and collect) some logs, maybe special build will do.
I had problems with image uploading on Xiaomi devices before.

charlag:
@connyduck you've increased timeout - I guess this is fixed now?

connyduck:
I think these are two different problems. I fixed the one where Tusky could not upload a file. Here the problem is that the file can't even be opened.

charlag:
Ah, I see, sorry, I even commented about it, right

manuel-uberti:
Is this still an issue? I am not using Mastodon anymore, but I was just wondering if we can close this.

manuel-uberti:
Is this still an issue? I am not using Mastodon anymore, but I was just wondering if we can close this.

charlag:
Oh, sorry to hear that
[...]

PR #426, opened 2017-10-28
Add media upload progress. Closes #412

### LINKED ISSUE #412: Media uploading progress
As media uploads are quite unstable it would be nice if we could show progress when we upload media.
It is doable, but not easy.

charlag:
There's an upload progress shown on the media now.
Had some fun with merging with myself, I hope everything is okay (also couldn't help myself but formatted ComposeActivity).
photo_2017-10-29_00-06-23

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/BaseActivity.java
+++ b/app/src/main/java/com/keylesspalace/tusky/BaseActivity.java
@@ -145,7 +145,7 @@ public class BaseActivity extends AppCompatActivity {
 
         if (BuildConfig.DEBUG) {
             okBuilder.addInterceptor(
-                    new HttpLoggingInterceptor().setLevel(HttpLoggingInterceptor.Level.BODY));
+                    new HttpLoggingInterceptor().setLevel(HttpLoggingInterceptor.Level.BASIC));
         }
 
         Retrofit retrofit = new Retrofit.Builder().baseUrl(getBaseUrl())
--- a/app/src/main/java/com/keylesspalace/tusky/ComposeActivity.java
+++ b/app/src/main/java/com/keylesspalace/tusky/ComposeActivity.java
@@ -86,6 +86,7 @@ import com.keylesspalace.tusky.entity.Account;
 import com.keylesspalace.tusky.entity.Media;
 import com.keylesspalace.tusky.entity.Status;
 import com.keylesspalace.tusky.fragment.ComposeOptionsFragment;
+import com.keylesspalace.tusky.network.ProgressRequestBody;
 import com.keylesspalace.tusky.util.CountUpDownLatch;
 import com.keylesspalace.tusky.util.DownsizeImageTask;
 import com.keylesspalace.tusky.util.IOUtils;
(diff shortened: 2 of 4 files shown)


connyduck:
Really nice, this is a feature a wanted for a long time too. 
Some remarks:
- please remove the logging of the progress value, it totally spams logcat
- wouldn't a progress bar (circular or other) look better than a number?
- the percent number is really small, if you stay with numbers please make them bigger
- when the screen is rotated while the file is uploaded, the upload seems to restart? (maybe we should suppress recreation of the activity on config change `android:configChanges="orientation|screenSize|keyboardHidden"` in manifest?)
- add the license information on top of your new files please

(while playing with the uploads, I think I might have found what the problem with #174 is - we need to increase the socket timeout)

moggers87:
If you do stick with text rather than a progress bar/graphic, what are you doing to do about contrast? It looks fine in the screenshot, but I can't imagine it works too well with a nearly white image.

charlag:
@moggers87 I've implemented exactly the same thing at work. The image is dimmed (colorFilter).
Anyway, I'm experimenting with drawing something like a circular progress, but I struggle with drawing non-full circle. @connyduck did you ever do it? I'm thinking about saveLayer (which is expensive) or about vector drawable with trimEnd

connyduck:
@charlag We could take a library like this https://github.com/dinuscxj/CircleProgressBar
(or you can check out how they do it)

charlag:
@connyduck thanks for the lead. They're basically drawing full circle and then another non-full one on top of it. I cannot do it because I have to save the image below.
https://github.com/dinuscxj/CircleProgressBar/blob/master/circleprogressbar/src/main/java/com/dinuscxj/progressbar/CircleProgressBar.java#L313

connyduck:
Ah yes I see the problem. What about overlaying a determinate ProgressBar? It won't be circular, but is that a problem?

charlag:
@connyduck oh, it would complicate things in other places and will be less efficient. If nothing will work I will resort to separate ProgressBar but I would like to stay with one view per image if possible.

connyduck:
Good look is worth an addidtional view imho, but you are the one implementing it, so you decide.
If you can't find an easy solution let us go with the percent for now

charlag:
So, I've updated things:
* HttpLoggingInterceptor level is set to BASIC, upload progress logging is removed
* I've implemented kind of a circular progress bar. ```Canvas#saveLayer()``` is not the cheapest operation but I couln't find another way (and it's still cheaper than another view). Also we can customize it in any way we want.
* Please, please, don't do an ```android:configChanges```. It's very tricky to handle it correctly. I say we either adopt some library for component, which survives (which also brings separation of concerns and testability) - like ViewModel from Support LIbrary or Mosby or whatever. The point is, Activities and Fragments are huge now. I'd go with ViewModel (and Rx but LiveData will do). In case we won't use anything we can use ```FragmentActivity#onRetainCustomNonConfigurationInstance()``` or [retain a special fragment](https://developer.android.com/guide/topics/resources/runtime-changes.html) 
* Thanks for telling about license - I've completely overlooked it.

charlag:
What about this? photo_2017-10-29_17-36-43

connyduck:
👌

PR #430, opened 2017-10-31
Media tab

charlag:
This is a tracking PR for account media tab.
API part is implemented as well as most of the adapter.
Infinite scrolling is not really infinite yet, but we're close.
We also need to open pictures so we should associate media with status somehow, maybe it should be boxed into another object with status ID.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -1,4 +1,5 @@
 apply plugin: 'com.android.application'
+apply plugin: 'kotlin-android'
 
 android {
     compileSdkVersion 26
@@ -58,6 +59,7 @@ dependencies {
     compile "com.mikepenz:google-material-typeface:3.0.1.0.original@aar"
     compile "com.theartofdev.edmodo:android-image-cropper:2.5.1"
     compile 'com.evernote:android-job:1.2.0'
+    implementation 'com.android.support.constraint:constraint-layout:1.0.2'
 
     //room
     compile "android.arch.persistence.room:runtime:1.0.0-rc1"
@@ -67,5 +69,9 @@ dependencies {
     androidTestCompile('com.android.support.test.espresso:espresso-core:2.2.2', {
         exclude group: 'com.android.support', module: 'support-annotations'
     })
+    compile "org.jetbrains.kotlin:kotlin-stdlib-jre7:$kotlin_version"
 
+}
+repositories {
+    mavenCentral()
 }
--- a/app/src/main/java/com/keylesspalace/tusky/AccountActivity.java
+++ b/app/src/main/java/com/keylesspalace/tusky/AccountActivity.java
@@ -41,6 +41,7 @@ import android.util.Log;
 import android.view.Menu;
 import android.view.MenuItem;
 import android.view.View;
+import android.view.ViewGroup;
 import android.widget.Button;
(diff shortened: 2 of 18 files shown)


charlag:
Or I may be wrong and we don't need status to open photo.

charlag:
@connyduck please, review! It also adds Kotlin, someone has to do this 😎

connyduck:
I am sorry, but this one does not (yet) work for me. 
screenshot_20171101-211150

- The tabs just do not look right when there are four of them and one has no number. And not sure about the auto sizing Textview. I really would prefer the media in the menu.
- I think on small screens three columns of images are to much. Can you make it a integer resource with default value 2 and with 3 for large screens?
- You forgot the license again

But Kotlin integration is awesome!

charlag:
* Yeah, I'm okay with that, no need to be sorry! Do you want to experiment with moving out followers/following tabs? If not, I will just do menu thing
* Oh, I didn't test it, you're right! Will do.
* Damn, license, sorry, I should change template I guess, thanks for pointing.

connyduck:
Can you put it in the menu for now? I think it is best to just follow web mastodon for now. 
(Another possibility would be to make it like the Twitter app - 3 tabs with Toots, Toots with replies and media, put Following/Followers above the tabs. )

charlag:
@connyduck yeah, I am talking exactly about this another possibility :) I will put it in the menu for this release then and we can move things around in the next one?

connyduck:
Both options are fine for me - just don't mix follow and media tabs

charlag:
@connyduck are you sure about 2 columns on small screens? 3 seems to be pretty standard (Instagram e.g.). I can try it out of course. The thing is, it is intended to be an overview of photos. It may be pretty longs list with 2 columns. Which screen should I test it one/you did test it one?

charlag:
@connyduck okay, I think this is it. I've moved counters to the separate buttons so there are only 2 tabs left now. Added license, set columns to "2" on the small screens.
I also found a bug with ```onlyMedia``` parameter. It shouldn't be present when it means 'false'.
Sorry, I screwed up with git and smashed everything into the one commit, while it should be at least two. Hope it's okay.

connyduck:
Ok now its almost perfect, definitive a improvement to Tusky 1.2.0, I would merge it.
Just two more things:
What about a "posts with replies" tabs? (we could do this later as well)
I think the Follower/Following/Posts Line should be visually different to the bio. What about using the lighter font color?

charlag:
Yeah, I totally agree! But we will have to filter posts with replies ourselves, right? I am for the "later" for this.

I thought that it is different but it became pretty close  in the process, yeah. Picking colors is the hardest :D Will look into this tomorrow.

connyduck:
no, there is an API for this
https://github.com/tootsuite/documentation/blob/master/Using-the-API/API.md#accounts
GET /api/v1/accounts/:id/statuses?exclude_replies=true

charlag:
A-ha, so its a piece of a cake then.

charlag:
I've changed buttons colors, looks fine to me (the last tab is pressed on the screenshots to show different states)

photo_2017-11-05_17-26-04
photo_2017-11-05_17-26-01

charlag:
🙇

Issue #417, opened 2017-10-26
Saved toots forget to which status they reply


connyduck:
When saving toots, only the media urls, the content and the content warning are saved. We should also save the id of the parent toot, if there is one.
To avoid any confusion, there should be a possibility to see the parent toot in ComposeActivity.

PR #432, opened 2017-11-01
Reply improvements

charlag:
* I refactored things with ```ComposeActivity``` to make it safer, got rid of random extras strings.
* I've added views which show you what status you're replying to. It can be hidden and shown. 
* I didn't start working on fixing #417 because it would be too big for one PR. I'll wait for this to be merged before working on it.


connyduck:
Wow. Really nice. And pretty 😍
Thank you!

charlag:
My pleasure 🙇 
Thanks for reviewing

PR #435, opened 2017-11-04
"load more" Placeholder

### LINKED ISSUE #399: unable to load toots in the middle of the timeline
when tusky is restored from a previous session, new toots commonly will not load automatically. on a pull-to-refresh, it fetches the latest 40 or so toots, and prepends them to the top of the list. it seems to be impossible to load toots between the newly loaded ones, and the ones from a previous session, leaving an unclosable gap in the timeline.

## to reproduce

1. open tusky
2. allow toots to load
3. switch apps so tusky is no longer focused, come back after more toots than the fetch amount (40 or so? i'm guessing) have been made
4. scroll to the top of the timeline, pull to refresh
5. scroll past the newly loaded toots

## observed behavior

there is no division between old and new toots. they are part of a contiguous timeline without the option to load those that are in the gap.

## desired behavior

a division between the old and new toots would be shown, with an option to load those that are in the gap

connyduck:
fixes #399 

screenshot_20171104-215535

I think I got all special cases & I tested it extensively so it should work fine.
Not quite happy with some constructs & variable names though.

please review @charlag

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/fragment/TimelineFragment.java
+++ b/app/src/main/java/com/keylesspalace/tusky/fragment/TimelineFragment.java
@@ -15,6 +15,7 @@
 
 package com.keylesspalace.tusky.fragment;
 
+import android.arch.core.util.Function;
 import android.content.Context;
 import android.content.SharedPreferences;
 import android.graphics.drawable.Drawable;
@@ -25,6 +26,7 @@ import android.support.annotation.Nullable;
 import android.support.design.widget.FloatingActionButton;
 import android.support.design.widget.TabLayout;
 import android.support.v4.content.LocalBroadcastManager;
+import android.support.v4.util.Pair;
 import android.support.v4.widget.SwipeRefreshLayout;
 import android.support.v7.widget.DividerItemDecoration;
 import android.support.v7.widget.LinearLayoutManager;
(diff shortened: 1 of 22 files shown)


charlag:
Wow, this is tremendous work, gold job!
On the first glance im quite sad that we had to pollute Status with that placeholder things. We need StatusViewData exactly for things like this. Maybe we should ditch PairedList then? Im not sure. We can refactor it later od course to push the release sooner.
I will try to review if properly tomorrow, thank you for this fix, it's not easy but it is quite important!

charlag:
What we also could do is encapsulate primary list items in something like Either<Notification, Placeholder> / Either<Status, Placeholder>. It is additional boxing so we should be careful with that but it's probably the cleanest we can come up with while keeping PairedList.

We could also regenerate list of ViewData on changes to the basic list. Downside is that it's eror prone and tedious (that's why I'be came up with that list). Regenerating the whole list is easy but expensive.

app/src/main/java/com/keylesspalace/tusky/fragment/TimelineFragment.java
```diff
@@ -344,6 +343,24 @@ public void onContentHiddenChange(boolean isShowing, int position) {
...
+            String toId = statuses.get(position + 1).id;
+            sendFetchTimelineRequest(fromId, toId, FetchEnd.MIDDLE, position);
+
+            StatusViewData newViewData = new StatusViewData.Builder(statuses.getPairedItem(position))
+                    .setPlaceholderLoading(true).createStatusViewData();
+            statuses.setPairedItem(position, newViewData);
+            adapter.changeItem(position, newViewData, false);
+
+        } else {
+            Log.d(TAG, "error loading more");
```
charlag: I think ```Log.w()```, ```Log.e()``` or throwing an error is more appropriate here. Just saying.

connyduck:
there was a true/false mixed up so you could not click a placeholder again after a load failed, I fixed it @charlag

PR #436, opened 2017-11-05
Make placeholder abstraction cleaner

charlag:
Okay, this may seem as an overkill but I've made it as clean as I could and it may save us some future headaches - programmer error will crash the app, not fail silently (which I would prefer in most cases).
I did this only for notifications but I would like to do the same for statuses, I just don't have time right now.
I modeled placeholder data using Either because Notification itself shouldn't be concerned about presentation so I think adding Placeholder as a notification type is not the best practice. ```Either``` shouldn't be huge overhead.
```NotificationVIewData``` is modeled using class hierarchy because it is okay for it to know that it may be a placeholder and it is cheaper.
I could implement ```Either``` using class hierarchy too but I just wrote it out of my mind. I am a little bit worried about ```equals()``` but it seems okay.
We could measure how bad is memory overhead and conclude if it is okay to do the same for statuses.
I didn't test it extensively but it seems to work.
I am also worried about one place in ```NotificationsFragment``` (I've left an "is it safe?" comment), I didn't dig into the logic deep enough, can you verify it please?
Thank you, once again, awesome work on logic (which is usually very hard for me) and I would like to help you clean it up a little now.

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/fragment/NotificationsFragment.java
+++ b/app/src/main/java/com/keylesspalace/tusky/fragment/NotificationsFragment.java
@@ -44,6 +44,8 @@ import com.keylesspalace.tusky.entity.Status;
 import com.keylesspalace.tusky.interfaces.ActionButtonActivity;
 import com.keylesspalace.tusky.interfaces.StatusActionListener;
 import com.keylesspalace.tusky.receiver.TimelineReceiver;
+import com.keylesspalace.tusky.util.CollectionUtil;
+import com.keylesspalace.tusky.util.Either;
 import com.keylesspalace.tusky.util.HttpHeaderLink;
 import com.keylesspalace.tusky.util.ListUtils;
 import com.keylesspalace.tusky.util.PairedList;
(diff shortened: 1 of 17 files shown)

### CODE REVIEW

**Comment by connyduck** [comment]
This is indeed cleaner, I like it. It may get proplematic once we implement a database cache, a List<Notification> is easy to persist with Room, a List<Either<Placeholder, Notification>> not so much.
Of course it also is a small performance hit, but I think its acceptable. (we should do some perf improvements like simplifying views soon anyway)

"is it safe?" to be completely sure, add a check if the element at `position` rellay is a placeholder. The method is directly called from the OnClickListener and not over a network callback so it would be really weird if something changed inbetween.

**Comment by charlag** [comment]
We don't need to persist EIther, we only need to persist ```Notification```s and then lift them if needed. If you think about it, persisting placeholders would be weird.
I am glad that we agree on this. I will try to transition posts as well then? (perhaps in another PR). I will also add checks here.
Thanks for looking at it!

**Comment by connyduck** [comment]
But how do we know where toots are missing is if we don't persist a placeholder? (but lets discuss this once we are implementing it)

Yes please do it for statuses as well

File: app/src/main/java/com/keylesspalace/tusky/fragment/NotificationsFragment.java [hunk 1]
```diff
@@ -613,17 +633,29 @@ private void insert(List<Notification> newNotifications, int pos) {
             return;
         }
 
-        if(newNotifications.size() >= LOAD_AT_ONCE) {
-            Notification placeholder = new Notification();
-            placeholder.type = Notification.Type.PLACEHOLDER;
-            newNotifications.add(placeholder);
+        List<Either<Placeholder, Notification>> lifedNew = liftNotificationList(newNotifications);
+
+        if (newNotifications.size() >= LOAD_AT_ONCE) {
```
charlag: @connyduck could you explain this one please? I don't understand this condition, if we loaded more than the limit, then we add the placeholder. How can we load more than the limit? I'm lost.

charlag:
@connyduck I've added another commit, this time for ```Status```. Works fine for me. Please, review.

charlag:
Nice thing is that while I was writing this I pretty much reviewed #435. Turns out it's not as complex as I thought it would be but I could still mess it up. Many thanks for implementing this, @connyduck.

connyduck:
Looks good, I will test it tomorrow 👍

PR #437, opened 2017-11-07
Russian language updated

MasterGroosha:


app/src/main/res/values-ru/strings.xml
```diff
@@ -144,7 +147,7 @@
     <string name="pref_title_notification_settings">Уведомления</string>
     <string name="pref_title_edit_notification_settings">Изменить уведомления</string>
     <string name="pref_title_notifications_enabled">Push-уведомления</string>
-	<string name="pref_title_pull_notification_check_interval">Интервал проверки</string>
+    <string name="pref_title_pull_notification_check_interval">Интервал проверки</string>
```
charlag: We should add that this is about notifications, don't you think?

MasterGroosha: @charlag Good question, Ivan. I think there shouldn't be any changes, because the subtitle says "Уведомления" and the very next option is obviously about notifications, so there shouldn't be any confusion at all.

image

But there are some another things I'd like to discuss:
1. The "time intervals" strings (1 minute, 2 minutes...) can't be translated, right?
2. What about changing "Изменить уведомления" to "Настройки уведомлений"? IMO it looks more "standard" for applications.

charlag: Ah, yes. you're absolutely right, I was on my phone and missed some context, thank you.

1. They can be, please look for "array.xml" in different "values" folders.
2. Yes, I totally agree, some far more natural.

Thank you for your translations, they're pretty good. I am always skeptical about translations (so all my gadgets and services are in English) but yours are natural.

PR #438, opened 2017-11-07
pt-rBR translation updated

contraexemplo:


charlag:
@contraexemplo thank you!
@connyduck I don't speak it - should we just merge it?

connyduck:
Yes, I think we can trust people with this

Issue #292, opened 2017-05-17
Display @ mentions outside the content warning


Vavassor:
This is how the website displays them. Also, it makes it clear that the post is a reply and who it's replying to without having to "show more".

charlag:
@connyduck still relevant? Did you decide against it?

connyduck:
Well I would like Tusky to be quite similar to the web, and this is definitely something we should do, but it doesn't have any priority.

ariasuni:
I thought this might not be too complicated to do so I checked the code a bit but I couldn’t find where toot displaying is done. Do you think it would be easy to do for someone not familiar with the code base? I’d like to have some pointers, and try to do that if I find the time.

charlag:
@ariasuni I think it's totally doable and I would like to help!
I think you need to look into [StatusViewHolder](https://github.com/tuskyapp/Tusky/blob/89f93eab853d99a92340dc2d5240192f098d4a7a/app/src/main/java/com/keylesspalace/tusky/adapter/StatusViewHolder.java)
If you need any additional help feel free to ping me here, on Masto or in our Riot room!

charlag:
(or base view holder)

ariasuni:
So I looked a bit at the code and managed to do something really buggy. I’m not sure how I can do that properly.

Here’s what I found:

- Content visibility is changed [here](https://github.com/tuskyapp/Tusky/blob/59257c7af4bbca6e068d4868126669df20294308/app/src/main/java/com/keylesspalace/tusky/adapter/StatusBaseViewHolder.java#L402), [here](https://github.com/tuskyapp/Tusky/blob/59257c7af4bbca6e068d4868126669df20294308/app/src/main/java/com/keylesspalace/tusky/adapter/StatusBaseViewHolder.java#L405) and [here](https://github.com/tuskyapp/Tusky/blob/59257c7af4bbca6e068d4868126669df20294308/app/src/main/java/com/keylesspalace/tusky/adapter/StatusBaseViewHolder.java#L412).
- Toot text is modified to make links/hashtags/mentions clickable [here](https://github.com/tuskyapp/Tusky/blob/59257c7af4bbca6e068d4868126669df20294308/app/src/main/java/com/keylesspalace/tusky/util/LinkHelper.java#L65).

I tried to change the text of the view instead of changing the visibility of the text. So here’s what I did:

- filter the toot text to only have mentions (it may it easiest to use the mentions array directly)
[...]

PR #439, opened 2017-11-07
Status notifications behavior (closes #322, partly #292)

### LINKED ISSUE #322: [Discussion] Notification tab actions
Currently, when you tap on notification in Notifications tab it opens the person notification is related to but there's no way to open the toot which notification is about.
I propose to change current behavior for fav/boost notifications and open the toot of the notification by tapping content of the notification while account would be opened by tapping the line on top of the toot (like XX favourited your toot/XX boosted your toot). Behavior for replies/follow notifications should stay the same.
I already implemented boost handling in #321 and changing behavior would be really easy and I am eager to implement it. Please, post your thought and opinions.

charlag:
I'm sorry, I'm doing non-planned things again but this was *so annoying*.

So, in summary:
* Notifications now work like boosts in timelines - clicking on a status opens the status, clicking on the message on the top opens account of the person who caused the notification.
* CWs now work like they do in the web version. Firstly, they're hidden in the notifications by default and user can trigger them. Secondly, when content is hidden mentions are still visible (when should do the same for timelines, code will stay pretty much the same).

Please, review!
(closes #322, partly #292)


charlag:
I forgot to set spoiler text color first, now it should be fine.

connyduck:
Yes this was annying, thank you very much!

charlag:
🙇

PR #445, opened 2017-11-10
UI Improvements

connyduck:
In this branch I will experiment with some ui stuff. Feedback welcome!


connyduck:
Do people like the SparkButtons? I don't. The animation always gets cut off, is not antialiased and annoying imho.

connyduck:
Before:
screenshot_20171110-125258

Changes so far: 
screenshot_20171110-125258x

imho looks much better

charlag:
@connyduck I kinda liked them at first but grew tired with time. It could be okay if they were faster, smaller and worked correctly (but they glich sometimes for me). I didn't experience any anti-aliasing or cutting problems for a long time, though. I am not sure how people will react but if there will be another kind of a callback it is fine.

charlag:
Nice job, by the way, looks tidy!

charlag:
I would get rid of the links underline too.

connyduck:
Great idea!
Next step: improving notifications the same way and polishing the detail status.

charlag:
Awesome!

charlag:
If we're already here... Do you like horizontal animations for opening activities? I don't like them much. They would also look weird with shared elements transitions which I would like to add at some point.

Also, I still forget to focus on field automatically in search. What do you think about it?

connyduck:
Well the activity animations don't really bother me personally but I heard it from other people as well that they should be removed.
And yes, the search field should focus automatically. This should be an easy fix.

connyduck:
New detail statuses! 
Before:
screenshot_20171111-133319

After:
screenshot_20171111-132333

Still missing: number of boosts/favs in the same colour as the icon, no boosts number when private.

charlag:
Awesome thank you!
I see a small problem on the picture - there's no bottom line on the expanded status.

charlag:
Or is it intended?

connyduck:
I am not sure what you mean exactly? Everything is intended, the thread line would have crossed the text.

charlag:
Yes, you're right, my apologies. I checked and Twitter works pretty much the same way

connyduck:
So now notifications are improved as well :)

Before:
screenshot_20171117-155839

After:
screenshot_20171117-155419

charlag:
Wow, this looks awesome! Thank you.

connyduck:
Less views in layouts -> more performance ✨

connyduck:
So this is now mostly done. Should fix https://github.com/Vavassor/Tusky/issues/117 https://github.com/Vavassor/Tusky/issues/446 https://github.com/Vavassor/Tusky/issues/192 and partly https://github.com/Vavassor/Tusky/issues/429
The only thing missing is that MediaActivity still behaves strange, if a post has gifs&plain images attached.
Using Glide instead of Picasso is NOT a solution, since mastodon "gifs" are in fact mp4s.

I will have another look if I find a better spark button library

connyduck:
So I decided that Tusky 1.4.0 should be released asap, main reason is that crash rate in play console is really high. Please review this pull request, it contains at least one fix for a crash (should have made it a separate commit). @charlag

charlag:
Will do! It's not too late, right?

app/src/main/res/layout/item_status.xml
```diff
@@ -318,14 +322,12 @@
 
         <com.varunest.sparkbutton.SparkButton
             android:id="@+id/status_reblog"
-            android:layout_width="40dp"
```
charlag: I thought you would keep button size the same and only make icon inside it smaller. It is okay for me, just worrying for people who claim themselves clumsy.

app/src/main/java/com/keylesspalace/tusky/TuskyApplication.java
```diff
@@ -39,6 +39,7 @@ public void onCreate() {
         Picasso.Builder builder = new Picasso.Builder(this);
         builder.downloader(new OkHttp3Downloader(OkHttpUtils.getCompatibleClient()));
         if (BuildConfig.DEBUG) {
+            builder.loggingEnabled(true);
```
charlag: https://square.github.io/picasso/2.x/picasso/com/squareup/picasso/Picasso.Builder.html#loggingEnabled-boolean-

"WARNING: Enabling this will result in excessive object allocation. This should be only be used for debugging purposes. Do NOT pass BuildConfig.DEBUG."

charlag: Ah, you already removed it, sorry

app/src/main/java/com/keylesspalace/tusky/adapter/NotificationsAdapter.java
```diff
@@ -377,46 +431,37 @@ void setAvatars(@Nullable String statusAvatarUrl, @Nullable String notificationA
...
         }
 
-        private void setupContentAndSpoiler(boolean shouldShowContentIfSpoiler) {
+        private void setupContentAndSpoiler(boolean shouldShowContentIfSpoiler, final LinkListener listener) {
+            this.linkListener = listener;
+
             boolean hasSpoiler = !TextUtils.isEmpty(statusViewData.getSpoilerText());
-            CharSequence content;
             if (!shouldShowContentIfSpoiler && hasSpoiler) {
-                if (statusViewData.getMentions() != null &&
```
charlag: Sorry, but I don't understand why you undid this. I've tried to match web version, there's even an issue somewhere asking for this behavior. Please, consider using this for the rest of the app.

charlag:
I am not sure if it's a bug or intended behavior but when I click on the notification text it doesn't get me anywhere. I have to aim at the empty space.

charlag:
There's no more "delete" button in the timeline, only in the detailed view. When I tried to delete my status from there it didn't disappear from the timeline (it was on the top).

charlag:
I think that font is statuses is a little bit too bit. I would do -1 sp or something.
Also font in detailed status is smaller than in the regular one?

app/src/main/java/com/keylesspalace/tusky/adapter/NotificationsAdapter.java
connyduck: Ok, but why was it only done for notifications? I found it highly confusing that in notifications mentions are outside spoiler, but not in normal toots

connyduck:
I will fix the notification click area, thx for noticing.

Afaik there never was a delete button in the timeline (at least not since I am maintainer). This should be consistent, yes, but I will make a seperate issue for it. That the staus does not disappear is a bug that we also have for a long time.

Well Material design recommended size for regular text is 16sp. I am now using the 15sp for some days and I really love it. I think we should add a option to change the font-size between 12-18.
I am not sure, I believe font size in detailed status only looks smaller, but is actually the same size, but will check again.

charlag:
Oh, sorry, my memory ia not really great and can give me false memories.
I've checked detailed and it looks the same, my bad.

app/src/main/java/com/keylesspalace/tusky/adapter/NotificationsAdapter.java
charlag: Yes, it was only done for notifications because I didn't have time to change it everywhere.

PR #447, opened 2017-11-11
Update strings.xml

BoFFire:
Updating arabic translation from Transifex.
https://www.transifex.com/opendz/tusky-in-arabic/language/ar/
Best regards,


charlag:
Thank you!

PR #450, opened 2017-11-12
Cancel notifications on opening notifications tab

### LINKED ISSUE #422: Notification about interactions could disappear after opening interactions tab

charlag:
This closes #422.


connyduck:
There is already a `clearNotifications` method in BaseActivity, I think it is best to make it static, put it in NotificationMaker (rename NotificationMaker to NotificationManager) and reuse that one method everywhere notifications get cancelled. 
Also, does the `NotificationClearBroadcastReceiver` get called when a notification is cancelled programmatically? If so `clearNotifications` doesn't need to change the SharedPreferences.

charlag:
@connyduck oh, thanks, will fix it!
I'm not sure, should check.

charlag:
I've updated it.
I've also checked ```NotificationClearBroadcastReceiver``` and it's not called when I cancel notifications programmatically.

PR #451, opened 2017-11-12
Don't load public timlines when offscreen

charlag:
This looks hacky but it worked suprisingly well for me.
I am not sure you will agree with this, because it takes time to load but I believe that it's better than constantly fetching timelines user doesn't see.

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/fragment/TimelineFragment.java
+++ b/app/src/main/java/com/keylesspalace/tusky/fragment/TimelineFragment.java
@@ -150,18 +150,25 @@ public class TimelineFragment extends SFragment implements
     }
 
     @Override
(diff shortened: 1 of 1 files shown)


connyduck:
Yes I am really not sure. I personally like it that timelines load in the background, but I can understand that not everybody wants that. I think we need a complete rework of this together with timeline caching.

charlag:
It's okay, I understand! But how would caching help? If would request the data from server every time the fragment is requested anyway.

connyduck:
Well we could display cached toots in the background and start a network load once in the foreground, that would hide the network latency better I suppose. 
Is it ok if I close this PR for now?

charlag:
I will close if, thanks for looking at it.

PR #454, opened 2017-11-13
Tweaks to Russian translation

MasterGroosha:
• Added missing translations to timestamps (1s, 2m, 3h etc);
• Renamed "bio" section (now it's translated as "about me" since "biography" means something large and detailed);
• Renamed "thread" sections (now it uses more native word "Discussion", rather than transliterated "Thread" word);
• Changed notification string about new notifications. This change may look a bit ugly, but it solves a problem with declension of numeric values.


charlag:
Thanks!

PR #456, opened 2017-11-13
simplify MainActivity back press handling

connyduck:
closes https://github.com/Vavassor/Tusky/issues/453


charlag:
@connyduck LGTM, I've also checked it on my phone and it works. Thank you.

PR #458, opened 2017-11-14
fix clear notifications crash

connyduck:
So this is how I would fix the crash, please test extensively before merging. (or wait until I find time for testing, probably tomorrow)


charlag:
Thanks for the fix! Sorry, It's very late here, cannot test today.

connyduck:
Tested it, could not find any problems!

PR #449, opened 2017-11-12
Save reply info in draft, refactor

charlag:
This fixes #417 and closes #440 and there's also a little bit of refactoring.
I actually did much more than I should have done first. I've tried saving ```Mention```s but turns out I didn't need them for this. It would be a good idea to save the toot we're replying to but the DB is not ready yet for such thing. So I've just added some fields and that's it.
I didn't understand the visibility logic first. I hope I didn't break anything.
I've tried to remove as much as possible from ```ComposeActivity``` because it's huge.
Sorry that it's all smashed together but it's mostly related things.

PR #459, opened 2017-11-14
[WIP] Multiaccount

charlag:
Tracking branch for multiaccount support.
Should be rebased (will probably do after #449 is merged)


connyduck:
Well I assigned the issue https://github.com/Vavassor/Tusky/issues/54 to me because I wanted to work on this xD
But it is ok if you are faster.

charlag:
I've missed it, sorry! I will not be able to work on it until Sunday, probably. So you're free to pick it from here or throw away or just leave to me, I am okay with any case

PR #469, opened 2017-11-28
Remove obsolete translate

nailyk-fr:
Solve build problems

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/ComposeActivity.java
+++ b/app/src/main/java/com/keylesspalace/tusky/ComposeActivity.java
@@ -916,7 +916,6 @@ public final class ComposeActivity extends BaseActivity
 
         if (response != null && inReplyToId != null && response.code() == 404) {
             new AlertDialog.Builder(this)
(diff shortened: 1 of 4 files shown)


connyduck:
Why are you removing the Message from the dialog? A dialog without message makes no sense.

connyduck:
If you are getting an build error try setting `abortOnError false` in `lintOptions` in `build.gradle`.
Sorry, but currently only tagged releases are guaranteed to be buildable/stable.
(Yes I know we should set up a buildserver to improve this, I will do so once I find time)

nailyk-fr:
Here you are https://releases.nailyk.fr/jenkins/job/F-Droid/job/k_tusky/ :)
Yes had no idea about the importance of this removal. 
So problem is `dialog_reply_not_found` is not in source string anymore: https://github.com/nailyk-fr/Tusky/commit/1c2a647569d29440c2b09d0d41aeddd6c67632bc#diff-01dafb3fd0217441330103cfc8300094L149

PR #470, opened 2017-11-28
Correct array translations

nailyk-fr:
Next to https://github.com/Vavassor/Tusky/pull/469
Should be a better fix.


connyduck:
Yes thank you! Seems like I made a mistake when merging.

Issue #130, opened 2017-04-12
[Feature request] Allow to configure a proxy in Tusky


cjeanneret:
Hello,

It would be marvelous if user could configure a proxy in the application. For example, that feature would allow users to make connections through Tor, or other services like that.

Thank you for your time, and a big, big :+1: for Tusky as my preferred app for Mastodon :)

Cheers,

C.

connyduck:
Proxy support was added in Tusky 1.5 :)

PR #489, opened 2017-12-18
implement support for HTTP proxy

slp:
This change allows the user to manually enter an unauthenticated proxy
configuration to be used for all API connections. This is mainly
intended for using Tusky with Tor (via Orbot or a local proxy).

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/BaseActivity.java
+++ b/app/src/main/java/com/keylesspalace/tusky/BaseActivity.java
@@ -137,8 +137,10 @@ public abstract class BaseActivity extends AppCompatActivity {
                 .registerTypeAdapter(Spanned.class, new SpannedTypeAdapter())
                 .create();
 
+        SharedPreferences preferences = PreferenceManager.getDefaultSharedPreferences(this);
+
         OkHttpClient.Builder okBuilder =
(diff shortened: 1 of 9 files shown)


slp:
Fixes #130

connyduck:
This is a really cool feature, thanks for your pull requests!

Some small things before I merge it: I think it would be better to put the proxy settings in their own fragment like notification and timeline filtering settings. In the main settings fragment a summary would be displayed.
Proxy server and proxy port settings should be disabled when the checkbox is not checked (you can do this with `android:dependency` in xml) OR remove the checkbox and only use the proxy when server&port are filled in.
What do you think?

slp:
@connyduck I'm OK with the changes you propose, but I think I'm going to preserve the checkbox, as it allows you to temporarily disable the proxy without removing the server/port configuration. Do you like this approach?

connyduck:
Well disabling the inputs won't delete the content, but it is just an idea, please do as you like :)

slp:
@connyduck I've just put the preferences on a separate fragment, with a summary in the preference box. Please let me know if you think something else can be improved. Thanks :)

connyduck:
It works, code looks good, super awesome, thank you!

Issue #411, opened 2017-10-24
Add caption functionality


connyduck:
Mastodon 2.0 feature for better accessibility
- [x] Add the ability to caption images for better accessibility
- [x] Set the imageDescription of the ImageViews to the caption, if one is given

connyduck:
done #500

PR #500, opened 2018-01-04
Set caption for media

charlag:
So, to finish #411 - setting caption.
There are some debatable UI choices. I went for popup menu and dialog, I think it works well but can be prettier.
I set dialog window to ```SOFT_INPUT_ADJUST_RESIZE``` because otherwise dialog buttons may be invisible. It doesn't always work perfectly, though.
I think I've also fixed a bug with not deleting media URLs properly

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/ComposeActivity.java
+++ b/app/src/main/java/com/keylesspalace/tusky/ComposeActivity.java
@@ -62,12 +62,15 @@ import android.text.style.URLSpan;
 import android.util.Log;
 import android.view.MenuItem;
 import android.view.View;
+import android.view.Window;
+import android.view.WindowManager;
 import android.webkit.MimeTypeMap;
 import android.widget.Button;
 import android.widget.EditText;
 import android.widget.ImageButton;
 import android.widget.ImageView;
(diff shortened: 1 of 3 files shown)


charlag:
Oh, crap, they're still separate commit, give me a sec, I will fix that

connyduck:
Well, we won't find the perfect solution, it is just so difficult to fit that many controls on a small screen.
What do you think about removing the popup and putting a delete button directly in the dialog?
Also the dialog could need some padding imo.
And for it to be perfect the captions should be saved in the database (but I will merge it without that as well)

charlag:
I can move it there if you think it will be better. As a third button?
Padding, yes, I agree

connyduck:
I asked what users think about it on the Tusky mastodon acc: https://mastodon.social/@Tusky/99299292541755594

connyduck:
So everybody seems to agree that click/long click is a bad idea. Lets stick with the popup. Can you add padding and resolve the merge conflict please?

charlag:
Yeah, sure, will do tomorrow.

charlag:
Done!

charlag:
It would clse #411

PR #501, opened 2018-01-04
Add basic lists support

charlag:
Bang! Lists list (no pun intended) and list timeline.
I miss Rx a lot but I am not sure that it is a good idea to pull it in yet.
I've decided to split Lists into ViewModel and View early, survives config changes without any dependencies by leveraging nonConfigurationInstance.
I also hate activities. Single Activity Applications FTW.

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -95,6 +95,8 @@
             </intent-filter>
             <meta-data android:name="android.app.searchable" android:resource="@xml/searchable" />
         </activity>
+        <activity android:name=".ListsActivity" />
+        <activity android:name=".ModalTimelineActivity" />
 
         <receiver android:name=".receiver.NotificationClearBroadcastReceiver" />
 
--- a/app/src/main/java/com/keylesspalace/tusky/ListsActivity.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/ListsActivity.kt
@@ -1,0 +1,199 @@
+package com.keylesspalace.tusky
+
+import android.content.Context
+import android.content.Intent
+import android.os.Bundle
+import android.support.v4.widget.TextViewCompat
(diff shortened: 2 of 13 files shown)


connyduck:
Wow, cool feature!

Things I noticed:
- The activities you created have no shadow under the toolbar. Have a look at the other activities, there is `@layout/toolbar_shadow_shim` included.
- The items in `ListsActivity` should be bigger and separated with a `DividerItemDecoration`. Maybe even add a list icon for decoration like the web ui does?
- I think we can move the menu item up above muted&blocked users?

Do you plan to extend this with more features?

charlag:
* Yes, you're right, I will add shadows
* Bigger like... bigger font? I can try adding an icon too.
* Makes sense!

connyduck:
Yes, bigger font, `?attr/status_text_medium` should be the right size. Also, give it a selectableItemBackground for touch feedback!

charlag:
Ah, crap, I always add it but now I've missed it! Thanks for noting, let's make this thing sexy.

About other features. I don't have any plans right now. It would be cool to check who's in the list and edit it but we have more high-priority tasks and I cannot promise anything. I would like it to be merge with the features it has now.

connyduck:
ok, fine for me

charlag:
I've updated it. Looks like this now.
I had to hardcore shim height in the layout, it doesn't work otherwise
screenshot_20180105-232223

connyduck:
looks good except for the icon not having the same color as the text and it probably should not be an imageview but an CompoundDrawable on the TextView

charlag:
Didn't ```CompoundDrawable``` bite you with compatibility things? Any advice?

connyduck:
If you take the IconicDrawable the xml inflating wont be a problem and you can get the color with ThemeUtils.getColor in code :)

charlag:
Fixed that too!

connyduck:
excellent

PR #502, opened 2018-01-14
Theming improvements

remi6397:
These patches make easier to add new themes in future. It also leverages [_UiModeManager_](https://developer.android.com/reference/android/app/UiModeManager.html) functionalities to allow automatic switching between day and night mode based on local sunset/~rise times.

I've tested these patches on Oreo 8.1 and they work flawlessly.


connyduck:
First things first: This is super awesome.

Now the details:

Yes, it works perfectly on my Android 8.1 device. But it does not work on my Android 5 device, there it always shows the light theme.
To change this, you have to use the AppCompat methods instead of the System ones: https://developer.android.com/reference/android/support/v7/app/AppCompatDelegate.html#setDefaultNightMode(int)

While I find the Tusky/Dusky thing a nice pun, I think it is more user friendly to call it Tusky dark and Tusky light. Also you should show the name of the currently selected theme in the preference summary (to do this, just add `android:summary="%s"` in xml).

app/src/main/java/com/keylesspalace/tusky/util/LinkHelper.java
```diff
@@ -172,8 +171,11 @@ public static void openLinkInBrowser(Uri uri, Context context) {
      * @param context context
      */
     public static void openLinkInCustomTab(Uri uri, Context context) {
-        boolean lightTheme = PreferenceManager.getDefaultSharedPreferences(context).getBoolean("lightTheme", false);
-        int toolbarColor = ContextCompat.getColor(context, lightTheme ? R.color.custom_tab_toolbar_light : R.color.custom_tab_toolbar_dark);
+        TypedValue toolbarColorTv = new TypedValue();
+        context.getTheme().resolveAttribute(
+                context.getResources().getIdentifier("custom_tab_toolbar", "attr", context.getPackageName()), toolbarColorTv, true);
+
+        int toolbarColor = toolbarColorTv.data;
```
connyduck: there should already be a helper method for this in com.keylesspalace.tusky.util.ThemeUtils

app/src/main/java/com/keylesspalace/tusky/fragment/AccountMediaFragment.kt
```diff
@@ -133,10 +133,12 @@ class AccountMediaFragment : BaseFragment() {
...
-        val lightThemeEnabled = PreferenceManager.getDefaultSharedPreferences(context)
-                .getBoolean("lightTheme", false)
-        val bgRes = if (lightThemeEnabled) R.color.window_background_light
-        else R.color.window_background_dark
+        val windowBackgroundTv = TypedValue()
+        activity?.theme?.resolveAttribute(
+                R.attr.window_background, windowBackgroundTv, true)
+
+        val bgRes = windowBackgroundTv.resourceId
+
```
connyduck: there should already be a helper method for this in com.keylesspalace.tusky.util.ThemeUtils (if not, add one)

app/src/main/java/com/keylesspalace/tusky/LoginActivity.java
```diff
@@ -234,15 +253,11 @@ private static String toQueryString(Map<String, String> parameters) {
...
-            toolbarColorRes = R.color.custom_tab_toolbar_light;
-        } else {
-            toolbarColorRes = R.color.custom_tab_toolbar_dark;
-        }
-        int toolbarColor = ContextCompat.getColor(context, toolbarColorRes);
+        TypedValue toolbarColorTv = new TypedValue();
+        context.getTheme().resolveAttribute(
+                context.getResources().getIdentifier("custom_tab_toolbar", "attr", context.getPackageName()), toolbarColorTv, true);
+
+        int toolbarColor = ContextCompat.getColor(context, toolbarColorTv.resourceId);
```
connyduck: there should already be a helper method for this in com.keylesspalace.tusky.util.ThemeUtils (if not, add one)

remi6397:
I think that everything is done… except theme name. Dusky/Tusky is a kind of compromise between a title that would mislead devs and a title that would mislead users. In code, it defaults to light theme, but dark is the default anyways. If you don't like this solution, I'll change this name.

remi6397:
When I changed night mode handler from UiModeManager to AppCompatDelegate, it looks like navbar buttons don't change color properly. Will try to fix them.

connyduck:
Thank you! 😊
Looks good & works well. If you change the names of the themes, I will merge it!

remi6397:
@connyduck, the only problem if users would like to enable Night mode, then they need to know somehow, that only the light one can be used with that option. Thats UX problem, and someone in near future would open an issue on that.

There is a second option - I could replace default values with night ones and move light ones to "-notnight". But maybe not today. I'll try, but I can't promise. I'm a bit tired. 😉

remi6397:
Also, I think that most new themes would be light by default, and that deviation would negatively affect maintainability of Tusky.

But again, that's author's choice. I can just indicate potential problems.

connyduck:
Well now it is definitely confusing 🤔 
Maybe we should remove the checkbox and have three themes Tusky dark, Tusky light and Tusky auto?
And I would add an explanation to that permission.

remi6397:
@connyduck Unfortunately, no. It's design problem. I would need to rewrite it, or the names will be misleading.

Currently, we have two choices:
* Split themes to default (fallback) and one alternative for night (when default is for daytime theme) or day. If default one is selected, we can enable auto night mode.
* Make just one variation of theme and define it as either day (light) or night (dark). Then, it couldn't be used with auto-daylight option, obviously.

So, please take a look on current implementation and indicate possible changes. ;-)

But wait… What about "Default" and "Default Dark"? I think it's a good compromise.

remi6397:
When it comes to permission, I have no idea how to add a description to dialog… Maybe I should display another dialog window before?

connyduck:
I am sorry, I don't understand why you think the names are misleading? Dark, light and automatic based on time seem perfectly understandably for me?

> When it comes to permission, I have no idea how to add a description to dialog… Maybe I should display another dialog window before?

yes thats what I meant

remi6397:
@connyduck No need to worry, just take look at `values*/styles.xml` in my revision. It explains everything. :-P When we add more themes, it would be impractical to display about three variants for every single theme. I'll think about that problem tomorrow.

Also, we can inform about need for permission in option summary. ;-)

remi6397:
Also, I need to adjust code to [avoid repeating myself](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself?wprov=sfla1). :stuck_out_tongue_closed_eyes:

remi6397:
What do you think about [this approach](https://github.com/remi6397/Tusky/tree/night-theme-default)? If it satisfies you, then I could merge that branch into this PR.

connyduck:
You mean the notnight change? I think both is fine. The internals are not that important to me, the user experience should be good and now thats not the case with the sometimes deactivated checkbox.

What do you think about two drop downs, one where the user selects the base theme, one to select dark-night-auto? This we we would avoid adding 3 variants for each theme

remi6397:
> What do you think about two drop downs, one where the user selects the base theme, one to select dark-night-auto? This we we would avoid adding 3 variants for each theme

Sounds good! I'll surely look into that. :blush:

remi6397:
Hmmm… :thinking: But what if some theme would be available only in single flavor? I'll think about how to detect theme variant availability.

remi6397:
Ufff… Now it looks better. @connyduck, take a look at newest changes and have a good week, as I'll probably be less available till Friday.

**I wanted to warn**, that I'm not sure whether this code contains bug or not, but we definitely need to do some tests before merge.

remi6397:
Ahh, and I need to remove this "default" option. It could be just computed after theme changes.

app/src/main/java/com/keylesspalace/tusky/LoginActivity.java
```diff
@@ -68,9 +75,16 @@
     protected void onCreate(Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
 
-        if (PreferenceManager.getDefaultSharedPreferences(this).getBoolean("lightTheme", false)) {
-            setTheme(R.style.AppTheme_Light);
-        }
+        preferences = PreferenceManager.getDefaultSharedPreferences(this);
+        String[] themeFlavorPair = preferences.getString("appTheme", "AppTheme:prefer:night").split(":");
+        String appTheme = themeFlavorPair[0], themeFlavorMode = themeFlavorPair[1], themeFlavorPreference = themeFlavorPair[2];
```
connyduck: themeFlavorMode is not used anywhere

app/src/main/res/values/strings.xml
```diff
@@ -147,7 +147,12 @@
     <string name="dialog_download_image">Download</string>
     <string name="dialog_message_follow_request">Follow request pending: awaiting their response</string>
     <string name="dialog_unfollow_warning">Unfollow this account?</string>
-    <string name="dialog_reply_not_found">Couldn\'t post this status. The status you\'re replying to might not be available. Remove reply info?</string> 
+    <string name="dialog_reply_not_found">Couldn\'t post this status. The status you\'re replying to might not be available. Remove reply info?</string>
+    <string name="dialog_optional_permission_request">Optional feature requires permission</string>
+    <string name="dialog_twilightmanager_coarse_location_permission_explanation">
+        Automatic theme adjustment option works best with location enabled.\n
+        We don't store or send your location anywhere.
```
connyduck: you need to escape the '

app/src/main/java/com/keylesspalace/tusky/PreferencesActivity.java
```diff
@@ -104,7 +155,32 @@ protected void onSaveInstanceState(Bundle outState) {
 
     public void onSharedPreferenceChanged(SharedPreferences sharedPreferences, String key) {
         switch (key) {
-            case "lightTheme": {
+            case "appTheme": {
+                String[] themeFlavorPair = sharedPreferences.getString("appTheme", "AppTheme:prefer:night").split(":");
+                String appTheme = themeFlavorPair[0], themeFlavorMode = themeFlavorPair[1], themeFlavorPreference = themeFlavorPair[2];
```
connyduck: themeFlavorMode and themeFlavorPreference are unused.

app/src/main/java/com/keylesspalace/tusky/PreferencesActivity.java
```diff
@@ -72,6 +77,50 @@ protected void onCreate(@Nullable Bundle savedInstanceState) {
...
+                        .setPositiveButton("OK", new DialogInterface.OnClickListener() {
+                            @Override
+                            public void onClick(DialogInterface dialog, int which) {
+                                ActivityCompat.requestPermissions(PreferencesActivity.this, permissions, 1);
+                                if (ContextCompat.checkSelfPermission(PreferencesActivity.this,
+                                        Manifest.permission.ACCESS_COARSE_LOCATION) != PackageManager.PERMISSION_GRANTED)
+                                    ThemeUtils.setAppNightMode(finalFlavor);
+                            }
+                        })
+                        .setNegativeButton("Ask me later", null)
```
connyduck: String needs to be translateable

app/src/main/java/com/keylesspalace/tusky/util/ThemeUtils.java
```diff
@@ -81,4 +88,34 @@ public static void setImageViewTint(ImageView view, @AttrRes int attribute) {
     public static void setDrawableTint(Context context, Drawable drawable, @AttrRes int attribute) {
         drawable.setColorFilter(getColor(context, attribute), PorterDuff.Mode.SRC_IN);
     }
+
+    public static boolean setAppNightMode(String flavor) {
+        switch (flavor) {
+            case "auto":
+                if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.M) {
+                    TuskyApplication.getUiModeManager().setNightMode(UiModeManager.MODE_NIGHT_AUTO);
```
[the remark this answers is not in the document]
remi6397: On Android M+, AppCompat implementation doesn't work.

connyduck:
I really like the day/night support, but I am still not a big fan of the themes. Themes currently will only switch colors. To make them better, they should be able to customize fonts and graphics and that does not play well with Tuskys leightweight approach unfortunately.

You know what would be super cool though?  🤔
A color choser in the settings where each user can pick their preferred primary color. This way we would get a great customization option instead of the themes.

app/src/main/java/com/keylesspalace/tusky/util/ThemeUtils.java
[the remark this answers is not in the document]
connyduck: you sure? It works on my device with 8.1

remi6397:
When it comes to advanced theming: this patchset isn't about user theme customization. I focused on making possible to add more themes in the future. And of course it makes possible to automatically switch between themes' variants, based on time of day. Maybe I'll try to implement more features in future, If I'll have some free time. :wink:

remi6397:
I think that I should remove that location popup as we now use UiModeManager for M+.

connyduck:
Can you please explain to me why AppCompatDelegate is not working for you? Because thats the very purpose of using the support library: You don't have to write api specific code.

Also, currently the Dialog is shown every time the preferences are opened while the permission is not granted and auto mode is selected. It should only be shown after changing the setting to auto.

remi6397:
When you use AppCompat night mode on M+, changing to auto night mode requires restart to make the changes appear on the screen.

Also, when you use native Android M+ night mode APIs, you don't need to ask for location permission - everything is handled by system. 

connyduck:
Why a new NightState class when you could do it with a static method?

remi6397:
Anonymous function would be the best solution, but AFAIR java doesn't support them. Private static method used by only one method… That idea doesn't sound good at all. :/

If you don't like this solution, I'll just move that logic to the end of that function.

connyduck:
Creating objects is just unnecessary overhead in this case.

charlag:
@remi6397 actually Java has anonymous objects & lambdas (which are the same thing on Android anyway) 😉
I see absolutely no need in creating a class and instances of it. What does it encapsulate?
There are couple of use cases for classes:
 * struct-like, keep related data together
 * as a collections of methods, keep related functions together
 * encapsulating some state

As far as I understand, you want to abstract that logic off night mode application, which is a great idea. It would be a great start for a new class called ThemeUtils... oh wait, we already have it. ThemeUtils is exactly the second case, being a namespace for related functions. I think the best solution would be the method (maybe even private one as it is an implementation detail).

charlag:
Anonymous classes are mostly used because previous maintainer did not use anything like retrolambda and there were no such functionality with default instruments. Since then Google worked on the compiler and made it cheap enough to not think about it and method count so we use lambdas now.

connyduck:
@remi6397 if it is ok for you I would merge it now?

remi6397:
I'm using Tusky with that patch and it works as intended. If you are OK with that code, you can merge it. I'm looking forward to seeing that functionality in next release of Tusky. 😊

PR #509, opened 2018-01-21
Privacy enhancement: option in preferences to enable Secure Screen

remi6397:
From user's perspective, Secure Screen prevents other apps (and users themselves) to take screenshots of Activities affected with that flag. It also hides application content preview from recents menu.

You'll still be able to launch application without authenticating.

This patch was tested under Android Lollipop 5.1 and Android Oreo 8.1. It's worth mentioning that on L it requires application to restart to apply that flag.

Feature preview


connyduck:
Sorry, but I don't think we need this. Tusky is not a banking app, its a social media app and mastodon is not suited for confidential communication anyway, this setting won't change anything about that.
Also, we are adding so many settings recently and I would really like to keep the settings count low.

PR #510, opened 2018-01-26
Update Polish translation

m4sk1n:


app/src/main/res/values-pl/strings.xml
```diff
@@ -153,37 +179,95 @@
...
     <string name="pref_title_notification_filter_mentions">wspomniano o mnie</string>
-    <string name="pref_title_notification_filter_follows">zaobserwowano mnie</string>
+    <string name="pref_title_notification_filter_follows">zaczęto mnie śledzić</string>
     <string name="pref_title_notification_filter_reblogs">moje posty zostaną podbite</string>
     <string name="pref_title_notification_filter_favourites">moje posty zostaną dodane do ulubionych</string>
     <string name="pref_title_appearance_settings">Wygląd</string>
     <string name="pref_title_app_theme">Motyw</string>
 
+
+    <string-array name="app_theme_names">
```
remi6397: Theme names don't need to be translated if they sound the same.

app/src/main/res/values-pl/strings.xml
```diff
@@ -57,8 +71,8 @@
     <string name="action_login">Zaloguj z Mastodon</string>
     <string name="action_logout">Wyloguj</string>
     <string name="action_logout_confirm">Czy chcesz wylogować się?</string>
-    <string name="action_follow">Obserwuj</string>
-    <string name="action_unfollow">Przestań obserwować</string>
+    <string name="action_follow">Śledź</string>
```
remi6397: "śledzić", "śledzenie", "śledzony", et cetera, mean "tracking" more than "following". "podążać" would be better translation, but in most cases "obserwować" sounds better. Anyways, most services (eg. Twitter) translate "follow" as "obserwować".

m4sk1n: But I’m using „śledzić” in Mastodon translation and some says it’s OK

remi6397: "Śledzony" might be confused with being tracked by some sort of "spy". Like NK's "śledzik" - people thought that someone was spying on them. Then that service died - mostly because of that [myth](https://proline.pl/?n=jak-usunac-sledzika-naszej-klasy). You may laugh, but many people (especially older people) have no idea how to properly "understand" that sort of things happening on the Internet.

m4sk1n: AFAIR there was discussion about it, it wasn’t just my opinion (but in fact I’m not getting any feedback about the translation)…

---

UWAGA! USUNĄĆ ŚLEDZIKA! Wklejcie to na swojego śledzika! Jeżeli
uzbieramy 70% użytkowników nk... to usuną śledzika!!

PR #522, opened 2018-02-08
notification improvements

connyduck:
I started this because of https://github.com/Vavassor/Tusky/issues/518
Now notification polling is disabled correctly on Android O when notifications are disabled.
I think it would be even better to implement [NotificationListenerService](https://developer.android.com/reference/android/service/notification/NotificationListenerService.html) so the Pull jobs get correctly enabled/disabled when the settings change even if the user does not open the app afterwards. 

Some other changes: 
- I renamed `NotificationManager` because I think it is confusing to use the same name as a system service.
- I deleted the list update in `AccountManager.saveAccount`, because the saved objects are in the list in all cases anyway and it was causing a crash, probably due to concurrency.

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/BaseActivity.java
+++ b/app/src/main/java/com/keylesspalace/tusky/BaseActivity.java
@@ -201,11 +201,11 @@ public abstract class BaseActivity extends AppCompatActivity {
     }
 
     protected void setPullNotificationCheckInterval(long minutes) {
-        JobManager.instance().cancelAllForTag(NotificationPullJobCreator.NOTIFICATIONS_JOB_TAG);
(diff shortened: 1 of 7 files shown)


connyduck:
Ok so the `NotificationListenerService` is apparently not intended for this use case. If a user has disabled Notifications on Android 8.0+, and then enables them again they might not receive notifications until the app has been openend again.

charlag:
Oh, crap, sorry, I totally forgot about it! In a few hours probs I will check it (I don't have an O device, though)

connyduck:
So as @nailyk-fr pointed out, there was no switch for the notification light on Oreo. It seems that setting a color on the led changes that.

PR #525, opened 2018-02-12
Rewrite EditProfileActivity in Kotlin

### LINKED ISSUE #504: Problems with Profile Editor's UI
There are several closely related problems:
- Right now, I can only remove my Bio by setting it to a whitespace (Mastodon strips whitespace from fields
- When I set my name to a whitespace, UI closes without errors, but name remains unchanged.

connyduck:
- fixes #504 
- makes the activity actually work again lol
- saves the image in the app cache dir and uses proper multipart encoding
- fixes a bug in MainActivity where profiles would duplicate
- probably fixes the TransactionTooLarge crashes I am seeing in the Play console, because the base64 shenanigans are gone
- small layout improvements
- Kotlin yeah yeah

the random filenames seems to be necessary so mastodon accepts the images


app/src/main/java/com/keylesspalace/tusky/EditProfileActivity.kt
```diff
@@ -0,0 +1,491 @@
...
+        }
+
+        val header = if(headerChanged) {
+            val headerBody = RequestBody.create(MediaType.parse("image/png"), getCacheFileForName(HEADER_FILE_NAME))
+            MultipartBody.Part.createFormData("header", getFileName(), headerBody)
+        } else {
+            null
+        }
+
+        if(displayName == null && note == null && avatar == null && header == null) {
```
charlag: Not clear why this check is needed. Perhaps a comment would help.

app/src/main/java/com/keylesspalace/tusky/EditProfileActivity.kt
```diff
@@ -0,0 +1,491 @@
...
+            EditProfileActivity.PickType.HEADER -> {
+                width = HEADER_WIDTH
+                height = HEADER_HEIGHT
+                cacheFile = getCacheFileForName(HEADER_FILE_NAME)
+            }
+            else -> {
+                throw AssertionError("PickType not set.")
+            }
+        }
+        ResizeImageTask(contentResolver, width, height, cacheFile, object : ResizeImageTask.Listener {
```
charlag: Of course it's not perfect that it captures activity but it shouldn't be too slow so I'm not sure if it's worth doing something with it.

app/src/main/java/com/keylesspalace/tusky/EditProfileActivity.kt
```diff
@@ -0,0 +1,491 @@
...
+
+        internal interface Listener {
+            fun onSuccess(resizedImage: Bitmap?)
+            fun onFailure()
+        }
+    }
+
+}
+
+fun saveBitmapToFile(bitmap: Bitmap, file: File): Boolean {
```
charlag: It's kinda weird that it's outside the class but inside the file.

app/src/main/java/com/keylesspalace/tusky/EditProfileActivity.kt
connyduck: If everything is null, nothing has changed and there is no need to make a network request :)
Will add a comment.

app/src/main/java/com/keylesspalace/tusky/EditProfileActivity.kt
connyduck: I am pretty sure that this AsyncTask does not have a reference to the activity. Thats the reason why the function below is outside the class, the function is not accessible if it is a member of the class (it doesnt compile).

PR #526, opened 2018-02-12
convert entity classes to Kotlin data classes

connyduck:
This is the foundation for the things I want to implement for Tusky 1.6. Probably will convert some more classes to Kotlin.

I used Kotlin `@Parcelize`, seems pretty stable but requires `experimental` flag in `build.gradle`. Also I am not yet satiesfied with the companion objects. Any ideas on how to avoid these are welcome.


app/src/main/java/com/keylesspalace/tusky/entity/Account.kt
```diff
@@ -0,0 +1,67 @@
...
+import android.text.Spanned
+
+import com.google.gson.annotations.SerializedName
+import com.keylesspalace.tusky.util.HtmlUtils
+import kotlinx.android.parcel.Parceler
+import kotlinx.android.parcel.Parcelize
+import kotlinx.android.parcel.WriteWith
+
+@Parcelize
+class Account (val id: String,
```
charlag: Why not data class? You could override `equals()` for it as well

app/src/main/java/com/keylesspalace/tusky/entity/Card.kt
```diff
@@ -13,21 +13,30 @@
...
  * see <http://www.gnu.org/licenses>. */
 
-package com.keylesspalace.tusky.entity;
+package com.keylesspalace.tusky.entity
+
+import android.os.Parcelable
+import kotlinx.android.parcel.Parcelize
+
+@Parcelize
+data class Card (val url: String,
```
charlag: Fields formatting? See 'Class header formatting'
https://kotlinlang.org/docs/reference/coding-conventions.html
"Classes with longer headers should be formatted so that each primary constructor parameter is in a separate line with indentation. Also, the closing parenthesis should be on a new line. If we use inheritance, then the superclass constructor call or list of implemented interfaces should be located on the same line as the parenthesis:"

app/src/main/java/com/keylesspalace/tusky/viewdata/StatusViewData.java
```diff
@@ -58,8 +58,8 @@ private StatusViewData() {
         private final String nickname;
         private final String avatar;
         private final Date createdAt;
-        private final String reblogsCount;
```
charlag: I am not sure about this one. For me ViewData is about minimum representation required to render a thing. I think it's views responsibility to decide how it wants to represent it. Is there a reason for this change?

app/src/main/java/com/keylesspalace/tusky/entity/Account.kt
connyduck: oh right, missed that

app/src/main/java/com/keylesspalace/tusky/viewdata/StatusViewData.java
connyduck: I don't understand why numbers where represented as String, does not make sense to me. 
Yes, I added number formatting in the views, looks much nicer now.

charlag: Oh, crap, sorry, it got reversed in my head, you've done it right.

connyduck:
I am also not happy with the three `var`s inside the status class. They should be `val`s, really

charlag:
I think it's possible now with `.copy()`

PR #528, opened 2018-02-13
Add Chinese Translations

aka-rin:
Although I tried by best to translate the strings manually to keep the readability, these translations still need to be improved, since I cannot test the translations on my phone and some terms in TW/HK may be not authentic...

It would be convenient for localization if Tusky has a functionality to load an external translation file to preview the result and switch between different languages. As far as I know, Telegram Android have this functionality. 😂 

Well it may be off the topic, but I did feel unnatural when I search in Tusky every time. It would be nice if Tusky could add a shortcut to [mastodonsearch](http://mastodonsearch.jp) in search menu... 😭

### DIFF
--- a/app/src/main/res/values-zh-rCN/strings.xml
+++ b/app/src/main/res/values-zh-rCN/strings.xml
@@ -1,6 +1,6 @@
 <resources>
 
-    <string name="error_generic">程序出现异常。</string>
+    <string name="error_generic">应用程序出现异常。</string>
     <string name="error_empty">内容不能为空。</string>
     <string name="error_invalid_domain">该域名无效</string>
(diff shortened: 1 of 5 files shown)


connyduck:
Wow thank you very much!
You can download the apk file created by bitrise from this branch here: https://www.bitrise.io/artifact/4396800/p/599b04b70413c00c57ac7c31cc677300
Try it out and then feel free to make additional changes!

aka-rin:
Thanks. I have tested it on my phone and everything seems OK. So maybe this is the last version of this pull request. Have a nice day too and happy Chinese new year! 😂

connyduck:
Thank you!
Yes, I will update the homepage with the next release!

PR #537, opened 2018-03-02
Translates: FR: Do not use plural for follows_you

nailyk-fr:
Probably another string to change. 
Gonna update the PR later.


nailyk-fr:
ohhhhh auto build on PR. Really nice!

PR #539, opened 2018-03-05
Compose close dialog + focus fix

charlag:
Here are two smol fixes for the 'compose' screen.
* If you press the close or back button it checks if something is not empty and asks you in a dialog if you want to really close everything
* Focusing on the text field by default - because if you reply to a toot with a CW it focuses on a CW field by default.

I've also refactored one dialog listener and formatted things a little.


connyduck:
Oh thank you very much. There are some more problems and I plan to redo a big part of the compose activity, but this is a nice start.

charlag:
Thank you for keeping it afloat!
I would redo it with Redukt/Knot but I think that at least pulling some state and logic out of the activity would be cool. I would let us write proper tests.

PR #540, opened 2018-03-06
Use the entire content warning text + button area as the hitbox for the "Show more/less" toggle

Tak:
Rationale: I'm always missing the actual button, sometimes getting three or four taps deep into the post view before actually hitting it.
The post view is still accessible by tapping on the expanded content (or anywhere in the post that's outside the CW header area).

This is my first Tusky PR; be gentle. :smiley:


charlag:
Hi! Welcome to the Tusky development.
Thank you for your PR.
I am quite clumsy as well but I didn't have problems with this button but I can easily imagine that people do.
Will there be enough space to open the toot itself? If yes, I am all for this.
Let's see what @connyduck will say about this.

connyduck:
I see the problem, but there might be better solutions. What about giving all toots & the content warning button a background that changes when touched so you better see where your finger is and can correct the position?

connyduck:
My concern right now is that if we make the entire content warning clickable people may complain about accidentally clicking it. I will later try to change it visually so it is more obvious what is clickable and what not.

connyduck:
The best solution to this problem is how Pawoo does it: Bigger button & on its own line. img_20180311_080930
Can you implement this @Tak? I can help you if you want.

charlag:
Maybe just a bigger button (on both axes) would be enough?

connyduck:
> Maybe just a bigger button (on both axes) would be enough?
Yes but I think a Button on its own line looks better

Tak:
Sure, I can do that.
I considered it initially, but I thought that changing the button dimensions and/or layout would be a more disruptive ui change.

While it is possible that making the whole description area activatable would lead to some accidental expansions, I would think that there would be very few cases where people would tap anywhere on the post if they didn't want to see the content. (i.e., there's not much point in going to the post view of a collapsed cw post.)

app/src/main/res/layout/item_status.xml
```diff
@@ -108,23 +108,42 @@
...
+                <ToggleButton
+                    android:id="@+id/status_content_warning_button"
+                    android:layout_width="wrap_content"
+                    android:layout_height="wrap_content"
+                    android:layout_centerHorizontal="true"
+                    android:background="?attr/content_warning_button"
+                    android:minHeight="0dp"
+                    android:minWidth="0dp"
+                    android:paddingBottom="6dp"
+                    android:paddingLeft="10dp"
```
connyduck: I would give the button a bit more left-right padding

app/src/main/res/layout/item_status.xml
```diff
@@ -108,23 +108,42 @@
...
+                    android:paddingRight="10dp"
+                    android:paddingTop="6dp"
+                    android:textAllCaps="true"
+                    android:textOff="@string/status_content_warning_show_more"
+                    android:textOn="@string/status_content_warning_show_less"
+                    android:textSize="?attr/status_text_large" />
```
connyduck: `?attr/status_text_medium` should be enough? Maybe add some more padding in exchange?

app/src/main/res/layout/item_status.xml
```diff
@@ -108,23 +108,42 @@
             android:textColor="?android:textColorPrimary"
             android:textSize="?attr/status_text_medium" />
 
-        <ToggleButton
-            android:id="@+id/status_content_warning_button"
+        <RelativeLayout
```
connyduck: It is now possible to ged rid of the `FlowLayout`, and the two `RelativeLayout`s are unnecessary as well. Please align the button within the toplevel `RelativeLayout`

Tak: I'll try removing the `FlowLayout`.
The extra `RelativeLayout` is there because the centering didn't work correctly unless I nested them like this.

app/src/main/res/layout/item_status.xml
```diff
@@ -87,71 +87,43 @@
...
-                    android:textOn="@string/status_content_warning_show_less"
-                    android:textSize="?attr/status_text_large" />
-
-            </RelativeLayout>
-        </RelativeLayout>
-    </com.keylesspalace.tusky.view.FlowLayout>
+            android:paddingLeft="16dp"
+            android:paddingRight="16dp"
+            android:paddingTop="6dp"
+            android:layout_marginVertical="4dp"
```
connyduck: `layout_marginVertical´ is only used on Api level 26+, please change it to top&bottom

app/src/main/res/layout/item_status.xml
```diff
@@ -87,71 +87,43 @@
...
             android:layout_height="wrap_content"
             android:layout_below="@+id/status_content_warning_description"
-            android:layout_marginBottom="4dp"
-            android:layout_toEndOf="@+id/status_avatar"
-            android:focusable="true"
-            android:paddingTop="6dp"
+            android:layout_centerHorizontal="true"
+            android:background="?attr/content_warning_button"
+            android:minHeight="0dp"
+            android:minWidth="0dp"
```
connyduck: I think a `minWidth` of about 150dp would help so the button stays the same size when the text changes (at least for most languages).

charlag:
Every time layout gets simplier I drop a tear of joy

app/src/main/res/layout/item_status.xml
Tak: This is just moved from the old version - do you still want me to change it?

app/src/main/res/layout/item_status.xml
Tak: Oops, thanks android studio!

app/src/main/res/layout/item_status.xml
connyduck: yes please

charlag:
Is it cool now or something else is needed?

connyduck:
> I think a minWidth of about 150dp would help so the button stays the same size when the text changes (at least for most languages).

Tak:
Oops, missed the notification for the response on that :neutral_face:

connyduck:
Looks good, thank you!

PR #542, opened 2018-03-06
Add tests

charlag:
Not sure if WIP or not, it's too late and I barely think.
We plan to refactor ComposeActivity, right? But we need tests to refactor. Instrumental tests is... not the most pleasant thing. So we need Robolectric.
Problems:
* Seems like it doesn't support some AppCompat things?
* AccountManager was hardcoded through the application.
Yes, I could use PowerMock to mock static methods but I didn't want to. It's better to get rid of such things.
I also couldn't just mock whole app because it's needed for the activity.

Anyway, this is a framework upon which we should build or tests to not break things all the time.


connyduck:
lgtm, I can work with this. I think we should introduce Dependency Injection to solve the problem with the account manager & with the Api/okhttp being created for every activity.

charlag:
Cool, thanks.
We totally should but it would a bigger PR.

PR #548, opened 2018-03-09
ComposeActivity improvements

connyduck:
Major changes incoming. I may even convert the activity to Kotlin.

Step one: Remove the media urls from status text to comply with Mastodon 2.3.


charlag:
Uh-oh, I would add some tests first but yeah.
Will we pull the logic into something which can survive config change (and which is easier to test)? We can do it with bare `onRetainCustomNonConfigurationInstance()` like in `ListsActivity` or we could use some Architecture Components for this.

Anyway, what are your plans on this now?

connyduck:
last commit solves the problem that on small screens editing of toots with content warnings and attached media is really difficult

my plans are mostly ui/ux related, but will add some tests as well

- have the keyboard stay open when toot privacy is changed or media is attached 
- make it more obvious that the "replying to" text can be expanded
- improve the toot privacy settings, probably move to bottom sheet
- remove the save toot button (need space for the emoji button), move saving of toots to the close dialog
- fix bugs with saved toots, mentions do not restore correctly, some crashes with attached media
- add emoji keyboard
- clean up the class, I hate files with > 1000 lines

charlag:
Cool, thanks! If it's mostly UI/UX than it's not too much pressure to add tests I guess? :D
I'm not quite sure what do you mean by toot privacy settings. We don't implement something?
I am not sure how we can clean it up without pulling the logic out (and for *this* we need tests.
@ me if you need help or anything! Thanks for explaining and for doing all this.

app/src/main/java/com/keylesspalace/tusky/fragment/ComposeOptionsView.kt
```diff
@@ -0,0 +1,72 @@
...
+import android.content.Context
+import android.util.AttributeSet
+import android.view.LayoutInflater
+import android.widget.LinearLayout
+import com.keylesspalace.tusky.R
+import com.keylesspalace.tusky.entity.Status
+import kotlinx.android.synthetic.main.view_compose_options.view.*
+
+
+class ComposeOptionsView @JvmOverloads constructor(context: Context, attrs: AttributeSet? = null, defStyleAttr: Int = 0) : LinearLayout(context, attrs, defStyleAttr) {
```
charlag: yay, Kotlin!
I would format constructor in multiple lines tbh

connyduck: yes

connyduck:
this currently crashes on Api < 21 and if all buttons are shown on a small device the toot button is pushed to the right.
I am quite pleased with the design otherwise.
Will add another BottomSheet for the attachment selection.

connyduck:
Note to self: There is also a crash when using the camera on old apis that needs investigation.

connyduck:
small style toot button for small displays

charlag:
I think it's such a huge PR, you should consider breaking them in parts for the future (I suck at writing/speaking, I hope it sounds okay)

nailyk-fr:
Out of topic: If you want I can add Tusky on my gerrit.A bit overkill for a 'single' project but have nice features like commits topic, or rework other people commits. Useful for important changes like this. 
Out of topic2: Any chance you have some matrix or IRC dev chan I can join to follow your work please ?

charlag:
Yeah, it doesn't belong here but...
Oh, I'm not sure about gerrit personally, I've been pretty happy with Github review system so far.
But chat room would a nice idea if we have want people who want to contribute to get help easily. Usually we discuss it on Masto (look for tusky and tuskydev tags). I would vote for Matrix or at least Gitter.

connyduck:
ToDos:
- cache for the emoji keyboard
- cleanup
- testing
- (maybe more improvements for small displays are needed)

> I think it's such a huge PR, you should consider breaking them in parts for the future
yes you are right, sorry about that

app/src/main/java/com/keylesspalace/tusky/ComposeActivity.java
```diff
@@ -913,49 +797,17 @@ private void sendStatus(String content, Status.Visibility visibility, boolean se
...
-
-            @Override
-            public void onFailure(@NonNull Call<Status> call, @NonNull Throwable t) {
-                onSendFailure(null);
-            }
-        };
-        mastodonApi.createStatus(content, inReplyToId, spoilerText, visibility.serverString(),
-                sensitive, mediaIds).enqueue(callback);
-    }
+        Intent sendIntent = SendTootService.sendTootIntent(this, content, spoilerText, inReplyToId,
```
charlag: If you're unhappy about too many parameters, maybe we should construct something on this side and pass it?

app/src/main/java/com/keylesspalace/tusky/ComposeActivity.java
```diff
@@ -1603,17 +1425,21 @@ private void handleCloseButton() {
...
-        ReadyStage readyStage;
-        byte[] content;
-        long mediaSize;
-        String description;
+    @Override
+    public void onEmojiSelected(@NotNull String shortcode) {
+        textEditor.getText().insert(textEditor.getSelectionStart(), ":"+shortcode+": ");
+    }
+
+    public static final class QueuedMedia {
```
charlag: If it's not exclusive to this activity anymore we could move it outside

app/src/main/java/com/keylesspalace/tusky/service/SendTootService.kt
```diff
@@ -0,0 +1,245 @@
...
+
+        if(account == null) {
+            tootsToSend.remove(tootId)
+            return
+        }
+
+        tootToSend.retries++
+
+        mastodonApi.createStatus(
+                "Bearer " + account.accessToken,
```
charlag: Why do you pass token here directly when you could leave it to interceptor?

app/src/main/java/com/keylesspalace/tusky/service/SendTootService.kt
```diff
@@ -0,0 +1,245 @@
...
+                    if(tootsToSend.isEmpty()) {
+                        ServiceCompat.stopForeground(this@SendTootService, ServiceCompat.STOP_FOREGROUND_DETACH)
+                        stopSelf()
+                    }
+
+                }
+
+            }
+
+            override fun onFailure(call: Call<Status>, t: Throwable) {
```
charlag: What a nice task for RxJava would it be ;)

app/src/main/java/com/keylesspalace/tusky/service/SendTootService.kt
```diff
@@ -0,0 +1,245 @@
...
+
+            return intent
+        }
+
+
+    }
+}
+
+@Parcelize
+data class TootToSend(val text: String,
```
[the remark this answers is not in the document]
charlag: Also I think it's fine that it has so many fields. It's plain data, it's not dependencies, so it's fine? We could group some fields but it wouldn't make it clearer.

app/src/main/java/com/keylesspalace/tusky/util/SaveTootHelper.java
```diff
@@ -0,0 +1,188 @@
...
+                             @NonNull Status.Visibility statusVisibility) {
+
+        if (TextUtils.isEmpty(content) && mediaQueued.isEmpty()) {
+            return false;
+        }
+
+        // Get any existing file's URIs.
+        ArrayList<String> existingUris = null;
+        if (!TextUtils.isEmpty(savedJsonUrls)) {
+            existingUris = new Gson().fromJson(savedJsonUrls,
```
charlag: ouch, let's pass gson here maybe?

app/src/main/java/com/keylesspalace/tusky/util/SaveTootHelper.java
```diff
@@ -0,0 +1,188 @@
...
+import com.keylesspalace.tusky.entity.Status;
+
+import java.io.File;
+import java.text.SimpleDateFormat;
+import java.util.ArrayList;
+import java.util.Date;
+import java.util.List;
+import java.util.Locale;
+
+public class SaveTootHelper {
```
charlag: Final class, final fields, tag can be private ^^

app/src/main/java/com/keylesspalace/tusky/util/SaveTootHelper.java
```diff
@@ -0,0 +1,188 @@
...
+             * can be deleted. */
+            deleteMedia(existingUris);
+        }
+        final TootEntity toot = new TootEntity(savedTootUid, content, mediaUrlsSerialized, contentWarning,
+                inReplyToId,
+                replyingStatusContent,
+                replyingStatusAuthorUsername,
+                statusVisibility);
+
+        new AsyncTask<Void, Void, Void>() {
```
charlag: I don't use AsyncTask much so I'm not sure if we have to execute them in a special way to make them parallel...
To silence the warning we should make a static class and pass DAO to it, right?

app/src/main/java/com/keylesspalace/tusky/util/SaveTootHelper.java
```diff
@@ -0,0 +1,188 @@
...
+            }
+        }.execute();
+        return true;
+    }
+
+    @Nullable
+    private List<String> saveMedia(@NonNull List<ComposeActivity.QueuedMedia> mediaQueued,
+                                   @Nullable List<String> existingUris) {
+
+        File imageDirectory = context.getExternalFilesDir(Environment.DIRECTORY_PICTURES);
```
charlag: Does user expect that we're going to save their pictures from the toot to the public directory where any app can access them?..

app/src/main/java/com/keylesspalace/tusky/util/SaveTootHelper.java
```diff
@@ -0,0 +1,188 @@
...
+    }
+
+    /**
+     * A∖B={x∈A|x∉B}
+     *
+     * @return all elements of set A that are not in set B.
+     */
+    private static List<String> setDifference(List<String> a, List<String> b) {
+        List<String> c = new ArrayList<>();
+        for (String s : a) {
```
charlag: This is some expensive stuff, like O(N^2)... it's probably doesn't matter here and I'm not sure that sorting lists in prior is cheaper

app/src/main/java/com/keylesspalace/tusky/service/SendTootService.kt
```diff
@@ -36,13 +38,15 @@ class SendTootService: Service(), Injectable {
...
-    private var notificationId = -1 // use negative ids to not clash with other notis
+    private lateinit var saveTootHelper: SaveTootHelper
+
     private val tootsToSend = mutableMapOf<Int, TootToSend>()
     private val timer = Timer()
 
 
     override fun onCreate() {
         AndroidInjection.inject(this)
+        saveTootHelper = SaveTootHelper(TuskyApplication.getDB().tootDao(), this)
```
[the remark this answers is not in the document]
connyduck: yes

app/src/main/java/com/keylesspalace/tusky/ComposeActivity.java
connyduck: good idea

app/src/main/java/com/keylesspalace/tusky/service/SendTootService.kt
connyduck: because otherwise the account can be switched while the status is in the retry-loop and the status will be sent with an account different from the one the user expects

app/src/main/java/com/keylesspalace/tusky/util/SaveTootHelper.java
connyduck: Thats no new code, its cut from `ComposeActivity`. And yes, its  probably a bad idea, I should change it.

app/src/main/java/com/keylesspalace/tusky/util/SaveTootHelper.java
connyduck: Thats no new code, its cut from `ComposeActivity`. And it is only called on attachment lists where n is at most 4.

app/src/main/java/com/keylesspalace/tusky/service/SendTootService.kt
charlag: Oh crap, it's probably my design mistake, we should have passed something like account id as the header, not the domain...
But then it wouldn't work for places like auth...
uuuuugh
I don't really like the idea of passing both everywhere but it seems it's the only flexible way?
Or we could teach interceptor about both domain and account id. The former would have higher priority and they probably won't be used in one request anyway.

charlag:
I've tested it and it added a "sensitive media" by default, is it intentional?

It would also be cool to show toast with button description on long press (like vanilla Android menus), I myself was confused by what a button does.

Another thing: I've turned on an airplane mode and sent a toot and it looks like it just sending it.
I am not sure that it is worth doing in this release but in the future we should probably tell user that it fails and give ability to restart it manually.

connyduck:
> I've tested it and it added a "sensitive media" by default, is it intentional?

the only thing that should be added automatically is sensitive media warning on toots that have a content warning (and media attached of course).

> It would also be cool to show toast with button description on long press (like vanilla Android menus), I myself was confused by what a button does.

yes I made those but only for Android 8+ https://developer.android.com/guide/topics/ui/tooltips.html

> Another thing: I've turned on an airplane mode and sent a toot and it looks like it just sending it.
I am not sure that it is worth doing in this release but in the future we should probably tell user that it fails and give ability to restart it manually.

yes it keeps retrying until status is sent

charlag:
Yea, but there were no CW... I guess so at least, I am not sure now. Don't bother ;)

Would you consider falling back to toasts on older versions? I could do it.

Maybe we should limit the number of retries? If user will send toot and then it will fail for some minutes we can drain battery really fast. It retries every second if I unserstand.

connyduck:
> Would you consider falling back to toasts on older versions? I could do it.

Yes please do if you want 

> Maybe we should limit the number of retries? If user will send toot and then it will fail for some minutes we can drain battery really fast. It retries every second if I unserstand.

No its backing off exponentially to max 1min interval, see line 190 in `SendTootService`. imho this should be fine but I will consider adding a maximum number of retries.

connyduck:
ok I think i fixed the notifications on older Androids. Found another problem with sending toots under low memory conditions, the `SendTootService` is crashing when recreating itself

PR #549, opened 2018-03-11
l10n: update Japanese translation

musashino205:
Update japanese translations for some new functions.

Signed-off-by: INAGAKI Hiroshi <musashino.open@gmail.com>


connyduck:
Thank you very much!

PR #554, opened 2018-03-24
Add Dagger

charlag:
So, this should've been a simple task but I ended up changing much more than I wanted. This PR adds Dagger Dependency Injection with Android plugins which should significantly improve architecture and testability in the future. For now it's mostly used for the networking stuff because it's the most obvious thing but it should also be used for the DB/Account manger and probably and more. Adding new modules as well as new screens should be trivial.

What we gain from this:
* Shared Gson caches, OkHttp pools/caches, Retrofit instances which should improve performance
* Testing things much simpler when they don't look for dependencies through static fields
* Better architecture - less coupling, less responsibilities

What we've lost:
* I deliberately omitted dispatcher canceling. Fragments which want canceling already add calls to the list and then cancel so I felt that it was redundant (in most other places this behavior is counter-intuitive).

Other changes:
* We don't create many Retrofit instances (which is not cheap because of the reflection) so we change domain URL dynamically. This has a nice side effects: it is not easy to implement "boost as" or "follow as" actions, because we can add header to the request which will tell `AuthInterceptor` to use specified account. Interceptor can strip this header afterwards (this technique was described by Jack Wharton - one of the OkHttp/Retrofit creators)
* I couldn't get rid of `ServiceLocator` ugliness because it is used in the `BaseActivity` and we cannot inject there (and I don't want to inject it in _every_ activity). I'm not sure why this check is there, maybe we don't really need it, maybe we should find some other way.
* We also cannot inject in `SFragment` really to I used this as a chance to move some of the functionality which doesn't belong to Fragments anyway to the separate class. It is not super clear abstraction but it should work, I think we could move even more things from `SFragment`. Subclassing should take place when there's a shared behavior, not when there's a shared code and you don't know where to put it. I still had to inject new things in every fragment but I think it's still a step in the right direction.
* I've also added missing licenses

I really don't like that we have many activities. Because of this this PR is much bigger.
While it seems redundant to add API to every activity/fragment manually it uncovers hidden dependencies. If dependency is hidden behind the parent it doesn't mean it's not there. Injecting != using constructor but it has similar effect.


charlag:
P.S.: I think I've checked all screens but I made some changes afterwards and I may have missed something

connyduck:
omneg this is wonderful <3
will have a look at it later or tomorrow

connyduck:
Can you check if we need to add anything to the proguard config? I am 99% sure DI needs some special proguard rules. And please do a production build and test it.

charlag:
Well, not really, at least not in my experience, because there's no reflection - everything is compiled. Their [website](https://google.github.io/dagger/android) doesn't really say anything either (more, they _rely_ on ProGuard to not blow up the method count) but I will check tomorrow, not a problem!

app/src/main/java/com/keylesspalace/tusky/network/AuthInterceptor.java
```diff
@@ -17,7 +34,6 @@
  */
 
 public final class AuthInterceptor implements Interceptor {
-
     AccountManager accountManager;
```
connyduck: can be private

app/src/main/java/com/keylesspalace/tusky/network/AuthInterceptor.java
```diff
@@ -1,10 +1,27 @@
...
+
+
 package com.keylesspalace.tusky.network;
 
 import android.support.annotation.NonNull;
 
 import com.keylesspalace.tusky.TuskyApplication;
 import com.keylesspalace.tusky.db.AccountEntity;
 import com.keylesspalace.tusky.db.AccountManager;
+import com.keylesspalace.tusky.db.AccountManager;
```
connyduck: unused imports

app/src/main/java/com/keylesspalace/tusky/network/AuthInterceptor.java
```diff
@@ -17,7 +34,6 @@
  */
 
 public final class AuthInterceptor implements Interceptor {
-
     AccountManager accountManager;
 
     public AuthInterceptor(AccountManager accountManager) {
```
connyduck: please give this class another name so it is clear it handles authentication and the switching of instances. I was very confused. Something like `InstanceSwitchAuthInterceptor` maybe?

app/src/main/java/com/keylesspalace/tusky/network/AuthInterceptor.java
```diff
@@ -31,8 +47,16 @@ public Response intercept(@NonNull Chain chain) throws IOException {
         AccountEntity currentAccount = accountManager.getActiveAccount();
 
         Request.Builder builder = originalRequest.newBuilder();
+        // In the future we could add a phantom header parameter to some requests which would
+        // signalise that we should override current account (could be useful for "boost as.."
+        // actions and the like
         if (currentAccount != null) {
-            builder.header("Authorization", String.format("Bearer %s", currentAccount.getAccessToken()));
+            // I'm not sure it's enough the hostname but should be good
```
connyduck: yes, is enough (you are missing a verb in that comment but you can remove it if you want)

app/src/main/java/com/keylesspalace/tusky/network/TimelineCases.kt
```diff
@@ -0,0 +1,100 @@
...
+            mastodonApi.unreblogStatus(id)
+        }
+        call.enqueue(callback)
+    }
+
+    override fun favouriteWithCallback(status: Status, favourite: Boolean, callback: Callback<Status>) {
+        val id = status.actionableId
+
+        val call: Call<Status>
+        call = if (favourite) {
```
connyduck: inline `val call =`

app/src/main/java/com/keylesspalace/tusky/MainActivity.java
```diff
@@ -82,6 +91,11 @@
     private static final long DRAWER_ITEM_SAVED_TOOT = 9;
     private static final long DRAWER_ITEM_LISTS = 10;
 
+    @Inject
+    public MastodonApi mastodonApi;
+    @Inject
+    public DispatchingAndroidInjector<Fragment> fragmentInjector;
+
```
connyduck: the comment in line 111 is no longer true (github only lets me comment changed lines 🙄)

app/src/main/java/com/keylesspalace/tusky/BaseActivity.java
```diff
@@ -123,44 +95,13 @@ protected SharedPreferences getPrivatePreferences() {
...
-                .addConverterFactory(GsonConverterFactory.create(gson))
-                .build();
-
-        mastodonApi = retrofit.create(MastodonApi.class);
-    }
-
     protected boolean redirectIfNotLoggedIn() {
-        if (accountManager.getActiveAccount() == null) {
+        // This is very ugly but we cannot inject into parent class and injecting into every
+        // subclass seems inconvenient as well. Do we really need this check in every activity?
```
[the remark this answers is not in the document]
charlag: But why? How can user be on some random screen without auth? It's either splash, login or settings (after which we redirect them to login I think)

app/src/main/java/com/keylesspalace/tusky/TuskyApplication.java
```diff
@@ -100,12 +113,31 @@ protected void initPicasso() {
         } catch (IllegalStateException e) {
             throw new RuntimeException(e);
         }
+
```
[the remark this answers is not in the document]
charlag: Yeah, sorry, bad merge

charlag:
Thanks for the review!
It should be better now. I've built it for release and it worked but I would test all the things additionally. I honestly didn't check notifications with different users because it's 00:51, lol.
I will be absent for the next 1.5 weeks so if you find something feel free to change it (or to wait till I return).

charlag:
Oh, crap, forgot to strip the header, not it's cool

charlag:
I've been using this build for a day and everything was great. So, did I miss something else? You wanna test it on your own probably?

app/src/main/java/com/keylesspalace/tusky/BaseActivity.java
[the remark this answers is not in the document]
connyduck: even the about screen makes network requests, without login we would not know to which instance to connect & the settings use the `AccountManager`

connyduck:
I made some small changes & tested it and it is really awesome, thank you so much

charlag:
Thank you for doing all of this!

PR #559, opened 2018-04-01
Compose: Make 'close unsaved toot' translatable 

nailyk-fr:
* Make 'Close unsaved toot' confirmation message translatable, 
* Translate it to French


connyduck:
Sry, this clashes with #548. The dialog should ask if the toot should be saved before closing.

Issue #471, opened 2017-11-29
Deleted toot does not get removed from timeline


connyduck:
Steps to reproduce:
- open detail view of own toot from timeline
- delete toot
- deleted toot is still visible in timeline

Issue #552, opened 2018-03-15
Thread/conversation not updating after replying


mbirth:
After I've sent a reply, the thread doesn't update to show my reply. I have to refresh manually to make it show up.

* * * *
- Tusky Version: 1.5.2
- Android Version: 7.1.1
- Android Device: BlackBerry KEYone
- Mastodon instance (if applicable): mastodon.social

- [X] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

charlag:
Hi,thanks for reporting!
You're right, it's not there yet, I want it myself.

charlag:
It does update as for 1.7.0 but it refreshes whole thread, #562 will improve things

PR #562, opened 2018-04-07
[WIP, Experimental] Add AppStore, add fav, reblog events

charlag:
This is an experiment on better handling of events like fav, boost, reply.
Basically we need something like EventBus but 
- it is limited
- I'm tired to see this verbose error handling across the project (onResponse ... if response.body ... onFailure ...) 
- I also consider LocalBroadcastManager a huge hack added for the sake of convenience 

So my solution is RxJava.
How is this better than broadcasts? Well, imagine we want to update a thread after posting a reply (#552). We would have to serialize/deserialize status somehow or make extra network request. But we can pass anything through RxJava so we don't have such a problem/overhead.

I didn't touch any other parts (like replacing all `Call`s with `Single`s or getting rid of broadcasts) just *yet* but that's what I would like to do in some point in the future if everyone agrees.

I've also converted `TimelineFragment` to Kotlin because at one point it just became unreadable. Sorry about that.

This is not complete in any sense (doesn't handle boosted things e.g.), just wanted to hear your opinion on fixing some things with RxJava.

(to make it clear: in `TimelineFragment` I just added a couple of methods on the bottom, not much else)


connyduck:
I like your initative on this <3 We definitely need something like this and might as well do it with RxJava.

Regarding your current code: when liking/boosting, `notifyItemChanged(position)` makes the item "flimmer" (you can see its updated), I really don't want this. (maybe its possible to tell RecyclerView to not play animation, but only updating the button is definitely preferable)

Please use Kotlin Android Extensions instead of `findViewById`

And can you merge master back in please?

I would really like it if you continue to work on this!

charlag:
Thanks for review! 🙇 
Yes, I agree, we should be able to disable animation in this case!
Indeed
I will probably work on this in May only, but we'll see! I will definitely rebase it.
Thanks a lot!

app/src/main/java/com/keylesspalace/tusky/fragment/NotificationsFragment.java
```diff
@@ -193,9 +199,48 @@ public View onCreateView(@NonNull LayoutInflater inflater, @Nullable ViewGroup c
...
+        appStore.getEvents()
+                .observeOn(AndroidSchedulers.mainThread())
+                .as(autoDisposable(from(this, Lifecycle.Event.ON_DESTROY)))
+                .subscribe(event -> {
+                    if (event instanceof FavoriteEvent) {
+                        handleFavEvent((FavoriteEvent) event);
+                    } else if (event instanceof ReblogEvent) {
+                        handleReblogEvent((ReblogEvent) event);
+                    }
+                    if (event instanceof BlockEvent) {
```
charlag: oops, missing `else` here

charlag:
Something is still wrong with boosted toots, needs more love, sigh

charlag:
Updating threads after reply is buggy too.

charlag:
I didn't find the bug with boosts bug I've fixed things with inserting new toots in threads and only in specific timelines. I'm not sure if we should insert "load more" in profiles because it doesn't seem to work there. Otherwise it should be good.

I've spent half a day trying to make Robolectric work but I've stumbled upon couple of [bugs](https://github.com/robolectric/robolectric/issues/3857) so I give up. It doesn't worth trying to make it work until we refactor logic into separate entity.
I am not sure if it's worth making things public and test them this way at least because everything is stateful.

charlag:
It crashed for me when the app did unload and then i tooted

charlag:
Will fix #471 #280

app/src/main/java/com/keylesspalace/tusky/ComposeActivity.java
```diff
@@ -169,6 +170,8 @@
     public MastodonApi mastodonApi;
     @Inject
     public AccountManager accountManager;
+    @Inject
```
connyduck: yes, more injecting 😎

app/src/main/java/com/keylesspalace/tusky/AccountActivity.java
```diff
@@ -91,6 +94,8 @@
     public AccountManager accountManager;
     @Inject
     public DispatchingAndroidInjector<Fragment> dispatchingAndroidInjector;
+    @Inject
+    public AppStore appstore;
```
connyduck: Is there a better name for this? It just reminds me so much of fdroid/google play 🙈

charlag: It's a Redux/Flux terminology to name such a thing Store. Can rename to GlobalStore or something

charlag:
About load more: the way we do it now will make timeline move on fetch result and it's very annoying to lose your postiion in timeline when you reply to someone. Do you have any idea how to fix this?

connyduck:
I think in `TimelineAdapter.update` we should not call `notifyDataSetChanged` but `notifyItemRangeInserted`, that should make the jumping go away?

charlag:
But we have to calculate what we inserted then because we may not have loaded anything?
I propose usage of `hasStableIds` until we migrate to `DiffUtils`. Need to give placesholders IDs tho.

charlag:
Anyway, @connyduck , if you insist we should actually refresh timeline on every post, I can try to make it work. It's kinda wasteful IMO but I see UX value in this.

connyduck:
I do not think its wasteful, the UX improvement is huge.

charlag:
I've made some progress on it.
I ended up re-doing part of timelines stuff with DiffUtil. I also think that when I'll finish with "load more" progress it will be better than now.

charlag:
I've stumbled upon some super weird bug, adding spinner as the last item crashes app with inconsistency exception. Will look at it tomorrow

charlag:
I've figured out the crash, it was a race condition because I used AsyncListDiffer incorrectly.
Now timelines work mostly correctly (I remove "load more" spinner in the wrong place). The real problem I have is that it gets into infinite loop on short lists because I changed the way InfiniteScrollListener (or whatever it's called) was only firing callbacks when amount of items increased. It caused unpleasant things (like if one timeline fetch failed we wouldn't load the rest). Now I fire them every time but I have to track when we stumbled upon bottom. Here is some problem, it doesn't work.

charlag:
I think that's it. It's a much bigger PR than I wanted but IMO it feels much better.
I'm mostly concerned about criteria for stopping loading things on the bottom but I didn't find a better one. It seems to work just fine.
Please check @connyduck and everyone else  :heart:

connyduck:
Works like a charm and code is way cleaner - really awesome work

One thing that personally bugs me with timelines is, when pulling down to refresh, it always jumps to the top instead of just scrolling down a little bit, enough to see that something new has loaded. The Twitter app does this really good. But we should probably handle this in a new pull request?

charlag:
Cool, thanks for the review
As discussed in a dedicated issue it is tricky to pick one option. I would love to handle this without any preferences and knobs. Probably keeping position would be the best but yeah, I think it's worth another PR, I tried to match current behavior.  Keeping position is trivial, we just need to delete a couple of lines.

connyduck:
ok you can merge this PR :)

PR #568, opened 2018-04-10
View links to statuses inside Tusky

Tak:
When opening a link, check whether it's a link to a Mastodon status, and if so, open it in Tusky's thread view instead of the external browser.


charlag:
Hi! I understand your desire but... web UI doesn't do it, right? They probably have a reason. And the reason is that we don't know if a link leads to a status or not. We can ask server to resolve the link but we cannot stop everything until it happens, right?

Tak:
The web ui doesn't do it, but it's not so jarring to have a browser tab created as it is to be shoved into a completely different app. Additionally, having the thread opened in Tusky allows you to interact with the content (favorite/follow/block) like normal.

The link resolution is completely asynchronous, if that's the concern.

charlag:
First of all, my apologies: I somehow missed that this is a PR and not just an issue.
I am yet to try it out but my concern is the same: we make user wait for the server search before opening a link. We're sacrificing experience of the most common case for the uncommon.
I believe that if this thing is to be done it should be done on the server side or prior to interacting with a link

Tak:
I don't feel that the current implementation detracts from the user experience. However, we could put it behind a preference if you think it would be contentious...

charlag:
I would ask for the @connyduck input here but he seems to be busy now so I cannot guarantee fast responses 😅

Tak:
We could also try a heuristic on the url, and only search urls that look like they could be mastodon statuses

connyduck:
I think we cant justify a network request for every click on a link :/
What about the Mastalab way: offering to open all links having an @ as the first character of the path, e.g. https://domain.xy/@test

see here: https://github.com/stom79/mastalab/blob/41d65e91206839f597b816211250a4b13241640b/app/src/main/AndroidManifest.xml#L96

This would have the advantage of making all Mastodon links openable in the app (even from Browser or other apps) & not needing an network request but the drawback of maybe opening non-mastodon links.

Tak:
How about this: we do the search, but only on "probably mastodon" links.
(FWIW, I think pleroma uses a different layout without the @ user chunk in the path, but the check could be relaxed later if desirable.)

If we added the path prefix for the url in the manifest, we would still have to perform the search for all statuses not originating in the current instance, and we would get a bunch of external false positives (e.g. medium.com often has paths that begin with `/@contributor`)

connyduck:
This seems really reasonable 🤔
I am now focused on getting Tusky 1.6, but I definitely want this feature in 1.7.

charlag:
We should consider release branches? Good for hotfixds too

connyduck:
Yes I will make one for 1.6, good point!

connyduck:
Ok, as I stated, I really like this feature, but we need to get this right before we ship it.
There are several problems with the implementation in this PR:

- Users expect an immeditate reaction to a click on a link. But when they have a slow internet connection, nothing happens for a while. Users could click other links in the meantime and then suddenly TWO links open. This is very bad user experience.

- In case Users have no internet connection, the link opens in the browser. But if they press back, they see an empty ViewThreadActivity. Very awkward.

- We should not only handle links to threads but also to user profiles.

- I think the check if a link looks like a Mastodon one should not be made in `LinkHelper.setClickableText` but when a link is actually clicked (better perf)

One possible solution could be: Upon clicking a link, that might be a status/profile, open a [bottom sheet](https://material.io/guidelines/components/bottom-sheets.html) with text "opening link" and/or a progress bar. Bottom sheets are non intrusive (not like dialogs), meaning the user sees something is going on, but can still use the rest of the ui. Once we resolved the status, we can open it or the browser. If the user clicked something else in the meantime, we need to cancel the resolving. I am not sure if thats the ideal solution though.

Tak:
Ok. I previously investigated a bit around popping up a busy spinner, but didn't come up with anything that seemed workable (I'm not a hardcore android ui developer ;-) ).
I'll look into addressing these.

Tak:
- Moved the "could be a mastodon status" check to the click handler
- Added a bottom sheet with a status message, search is canceled when it's dismissed
- Suppressed opening the thread view activity while searching

app/src/main/java/com/keylesspalace/tusky/fragment/SFragment.java
```diff
@@ -269,10 +274,17 @@ private static boolean looksLikeMastodonUrl(String urlString) {
...
-                uri.getPath() != null &&
-                uri.getPath().matches("^/(@|notice).*/\\d+$"));
+        if (uri.getQuery() != null ||
+                uri.getFragment() != null ||
+                uri.getPath() == null) {
+            return false;
+        }
+
+        String path = uri.getPath();
+        return path.matches("^/@[^/]*$") ||
```
charlag: I think it could be more efficient to smash them all into one regex but I won't be able to do such wizardry nor I do I ask you, just saying

app/src/main/res/layout/fragment_timeline.xml
```diff
@@ -1,13 +1,22 @@
 <?xml version="1.0" encoding="utf-8"?>
-<android.support.v4.widget.SwipeRefreshLayout
+<RelativeLayout
```
charlag: Why `RelativeLayout` here?
I would either:
1. Make it a `FrameLayout`
2. Make it a `ConstraintLayout`
3. Make the `CoordinatorLayout` a parent and put `SwipeRefreshLayout` inside alongside that include (probably the best choice)

app/src/main/java/com/keylesspalace/tusky/fragment/SFragment.java
```diff
@@ -235,4 +259,140 @@ protected void openReportPage(String accountId, String accountUsername, String s
...
+    private void onBeginSearch(@NonNull String url) {
+        searchUrl = url;
+        showQuerySheet();
+    }
+
+    private boolean getCancelSearchRequested(@NonNull String url) {
+        return !url.equals(searchUrl);
+    }
+
+    private boolean getIsSearching() {
```
charlag: I'm kinda late for this but why `get` prefix? Kotlin is the only reason I can think of but it's private anyway

app/src/main/java/com/keylesspalace/tusky/fragment/NotificationsFragment.java
```diff
@@ -379,6 +376,11 @@ public void onViewThread(int position) {
         super.viewThread(notification.getStatus());
     }
 
+    @Override
+    public void onViewURL(String url) {
```
charlag: Why do we need this overridden?

app/src/main/java/com/keylesspalace/tusky/fragment/TimelineFragment.java
```diff
@@ -455,6 +453,11 @@ public void onViewThread(int position) {
         super.viewThread(statuses.get(position).getAsRight());
     }
 
+    @Override
+    public void onViewURL(String url) {
```
charlag: And this one?

app/src/main/java/com/keylesspalace/tusky/fragment/ViewThreadFragment.java
```diff
@@ -260,6 +257,11 @@ public void onViewThread(int position) {
         super.viewThread(status);
     }
 
+    @Override
+    public void onViewURL(String url) {
```
charlag: And this one

app/src/main/res/layout/fragment_view_thread.xml
```diff
@@ -1,12 +1,22 @@
 <?xml version="1.0" encoding="utf-8"?>
-<android.support.v4.widget.SwipeRefreshLayout xmlns:android="http://schemas.android.com/apk/res/android"
-    android:id="@+id/swipe_refresh_layout"
-    android:layout_width="match_parent"
-    android:layout_height="match_parent">
+<RelativeLayout
```
charlag: The same thing here with layout

app/src/main/res/values/strings.xml
```diff
@@ -303,5 +303,6 @@
 
     <string name="error_no_custom_emojis">Your instance %s does not have any custom emojis</string>
     <string name="copy_to_clipboard_success">Copied to clipboard</string>
+    <string name="performing_lookup_title">Performing lookup...</string>
```
charlag: This sounds a little tech-ish too me, maybe something more friendly like "looking it up?" Not sure what is easier to translate too.
Somebody ask people please, I'm not good at this
Also there's a symbol for ellipsys: … :3

app/src/main/java/com/keylesspalace/tusky/fragment/SFragment.java
Tak: I could make them a single regex, but I thought this was more readable/debuggable. The time taken to match against all the regexes in the worst case is still tiny compared to the actual network request (either for the search or for the browser).

app/src/main/res/layout/fragment_timeline.xml
Tak: I guess because I don't have strong feelings about it - I can adapt it to whichever of the above seems best.

app/src/main/java/com/keylesspalace/tusky/fragment/SFragment.java
charlag: Yeah, you're right.

app/src/main/java/com/keylesspalace/tusky/fragment/SFragment.java
Tak: Good point, this naming is just left over from earlier iterations.

app/src/main/java/com/keylesspalace/tusky/fragment/NotificationsFragment.java
Tak: Good point, leftover

app/src/main/res/values/strings.xml
Tak: I have no strong feelings; it can be changed to whatever seems best.
(We're not using the ellipsis character in "Sending Toot..." :wink: )

connyduck:
Ok I just tried it out and it is perfect 👌
We may need to do some tuning to the message, and maybe it makes sense to delay the bottom sheet a bit so it doesnt show up on fast connections, but lets wait for user feedback.

Mergy-merge?

charlag:
I still want button on the sheet to stop search but I think we can add it later in another PR.
Would be cool to make a build with this and collect some feedback, huh?
Let's merge!

connyduck:
Lets see if we get any feedback: https://mastodon.social/@ConnyDuck/99916304078357678

Tak:
It _does_ stop the search if you swipe the sheet away...maybe that's not obvious UX?

charlag:
But does it open the status after swiping the sheet away or it will abort the whole process? Cannot really test it

Tak:
It cancels the search and does not open the status.

PR #571, opened 2018-04-13
Query instance for toot character limit

### LINKED ISSUE #393: Instance-specific Character Limit.
I'm on a Mastodon instance that offers a 750 characters-limit, but on Tusky I can only post toots 500 characters long. When I was using Mastalab, it counted characters from 0, not from 500 like Tusky and Twidere. Maybe that's one way of giving Tusky an instance-specific character limit? I don't know. But Tusky's limitation definitely defeats the purpose of joining an instance that offers more characters. Please look into this.

Tak:
Fixes #393


app/src/main/java/com/keylesspalace/tusky/ComposeActivity.java
```diff
@@ -164,6 +165,8 @@
     private static final String REPLYING_STATUS_AUTHOR_USERNAME_EXTRA = "replying_author_nickname_extra";
     private static final String REPLYING_STATUS_CONTENT_EXTRA = "replying_status_content";
 
+    private static int maximumTootCharacters = STATUS_CHARACTER_LIMIT;
```
charlag: I think it doesn't have to be static?.. At least it seems hacky to me

charlag:
That's pretty cool, thanks!
but character limit is not in the vanilla servers, right? Only glitchsoc?

app/src/main/java/com/keylesspalace/tusky/ComposeActivity.java
Tak: I made it static as a simple way of caching it between instances.
If you'd rather, I can make it an instance field instead.

Tak:
Not sure what servers have it, tbh.
I made all the Instance fields that aren't documented as non-nullable in the tootsuite api doc nullable - my instance doesn't support `max_toot_chars`, but the request is still handled correctly.

app/src/main/java/com/keylesspalace/tusky/ComposeActivity.java
charlag: I appreciate the idea but maybe we should cache it some other way (e.g. with account data). Shared and mutable are okay by itself and shared and mutable is a recipe for an untestable mess ;)
Let's think about the best way to do it either here or
https://matrix.to/#/#Tusky:matrix.org

Tak: The caching isn't super important, I just thought it would be a nice side effect, and I didn't think too much about it since there can only be one compose activity at a time (I think?).

But it can just be an instance field; I don't think it's a big deal.

charlag: Yeah, I just like the idea of caching too but I don't like static fields :sweat_smile: 
Here's a real-life scenario: you change your account and it has another char limit and now we show wrong limit.
It would be cool to cache it while keeping it maintainable but I would approve it even without cache, tbh.
Maybe store the limit in SharedPreferences or make a Room table for instances. We could use some instance info in the future and it would be nice to have it cached.
Uuuugh, @connyduck can you give your opinion on this pls since you will be deciding on it?..

Tak: If I understand the life cycle correctly, if you change your account, the value would be requeried, and it could only be incorrect if the request failed (which is also true in the normal case). 

But it's not a problem; I'll just change it. ☺

connyduck:
Can you cache this the same way we cache the emojis? Maybe even use the same database table?

Tak:
Sure, I can take a look (I made the initial version of this before the emoji stuff landed in master)

Tak:
Ok, I expanded on the emoji list table, which makes sense since they're both keyed on the instance (although that makes me wonder why the emoji list is a separate rest api point...)

app/src/main/java/com/keylesspalace/tusky/db/InstanceEntity.kt
```diff
@@ -25,8 +25,9 @@ import com.keylesspalace.tusky.entity.Emoji
 
 @Entity
 @TypeConverters(Converters::class)
-data class EmojiListEntity(@field:PrimaryKey var instance: String,
-                           val emojiList: List<Emoji>)
+data class InstanceEntity(@field:PrimaryKey var instance: String,
```
charlag: Please format it like you formatted `Instance` with each field on its own like thx :heart:

app/src/main/java/com/keylesspalace/tusky/ComposeActivity.java
```diff
@@ -1429,6 +1436,26 @@ public void onEmojiSelected(@NotNull String shortcode) {
...
+            Integer max = instanceEntity.getMaximumTootCharacters();
+            maximumTootCharacters = (max == null ? STATUS_CHARACTER_LIMIT : max.intValue());
+            emojiList = instanceEntity.getEmojiList();
+            if (emojiList != null) {
+                emojiView.setAdapter(new EmojiAdapter(emojiList, ComposeActivity.this));
+            }
+        }
+    }
+
+    private void cacheInstanceMetadata(@NotNull AccountEntity activeAccount)
```
charlag: I wonder why it works.
You call it from requests callbacks which are called... on the main thread I guess?
Shouldn't Room complain that you access DB from UI thread?

connyduck: We set `allowMainThreadQueries()` in `TuskyApplication`

charlag: Uuuugh, afaik it's mostly for debugging... I wish we did it asynchronously but okay

charlag:
Now it looks like a solid PR, thank you! Would be nice to have some tests but it's good-to-merge already to me.
But CI fails 🤔
(It is kinda annoying that it asks you ti login to check what's wrong)

Tak:
Quality CI :wink:

Tak:
I made a first pass at adding some tests: https://github.com/Tak/Tusky/commit/83f5ec128058f0a81f2053db9df2254653b6cd33
(on a separate branch, since I suspect there will be some discourse...)

app/src/main/java/com/keylesspalace/tusky/ComposeActivity.java
```diff
@@ -222,6 +226,7 @@ public void onCreate(Bundle savedInstanceState) {
         emojiButton = findViewById(R.id.composeEmojiButton);
         hideMediaToggle = findViewById(R.id.composeHideMediaButton);
         emojiView = findViewById(R.id.emojiView);
+        emojiList = new ArrayList<>();
```
charlag: 1. You could do `Collections.emptyList()`. Not that important but good-to-know
2. We have redundant checks for it to be null in the fragment

Tak: The json converter requires it to be nonnull, however - I hit this while working on tests

charlag: Yes, I I'm cool with making it non-null at all times, I'm just saying that we have checks for null which we don't need now

app/src/main/java/com/keylesspalace/tusky/ComposeActivity.java
connyduck: Yes its ugly. We will get rid of it eventually.

connyduck:
Your database upgrade does not work

```
android.database.sqlite.SQLiteException: table InstanceEntity has 3 columns but 2 values were supplied (code 1): , while compiling: INSERT INTO `InstanceEntity` SELECT * FROM `EmojiListEntity`;
        at android.database.sqlite.SQLiteConnection.nativePrepareStatement(Native Method)
        at android.database.sqlite.SQLiteConnection.acquirePreparedStatement(SQLiteConnection.java:890)
        at android.database.sqlite.SQLiteConnection.prepare(SQLiteConnection.java:501)
        at android.database.sqlite.SQLiteSession.prepare(SQLiteSession.java:588)
        at android.database.sqlite.SQLiteProgram.<init>(SQLiteProgram.java:58)
        at android.database.sqlite.SQLiteStatement.<init>(SQLiteStatement.java:31)
        at android.database.sqlite.SQLiteDatabase.executeSql(SQLiteDatabase.java:1752)
        at android.database.sqlite.SQLiteDatabase.execSQL(SQLiteDatabase.java:1682)
        at android.arch.persistence.db.framework.FrameworkSQLiteDatabase.execSQL(FrameworkSQLiteDatabase.java:240)
        at com.keylesspalace.tusky.db.AppDatabase$5.migrate(AppDatabase.java:90)
        at android.arch.persistence.room.RoomOpenHelper.onUpgrade(RoomOpenHelper.java:73)
```

charlag:
It happens :sweat_smile:

PR #586, opened 2018-04-17
Translation to Dutch

littlebelgianwriter:
Is it possible to continue the translation later on, like bit by bit?
For the moment, I don't see any issues with regards to the translation. 
I should be able to do this.


connyduck:
Thanks for your help!
But you are currently editing the wrong file, Dutch translations should go into `app/src/main/res/values-nl/strings.xml` and not into `app/src/main/res/values/strings.xml` (thats the default English translation, please don't override that).

PR #587, opened 2018-04-18
Notification tweaks: Grouping and Quick Reply button

### LINKED ISSUE #535: Group notifications, Android N style
Pretty self-explanatory - notifications should be separate but grouped if there are more than a couple. I'd offer help but I haven't worked with Android in over a year and have no idea how to work with the new APIs ¯\\\_(ツ)\_/¯

remi6397:
Fixes #535

This PR adds [Notification Grouping](https://developer.android.com/training/notify-user/group.html) and [Quick Reply button](https://developer.android.com/training/notify-user/build-notification.html#reply-action) to notifications.

I'm submitting a PR, because I want you to let me know about your opinions on the UX. It looks like everything is working properly, but it may need some more tests.


app/src/main/java/com/keylesspalace/tusky/receiver/SendStatusBroadcastReceiver.kt
```diff
@@ -0,0 +1,91 @@
...
+            val notification = intent.getStringExtra(NotificationHelper.KEY_NOTIFICATION_ID)
+            val sender = intent.getLongExtra(NotificationHelper.KEY_SENDER_ACCOUNT, -1)
+            val citedStatus = intent.getLongExtra(NotificationHelper.KEY_CITED_STATUS, -1)
+            val visibility = intent.getSerializableExtra(NotificationHelper.KEY_VISIBILITY) as Status.Visibility
+            val spoiler = intent.getStringExtra(NotificationHelper.KEY_SPOILER)
+            val mentions = intent.getStringArrayExtra(NotificationHelper.KEY_MENTIONS)
+
+            val account = accountManager.getAccountById(sender)
+
+            val sendIntent = SendTootService.sendTootIntent(context!!, mentions.map { "@$it" }.joinToString(" ") + " " + message.toString(), spoiler,
```
charlag: It looks nice but it seems expensive
Check `Sequence.joinToString()` methods, that's exactly what we need (or maybe even `joinTo()`)

app/src/main/java/com/keylesspalace/tusky/receiver/SendStatusBroadcastReceiver.kt
```diff
@@ -0,0 +1,91 @@
...
+                    visibility, false, Arrays.asList(), Arrays.asList(), citedStatus.toString(),
```
charlag: I'd put arguments names here for readability.

app/src/main/java/com/keylesspalace/tusky/util/NotificationHelper.java
```diff
@@ -104,75 +133,145 @@ public static void make(final Context context, Notification body, AccountEntity
...
+        summaryBuilder.setCategory(NotificationCompat.CATEGORY_SOCIAL);
+        summaryBuilder.setOnlyAlertOnce(true);
+        summaryBuilder.setGroupSummary(true);
+
+        NotificationManagerCompat notificationManager = NotificationManagerCompat.from(context);
+
+        //noinspection ConstantConditions
+        notificationManager.notify(Integer.valueOf(body.getId()), builder.build());
+        if (currentNotifications.length() == 1) {
+            notificationManager.notify((int) account.getId(), builder.setGroupSummary(true).build());
```
charlag: Is it safe?..

charlag: I mean casting to int

app/src/main/java/com/keylesspalace/tusky/receiver/SendStatusBroadcastReceiver.kt
remi6397: How exactly would you rewrite that to use lazy sequences?

charlag: I think the way you did it is already cool

app/src/main/java/com/keylesspalace/tusky/util/NotificationHelper.java
remi6397: It's okay, because row identifiers in database aren't random, but incremented by one for each new account, and I guess that no one would exceed the "limit" of 65535 accounts. :wink:

app/src/main/java/com/keylesspalace/tusky/receiver/SendStatusBroadcastReceiver.kt
remi6397: I think that it's OK now. Better not to exaggerate; other developers would probably check function signature anyway. :wink:

app/src/main/java/com/keylesspalace/tusky/util/NotificationHelper.java
charlag: Ah, it's a local id, thx, I think it's perfectly legal

app/src/main/java/com/keylesspalace/tusky/receiver/SendStatusBroadcastReceiver.kt
```diff
@@ -0,0 +1,100 @@
...
+                    text,
+                    spoiler,
+                    visibility,
+                    false,
+                    Arrays.asList(),
+                    Arrays.asList(),
+                    citedStatus.toString(),
+                    null,
+                    null,
+                    null, account!!, 0)
```
connyduck: please add a check that `account != null`, just to be save

remi6397:
@connyduck, It would be best to ask community whether they want that feature or not. :wink:

charlag:
I'm also for the opening the compose screen I guess.

remi6397:
It looks like most of people agree with @connyduck and @charlag here. However, I liked the idea of making it an optional feature brought by @unfld@mastodon.social. When combined with idea of opening ComposeActivity anyway for restricted toots (which was brought by @adamk678@mastodon.social
), it makes a perfect compromise as for me. What do you think? :grin:

charlag:
I'm okay with the option

remi6397:
@connyduck, that's definitely the best options of all. However, don't you think that adding more buttons makes application's interface less intuitive?

What about just jumping to the toot when notification area is clicked, and leaving "Quick Reply" button as is? That would make it easy to favorite/boost the toot before mentioning. Also, we can disable it for toots with CW and display privacy settings right in the notification.

As for legacy API support, we can make the reply button fall back to ComposeActivity on older devices.

connyduck:
> @connyduck, that's definitely the best options of all. However, don't you think that adding more buttons makes application's interface less intuitive?

Yes it does, so my preferred option is still no quick reply. It needs the same amount of clicks to send a reply when we open the ComposeActivity.

What about a setting so users can change it to their liking?

connyduck:
Ok so I think the best compromise we will get is the following:
- clicking notification content takes user to notification tab as it always was
- mention notifications get a "reply" button that opens up ComposeActivity
- on Api level >= N we add a "quick reply" button that starts quick reply

Needs no settings, improves features on all api levels and power users get their quick reply 🙂

remi6397:
I think it should be okay now. What do you think? @connyduck @charlag

app/src/main/java/com/keylesspalace/tusky/util/NotificationHelper.java
```diff
@@ -104,75 +139,168 @@ public static void make(final Context context, Notification body, AccountEntity
...
+        summaryBuilder.setOnlyAlertOnce(true);
+        summaryBuilder.setGroupSummary(true);
 
-        builder.setVisibility(NotificationCompat.VISIBILITY_PRIVATE);
-        builder.setCategory(NotificationCompat.CATEGORY_SOCIAL);
+        NotificationManagerCompat notificationManager = NotificationManagerCompat.from(context);
 
-        builder.setOnlyAlertOnce(true);
+        //noinspection ConstantConditions
+        notificationManager.notify(Integer.valueOf(body.getId()), builder.build());
```
connyduck: I don't think it is a good idea to use Mastodon notification ids as Android notification ids.
- they might not be unique since we support multiple accounts (extremly unlikely, but still)
- they might be way bigger than Integer.MAX_VALUE

app/src/main/java/com/keylesspalace/tusky/receiver/SendStatusBroadcastReceiver.kt
```diff
@@ -0,0 +1,140 @@
...
+                val text = mentions.joinToString(" ", postfix = " ") { "@$it" } + message.toString()
+
+                val sendIntent = SendTootService.sendTootIntent(
+                        context,
+                        text,
+                        spoiler,
+                        visibility,
+                        false,
+                        Arrays.asList(),
+                        Arrays.asList(),
```
connyduck: you can use `emptyList()` here thats more efficient

charlag: `listOf` would do the same

app/src/main/java/com/keylesspalace/tusky/util/NotificationHelper.java
remi6397: @connyduck I don't like that solution too. Do you have any other proposals?

connyduck:
People will complain that they get spammed with so many notifications at once, but I don't think there is much we can do about that 🤔

The last problem I found is that you don't make sure the compose activity opens with the account that received the mention. (yes there may be usecases where the user would like to answer with a different account, but we need to solve that within the ComposeActivity)

And Tusky crashes when cancelling a notification, but that is my fault because I messed up the dependency injection somewhere I think, I will fix it once this is merged.

connyduck:
oh and I just made a quick reply where the original notification stayed on forever
screenshot_20180503-195249

connyduck:
oh and I think the switching of the account is also necessary before opening a thread

remi6397:
@connyduck Where exactly? I don't see any problems in there…

charlag:
I'm not sure why, but master was building while this branch was not. Some problem with licenses.
I'm trying a newer container (api-27) in case it was because of that

connyduck:
> @connyduck Logcat would be very welcome ^^

I could not reproduce it a second time, will post logcat once it happens again

remi6397:
The same happened to me, however I have both my accounts on same instance. My logcat: gist.github.com/remi6397/37d3f287e8ce9885b3228cc82c9157e8
I'm tired now, so I'll try to fix that tomorrow :)

charlag:
@connyduck are you waiting for my approval or?..

PR #590, opened 2018-04-19
strings: Add spanish translations

Eskuero:
I kept more or less the same length as the original english translations so everything should fit, although I'm unable to test them at the moment.
For every string I wasn't sure how to translate I checked it against the web interface equivalent from mastodon.social
For every alert or information provided by the app I used a polite verb form, while keeping casual ones for interactions between users.

### DIFF
--- a/app/src/main/res/values-es/strings.xml
+++ b/app/src/main/res/values-es/strings.xml
@@ -231,7 +231,7 @@
 
     <string name="description_account_locked">Cuenta protegida</string>
 
-    <string name="about_title_activity">About</string>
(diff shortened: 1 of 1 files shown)


charlag:
Cool, thank you! Would be nice if someone had a look at it but if not we'll merge it as-is

Eskuero:
I forgot to translate one string which I just pushed. I built the app and everything else seems to fit just fine.

screenshot_20180419-165000screenshot_20180419-165418screenshot_20180419-165351screenshot_20180419-165423screenshot_20180419-165454

connyduck:
Super awesome, thank you very much!

PR #591, opened 2018-04-19
Add definition of Free Software in strings.xml to prevent mistakes in farther translations

remi6397:
It may help to fix some inconsistencies in translations. Consider "бесплатное" ("gratis") from Russian translation and "wolnym" ("libre") from Polish translation.


charlag:
Oh, good catch! How did I miss this in the Russian translation x_x (I don't use it in Russian tho)

PR #593, opened 2018-04-20
New Dutch Translation

littlebelgianwriter:
I replaced the old and incomplete translation by my new translation.


charlag:
Neat, so we can close previous pull request?

littlebelgianwriter:
I hope so, for some reason, I see now the 'All checks have failed message here above.

connyduck:
Looks like the `</resources>` tag at the start of the file is missing

charlag:
I've tried to fix it directly on GitHub, let's see how it goes

littlebelgianwriter:
Thank you! I shouldn't have missed that.

connyduck:
Thank you very much! Your changes will be in the next release!

PR #596, opened 2018-04-20
Dependency injection improvement

connyduck:


charlag:
This warms my heart
DI all the things!

connyduck:
I don't really know how to inject into `PreferencesFragment` because it does not extend from `android.support.v4.app.Fragment`. Changing to support Fragment fucks up the code for regex validating...
But its the only thing stopping me from removing the ugly `getInstance` from `TuskyApplication` :/

charlag:
You need another `DispatchingAndroidInjector` I believe? One for platform `Fragment`s.
And then you probably need to register lifecycleCallbacks... (but that's API 26 so we're out of luck) or use AndroidInjection manually (I don't really remember how to).

You can do it in a "traditional" way: Make a new Component which can inject into fragment. Construct it in the fragment (`DaggerPreferecesComponent.Builder()...`) and use it. This component will need an AppComponent to inheret stuff from it. But wait, `AppComponent` cannot expose dependencies because it injects... and you would need an instance of the app anyway.

So, upgrading it to support fragment would be the least hacky way imo but I've never used `PreferencesFragment` so I cannot tell much about it

charlag:
Also check out this
https://stackoverflow.com/a/32328219/1923879

connyduck:
last commits make sure there is always only one `OkHttpClient` & fix the logging

charlag:
Btw, why cache is 10 Mb? I read that by default
"Disk cache of 2% storage space up to 50MB but no less than 5MB."
I don't mind having a few hundred mb cache.
It would be really great to cache userpics more aggressively but I'm not sure that event newest versions of Picasso allow that.

charlag:
It's from here:
https://futurestud.io/tutorials/picasso-influencing-image-caching
on their website it also says "Automatic memory and disk caching."

My biggest concern is bad network conditions but it's okay I guess. I just compare it with other social apps and I think they have a large cache afaik.

I think the trick with userpics is only possible with custom cache class and it can be done sometime in the distant future

connyduck:
Ok we need to get rid of the hacky regex validation in `PreferencesFragment.onCreate` because it prevents us from using the support library for preferences (which is needed for injecting).

PR #599, opened 2018-04-22
Add tests for the custom toot character limit functionality

Tak:


connyduck:
The tests fail for me with 
`java.lang.VerifyError: class org.robolectric.android.fakes.RoboMonitoringInstrumentation overrides final method specifyDexMakerCacheProperty.()V`

I guess some dependency versions have to be changed? https://github.com/robolectric/robolectric/issues/3839

Tak:
According to https://github.com/robolectric/robolectric/issues/3751#issuecomment-380014021 , it's a bug in Android Studio 3.1.1.
They do pass if you run them at the command line. (`./gradlew test`)

PR #600, opened 2018-04-24
EmojiCompat support

C1710:
So I finally added EmojiCompat :D 
Please note that I don't know anything about license compatibility...
There are currently two fonts available: Twemoji and Blobmoji.
The font is loaded from the apps private external storage folder (`/storage/emulated/0/Android/data/com.keylesspalace.tusky/files/EmojiCompat.ttf`).
This method ensures that using EmojiCompat is optional (it will use a dummy configuration if no font is present) and customizable.  
It should work on any component expect for the Account Header in the Navigation Drawer.


charlag:
Wow, that's wonderful, thank you!
I'm not sure I understand it completely.
There are two configurations - assets and file and I am confused.
I'm also confused about how we should ship this - make two builds?
I'm probably missing something, I would be glad if you could explain it, thx!

C1710:
These two configurations are used because of some issues if the EmojiCompat font is missing.  
Usually it will try to use `FileEmojiCompatConfig`.  
It will look into the `file`-directory and look for the font file called `EmojiCompat.ttf` - it can be downloaded manually into this directory.  
Since the app will crash if `EmojiCompat` isn't initialized properly, I needed to add a fallback solution.  
So I just made a minimal (and very small) EmojiCompat font which only consists of 6(?) emojis and won't replace them if they are present on the device (which is like always).  
So in this fallback case it will use `AssetEmojiCompatConfig` to load this dummy font.  

This solution ensures that the user is able to
- use the app without a large APK to download
- use the app with their favourite emoji font (as of now I only made a font for [Twemoji](https://github.com/twitter/twemoji) and [Blobmoji](https://github.com/c1710/blobmoji)
- use the app without any EmojiCompat font (this is the fallback solution I mentioned earlier) - although EmojiCompat is actually enabled and "working", but it doesn't affect anything.

A more user-friendly solution would be to make the fonts available online and to add an option (e.g. in Settings) to download one of them.  
As far as I understood, you can even refresh EmojiCompat at runtime but I think restarting the app shouldn't be too difficult.  
So all in all this solution won't do _anything_ by default, but it offers the possibility to add EmojiCompat without having to create a new fork of this app.

connyduck:
Ok this is super awesome, but not yet userfriendly enough. Users should be able to select their preferred style in the settings + then we download the file to the device. I can implement that if you dont want to.

C1710:
@connyduck I can try adding this option, but I'm not very sure about these things 😅

connyduck:
> I don't know how Tusky handles all these things with settings and so on😂 - integrating EmojiCompat was relatively easy though...

I would do it the following way: Add a new `Preference` between theme and font size. On click, show a dialog with the options (maybe even example emojis) and the download size. Add a text like "Tusky needs to download the emojis to your device in order for this to work". 
After users selected a style, show a progressbar (this may be tricky to do with okhttp, indefinite progressbar is ok for now) and then just download the file, a subfolder `Tusky` of `Environment.getExternalStorageDirectory()` is probably best. After the download finished, verify the file is there (bonus: check hashsum) and save the selected style to SharedPreferences. On error show an error message and revert to default style.

On appstart, check the shared preferences for the selected style and then check if the required font file exists. If so, use it, otherwise fall back to default.

I can help you if you have questions about implementation details

> The question is: Where should the fonts be hosted?

We could put it on https://tuskyapp.github.io/ ?

C1710:
I just finished the implementation of EmojiCompat!  
Remaining issues:
- I didn't comment the code yet (I'll do this in the next few hours)
- There's no option to stop a download
- Some weird positioning of the radio buttons
- You can't update a font or the font list
- The code might be a bit chaotic - I'll need to do some cleanup
- There's no progress shown
- No checksums
- There's no link to the font's sources. It is available in the files though to be easily implemented
- I'll need to sync the fork in order to prevent Merge conflicts.

But there are a few things I can't do:
1. The fonts still need to be hosted somewhere
2. The font configuration is based on the JSON format. This includes a list of the available fonts which is downloaded as well. I added a [sample version](https://github.com/C1710/Tusky/blob/emojiSettings/fonts.json) of it. (It's actually just missing the real links. I used the Blobmoji links for debugging)

Please be aware that I currently use a different branch than the one I started the pull request from

C1710:
Oh and another issue is that you'll need to restart the app to select a new emoji font since @Google blocked classes from outside the EMojiCompat library itself resetting the configuration 🤷‍♀️

C1710:
Okay. I solved the following issues:
- There's now documentation and comments
- Everything is ready to merge

But I don't know how to properly set the text color depending on the app theme. So I haven't changed it yet.

And you would need to insert the final URLs into the font list as well as in `EmojiPreference` 😅

C1710:
Oh - it looks like something wnt wrong when I merged the emojiSettings branch back into fEmojiCompat...
So I guess I'll simply do another Pull request based ont that branch. Sorry!

C1710:
Okay. Everything seems broken now 😣

charlag:
Sorry! Shit happens.

C1710:
Okay. It seems to work now. I had some issues with Gson and with a variable which should have been shadowed...  
And it seems like I finally overwrite-merged everything into fEmojiCompat now.

app/build.gradle
```diff
@@ -71,6 +73,10 @@ dependencies {
     }
     implementation 'com.evernote:android-job:1.2.5'
     implementation 'com.android.support.constraint:constraint-layout:1.1.0'
+    // EmojiCompat
+    implementation "com.android.support:support-emoji:$supportLibraryVersion"
+    implementation "com.android.support:support-emoji-bundled:$supportLibraryVersion"
```
connyduck: NOPE this bundles the noto font into the apk we do not want this

app/src/main/java/com/keylesspalace/tusky/EmojiPreference.java
```diff
@@ -0,0 +1,268 @@
...
+                // Start download
+                Response response = client.newCall(request).execute();
+                // Open up the writing part
+                sink = Okio.buffer(Okio.sink(fontList));
+                try {
+                    if (response.body() != null) {
+                        // GOGOGO! DOWNLOAD!
+                        sink.writeAll(response.body().source());
+                    } else {
+                        Log.e("FUCK", "downloadFonts: Source empty");
```
connyduck: lol. Please use a proper logging tag

app/src/main/res/layout/emoji_pref_item.xml
```diff
@@ -0,0 +1,119 @@
...
+        app:layout_constraintEnd_toEndOf="parent"
+        app:layout_constraintTop_toTopOf="parent"
+        app:srcCompat="@drawable/ic_file_download_black_24dp"
+        android:tint="@color/colorAccent"
+        app:layout_constraintVertical_bias="0.5"
+        android:focusable="true"
+        android:contentDescription="@string/download_fonts"
+        android:visibility="visible"/>
+
+    <!--TODO: enable this button again-->
```
connyduck: yes please :D

app/src/main/res/layout/emoji_pref_item.xml
```diff
@@ -0,0 +1,119 @@
...
+        android:layout_width="72dp"
+        android:layout_height="72dp"
+        android:paddingBottom="36dp"
+        android:paddingTop="36dp"
+        android:paddingEnd="36dp"
+        android:paddingStart="36dp"
+        android:layout_marginEnd="-20dp"
+        android:layout_marginStart="-20dp"
+        style="?android:attr/borderlessButtonStyle"
+        android:background="@drawable/button_ripple"
```
connyduck: should be `?attr/selectableItemBackgroundBorderless`, delete the `button_ripple` drawable

app/src/main/res/layout/emoji_pref_item.xml
```diff
@@ -0,0 +1,119 @@
...
+    <ImageButton
+        android:id="@+id/download_emoji_button"
+        android:layout_width="72dp"
+        android:layout_height="72dp"
+        android:paddingBottom="36dp"
+        android:paddingTop="36dp"
+        android:paddingEnd="36dp"
+        android:paddingStart="36dp"
+        android:layout_marginEnd="-20dp"
+        android:layout_marginStart="-20dp"
```
connyduck: negative margins ouch

app/src/main/res/layout/emoji_pref_item.xml
```diff
@@ -0,0 +1,119 @@
+<?xml version="1.0" encoding="utf-8"?>
```
connyduck: file should be called `item_emoji_pref.xml`

app/src/main/res/layout/emojicompat_dialog.xml
```diff
@@ -0,0 +1,51 @@
+<?xml version="1.0" encoding="utf-8"?>
```
connyduck: file should be called `dialog_emojicompat.xml`

app/src/main/res/xml/preferences.xml
```diff
@@ -12,6 +12,11 @@
             android:summary="%s"
             android:title="@string/pref_title_app_theme" />
 
+        <com.keylesspalace.tusky.EmojiPreference
+            android:defaultValue="@string/system_default"
+            android:key="emojiCompat"
+            android:title="@string/emoji_style"/>
+
```
connyduck: the summary of the Preference should be the currently selected style

app/src/main/java/com/keylesspalace/tusky/FileEmojiCompatConfig.java
```diff
@@ -0,0 +1,170 @@
...
+            // We'll need the context for the fallback solution
+            this.mContext = context.getApplicationContext();
+            // NEW
+            this.file = fontFile;
+        }
+
+
+        // Copied from BundledEmojiCompatConfig
+        @Override
+        @RequiresApi(19)
```
connyduck: no need for this, we are on Api 19

app/src/main/java/com/keylesspalace/tusky/adapter/EmojiFontAdapter.java
```diff
@@ -0,0 +1,317 @@
...
+     * @param viewType ????
+     * @return An empty font entry
+     */
+    @NonNull
+    @Override
+    public EmojiFontViewHolder onCreateViewHolder(@NonNull ViewGroup parent, int viewType) {
+        View item = LayoutInflater.from(parent.getContext())
+                .inflate(R.layout.emoji_pref_item, parent, false);
+        // All radio buttons need to be stored to uncheck them later
+        radioButtons.add(item.findViewById(R.id.emoji_radio));
```
connyduck: I think this does not work with views that are recycled

app/src/main/java/com/keylesspalace/tusky/entity/EmojiCompatFont.java
```diff
@@ -0,0 +1,321 @@
...
+import java.io.IOException;
+
+import okhttp3.OkHttpClient;
+import okhttp3.Request;
+import okhttp3.Response;
+import okio.BufferedSink;
+import okio.Okio;
+
+
+// TODO: It might be possible that this is not a class that belongs to the entity-package...
```
connyduck: put it into the util package for now until we figure out a better package structure

app/build.gradle
C1710: Oh. I thought I deleted that line 😂

app/src/main/java/com/keylesspalace/tusky/EmojiPreference.java
C1710: This is something I forgot to remove. I did this to make debugging the app a little bit easier since I could just search the logcat 😅

app/src/main/java/com/keylesspalace/tusky/FileEmojiCompatConfig.java
C1710: It's there because I simply copy-pasted BundledEmojiCompatConfig. But I'll remove it of course

app/src/main/java/com/keylesspalace/tusky/adapter/EmojiFontAdapter.java
C1710: As far as I remember onCreateViewholder is never called on recycled Views as this creates these Views.  
There's another method for recycled Views which I don't remember...

app/src/main/res/layout/emoji_pref_item.xml
C1710: This was a solution I made in another (not published) app.  
I can try to find another solution...

C1710: Seems like these don't do anything anyway 😅

C1710:
So for now I [implemented](https://github.com/C1710/Tusky/commit/bdeccd48cafc68d8669c268430f3f8a825602676) all changes (well, I'm working on cleaning up all those comments) except for the removal of the font list feature as I got a few questions on how you imagine an alternative solution:
1. What do you think should be used instead of the RecyclerView? Just some hardcoded layout?
2. Does "hardcode" mean, putting everything right into the app or simply replacing the downloadable list with a JSON file stored locally in the `assets`-folder?
3. What about the thumbnail files? Put them into `res/drawable`, `assets` or download them like it is done now?  

Cancelling a font download should work now btw.

connyduck:
> What do you think should be used instead of the RecyclerView? Just some hardcoded layout?

Yes, why not
> Does "hardcode" mean, putting everything right into the app or simply replacing the downloadable list with a JSON file stored locally in the `assets`-folder?

Just put it right into the code
> What about the thumbnail files? Put them into `res/drawable`, `assets` or download them like it is done now?

Please use vector drawables. I can help you creating those.

You can send a PR to https://github.com/tuskyapp/tuskyapp.github.io for hosting the font files.
To fix the bitrise build you probably have to merge master.

C1710:
> Please use vector drawables. I can help you creating those.

For Blobmoji I already have a VectorDrawable available and for Twemoji you could simply convert one of their SVG assets 😅

C1710:
Okay. The latest commit includes all changes you requested (at least I hope so).  
And it adds a real progress bar (which was rather easy to implement) to the downloads as well as a confirmation dialog asking the user to restart the app.  
The new dialog required several new strings (as far as I could see) which are only available in English and German for now...

app/src/main/java/com/keylesspalace/tusky/util/EmojiCompatFont.java
```diff
@@ -0,0 +1,309 @@
...
+        this.src = src;
+    }
+
+    /**
+     * Returns the Emoji font associated with this ID
+     * @param id the ID of this font
+     * @return the corresponding font. Will default to SYSTEM_DEFAULT if not in range.
+     */
+    public static EmojiCompatFont byId(int id) {
+        if(id >= 0 && id < FONTS.length) {
```
charlag: This formatting of if/else makes me sad

C1710:
About Kotlin: I think I'm able to _read_ code written in Kotlin, but I'm not sure that I'm able to actually _write_ code in Kotlin.  
About the fonts: I made a pull request for [tuskyapp.github.io](https://github.com/tuskyapp/tuskyapp.github.io) as @connyduck suggested in order to host the files there.  
At least for Blobmoji I can say that it's either licensed under the SIL OFL or Apache 2.0 license, but I'd go for SIL OFL (I'm not sure about all this stuff as I didn't create most of the assets of it which makes it some kind of derivate work. The original NotoColorEmoji _font_ is licensed under the OFL though. I think I should ask them on how licensing works in this case).  
Twemoji is licensed under CC-BY-4.0, but they got some more permissive [Attribution requirements](https://github.com/twitter/twemoji#attribution-requirements).  
And I simply forgot to remove the font files from this repo 😅

app/build.gradle
```diff
@@ -50,6 +50,8 @@ dependencies {
     implementation('com.mikepenz:materialdrawer:6.0.7@aar') {
         transitive = true
     }
+    // GSON is used to parse the EmojiCompat font-list
```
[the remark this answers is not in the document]
charlag: Are you sure that this is a good idea? If we're depending on Gson not only for network than it's not bad to mention it explicitly?
Theoretically: we're getting rid of convertor. The person removes it. Other part of the app breaks. No good.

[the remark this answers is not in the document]
connyduck: Yes I agree, but GSON usage has been removed from this PR I cant find it anymore?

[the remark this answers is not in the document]
charlag: Oops, sorry, my bad, I fail to keep up with this fast-changing PR :sweat_smile:

C1710:
I just added all these changes including the removal of the GSON dependency.  
I already included two files which can be used for some credits in the About screen, but since my train is just arriving at its destination, I don't have the time to implement it now

C1710:
Information about the emoji font's licensing is now available in the About screen.  
It should now comply with both licenses.

PR #608, opened 2018-04-26
Fix SparkButtons

connyduck:
- sparks no longer cut off
- higher perf (~50% faster view inflation on my devices)
- antialiased sparks
- bigger hit area for the buttons
- fixed the invalidation of the detailed status

I needed to fork the library because my PRs did not get accepted ¯\\\_(ツ)_/¯
But these things annoyed me for so long, finally I came around to fix them

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/fragment/ViewThreadFragment.java
+++ b/app/src/main/java/com/keylesspalace/tusky/fragment/ViewThreadFragment.java
@@ -191,7 +191,7 @@ public class ViewThreadFragment extends SFragment implements
                     StatusViewData.Concrete newViewData = viewDataBuilder.createStatusViewData();
 
                     statuses.setPairedItem(position, newViewData);
-                    adapter.setItem(position, newViewData, true);
+                    adapter.setItem(position, newViewData, false);
                 }
             }
 
@@ -224,7 +224,7 @@ public class ViewThreadFragment extends SFragment implements
                     StatusViewData.Concrete newViewData = viewDataBuilder.createStatusViewData();
 
                     statuses.setPairedItem(position, newViewData);
-                    adapter.setItem(position, newViewData, true);
+                    adapter.setItem(position, newViewData, false);
                 }
             }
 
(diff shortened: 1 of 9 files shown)


charlag:
This sounds awesome!
Where can we find fork source? I don't see it on Github but it's probably my fault

connyduck:
https://github.com/connyduck/SparkButton

connyduck:
Merging this will also make the build error on master go away, because the old library included `android:allowBackup` in its manifest :D

charlag:
I'm looking at it rn, give me a sec...

app/src/main/java/com/keylesspalace/tusky/fragment/ViewThreadFragment.java
```diff
@@ -191,7 +191,7 @@ public void onResponse(@NonNull Call<Status> call, @NonNull Response<Status> res
                     StatusViewData.Concrete newViewData = viewDataBuilder.createStatusViewData();
 
                     statuses.setPairedItem(position, newViewData);
-                    adapter.setItem(position, newViewData, true);
```
charlag: I was thinking that this is a bad idea because adapter wouldn't new about the changes and I tried to hack around with ItemAnimator but then I realized that it's actually cool because we still update adapter's collection

app/src/main/res/layout/item_status_detailed.xml
```diff
@@ -1,10 +1,13 @@
 <?xml version="1.0" encoding="utf-8"?>
 <RelativeLayout xmlns:android="http://schemas.android.com/apk/res/android"
     xmlns:app="http://schemas.android.com/apk/res-auto"
+    xmlns:sparkbutton="http://schemas.android.com/apk/res-auto"
     xmlns:tools="http://schemas.android.com/tools"
     android:id="@+id/status_container"
     android:layout_width="match_parent"
     android:layout_height="wrap_content"
+    android:clipChildren="false"
```
charlag: It still clips animation for me...

charlag: Okay, I see, before it was clipped on top and not it's clipped only at the bottom. Weird that it still does that but it's better ofc

connyduck: I guess we would need to set it on the containing view via code if we want no clipping at all

charlag: I think it's good as-is, we can improve it later

charlag:
Anyway, I didn't test it on the old devices because my laptop would play dead but otherwise it looks really nice - good job!
I approve!

PR #609, opened 2018-04-26
Add expand/collapse button for threads

charlag:
Button in the toolbar which expands/collapses all posts in a thread.

It's not really tested (and not tested at all on old devices because my laptop hangs even without en emulator) but it seems to work (even with rotation for some reason).

It tries to match web interface behavior.

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/ViewThreadActivity.java
+++ b/app/src/main/java/com/keylesspalace/tusky/ViewThreadActivity.java
@@ -35,8 +35,16 @@ import dagger.android.support.HasSupportFragmentInjector;
 
 public class ViewThreadActivity extends BaseActivity implements HasSupportFragmentInjector {
 
+    public static final int REVEAL_BUTTON_HIDDEN = 1;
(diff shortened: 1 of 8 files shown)


app/src/main/java/com/keylesspalace/tusky/ViewThreadActivity.java
```diff
@@ -35,9 +35,17 @@
 
 public class ViewThreadActivity extends BaseActivity implements HasSupportFragmentInjector {
 
+    public static final int REVEAL_BUTTON_HIDDEN = 1;
+    public static final int REVEAL_BUTTON_REVEAL = 2;
+    public static final int REVEAL_BUTTON_HIDE = 3;
```
Tak: Why not an enum?

charlag: Hard question.
Probably because it's such a small feature and it's pretty much local I didn't want to add an overhead of a class.

connyduck:
nice :3

PR #617, opened 2018-04-30
Add tests for search functionality in SFragment

Tak:


app/src/test/java/com/keylesspalace/tusky/SFragmentTest.kt
```diff
@@ -0,0 +1,301 @@
...
+            null
+    )
+    private val statusCallback = FakeSearchResults(status)
+
+    @Before
+    fun setup() {
+        fragment = FakeSFragment()
+
+        apiMock = Mockito.mock(MastodonApi::class.java)
+        `when`(apiMock.search(eq(accountQuery), ArgumentMatchers.anyBoolean())).thenReturn(accountCallback)
```
charlag: We should include mockito-kotlin finally...

charlag:
FIrst, I :heart: more tests, thank you a lot for doing this non-sexy work.

What I learned just recently is that if you want to test private methods/fields, then you're doing it wrong. We should test behavior, not implementation. Otherwise we have to maintain tests a lot.

Some things are maybe not tightly related to fragment: that regex e.g. Maybe we should move it somewhere and test separately? I'm not sure. We can test it as part of other things but maybe that's unwieldy.

Some things are easy to test without exposing private things: you don't need to expose `onBeginSearch`if `onViewUrl` is public anyway.

Some things are hard to test without exposing. Canceling is one of them. We could in theory Hide the sheet programmatically but we would need Robolectric for that.
That's why I prefer to put all the view logic in something like VIewModel/Presenter but it's not a priority for now.

Anyway, I believe if we're going to test it not by pulling private fields but by emulating failed network requests tests are going to be more robust and really trustworthy.

My only concern is the bottom sheet. If it's too unwieldy to set up a Robolectric sheet I would make a method `onBottomSheetClosed()` and put `@VisibleForTesting` on it.

I've found this article super useful.
http://blog.codepipes.com/testing/software-testing-antipatterns.html#anti-pattern-5---testing-internal-implementation

I hope my feedback is useful! Please, let's discuss what you're going to do before you make any future steps (otherwise I feel like I'm demanding changes).

Tak:
> if you want to test private methods/fields, then you're doing it wrong. We should test behavior, not implementation
I know that this is a widely held opinion; however, it's one with which I disagree in many cases.
I actually want to test the _decision mechanism_. Just testing the observable behavior is subject to too many things outside of my control (android software stack, mocks, etc.).
Additionally, things are often private just because there's no need to expose them. If we had kept the `SearchManager` interface after moving everything into `SFragment`, then all this API would be public anyway.

Regardless, all that stuff is just my opinion - if you don't want tests done like that in Tusky, I'm ok with that. (But I probably wouldn't be adding any more tests in the future unless there are clear guidelines...)

charlag:
I see your point!

I am not good enough in testing to provide clear guidelines, I'm sorry - just giving my feedback. Please, don't let me stop you from contributing.

I could argue that if you test things through class boundaries (e.g. network response) you're killing two birds with one stone - e.g. you would also test logic inside `onResponse()`.
If I moved such code to SearchManager I still would not check it's internal flags - I would check it's behavior.

I think value of these test tests is much more than potential burden of maintaining them - if we will notice some problems we can refactor them.

I didn't run them yet but I think they're really useful and we should merge them.
You can leave them as is if you think this is the best.

app/src/test/java/com/keylesspalace/tusky/SFragmentTest.kt
```diff
@@ -0,0 +1,301 @@
...
+        override fun execute(): Response<SearchResults> { TODO("not implemented") }
+        override fun request(): Request { TODO("not implemented") }
+    }
+
+    class FakeSFragment : SFragment {
+        var status: Status? = null
+        var accountId: String? = null
+        var url: String? = null
+
+        constructor(): super() {
```
charlag: No need for empty constructor in Kotlin. Change it to `init` please

app/src/test/java/com/keylesspalace/tusky/SFragmentTest.kt
```diff
@@ -0,0 +1,301 @@
...
+
+        override fun viewAccount(id: String?) {
+            accountId = id
+        }
+
+        override fun viewThread(status: Status?) {
+            this.status = status
+        }
+
+        override fun removeItem(position: Int) { TODO("not implemented") }
```
charlag: Android Studio will think that these are real TODOS and will show them in the TODO list. Maybe change it to `AssertionError` or something? Same for `FakeSearchResults`

app/src/test/java/com/keylesspalace/tusky/SFragmentTest.kt
```diff
@@ -0,0 +1,301 @@
...
+                "https://pleroma.foo.bar/objects/xabcdef-123-abcd-9876543/",
+                "https://pleroma.foo.bar/objects/xabcdef-123-abcd_9876543"
+        )
+
+        for(url in urls)
+            Assert.assertFalse("Matched invalid url $url", SFragment.looksLikeMastodonUrl(url));
+    }
+
+    @Test
+    fun beginEndSearch_setIsSearching() {
```
charlag: All your tests have multiple `assert`s. It may make it unclear which test case fails. I'd rather see them as separate tests in each case. This way we're also sure that one test doesn't affect another one (which is kinda really important)

Tak: Just to be clear, you'd be ok with me making these parameterized?
I.e. the url tests would all be a single parameterized test class with one test per url, and the search ones would be the same tests run on account, status, and non-mastodon urls.

PR #628, opened 2018-05-03
move bottom sheet from fragments to activities

### LINKED ISSUE #612: Scrolling in profiles does not pull up the header (in current master build)
So I just found a bug that if you go to a user page and try to scroll through the user's toots without raising the finger, instead of pushing the header image, etc. up, you'll just scroll thorugh that small part in which the toots are shown.  
This is not the case in the 1.7 build.
Here's a screencast of it (sorry for the bad quality 😅):
https://drive.google.com/open?id=1LLD5HkDAAvhCA60-8xKdYgClP0G4rZgb

* * * *
- Tusky Version: Current Master build (3dfe43dfb2b7712ae9eceff18c07b534b9f63d65)
- Android Version: 7.1.2
- Android Device: Moto Z2 Play
- Mastodon instance (if applicable): mastodon.xyz

connyduck:
closes #612 

seems to work fine and should also be a bit more efficient since there are less bottom sheets and CoordinatorLayouts inflated

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/AccountActivity.java
+++ b/app/src/main/java/com/keylesspalace/tusky/AccountActivity.java
@@ -61,8 +61,6 @@ import com.keylesspalace.tusky.util.ThemeUtils;
 import com.pkmmte.view.CircularImageView;
 import com.squareup.picasso.Picasso;
 
-import org.jetbrains.annotations.NotNull;
-
 import java.text.NumberFormat;
 import java.util.ArrayList;
 import java.util.Collections;
@@ -716,7 +714,7 @@ public final class AccountActivity extends BottomSheetActivity implements Action
         return dispatchingAndroidInjector;
     }
 
-    @NotNull
+    @NonNull
     @Override
     public MastodonApi getMastodonApi() {
         return mastodonApi;
--- a/app/src/main/java/com/keylesspalace/tusky/AccountActivity.java
+++ b/app/src/main/java/com/keylesspalace/tusky/AccountActivity.java
@@ -61,6 +61,8 @@ import com.keylesspalace.tusky.util.ThemeUtils;
 import com.pkmmte.view.CircularImageView;
 import com.squareup.picasso.Picasso;
 
+import org.jetbrains.annotations.NotNull;
+
 import java.text.NumberFormat;
 import java.util.ArrayList;
 import java.util.Collections;
@@ -75,7 +77,7 @@ import retrofit2.Call;
 import retrofit2.Callback;
 import retrofit2.Response;
 
-public final class AccountActivity extends BaseActivity implements ActionButtonActivity,
+public final class AccountActivity extends BottomSheetActivity implements ActionButtonActivity,
         HasSupportFragmentInjector {
     private static final String TAG = "AccountActivity"; // logging tag
 
@@ -331,8 +333,8 @@ public final class AccountActivity extends BaseActivity implements ActionButtonA
             }
 
             @Override
-            public void onViewURL(String url) {
-                LinkHelper.openLink(url, note.getContext());
+            public void onViewUrl(String url) {
+                viewUrl(url);
             }
         });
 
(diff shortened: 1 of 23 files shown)


connyduck:
this now also fixes the issue where mentions in account notes did not get openend in Tusky

what do you think?

app/src/main/java/com/keylesspalace/tusky/AccountActivity.java
```diff
@@ -61,6 +61,8 @@
 import com.pkmmte.view.CircularImageView;
 import com.squareup.picasso.Picasso;
 
+import org.jetbrains.annotations.NotNull;
```
charlag: Why not Android support one?

app/src/main/java/com/keylesspalace/tusky/BottomSheetActivity.kt
```diff
@@ -0,0 +1,198 @@
...
+
+    override fun onPostCreate(savedInstanceState: Bundle?) {
+        super.onPostCreate(savedInstanceState)
+
+        val bottomSheetLayout: LinearLayout = findViewById(R.id.item_status_bottom_sheet)
+            bottomSheet = BottomSheetBehavior.from(bottomSheetLayout)
+            bottomSheet.state = BottomSheetBehavior.STATE_HIDDEN
+            bottomSheet.setBottomSheetCallback(object : BottomSheetBehavior.BottomSheetCallback() {
+                override fun onStateChanged(bottomSheet: View, newState: Int) {
+                    when (newState) {
```
charlag: when -> if ?

app/src/main/java/com/keylesspalace/tusky/BottomSheetActivity.kt
```diff
@@ -0,0 +1,198 @@
...
+                        }
+                    }
+                }
+
+                override fun onSlide(bottomSheet: View, slideOffset: Float) {}
+            })
+
+    }
+
+    open fun viewUrl(url: String) {
```
charlag: Why methods are open? If it's for mocking, we use 'mockito-inline' which doesn't care

app/src/test/java/com/keylesspalace/tusky/BottomSheetActivityTest.kt
```diff
@@ -283,29 +283,32 @@ class SFragmentTest {
...
+    class FakeBottomSheetActivity(val api: MastodonApi) : BottomSheetActivity() {
+
         var status: Status? = null
         var accountId: String? = null
-        var url: String? = null
+        var link: String? = null
 
         init {
-            callList = mutableListOf()
+            bottomSheet = Mockito.mock(BottomSheetBehavior::class.java) as BottomSheetBehavior<LinearLayout>
```
charlag: I think we need to suppress this

charlag:
Also tests are failing because of that bottom sheet... I wonder if suppress will help

app/src/main/java/com/keylesspalace/tusky/AccountActivity.java
connyduck: sry probably autocomplete

app/src/main/java/com/keylesspalace/tusky/BottomSheetActivity.kt
connyduck: how can I do this without complicating the code a lot?

app/src/main/java/com/keylesspalace/tusky/BaseActivity.java
```diff
@@ -37,12 +37,17 @@
...
 
+import javax.inject.Inject;
+
 import retrofit2.Call;
 
 public abstract class BaseActivity extends AppCompatActivity {
 
     protected List<Call> callList;
 
+    @Inject
```
charlag: Huh? Does it work like that? I thought you can inject only in concrete classes?

app/src/main/java/com/keylesspalace/tusky/BottomSheetActivity.kt
charlag: Wait, I don't understand. You made them open for mocking, right?
You just do `when(activity.methodCall()).doReturn(restult)`. Nothing changes with them being open or not.
I feel like I don't understand something.

connyduck: no, the class is not mocked mith Mockito, but subclassed as FakeBottomSheetActivity

charlag: Ah, I see now, it's a poor man's Spy. We'll fix it sometime

Issue #573, opened 2018-04-14
Remaining characters are calculated wrongly


connyduck:
Links in Mastodon statuses are only counted as 23 characters, no matter how long they are. Tusky ignores this, which means you have less characters in Tusky than on Mastodon web.

charlag:
This is basically fixed by #629 right?

connyduck:
yes

PR #629, opened 2018-05-03
Behave like Mastodon web ui and only count URLs as 23 characters when composing

Tak:
Issue #573 

This also adds URL highlighting in the compose activity
Includes refactor of `SpanUtils` and test coverage for spanning and remaining character count


app/src/main/java/com/keylesspalace/tusky/ComposeActivity.java
```diff
@@ -771,12 +774,23 @@ public void onVisibilityChanged(@NonNull Status.Visibility visibility) {
...
     }
 
-    private void updateVisibleCharactersLeft() {
-        int charactersLeft = maximumTootCharacters - textEditor.length();
+    int calculateRemainingCharacters() {
+        int offset = 0;
+        URLSpan[] urlSpans = textEditor.getUrls();
+        if (urlSpans != null) {
+            for (URLSpan span : urlSpans) {
+                offset += Math.max(0, span.getURL().length() - MAXIMUM_URL_LENGTH);
```
charlag: Sorry, I don't get it - you subtract maximum toot length from actual link length?..

app/src/main/java/com/keylesspalace/tusky/util/SpanUtils.kt
```diff
@@ -0,0 +1,132 @@
...
+
+        private enum class FoundMatchType {
+            HTTP_URL,
+            HTTPS_URL,
+            TAG,
+            MENTION,
+        }
+
+        private class FindCharsResult {
+            lateinit var matchType: FoundMatchType
```
charlag: Why latenit? This could be even a data class I guess

charlag:
Awesome, thank you!
I'll try to check it today or tomorrow

app/src/test/java/com/keylesspalace/tusky/SpanUtilsTest.kt
```diff
@@ -0,0 +1,159 @@
...
+
+    class FakeSpannable(val text: String) : Spannable {
+        var spans = mutableListOf<BoundedSpan>()
+
+        override fun setSpan(what: Any?, start: Int, end: Int, flags: Int) {
+            spans.add(BoundedSpan(what, start, end))
+        }
+
+        override fun <T : Any?> getSpans(start: Int, end: Int, type: Class<T>?): Array<T> {
+            var tSpans = mutableListOf<T>()
```
charlag: You could
```kotlin
return spans.filter { it.start >= start && it.end <= end }
        .filterIsInstance(type)
        .let { ArrayList(it) }
        .toArray() as Array<T>
```

app/src/test/java/com/keylesspalace/tusky/SpanUtilsTest.kt
```diff
@@ -0,0 +1,159 @@
...
+        }
+
+        override fun toString(): String {
+            return text
+        }
+
+        override val length: Int
+            get() = text.length
+
+        class BoundedSpan(val span: Any?, val start: Int, val end: Int) { }
```
charlag: Extra field and braces

app/src/main/java/com/keylesspalace/tusky/util/SpanUtils.kt
```diff
@@ -0,0 +1,132 @@
...
+            var start: Int = -1
+            var end: Int = -1
+        }
+
+        private class PatternFinder {
```
charlag: A lot of quick-fixes for this class

app/src/main/java/com/keylesspalace/tusky/util/SpanUtils.kt
charlag: Okay, it is used in imperative code which is imperative for the sake of performance I guess but I would avoid lateinit anyway, if it's optional - let it be so

app/src/main/java/com/keylesspalace/tusky/util/SpanUtils.kt
```diff
@@ -0,0 +1,132 @@
...
+            for (spanClass in spanClasses) {
+                clearSpans(text, spanClass)
+            }
+
+            // Colour the mentions and hashtags.
+            val string = text.toString()
+            val length = text.length
+            var start = 0
+            var end = 0
+            while (end >= 0 && end < length && start >= 0) {
```
charlag: Studio proposed could of quick-fixes

app/src/main/java/com/keylesspalace/tusky/ComposeActivity.java
Tak: `MAXIMUM_URL_LENGTH` is the length that mastodon uses (23)

app/src/main/java/com/keylesspalace/tusky/util/SpanUtils.kt
Tak: I skipped most of the suggestions (e.g. convert lessthan/greaterthan to range) because this is a hot path and I didn't want to affect the performance too much.

charlag: Honestly, I didn't check the bytecode but it should be fine. I'm okay with you not changing to ranges

charlag:
So, would you apply quick-fixes for the classes and not for ranges or you would leave them as-is?
Also, I've noticed that we're a little more relaxed to the URLs than mastodon front-end. With our current regex you can get counted as URL without domain name.
Here's what Masto frontend does:
https://github.com/tootsuite/mastodon/commit/53b2b1b2389c0ea941c50d4a9c726539d808c384#diff-3844faa5b838d674cbd72e3c1a1b477dR5

Tak:
I can go back and audit the suggestions again; maybe I missed something.
I can adopt the regex from mastodon as well.

charlag:
Cool, thx :3

Tak:
Oh, but the link you posted is the _old_ implementation.
Here's the new one! :joy: 
https://github.com/tootsuite/mastodon/blob/master/app/javascript/mastodon/features/compose/util/url_regex.js

Tak:
What about keeping the simple/more permissive regex and validating with `URI.parse`?

app/src/main/java/com/keylesspalace/tusky/util/SpanUtils.kt
```diff
@@ -86,13 +79,21 @@ class SpanUtils {
...
                 result.start = matcher.start()
                 if (Character.isWhitespace(string.codePointAt(result.start))) {
                     ++result.start
                 }
-            } else {
-                result.end = -1
+                when(result.matchType) {
+                    FoundMatchType.HTTP_URL, FoundMatchType.HTTPS_URL -> {
+                        // Preliminary url patterns are fast/permissive, now we'll do full validation
+                        if (WEB_URL == null || // This doesn't get mocked for tests :-|
```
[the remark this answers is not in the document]
charlag: On the contrary, we can one implementation across all versions and our texts actually test things

app/src/test/java/com/keylesspalace/tusky/SpanUtilsTest.kt
```diff
@@ -93,34 +93,26 @@ class SpanUtilsTest {
...
-            var tSpans = mutableListOf<T>()
-            for (span in spans) {
-                val tSpan = span as T
-                if (tSpan != null && span.start >= start && span.end <= end) {
-                    tSpans.add(tSpan)
-                }
+            val matching = if (type == null) {
+                ArrayList<T>()
+            } else {
+                spans.filter ({ it.start >= start && it.end <= end && (it.span as T) != null })
```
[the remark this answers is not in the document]
charlag: `as` won't return you null, it will throw. Or I am wrong?

app/src/main/java/com/keylesspalace/tusky/util/SpanUtils.kt
```diff
@@ -0,0 +1,133 @@
...
+
+import android.text.Spannable
+import android.text.Spanned
+import android.text.style.CharacterStyle
+import android.text.style.ForegroundColorSpan
+import android.text.style.URLSpan
+import android.util.Patterns.WEB_URL
+import java.util.regex.Pattern
+
+class SpanUtils {
```
connyduck: This is Kotlin, not Java, no need for a class when everything is static anyway. You can remove class/companion object and @JvmStatic, reference the functions in Java as `SpanUtilsKt` afterwards.

connyduck:
I tested it on my crappy old device, performance is fine.
But I just noticed, mentions & urls have a different highlighting color on the light theme. I like the one of urls more, can you set that for all please?

PR #634, opened 2018-05-06
[RU] New strings for 1.8

MasterGroosha:
I'm not sure whether we should use word "статусы" (statuses) or "посты" (posts). Or how to mix them in the right way
Anyway, here are the new strings.


MasterGroosha:
I don't understand what's wrong with CircleCI test :(

charlag:
It's out of memory again 🙄
I've told Gradle not to eat too much but there's something wrong still...

charlag:
As long as it builds (and it did) I think your changed are fine.
We confuse statuses and posts and toots on English strings as well all the time too so...

PR #637, opened 2018-05-07
1.8 fixes

connyduck:
just some small things, mostly deleting unneeded stuff

noticed the dependency injection setup I did with BottomSheetActivity can be simplified
also notifications should cancel themselfes when clicked


charlag:
I wonder how it manages to inject into base classes... probably because it is Android injection and it does some things at runtime

connyduck:
no, injecting into the concrete class injects fields of the base class as well, have a look at the generated code

charlag:
Okay, I just remember it non-working (but maybe I'm wrong). I'm probably getting old, lol
Anyway, glad that it's simpler now, LGTM?

app/src/main/java/com/keylesspalace/tusky/BaseActivity.java
```diff
@@ -111,20 +112,16 @@ protected SharedPreferences getPrivatePreferences() {
         return getSharedPreferences(getString(R.string.preferences_file_key), Context.MODE_PRIVATE);
     }
 
-    protected boolean redirectIfNotLoggedIn() {
+    protected void redirectIfNotLoggedIn() {
         // This is very ugly but we cannot inject into parent class and injecting into every
```
charlag: Seems like this comment is not true anymore, huh?

app/src/main/java/com/keylesspalace/tusky/fragment/ViewThreadFragment.java
```diff
@@ -532,20 +532,14 @@ private void showCard(Card card) {
...
     private void updateRevealIcon() {
         ViewThreadActivity activity = ((ViewThreadActivity) getActivity());
         if (activity == null) return;
 
         boolean hasAnyWarnings = false;
         // Statuses are updated from the main thread so nothing should change while iterating
         for (int i = 0; i < statuses.size(); i++) {
-            if (statuses.get(i).getSpoilerText() != null
-                    && !statuses.get(i).getSpoilerText().isEmpty()) {
+            if (!TextUtils.isEmpty(statuses.get(i).getSpoilerText())) {
```
charlag: TextUtils is a platform class which will make things hard to test. Maybe use StringsKt or something?

PR #639, opened 2018-05-08
[CI] Build release apk's

charlag:
Adds additional step to CicleCI which build release version of the app and stores it in /apk folder in artifacts.

It would be nice if we could be them separate jobs (see Mastodon repo) so we could have tests feedback faster because with release builds it takes 2x longer to complete but that's not crucial.


charlag:
Commit history is somehow broken, don't mind it

charlag:
I've tried to install an APK from CI and it says that APK is malformed? Is it because it's release build but unsigned?

PR #640, opened 2018-05-09
Separate display name from notification summary text for favorite/boost/follow

Tak:
Fixes notification display when display name contains RTL text

Before
image

After
image


connyduck:
Sorry, but no. We are not adding two additional Views just to solve a relatively minor display bug.

There must be a better solution for this. 
Please try this: https://stackoverflow.com/questions/6302221/android-setting-with-textview-for-hebrew-text https://www.w3.org/TR/WCAG20-TECHS/H34.html

charlag:
Would it be okay if it was just one additional view?
We really should move from RelativeLayout to ConstraintLayout by the way...

Issue #636, opened 2018-05-07
Replies should default to the lowest visibility of the set default and the post being replied to


ashkitten:
I have my default visibility set to followers only, but when I reply to a post it inherits that post's privacy setting unlike the web client, which takes the lowest privacy setting of the two. (e.g. if the default is followers-only and the toot is public, the reply should default to followers-only)

* * * *
- Tusky Version: 1.7.0
- Android Version: 8.1.0
- Android Device: Google Pixel XL
- Mastodon instance (if applicable): dev.glitch.social

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

charlag:
Hmm, I think we had this algorithm but dropped it because it was confusing.
@connyduck do you remember?

connyduck:
I thought we do it like the web client? Seems like we missed this detail & we should change it. Thx for reporting!

connyduck:
fixed in #641

PR #641, opened 2018-05-09
When replying, set default post visibility to the most restrictive of : original post visibility, default visibility preference.

Tak:
Issue #636


connyduck:
Looks good, thank you.

PR #643, opened 2018-05-09
Add CI workflow to separate out test and build

technomancy:
As discussed in [this thread](https://birb.site/@charlag/99995654725177209)

Is that what you had in mind?


charlag:
I think this is it, thanks a lot!
But for some reason Circle doesn't detect this branch now?

technomancy:
There's a setting to build PRs which come from forks (vs just PRs from the same repo) that you might need to turn on within CircleCI; it's off by default for security reasons. Check under "build settings > advanced settings"

charlag:
Yes, this is true put check for example this PR
https://github.com/tuskyapp/Tusky/pull/641

it builds and it comes from another repo (sorry for being unclear)

charlag:
I double checked it and "Build forked pull requests" is selected so either Circle won't change workflow from forks or there's something wrong with config.yml

charlag:
screenshot from 2018-05-11 00-29-46
screenshot from 2018-05-11 00-29-52
okay, there's something...
"We weren't able to start this workflow.
The following jobs are unreachable: build
For more examples see the Workflows documentation."

charlag:
I suspect that's because of the
```
     - build:
         requires:
           - build
```
I'm not sure but you probably meant `test` instead of the second `build`?

technomancy:
Haha; oops. Made a last-minute change and didn't update consistently; my bad.

charlag:
I am just not really familiar with this and I wasn't sure that what I was saying makes sense, thanks!

charlag:
Would you mind if me or you removed only-master filter? this way it will trigger it for this branch. I think I would prefer it to build APKs for forks as well

technomancy:
Do you think it's a good idea to have untrusted APKs mixed in with the ones from the official master branch?

charlag:
Well these are not release builds anyway (not going to be signed with release key) - they're just needed to quickly test PRs. We have no plans to deploy and distribute from the CI for now, it's all testing

charlag:
I wasn't sure how to merge our 2 PRs properly so I merged this one into #639, hope you don't mind.
Thank you a lot for your help, it works as intended!

technomancy:
I see; that sounds reasonable. Glad I could help.

PR #647, opened 2018-05-10
some Optimizations for tablets

### LINKED ISSUE #421: Tablet optimization
Thats a huge Tasks. But I think it should be done eventually so Tusky is really awesome on all devices.

connyduck:
resolves #421 
Step one: Show ComposeActivity as a dialog on tablets!

 vs.

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -15,7 +15,7 @@
         android:label="@string/app_name"
         android:name=".TuskyApplication"
         android:supportsRtl="true"
-        android:theme="@style/AppTheme">
+        android:theme="@style/TuskyTheme">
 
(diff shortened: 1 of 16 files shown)


connyduck:
Step two: limit timeline width on larger devices. Onfurtunately, I found no way of doing this without an additional `FrameLayout`.

 vs.

trwnh:
Rather than limiting timeline width, perhaps try for a multi-column approach? If not that, then maybe figure out a better use for the whitespace around the edges (putting notifications on the side of the timelines?)

connyduck:
I will only do stuff that can be achieved with layout/style changes because I do not want to add additional code complexity or spend much time on this. Yes, multi column would be awesome, but its too big of a change for now.

PR #659, opened 2018-05-23
Isolate usernames when formatting, to improve interaction of RTL usernames with LTR locales (and vice versa)

Tak:
Fixes notification display when display name contains RTL text, redux :duck: 

Before
image

After
image


connyduck:
Oh very nice, I did not know about BidiFormatter before.

Can you fix this for system notifications as well? should be in `NotificationHelper` line 495, `joinNames` and `titleForType` methods.

`BidiFormatter.getInstance()` seems to create a new Object everytime it is called, I think it is a good idea to cache the instance in an static field.

charlag:
Wow, cool find!
I'd maybe inject it or save it just as instance field, but maybe not static. It may impose difficulties when changing the language or when testing. Luckily it's part of the support lib and not the system itself so it won't bark during testing.

Tak:
Hm, maybe it makes more sense to wrap the `displayName` getter on `Account`?

charlag:
I'm not sure, it's more of a view concern than data

PR #660, opened 2018-05-26
Add smaller and larger font options (fix #657)

### LINKED ISSUE #657: Add a "Smaller" and a "Larger" font size option
These have both been requested quite often.

charlag:
No translations yet. Maybe it breaks defaults. Did I do the right thing at all?...


connyduck:
Almost 😅
I was thinking of Smaller - Small - Medium - Large - Larger
but maybe we should call it Smallest - Small - Medium - Large - Largest if it is unclear?

the new settings should be 10-12-14dp and 18-20-22dp
Or should we use sp? No idea what is better

Comment out other translations, or it will break for them

charlag:
Ah, okay, no problems, maybe it will be event better name. Will change today later.
Sp should be better.

charlag:
Seems to be according to the plan now :wink:

PR #663, opened 2018-05-26
Timeline cache

charlag:
(this is branched from the #562 )
This is PR aims to add kind of a an offline cache to the timelines. Something like Mastalab and Pinafore have.
It is a *super* tricky topic (cache invalidation, you know) but the potential benefit is huge.
I was surprised that I was able to make this initial implementation in a pretty efficient way (no N+1 requests, two JOINs).
The hardest part is reblogs. The easy way would be to seialize reblogged toot but I think I've came up with something much more efficient, see `TimelineStatusEntity`.

We should thoughtfully discuss caching strategy. [There's a wonderfully illustrated explanation on the topic](https://jakearchibald.com/2014/offline-cookbook/#serving-suggestions-responding-to-requests). I'd say we use "cache than network". It would updating timeline two times but that's the hardest part, actual request race is trivial with Rx.

Another thing is that we should stop relying on headers for timeline paging because it's incompatible with offline caching.

Anyway, it's not ready but I would like to hear your feedback on this.


charlag:
Okay, let's track things
- [x] Tables for statuses and accounts
- [x] Query for all of it
- [x] Conversion between DB entities and objects
- [x] Remembering timeline position and restoring it
- [ ] Correct infinite loading and gap handling (still glitchy/works not in all cases)
- [ ] Multiple accounts (it should work but it's not really tested yet)
- [ ] Handling deleted statuses (maybe in the future)

mal0ki:
There is a new PR on this, rebased by charlag 
https://github.com/tuskyapp/Tusky/pull/809

PR #681, opened 2018-06-10
Do not scroll to top on timeline refresh

charlag:
Unconditionally scroll timeline a bit on insertion on the top.
Not sure if I should use dp here.

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/fragment/TimelineFragment.java
+++ b/app/src/main/java/com/keylesspalace/tusky/fragment/TimelineFragment.java
@@ -78,6 +78,7 @@ import java.util.regex.Pattern;
 
 import javax.inject.Inject;
 
+import at.connyduck.sparkbutton.helpers.Utils;
 import io.reactivex.android.schedulers.AndroidSchedulers;
 import retrofit2.Call;
 import retrofit2.Callback;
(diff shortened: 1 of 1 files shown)


connyduck:
yes please use dp, 40px is a lot on low res devices and almost nothing on hi res devices

app/src/main/java/com/keylesspalace/tusky/fragment/TimelineFragment.java
```diff
@@ -1055,11 +1056,8 @@ private void updateAdapter() {
         @Override
         public void onInserted(int position, int count) {
             adapter.notifyItemRangeInserted(position, count);
-            if (position == 0
-                    && layoutManager.findFirstVisibleItemPosition() == 0
-                    && (swipeRefreshLayout.getVisibility() == View.VISIBLE
-                    || progressBar.getVisibility() == View.VISIBLE)) {
-                recyclerView.post(() -> layoutManager.scrollToPosition(0));
+            if (position == 0) {
+                recyclerView.scrollBy(0, Utils.dpToPx(Objects.requireNonNull(getContext()), -30));
```
connyduck: you could just use requireContext() instead of wrapping it into requireNonNull

charlag: Sure thing

PR #684, opened 2018-06-14
Fix account media progress indicator

charlag:
Things were added in the layout for timelines but I forgot to change AccountMediaFragment. Should be fixed now.


connyduck:
can you switch to Kotlin Android Extensions instead of `findViewById`?

PR #686, opened 2018-06-16
fix media resizing

### LINKED ISSUE #677: Image resizing
When I try to post a toot containing an image, the image upload progress bar gets all the way to the end, then I get the error: "The upload failed."

This appears to be related to mastodon.social limits on image upload sizes.  In Mastalab, I can configure it to resize to ex. 2mb, which works (even though it looks awful).  Tusky doesn't appear to have a similar setting, and it should.  The only way I can post an image with Tusky is to resize it in an image editor, which is so annoying that I just don't post at all.

* * * *
- Tusky Version: 1.8.0
- Android Version: 8.0.0
- Android Device: Sony Xperia XZ1 Compact

Possibly related: #223

connyduck:
fix #677

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/ComposeActivity.java
+++ b/app/src/main/java/com/keylesspalace/tusky/ComposeActivity.java
@@ -85,7 +85,6 @@ import com.keylesspalace.tusky.adapter.EmojiAdapter;
 import com.keylesspalace.tusky.adapter.MentionAutoCompleteAdapter;
 import com.keylesspalace.tusky.adapter.OnEmojiSelectedListener;
 import com.keylesspalace.tusky.db.AccountEntity;
-import com.keylesspalace.tusky.db.AccountManager;
 import com.keylesspalace.tusky.db.AppDatabase;
 import com.keylesspalace.tusky.db.InstanceEntity;
 import com.keylesspalace.tusky.di.Injectable;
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/keylesspalace/tusky/ComposeActivity.java
```diff
@@ -1063,7 +1063,9 @@ private void addMediaToQueue(@Nullable String id, QueuedMedia.Type type, Bitmap
 
         if (item.readyStage != QueuedMedia.ReadyStage.UPLOADED) {
             waitForMediaLatch.countUp();
-            if (mediaSize > STATUS_MEDIA_SIZE_LIMIT && type == QueuedMedia.Type.IMAGE) {
+
+            if (type == QueuedMedia.Type.IMAGE &&
+                    (mediaSize > STATUS_MEDIA_SIZE_LIMIT || MediaUtils.getImageSquarePixels(getContentResolver(), item.uri) > STATUS_MEDIA_PIXEL_SIZE_LIMIT)) {
```
charlag: I don't quite understand why we need to query media size again tbh

connyduck: There are 2 limits: filesize (8MB) and pixel count (16 mega pixels) and thats two totally different things. Files with less than 8MB but over 16 megapixels caused the bug.

app/src/main/java/com/keylesspalace/tusky/util/MediaUtils.java
```diff
@@ -126,4 +126,21 @@ public static Bitmap getVideoThumbnail(Context context, Uri uri, @Px int thumbna
         source.recycle();
         return bitmap;
     }
+
+    public static long getImageSquarePixels(ContentResolver contentResolver, Uri uri) {
+        InputStream input;
+        try {
+            input = contentResolver.openInputStream(uri);
+        } catch (FileNotFoundException e) {
+            return -1;
```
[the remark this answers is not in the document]
connyduck: yes you are right

Issue #631, opened 2018-05-06
Enable custom emojis in profiles (notes, field values, display names)


charlag:
See https://github.com/tootsuite/mastodon/pull/7374

connyduck:
Custom emoji support should now be complete, except for users own name in the Drawer. That one is tricky because it is not possible with the current Drawer api. I will look at it for the next release, maybe we need to subclass something or send a PR to the drawer library.

connyduck:
I think we have to cache an accounts emojis in database in order to have them display correctly in the side drawer 🤔

charlag:
You mean shortcode -> url?

connyduck:
Yes. Or we could use the cache we use for the emoji keyboard. But that contains ALL emojis of an instance and that could make `CustomEmojiHelper.emojifyText` really slow 🤔

PR #687, opened 2018-06-18
improve custom emoji support

connyduck:
see #631 

Except for ones own username in the drawer, that is not possible with the Api of the drawer. So we probably have to subclass something and/or send a PR to the drawer lib, but lets save that for another relase.


charlag:
Didn't launch but changes LGTM

PR #694, opened 2018-06-26
Updated pt-rBR (Brazilian Portuguese) translation

contraexemplo:


contraexemplo:
Seems like Stringlate got a lot file changes when automatically creating this PR. Should I close it and open a new one with my modifications only? Or can you are able to clean this mess that I made (sorry!)?

connyduck:
Can you take the strings.xml file and make a PR with just that file? I have no idea what stringlate is doing here

PR #696, opened 2018-06-27
🌎: Updated Brazilian Portuguese translation

contraexemplo:
- Added missing strings and their translations.


connyduck:
Thank you very much!

PR #704, opened 2018-07-02
Fix crash on receiving event in timeline

charlag:
As per this stack trace
https://gist.github.com/connyduck/fec040ac63c3615f9ee8e6c0758686bc
it seems like we're trying to handle event after `onCreate` but before `onCreateView` and we don't have a kind at that moment of time yet. So it's hard to reproduce (and I didn't really but I had this crash I think) but I see no other reason.


connyduck:
Yes, thats also the conclusion I came to. Thanks!

PR #711, opened 2018-07-10
Updated sv (Swedish) translation

iah:


iah:
Is this right?

PR #722, opened 2018-07-23
Improve media resizing

connyduck:
This should fix OOMs when uploading images: Only sampled bitmaps are loaded into memory when creating thumbnails, so OOMs should be far less likely, and it will no longer crash in case there is still not enough memory.
Also fixes the problem where thumbnails were super blurry.

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/util/MediaUtils.java
+++ b/app/src/main/java/com/keylesspalace/tusky/util/MediaUtils.java
@@ -27,6 +27,7 @@ import android.provider.OpenableColumns;
 import android.support.annotation.NonNull;
 import android.support.annotation.Nullable;
 import android.support.annotation.Px;
+import android.util.Log;
 
 import java.io.ByteArrayOutputStream;
 import java.io.FileNotFoundException;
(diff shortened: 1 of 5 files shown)


app/src/main/java/com/keylesspalace/tusky/util/MediaUtils.java
```diff
@@ -88,30 +88,51 @@ public static long getMediaSize(@NonNull ContentResolver contentResolver, @Nulla
...
+
+        // Decode bitmap with inSampleSize set
+        options.inJustDecodeBounds = false;
+        try {
+            stream = contentResolver.openInputStream(uri);
+            return BitmapFactory.decodeStream(stream, null, options);
+        } catch (FileNotFoundException e) {
+            Log.w(TAG, e);
+            return null;
+        } catch (OutOfMemoryError e) {
```
charlag: As far as I know, trying to catch OOM is pointless because nothing else can be done

charlag: I am kinda torn on this one
[Here](https://stackoverflow.com/questions/2679330/catching-java-lang-outofmemoryerror) people say it's only for cleanup/
[But here](https://stackoverflow.com/questions/7138645/catching-outofmemoryerror-in-decoding-bitmap#7138785) people say it's okay when decoding Bitmap
So I guess this is fine?

connyduck: Yes its fine, I use this in multiple apps. And I tested it with my slowest phone and an 70 megapixel photo xD

Thx for reviewing!

Issue #669, opened 2018-05-31
Feature request: open avatars to see them in bigger size


moshpirit:
Right now we can see avatars and that's nice but if we go (on web UI) to a profile an copy the avatar's image URL we can see in another tab the avatar in a better resolution.

 It's a cool feature I miss from GNU social that would be nice to have on Mastodon (I'll check if this is also requested on Mastodon's web UI)

charlag:
It's actually a pretty good idea, thanks!

PR #724, opened 2018-07-23
add possibility to view avatars fullscreen

connyduck:
Closes https://github.com/tuskyapp/Tusky/issues/669

Not my prettiest code, but does the job.
I would love to have a ripple effect on the avatar, but no idea how to achieve that 🤔

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/AccountActivity.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/AccountActivity.kt
@@ -146,8 +146,6 @@ class AccountActivity : BottomSheetActivity(), ActionButtonActivity, HasSupportF
         val intent = intent
         accountId = intent.getStringExtra(KEY_ACCOUNT_ID)
 
-        loadedAccount = null
-
         // set toolbar top margin according to system window insets
         ViewCompat.setOnApplyWindowInsetsListener(accountCoordinatorLayout) { _, insets ->
             val top = insets.systemWindowInsetTop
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/keylesspalace/tusky/AccountActivity.kt
```diff
@@ -321,6 +322,15 @@ class AccountActivity : BottomSheetActivity(), ActionButtonActivity, HasSupportF
                     .load(account.header)
                     .into(accountHeaderImageView)
 
+            accountAvatarImageView.setOnClickListener {
+                val intent = ViewMediaActivity.newAvatarIntent(it.context, account.avatar)
+
+                ViewCompat.setTransitionName(it, account.avatar)
```
charlag: IMHO this lambda is too big to use `it`, it takes some effort to understand what it means

app/src/main/java/com/keylesspalace/tusky/AccountActivity.kt
```diff
@@ -321,6 +322,15 @@ class AccountActivity : BottomSheetActivity(), ActionButtonActivity, HasSupportF
...
+                val options = ActivityOptionsCompat.makeSceneTransitionAnimation(this@AccountActivity, it, account.avatar)
```
charlag: You don't need a `this` qualifier here, if I understand correctly

connyduck:
thx @charlag

moshpirit:
In what release will we see this feature? it's not in F-droid v. 2.3

connyduck:
It will be in Tusky 3. You can try it in the nightly right now: https://play.google.com/store/apps/details?id=com.keylesspalace.tusky.test

moshpirit:
Great! Thanks a lot!

PR #726, opened 2018-07-24
Add black theme

qwazix:
On amoled screens a completely black theme helps save battery, besides looking awesome!

I made this for my own use but I hope others could find it useful, so I hope you might accept the PR.


connyduck:
To be honest, I am really against another theme in Tusky. The main reason is that it adds significant overhead to further development because every new feature has to be implemented and tested with all themes.

app/src/main/res/values/strings.xml
```diff
@@ -176,6 +176,7 @@
     <string-array name="app_theme_names">
         <item>Dark</item>
         <item>Light</item>
+        <item>Black</item>
         <item>Automatic at sunset</item>
     </string-array>
```
connyduck: all translations need to be updated or this will only work in English

connyduck:
And you need to set the theme in LoginActivity as well (it does not inherit from BaseActivity)

qwazix:
> there is no TuskyDialogActivityBlackTheme in values-large so the compose activity is no dialog with the black theme

I can add a TuskyDialogActivityBlackTheme but can you explain what you mean with no dialog? The compose activity looks the same to me in both cases with the difference of the background color.

>  I think you can have the black theme inherit from the dark theme and remove all attributes that are the same

Yes, I will do that. It's a better approach. I will try to only change the background and action bar color.

>  The line above the nav bar looks a bit strange (Android 8.1+ only)

Ok I'll have a look at it

>    Im not sure if its a good idea to have the status bar in a lighter color than actionbars (all activities except the main one)
It's not. I was going for the same color as the action bar like the main activity but it slipped. I was pondering making it all black but in that case when you scroll up the text is cut off without an obvious reason so I opted for a slightly lighter color, what do you think? Twidere has it all black.

screenshot_twidere_20180726-103208

> Can you support the theme after it is merged?

I'd love to

qwazix:
I fixed all the issues except for the login activity. Actually I did change it but it's ugly and the webview shows the dark grey/blue color (on the instance) so I think it's better to keep the login activity in dark mode and avoid black.

If you still prefer it tell me and I'll commit the change

screenshot_1532686331

I did the translations as best I could with Google Translate and Wikipedia, but of course I am not sure if everything is fine.

connyduck:
I think black background works on the LoginActivity.

qwazix:
Done

connyduck:
Looks good, thank you! 
I will contact you on Mastodon in case any problems with the theme arise.

qwazix:
Feel free and thanks. Proud to be even a small part of this.

PR #727, opened 2018-07-24
try to fix Circle CI

connyduck:
lets see if this does the trick

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -9,7 +9,7 @@
 
 # Specifies the JVM arguments used for the daemon process.
 # The setting is particularly useful for tweaking memory settings.
-org.gradle.jvmargs=-Xmx4096m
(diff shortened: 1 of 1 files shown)


connyduck:
looks like it does not help?

charlag:
It may still fail because it cannot find APK because of the flavor?

charlag:
No, sorry, it's 137

charlag:
We should also look for an APK in another folder if we build flavor

charlag:
It was buildc correctly last time (by accident I guess) but
`Unable to open 'app/build/outputs/apk/release/app-release-unsigned.apk' as zip archive`
because the folder should be different (this is a signing step before your last change of the destination)

charlag:
Even tests failed with 137 so I restarted it but wth

PR #729, opened 2018-07-25
fix bug where home timeline was never showing "load more" button

connyduck:
if we dont send the number of toots we want to get to the api, the check if we actually got that number does not work.

(Im pretty sure the timeline logic has some other problems as well, but this fixes this annoying problem at least most of the time)


charlag:
LGTM, I'm not sure why we didn't do this before

PR #731, opened 2018-07-26
work around crash on Android 8

connyduck:
Seeing this crash in Play Store logs & getting it reported by multiple people. 

to reproduce, paste the following text into the compose TextEditor while custom emoji style is active, then add somemthing between the second emoji and the c -> 💥

> now Im sure that it's true 🎶
🎵cause think you're so good
and I'm nothing like you ✨

Its a platform bug, tracked by Google here: https://issuetracker.google.com/issues/67102093
Only crashes in production apk.

It does not crash anymore with this fix, I hope it doesnt have any side effects >_<

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/ComposeActivity.java
+++ b/app/src/main/java/com/keylesspalace/tusky/ComposeActivity.java
@@ -31,6 +31,7 @@ import android.graphics.PorterDuff;
 import android.graphics.drawable.Drawable;
 import android.net.Uri;
 import android.os.AsyncTask;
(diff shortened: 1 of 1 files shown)


Tak:
Should a similar workaround be applied to any other text editors?

connyduck:
Its the only subclassed `AppCompatMultiAutoCompleteTextView` Tusky has, it has the heaviest usage of Text styling and is the only place where users enter longer text. So I would leave it like this for now and only fix other `EditTexts` if crashes are getting reported.

PR #733, opened 2018-07-28
Update strings.xml

silkevicious:
added missing strings manually


connyduck:
looks good, thnak you!

PR #735, opened 2018-07-28
Added the string-arrays etc.

iah:
This should to it.


connyduck:
Ok perfect!
Thank you very much!

PR #737, opened 2018-07-28
Custom emojis in drawer

### LINKED ISSUE #631: Enable custom emojis in profiles (notes, field values, display names)

connyduck:
closes #631

- this is now possible since my PR was merged https://github.com/mikepenz/MaterialDrawer/pull/2333
- I did not want to mess with the internals of the drawer, so I invalidate the whole header, as @C1710 suggested, works like a charm
- while measuring the perf impact this has, I found it is not much worse than before (some additional time spent in json de/serializing)
- but I found updating profile drawer info is quite expensive, so I made sure it is only updated once and not (2*number of profiles) times.

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/MainActivity.java
+++ b/app/src/main/java/com/keylesspalace/tusky/MainActivity.java
@@ -511,7 +511,7 @@ public final class MainActivity extends BottomSheetActivity implements ActionBut
             CharSequence emojifiedName = CustomEmojiHelper.emojifyString(acc.getDisplayName(), acc.getEmojis(), headerResult.getView());
             emojifiedName = EmojiCompat.get().process(emojifiedName);
 
-            profiles.add(
+            profiles.add(0,
                     new ProfileDrawerItem()
                             .withName(emojifiedName)
                             .withIcon(acc.getProfilePictureUrl())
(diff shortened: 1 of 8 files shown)


connyduck:
whoops that broke account switching

app/src/main/java/com/keylesspalace/tusky/db/Converters.kt
```diff
@@ -0,0 +1,34 @@
...
+import android.arch.persistence.room.TypeConverter
+import com.google.gson.Gson
+import com.google.gson.reflect.TypeToken
+import com.keylesspalace.tusky.entity.Emoji
+
+class Converters {
+
+    @TypeConverter
+    fun jsonToEmojiList(emojiListJson: String?): List<Emoji>? {
+        return Gson().fromJson(emojiListJson, object : TypeToken<List<Emoji>>() {}.type)
```
charlag: Creating GSON every time is suboptimal, it won't be able to cache anything

app/src/main/java/com/keylesspalace/tusky/MainActivity.java
```diff
@@ -502,24 +499,28 @@ private void updateProfiles() {
...
 
-        //remove profiles before adding them again to avoid duplicates
-        List<IProfile> profiles = new ArrayList<>(headerResult.getProfiles());
-        for (IProfile profile : profiles) {
-            if (profile.getIdentifier() != DRAWER_ITEM_ADD_ACCOUNT) {
-                headerResult.removeProfile(profile);
+        // reuse the already existing "add account" item
+        List<IProfile> profiles = new ArrayList<>(allAccounts.size()+1);
+        for (IProfile profile: headerResult.getProfiles()) {
+            if (profile.getIdentifier() == DRAWER_ITEM_ADD_ACCOUNT) {
```
charlag: You could `break` after finding the right profile but shouldn't it be always the last one?

PR #741, opened 2018-08-02
add more aggressive proguard config

connyduck:
This is a more aggressive proguard config than the default optimizing Android config. 
Specifically, it turns on the `code/simplification/arithmetic` optimization and makes one more pass.
Shrinks the apk about 15kb. I could not find any problems with this config, and I bet it feels a tad faster than before.
I will be deploying this to nightly and merge it if no problems occur.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -28,7 +28,7 @@ android {
         release {
             minifyEnabled true
             shrinkResources true
-            proguardFiles getDefaultProguardFile('proguard-android-optimize.txt'), 'proguard-rules.pro'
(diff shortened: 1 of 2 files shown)


connyduck:
Probably cant merge it because it takes way longer to build and exceeds bitrise build time 🙄

connyduck:
seems like there are no issues with this config

charlag:
I don't know if it's worth it - you say that the benefit is minimal but build time is higher. It's only for relese builds tho :thinking:

Issue #745, opened 2018-08-10
AccountActivity enhancements


connyduck:
Just some small things to improve the AccountActivity

- [ ] add pull to refresh to the header, remove it from tabs
- [ ] use the "follow" button as an "unblock" button on the profiles of blocked users
- [ ] add an icon to the profiles of muted users that can be clicked to unmute the user
- [ ] mark suspended profiles and profiles that have blocked the current user
- [x] improve the layout of the AccountFields
- [ ] dont remove the tabs from "moved" profiles

pandasoft0:
@connyduck Does anybody work on this issue? I would prefer to start to work on this feature if nobody work on it.

connyduck:
there has been some work done in #755, but no, nobody is working on it, feel free to pick it up

pandasoft0:
@connyduck i have some questions about this task
1. "add pull to refresh to the header, remove it from tabs
Should app refresh a full content of the account page on swipe down (include all tabs) or just an account info?
2. "mark suspended profiles and profiles that have blocked the current user"
Any ideas how can i get "suspended" or "blocked by user" statuses? I didn't find this info at the API responses
3. "dont remove the tabs from "moved" profiles" 
Sorry, i am newbie at Mastodon. Could you please say me how to do the "moved" status for the profile?

connyduck:
1. It should refresh everything including tabs, otherwise it won't make sense to move it
2. no but should be easy to try out, maybe with the help of an admin
3. My old account is a moved one, for example https://mastodon.social/@ConnyDuck

PR #755, opened 2018-08-14
Work on AccountActivity enhancements #745

eleete0712:
Current state:
- [ ] add pull to refresh to the header, remove it from tabs
- [x] use the "follow" button as an "unblock" button on the profiles of blocked users
- [x] add an icon to the profiles of muted users that can be clicked to unmute the user
 - Actually mute/unmute button (always visible). What do you think?
- [ ] mark suspended profiles and profiles that have blocked the current user
- [ ] improve the layout of the AccountFields
- [x] dont remove the tabs from "moved" profiles
 - General question: removing "moved" status from profile doesn't reset "moved" status when others (from other instances?) look at profile. Is that a Mastodon bug?

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/AccountActivity.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/AccountActivity.kt
@@ -379,6 +379,10 @@ class AccountActivity : BottomSheetActivity(), ActionButtonActivity, HasSupportF
                     startActivityForResult(intent, EDIT_ACCOUNT)
                     return@setOnClickListener
                 }
+                if (blocking) {
+                    viewModel.changeBlockState(accountId)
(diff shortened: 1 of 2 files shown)


connyduck:
> Actually mute/unmute button (always visible). What do you think?

Good idea!



On small screens the mute button slides below the avatar :/


> General question: removing "moved" status from profile doesn't reset "moved" status when others (from other instances?) look at profile. Is that a Mastodon bug?

Yes, this is the reason why we need to activate the tabs again, some people use a account that is marked as moved.

charlag:
@eleete0712 hi! Do you still work on this?

connyduck:
-> https://github.com/tuskyapp/Tusky/pull/1196

PR #757, opened 2018-08-15
add new elephant friends

connyduck:


app/src/main/res/values/colors.xml
```diff
@@ -84,4 +84,15 @@
 
     <color name="compose_media_visible_button_disabled_blue">#8c2b90d9</color>
 
+    <!-- colors used in the elpahnt friend drawables -->
```
digitalcircuit: **Minor:** `elpahnt` → `elephant`

connyduck: thx

PR #759, opened 2018-08-15
[RU] Language update 20180815

MasterGroosha:
Added new licence string, added metadata strings.

Upd: From what I see there are some "extra" lines, so I think this PR should be squashed into 1 commit, right?


connyduck:
Thank you vrey much! <3

PR #766, opened 2018-08-16
Show absolute time on timelines

kyori19:
Add an option in the settings to show the absolute time of every user’s post on timeline instead of relative time, as there frequently happens time delay if we use current system depending on refresh time of the timeline. This PR is made for those users who want to know the time by the default time stamp.
c23388134cd7af55
7f622e30147c61fd

Same as for now, we can use relative time displaying by disable this settings. See the attached image:
e3f677a4fd7ac0aa

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/adapter/NotificationsAdapter.java
+++ b/app/src/main/java/com/keylesspalace/tusky/adapter/NotificationsAdapter.java
@@ -16,12 +16,10 @@
 package com.keylesspalace.tusky.adapter;
 
 import android.content.Context;
-import android.content.SharedPreferences;
 import android.graphics.Color;
 import android.graphics.PorterDuff;
 import android.graphics.Typeface;
 import android.graphics.drawable.Drawable;
-import android.preference.PreferenceManager;
 import android.support.annotation.NonNull;
 import android.support.annotation.Nullable;
 import android.support.v4.content.ContextCompat;
@@ -70,6 +68,7 @@ public class NotificationsAdapter extends RecyclerView.Adapter {
     private NotificationActionListener notificationActionListener;
     private FooterViewHolder.State footerState;
     private boolean mediaPreviewEnabled;
+    private boolean useAbsoluteTime;
     private BidiFormatter bidiFormatter;
 
     public NotificationsAdapter(StatusActionListener statusListener,
@@ -80,6 +79,7 @@ public class NotificationsAdapter extends RecyclerView.Adapter {
         this.notificationActionListener = notificationActionListener;
         footerState = FooterViewHolder.State.END;
         mediaPreviewEnabled = true;
+        useAbsoluteTime = false;
         bidiFormatter = BidiFormatter.getInstance();
     }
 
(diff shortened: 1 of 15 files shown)


kyori19:
@connyduck 
I made some changes. Check please.

app/src/main/java/com/keylesspalace/tusky/adapter/NotificationsAdapter.java
```diff
@@ -365,26 +377,40 @@ private void setUsername(String name) {
...
-            if (createdAt != null) {
-                long then = createdAt.getTime();
-                long now = new Date().getTime();
-                readout = DateUtils.getRelativeTimeSpanString(timestampInfo.getContext(), then, now);
-                readoutAloud = android.text.format.DateUtils.getRelativeTimeSpanString(then, now,
-                        android.text.format.DateUtils.SECOND_IN_MILLIS,
-                        android.text.format.DateUtils.FORMAT_ABBREV_RELATIVE);
+        protected void setCreatedAt(@Nullable Date createdAt) {
+            if (useAbsoluteTime) {
+                String time = "ERROR!";
```
connyduck: the String doesnt need to be initialized here

app/src/main/java/com/keylesspalace/tusky/adapter/NotificationsAdapter.java
```diff
@@ -365,26 +377,40 @@ private void setUsername(String name) {
...
+                if (createdAt != null) {
+                    SimpleDateFormat sdf;
+                    if (new Date().getTime() - createdAt.getTime() > 86400000L) {
+                        sdf = new SimpleDateFormat("MM/dd HH:mm:ss", Locale.getDefault());
+                    } else {
+                        sdf = new SimpleDateFormat("HH:mm:ss", Locale.getDefault());
```
connyduck: make the `SimpleDateFormat`s class members then they dont need to be created for every timestamp. (Like the `BidiFormatter`)

app/src/main/java/com/keylesspalace/tusky/adapter/NotificationsAdapter.java
```diff
@@ -365,26 +377,40 @@ private void setUsername(String name) {
...
+                    }
+                    time = sdf.format(createdAt);
+                    timestampInfo.setText(time);
+                }
```
connyduck: there is no `else` branch for when the `createdAt` is null (just set the timestamp to ? or something)

app/src/main/java/com/keylesspalace/tusky/adapter/StatusBaseViewHolder.java
```diff
@@ -126,25 +132,39 @@ void setAvatar(String url, @Nullable String rebloggedUrl) {
     }
 
     protected void setCreatedAt(@Nullable Date createdAt) {
```
connyduck: same as in `NotificationsAdapter`

kyori19:
@connyduck Fixed problems.

connyduck:
Almost! Use two `SimpleDateFormat` members and remove the `passed1day` and the it should be perfect. 
pseudocode:
```
private SimpleDateFormat sameDayFormat
private SimpleDateFormat differentDayFormat   

....

if (new Date().getTime() - createdAt.getTime() > 86400000L) {
   time = differentDayFormat.format(createdAt);
} else {
   time = sameDayFormat.format(createdAt)
    }
```

also thinking of it `System.currentTimeMillis()` should be marginally faster than `new Date().getTime()` because it doesnt create a new object.

kyori19:
@connyduck I could build apk on my PC but CircleCI was failed...
Please retry it...

connyduck:
Ok now the code looks fine ^__^
I will merge it later for Tusky 4.
Don't worry about CircleCI.

PR #777, opened 2018-08-19
notification tab - fix progressbars of footer and fragment overlapping

connyduck:
There where two progressbars showing when the notification tab did its initial load.
I removed the progressbar of the footer.

Still missing - loading bar at the bottom when scrolling completely down

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/fragment/NotificationsFragment.java
+++ b/app/src/main/java/com/keylesspalace/tusky/fragment/NotificationsFragment.java
@@ -132,9 +132,7 @@ public class NotificationsFragment extends SFragment implements
     private TabLayout.OnTabSelectedListener onTabSelectedListener;
     private boolean hideFab;
     private boolean topLoading;
-    private int topFetches;
     private boolean bottomLoading;
-    private int bottomFetches;
     private String bottomId;
     private String topId;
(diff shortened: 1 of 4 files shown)


connyduck:
ok so now the progress bar on the bottom is here again. For some reason loading at the bottom sometimes gets stuck, will investigate that later.

connyduck:
ok now this works fine

app/src/main/java/com/keylesspalace/tusky/fragment/NotificationsFragment.java
```diff
@@ -202,15 +200,15 @@ public View onCreateView(@NonNull LayoutInflater inflater, @Nullable ViewGroup c
...
-        topFetches = 0;
         bottomLoading = false;
-        bottomFetches = 0;
         bottomId = null;
         topId = null;
 
         ((SimpleItemAnimator) recyclerView.getItemAnimator()).setSupportsChangeAnimations(false);
         setupNothingView();
 
+        sendFetchNotificationsRequest(null, topId, FetchEnd.TOP, -1);
```
charlag: Why do we have to request it now but we didn't need to before ?

connyduck: because previously it was triggered by the scroll listener (99% sure)

PR #778, opened 2018-08-19
fix crash when EmojiCompat was not initialized

connyduck:
Seems like this time its a racecondition because it does not crash every time like when we had this Problem in `AccountActivity` and the font did not work?  Any ideas @C1710?

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/MainActivity.java
+++ b/app/src/main/java/com/keylesspalace/tusky/MainActivity.java
@@ -516,7 +516,11 @@ public final class MainActivity extends BottomSheetActivity implements ActionBut
 
         for (AccountEntity acc : allAccounts) {
             CharSequence emojifiedName = CustomEmojiHelper.emojifyString(acc.getDisplayName(), acc.getEmojis(), headerResult.getView());
-            emojifiedName = EmojiCompat.get().process(emojifiedName);
+            try {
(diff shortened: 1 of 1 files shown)


C1710:
My guess is that this could have something to do with the EmojiCompat library itself.  
I know that there is one race condition in `FilemojiCompatConfig`, but that's not something that causes an instable state.  
However the font indeed __is__ loaded in a seperate Thread, but this is not something I can change.  
The only solution would be to try the latest EmojiCompat version (Google released a new version with Android 9. I know that because Twitter had to wait for the new version since they had issues with the old EmojiCompat version in conjunction with Emoji 11)

C1710:
Wait a minute....
This is an excerp from the current EmojiCompat's javadoc:  
```
During initialization information about emojis is loaded on a background thread. Before the
EmojiCompat instance is initialized, calls to functions such as {@link
EmojiCompat#process(CharSequence)} will throw an exception. You can use the {@link InitCallback}
class to be informed about the state of initialization.
```

It looks like this might be the cause for this problem...

connyduck:
> However the font indeed is loaded in a seperate Thread

Isnt't that the exact problem here? `EmojiCompat.get().process()` is called before the other Thread finished loading?

connyduck:
Ok so I guess I will fix this differently

PR #780, opened 2018-08-19
Address overlap of name/time in status display

gtcompscientist:
Use a ConstraintLayout and LinearLayout to better manage the size of the names.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -1,7 +1,7 @@
 apply plugin: 'com.android.application'
-apply plugin: 'kotlin-kapt'
 apply plugin: 'kotlin-android'
 apply plugin: 'kotlin-android-extensions'
+apply plugin: 'kotlin-kapt'
 
 def getGitSha = { ->
     def stdout = new ByteArrayOutputStream()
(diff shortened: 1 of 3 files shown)


gtcompscientist:
Fair enough. I've dug in and I think I have a way to lay out the entire thing as a flat ConstraintLayout, I'll let you know when that is done.

gtcompscientist:
@connyduck So, I was able to get the layout entirely flat, or nearly flat. The two ways break down like this:

1. Entirely flat requires setting the display name/username as a `SpannedString` (to bold the display name)
1. Nearly flat flattens the entire rest of the layout but leaves the `LinearLayout` to manage the ellipsizing.

Happy to go with whichever you prefer.

connyduck:
Awesome. I think I prefer the option with the `LinearLayout` as we end up with less code in the adapter?
But please do as you want, I will merge both!

charlag:
@gtcompscientist do you plan working on this further? (no rush, just going through everything I've missed)

connyduck:
-> #967

PR #783, opened 2018-08-20
Added Korean Translation

BasixKOR:
Files made via Stringlate. Needs to be reviewed before merge. Continued from #782.


Kjwon15:
I reviewed on #782 
https://github.com/tuskyapp/Tusky/pull/782#pullrequestreview-147631374

PR #785, opened 2018-08-20
values-es: Update translations

Eskuero:
I made the original spanish translation months ago so there was plenty of new strings to translate. I added those.
I also changed a bit some others because the translated sentences ended being too large in spanish due to words being bigger and had to be cut down.
Per example:
"user favourited your toot"
was translated into:
"user marcó como favorito tu estado"
and didn't fit on the screen so I now changed it into:
"user marcó favorito"

Also I've seen a rise in the usage of other software server like Pleroma so I tried to avoid branding as much as possible and use "shared/repost" or "status/post" rather than "boosts" and "toots"


connyduck:
Looks good, thank you!

Issue #772, opened 2018-08-18
Post visibility isn't preserved when saving draft


Tak:
Post visibility isn't preserved when saving a draft.

Steps to reproduce:
- Begin drafting a post
- Set the visibility to something other than your default
- Close the post and save it
- Begin editing the resulting draft again

Result: Post visibility is reset to the user default
Expected: Post visibility of the draft was preserved

* * * *
- Tusky Version: 3.0b1
- Android Version: 6.0.1
- Android Device: Yotaphone 2
- Mastodon instance (if applicable): cybre.space , probably not relevant

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

connyduck:
closed by #772

PR #787, opened 2018-08-21
Separate saved draft visibility from default reply visibility: #772

Tak:


charlag:
Ah, so we additionally save visibility now.
Should be backward compatible, right, it was in the DB already if I understand correctly?

Tak:
Yes, we were already persisting the visibility, it's just that we were using the same field in the compose activity for the visibility of the post being replied to and the visibility of the stored draft, with no real way for the initialization code to know which case it was in.

PR #807, opened 2018-08-26
fix notification tab loading bottom notifications forever

connyduck:


charlag:
I am never sure if `fromId` means what I want but otherwise lgtm

mal0ki:
I was just about to report this issue, glad I wasn't the only one to notice.

PR #825, opened 2018-08-31
[Feature] Add support for collapsible statuses when they exceed 500 characters

### LINKED ISSUE #748: Auto collapse posts with more than 500 characters
Other ActivityPub implementations (like Pleroma or GNU/Social) support more than 500 characters (Mastodon's recommended character length) in a status. Whichs results in long statuses appearing in their entire length on Tusky.  

I suggest the implementation of an auto-collapse feature, that only shows a preview of the status first (500 characters?), but can be expanded using a button. 

* * * *
- Tusky Version: 2.3-d29a123
- Android Version: 6.0.1
- Android Device: Meizu MX4

- [Y] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

HellPie:
### Description
Supports showing a button to collapse/expand long posts. Fixes #748.

### Features
- Toggleable by the user
- Limit set to 500 characters (like Pleroma and other existing implementations)
- It works in:
    - Timelines
    - Notifications
- It works by:
    - Showing a custom button below the content
    - Hiding the button if the post is CW-ed until content is shown

### Concerns
The way it was implemented has proven to be potentially expensive if more and more features will be added overtime to the timeline. This is mostly the fault of the current design within Tusky, which reloads layouts rather than just content.

A possible solution would be to redesign the core data models to store a full and a compressed version of the content but that would require extensive rewrite of several parts of the codebase which would not need to know about it, again, due to how Tusky's codebase is designed.

It has not proven to be an issue on my Nexus 6 (Lineage OS 15.1), OnePlus 3 (Lineage OS 15.1) and Moto G (1st gen, LineageOS 14.1) yet so I did not want to enforce the feature so deep into the internals of the application.

### Screenshots

Screenshot of the feature being shown in all themes, for both timelines and notifications

#### Video of the feature is available [on Imgur](https://i.imgur.com/DR7NFhx.gifv), due to restrictions with file types allowed on GitHub.


HellPie:
Commit a0a41ca and 0ee004d will be reverted to the original settings once this is approved, for as long as I'll have to work on it, I won't change those due to issue I had with Android Studio 3.1.4 not rendering previews for certain layouts.

Commit 8f6c327 may be reverted although it has proven to be very useful to me over the years and is suggested by the [Android Developers website](https://developer.android.com/studio/build/optimize-your-build#optimize) so I would discourage from doing so, while I would much rather encourage implementing it properly as a custom `productFlavor` as suggested by the article linked.

app/src/main/java/com/keylesspalace/tusky/adapter/NotificationsAdapter.java
```diff
@@ -337,6 +347,8 @@ void setupButtons(final NotificationActionListener listener, final String accoun
             message.setOnClickListener(this);
             statusContent.setOnClickListener(this);
             contentWarningButton.setOnCheckedChangeListener(this);
+
+            contentCollapseButton = itemView.findViewById(R.id.button_toggle_notification_content);
```
charlag: please move it above to other `findViewById()`

app/src/main/java/com/keylesspalace/tusky/adapter/NotificationsAdapter.java
```diff
@@ -488,11 +499,58 @@ private void setupContentAndSpoiler(NotificationViewData.Concrete notificationVi
             Spanned content = statusViewData.getContent();
             List<Emoji> emojis = statusViewData.getStatusEmojis();
 
-            Spanned emojifiedText = CustomEmojiHelper.emojifyText(content, emojis, statusContent);
+            if(contentCollapseButton != null && statusViewData.isCollapsible() && (notificationViewData.isExpanded() || !hasSpoiler)) {
```
charlag: Why do you need to check for null?

app/src/main/java/com/keylesspalace/tusky/adapter/NotificationsAdapter.java
```diff
@@ -488,11 +499,58 @@ private void setupContentAndSpoiler(NotificationViewData.Concrete notificationVi
...
+                    int position = getAdapterPosition();
+                    if(position != RecyclerView.NO_POSITION && notificationActionListener != null) {
+                        notificationActionListener.onNotificationContentCollapsedChange(isChecked, position);
+                    }
+                });
+
+                contentCollapseButton.setVisibility(View.VISIBLE);
+                if(statusViewData.isCollapsed()) {
+                    contentCollapseButton.setChecked(true);
+                    statusContent.setFilters(new InputFilter[]{(source, start, end, dest, dstart, dend) -> {
```
charlag: Please, extract this

app/src/main/java/com/keylesspalace/tusky/adapter/NotificationsAdapter.java
```diff
@@ -488,11 +499,58 @@ private void setupContentAndSpoiler(NotificationViewData.Concrete notificationVi
...
+                            if(keep == start) return "";
+                        }
+
+                        return source.subSequence(start, keep) + "…";
+                    }});
+                } else {
+                    contentCollapseButton.setChecked(false);
+                    statusContent.setFilters(new InputFilter[]{});
+                }
+            } else if(contentCollapseButton != null) {
```
charlag: Null check again?

app/src/main/java/com/keylesspalace/tusky/adapter/StatusBaseViewHolder.java
```diff
@@ -499,7 +501,54 @@ void setupWithStatus(StatusViewData.Concrete status, final StatusActionListener
         } else {
             setSpoilerText(status.getSpoilerText(), status.getStatusEmojis(), status.isExpanded(), listener);
         }
-    }
 
+        if(contentCollapseButton != null && status.isCollapsible() && (status.isExpanded() || status.getSpoilerText() == null || status.getSpoilerText().isEmpty())) {
```
charlag: Same comments here as in NotificationsAdapter

app/src/main/java/com/keylesspalace/tusky/fragment/NotificationsFragment.java
```diff
@@ -491,6 +498,84 @@ public void onLoadMore(int position) {
...
+    public void onContentCollapsedChange(boolean isCollapsed, int position) {
+        if(position < 0 || position >= notifications.size()) {
+            Log.e(TAG, String.format("Tried to access out of bounds status position: %d of %d", position, notifications.size() - 1));
+            return;
+        }
+
+        NotificationViewData notification = notifications.getPairedItem(position);
+        if(!(notification instanceof NotificationViewData.Concrete)) {
+            if(notification == null) {
+                Log.e(TAG, String.format(
```
charlag: Why would it be `null`?

app/src/main/java/com/keylesspalace/tusky/fragment/NotificationsFragment.java
```diff
@@ -491,6 +498,84 @@ public void onLoadMore(int position) {
...
+                        position,
+                        notifications.size() - 1
+                ));
+            }
+
+            return;
+        }
+
+        StatusViewData.Concrete status = ((NotificationViewData.Concrete) notification).getStatusViewData();
+        if(status == null) {
```
charlag: Why would it be `null`?

app/src/main/java/com/keylesspalace/tusky/fragment/NotificationsFragment.java
```diff
@@ -491,6 +498,84 @@ public void onLoadMore(int position) {
...
+                concreteNotification.isExpanded()
+        );
+        notifications.setPairedItem(position, updatedNotification);
+        adapter.updateItemWithNotify(position, updatedNotification, false);
+
+        // Since we cannot notify to the RecyclerView right away because it may be scrolling
+        // we run this when the RecyclerView is done doing measurements and other calculations.
+        // To test this is not bs: try getting a notification while scrolling, without wrapping
+        // notifyItemChanged in a .post() call. App will crash.
+        recyclerView.post(() -> adapter.notifyItemChanged(position, notification));
```
charlag: just wrap whole `updateItemWithNotify` in `post`

app/src/main/java/com/keylesspalace/tusky/fragment/NotificationsFragment.java
```diff
@@ -491,6 +498,84 @@ public void onLoadMore(int position) {
...
+
+    /**
+     * Called when the status {@link ToggleButton} responsible for collapsing long
+     * status content is interacted with.
+     *
+     * @param isCollapsed Whether the status content is shown in a collapsed state or fully.
+     * @param position    The position of the status in the list.
+     */
+    @Override
+    public void onNotificationContentCollapsedChange(boolean isCollapsed, int position) {
```
charlag: Not sure if it makes sense to copy javadoc for overrides

app/src/main/java/com/keylesspalace/tusky/fragment/TimelineFragment.java
```diff
@@ -262,6 +267,8 @@ private void setupTimelinePreferences() {
             filterRemoveRegexMatcher = Pattern.compile(regexFilter, Pattern.CASE_INSENSITIVE)
                     .matcher("");
         }
+
+        collapseLongStatusContent = preferences.getBoolean("collapseLongStatuses", true);
```
charlag: Extract preference key constant please

app/src/main/java/com/keylesspalace/tusky/fragment/TimelineFragment.java
```diff
@@ -561,6 +568,48 @@ public void onLoadMore(int position) {
...
+        }
+
+        StatusViewData status = statuses.getPairedItem(position);
+        if(!(status instanceof StatusViewData.Concrete)) {
+            // Statuses PairedList contains a base type of StatusViewData.Concrete and also doesn't
+            // check for null values when adding values to it.
+
+            // TODO: Implement @NonNull/@Nullable in PairedList insert methods
+
+            if(status == null) {
```
charlag: No need to null check

app/src/main/java/com/keylesspalace/tusky/fragment/TimelineFragment.java
```diff
@@ -561,6 +568,48 @@ public void onLoadMore(int position) {
...
+     */
+    @Override
+    public void onContentCollapsedChange(boolean isCollapsed, int position) {
+        if(position < 0 || position >= statuses.size()) {
+            Log.e(TAG, String.format("Tried to access out of bounds status position: %d of %d", position, statuses.size() - 1));
+            return;
+        }
+
+        StatusViewData status = statuses.getPairedItem(position);
+        if(!(status instanceof StatusViewData.Concrete)) {
```
charlag: merge this check into check above, there's no need to be so precise in errors in this case

app/src/main/java/com/keylesspalace/tusky/util/ViewDataUtils.java
```diff
@@ -58,12 +59,24 @@
                 .setApplication(visibleStatus.getApplication())
                 .setStatusEmojis(visibleStatus.getEmojis())
                 .setAccountEmojis(visibleStatus.getAccount().getEmojis())
+                .setCollapsible(collapseLongStatusContent && status.getContent().length() > 500)
```
charlag: This value is hardcoded in multiple places

charlag:
There's one minor thing I noticed: when post is barely more than 500 symbols, the button just reveals some more symbols on the same line, making it useless.
Quick hack would be showing button for posts with 550+ symbols but cut the text to 500 or something like that

app/src/main/java/com/keylesspalace/tusky/adapter/NotificationsAdapter.java
HellPie: Would it be ok if I were to move it to SmartLengthFilter in the util package?
Also will I need to convert that to Kotlin or is Java ok? I don't have much experience with Kotlin.

app/src/main/java/com/keylesspalace/tusky/fragment/NotificationsFragment.java
HellPie: I just wanted to avoid adding an entire level of nesting, since it's needed only for one method.

app/src/main/java/com/keylesspalace/tusky/adapter/NotificationsAdapter.java
HellPie: I was working on some experiments to reuse some code between notifications and timeline but it didn't work out due to different layouts and data holders. Didn't work out.

app/src/main/java/com/keylesspalace/tusky/fragment/NotificationsFragment.java
HellPie: I had that because nowhere were get or set operations marked as `@NonNull` for the provided elements so I picked the "better safe than sorry" route around it.
I can remove it but PairedList (and the used List implementers in it) doesn't guarantee null safety.

app/src/main/java/com/keylesspalace/tusky/fragment/NotificationsFragment.java
HellPie: I had that because nowhere were get or set operations marked as `@NonNull` for the provided elements so I picked the "better safe than sorry" route around it.
I can remove it but PairedList (and the used List implementers in it) doesn't guarantee null safety.

app/src/main/java/com/keylesspalace/tusky/fragment/NotificationsFragment.java
HellPie: Because, again, lack of `@NonNull` annotations makes me feel very uncomfortable, especially since the almost excessive amount of crashes after the codebase switched to Kotlin, which usually suggests NPEs issues. I annotated constructor, getter and property for `NotificationViewData::statusViewData` as `@NonNull` and removed the check.

app/src/main/java/com/keylesspalace/tusky/fragment/TimelineFragment.java
HellPie: Is there a class that is responsible for storing these accessible by TimelineFragment? The previews preferences are also setup like this, it's where I looked to decide how to do it.
Should I move them all to the `notranslate.xml` resource file and use `getString()` instead?

app/src/main/java/com/keylesspalace/tusky/util/ViewDataUtils.java
HellPie: The issue I ran into is most of those places don't have easy access to preferences (which would be the ideal implementation). I'm looking into what I can access to see if I can easily get the value from the common available items in each method to properly implement this.

app/src/main/java/com/keylesspalace/tusky/fragment/NotificationsFragment.java
HellPie: oh yeah I just smash Alt+Ins, "Implement..." and then smash enter. It turned out useful to understant which method I was working on, I am in the process of removing it from the various places where this has happened rn

HellPie:
I left some changes unresolved, awaiting further suggestions to avoid having to change the code again later.

clarcharr:
I'd personally like this configurable-- for example, cybre.space has 512 chars max and I'd rather not have posts with 512 characters get collapsed. Maybe 600 might be a good default to cover that case specifically.

HellPie:
I was really hoping to not see that requested until after the merge, and I would like to get this to work at least in a very basic way to ship it and start getting bug reports before implementing extra code that will only turn out to make debugging person-specific and end up in some sort of bug micromanagement hell.

My ideal design for how this should work at its best possible state is:
[x] Collapse at 500 characters by default (although 500 is a lot on mobile, two posts can take up a screen)
[ ] Allow some "runway" (10 characters in excess) to collapse after sentences, punctuation, words
[ ] Do not collapse if the collapsed/visible ratio is not worth the effort (25% maybe)

And for the settings:
[x] Allow the user to toggle the feature entirely
[ ] Allow the user to disable "smart collapse" and collapse the post at the exact character limit
[ ] Allow the user to specify their own, preferred collapsible length

And then it would be "peak implementation" if we could:
- Get Gargron to aknowledge this damn feature as one, once and for all
- Get Pleroma and Mastodon to agree on a key-value in `api/v1/instance` like we have `toot_max_length`
- Change the "selection" process to: user settings, instance settings, 500

Now there are issues:
The current implementation is nearing the border of what I consider smooth performances, and that is what is stopping me from considering those as worth implementing into this single PR, without a wider userbase having tested the feature first.
I already explained my concerns in the PR description in the "Concerns" section, I want to make them clear again: adding those features will have a noticeable impact in performance, as of now the app is not going to die, it will run well with such a "simple" (imo it's already advanced compared to an MVP) implementation, but the cost in performance of adding everything else will require a deeper integration of the feature in the app.

clarcharr:
Ah, makes sense! I appreciate the thought you put into the design already. :)

Allowing runaway characters seems to fit the use case very well. In that case, config can come later. :)

charlag:
No rush
I still don't see where performance bottleneck should be

connyduck:
- I dont like the look of the button, I feel like it should be aligned to the start of the text, not stretch the whole width but have a bigger vertical padding
- please remove the changes to build files from this PR
- I don't think this needs to be configurable. 
- Another possibility would be to just open the detail view on button click and not expanding it inside the timeline
- your implementation does not work with custom emojis ❗

Getting this crash when clicking on a shortened status: ❗
   ```
 java.lang.NullPointerException: Attempt to invoke virtual method 'void android.widget.ToggleButton.setOnCheckedChangeListener(android.widget.CompoundButton$OnCheckedChangeListener)' on a null object reference
        at com.keylesspalace.tusky.adapter.StatusBaseViewHolder.setupWithStatus(StatusBaseViewHolder.java:506)
        at com.keylesspalace.tusky.adapter.StatusDetailedViewHolder.setupWithStatus(StatusDetailedViewHolder.java:91)
        at com.keylesspalace.tusky.adapter.ThreadAdapter.onBindViewHolder(ThreadAdapter.java:71)
        at android.support.v7.widget.RecyclerView$Adapter.onBindViewHolder(RecyclerView.java:6673)
...
```

**Comment by HellPie** [comment]
After @connyduck pointed out custom emojis not working I noticed using an InputFilter on a TextBox doesn't actually keep the Spannable, Android instead flattens it to a CharSequence and expects it to be rebuilt. I will have to refactor some of the code so the method that handles custom emojis, tags etc and embeds them in the TextView is also called using the cropped text.

Regarding @connyduck's comments, I have some other comments, ragrding the crash, I'm pretty sure it's because I removed one of the null checks and that's why I'm paranoid about those, I'll have to take a look and see if that's a special case that I don't handle.

- Since nobody was willing to implement it, I did it and this is my ideal design for it, it's also matching what Pleroma has been using for over a year and what tootsuite/mastodon#8205 has been using: full width, centered text.
- I'll remove those but I'll have to keep them locally as, like the commits explain, I refuse to use Studio if I have to compile every change when a layout is involved to see the result. I hope it won't affect rebasing on master.
- I think especially on mobile personalization should be a must. I am personally running a third version of the app built specifically to give me collapsibility at 300 characters since two 500 characters long statuses take up most of the screen on my OnePlus 3 and with three even my Nexus 6 is filled. For all I care about, Mastodon's design itself on mobile should have this feature. I also run a fourth build on my Moto G 2013 capped at 140 characters, so far it has been my favorite size, it allows me to skim through statuses and only care about those I like to care about.
- Opening the thread/detail view for each post that one wants to read seems overkill. I would implement that as another toggle, but it disrupts the flow of the user through the timeline in ways that ruin the experience more than enhancing it. It would be like being presented with a list of apps with extremely similar icons and having to open each one of them to check which one is the one the person wants. It's just "too much effort" and "too many taps" for what I consider a sane UX, although I won't disagree for an MVP that works.
- I commented on that already, and I need to work on the InputFilter, but regarding the recent discovery of emojis, I'd say first I fix that, then I can work on the filter as much as I want. For an MVP it looks weird, I'd probably stop at implementing a "cut at word" filter and wait further bug requests and suggestions from a wider userbase before implementing features that sure, are nice but have near to no UX meaning other than being good to look at.

Regarding @charlag's question about performance, I have to say I didn't expect a 2013 device to run Oreo as well as my Moto G does, and I'm happy it does, but ever since I updated to LOS 15.1 this change cripples the framerate of the phone when building the expanded and collapsed statuses when the user clicks the button. It only lasts between a quarter of a second to half a second in the noticeable scenarios, based on the length and complexity of the collapsed post, but the heavy usage of layout re-inflating that Tusky leverages to apply changes to posts and which I leverage to update the post's content are not something I'd like to expand upon as it is. Granted, there is no reason why a 2013 32bit, low end device should run an OS that is targeting 2016, 64bit-only middle/high end specs but it does, I'm not the only one and there are plenty of devices coming up in several markets with worse specifications than those my Moto G has.

**Comment by HellPie** [comment]
Also I'd like @charlag's input on the changes I left unresolved, since I don't really like the idea of nesting the entire block of code and I'll happily avoid complying if possible, but I mostly don't know how he wants me to handle the other changes, since I don't really see many examples in the interested source files.

**Comment by mal0ki** [comment]
Thank you @HellPie for this PR, and everyone else for the constructive discussion.

File: app/src/main/java/com/keylesspalace/tusky/adapter/NotificationsAdapter.java (hunk 3 above)
Comment (HellPie) [reply app/src/main/java/com/keylesspalace/tusky/adapter/NotificationsAdapter.java]: I implemented it this way, so once I'm done complaining about how bad it works on devices it shouldn't even run on (Moto G 2013), I (or someone else) will start working on the smart cutting features I discussed in [the comment below](https://github.com/tuskyapp/Tusky/pull/825#issuecomment-417800580).

**Comment by HellPie** [comment]
I just realized maybe by padding you mean the button is too think? Because I set its height to `32dp` I think rather than the default `48dp` because it looked kinda big on my Nexus 6 but yeah it kinda is too small. I don't think real padding is needed because spacing the button will either make the media look disconnected from the text or the text from the button itself.

File: app/src/main/java/com/keylesspalace/tusky/fragment/TimelineFragment.java (hunk 10 above)
Comment (HellPie) [reply app/src/main/java/com/keylesspalace/tusky/fragment/TimelineFragment.java]: I'll just open a new PR later and take care of moving all the string resources to the `notranslate.xml` if you prefer, to not pollute this PR with changes regarding other configurations, since if I do it for mine I might as well do it for all of them.

Comment (charlag) [reply app/src/main/java/com/keylesspalace/tusky/fragment/TimelineFragment.java]: Sounds great, thanks

**Comment by HellPie** [comment]
With this I believe to have fixed everything. Video of the feature as it is right now: https://weeaboo.space/notice/1637953

**Comment by charlag** [comment]
I will try to look at this again tomorrow or at the weekend, thanks for the changes, looks pretty good now IMO

**Comment by HellPie** [comment]
I did that merge from GitHub's web interface so tbh I have no idea how that will end up going.
Worst case I'll sacrifice a goat and hope everything gets fixed.

**Comment by HellPie** [comment]
Sorry for that. I wanted to comment that I still need to fix things because that merge obviously didn't go as cleanly as GitHub mentioned and then Chrome decided to scroll the page after loading idk what. tl;dr: Misclick D: this is frustrating.

File: app/src/main/java/com/keylesspalace/tusky/adapter/StatusBaseViewHolder.java [hunk 14]
```diff
@@ -503,23 +503,28 @@ void setupWithStatus(StatusViewData.Concrete status, final StatusActionListener
...
+                    int position = getAdapterPosition();
+                    if(position != RecyclerView.NO_POSITION)
+                        listener.onContentCollapsedChange(isChecked, position);
+                });
+
+                contentCollapseButton.setVisibility(View.VISIBLE);
+                if(status.isCollapsed()) {
+                    contentCollapseButton.setChecked(true);
+                    content.setFilters(new InputFilter[]{
+                            new SmartLengthInputFilter(SmartLengthInputFilter.LENGTH_DEFAULT)
```
charlag: I don't remember if I asked if we can reuse the filter or not

app/src/main/java/com/keylesspalace/tusky/util/SmartLengthInputFilter.java
```diff
@@ -0,0 +1,145 @@
...
+
+	/** {@inheritDoc} */
+	@Override
+	public CharSequence filter(CharSequence source, int start, int end, Spanned dest, int dstart, int dend) {
+		// Code originally imported from InputFilter.LengthFilter but heavily customized.
+		// https://android.googlesource.com/platform/frameworks/base/+/master/core/java/android/text/InputFilter.java#175
+
+		int sourceLength = source.length();
+		int keep = max - (dest.length() - (dend - dstart));
+		if(keep <= 0) return "";
```
charlag: Not sure if we need to check that but can be left for keeping it generic I guess
and put spaces after `if`s please

app/src/main/java/com/keylesspalace/tusky/util/SmartLengthInputFilter.java
```diff
@@ -0,0 +1,145 @@
...
+			int boundary;
+
+			// Android N+ offer a clone of the ICU APIs in Java for better internationalization and
+			// unicode support. Using the ICU version of BreakIterator grants better support for
+			// those without having to add the ICU4J library at a minimum Api trade-off.
+			if(android.os.Build.VERSION.SDK_INT >= android.os.Build.VERSION_CODES.N) {
+				android.icu.text.BreakIterator iterator = android.icu.text.BreakIterator.getWordInstance();
+				iterator.setText(source.toString());
+				boundary = iterator.following(keep);
+				if(keep - boundary > 10) boundary = iterator.preceding(keep);
```
charlag: not clear what is `10`.  Please extract it

app/src/main/java/com/keylesspalace/tusky/util/SmartLengthInputFilter.java
HellPie: That was actually checked in Google's original `LengthFilter` code. I left it in as it catches one more particular case without having to go through the logic below which would become quite expensive in comparison.

I didn't notice you styled if blocks spaced, I may have to go through all of the code I touched and manually put them in if that's the case because I don't have that set in my style settings.

HellPie:
If you change idea on the configurability please feel free to revert b5d13e but even if you don't plan to, I'd argue that asking the community about their preference regarding having or not a toggle in the settings fragment would be a much better way to understand if it's worth removing or if it's going to cripple user experience.
In my opinion the toggle is going to be the only part of the code that will be easy to maintain, that has near to no overhead from both a runtime and a developer's perspective and also it's probably the only UI element that looks exactly like you and @connyduck would like it to look, out of the box 😛

charlag:
That's the thing about both configurability and design: everyone has different opinions. We can ask, of course, but my guess is 90% of the people would agree with this being on. I see no reason to scroll through random 10k character post on mobile. 
To get real UI feedback from a large crowd we need to publish it as nightly or even as a release.
I think making it a preference if there's a huge demand for it will be easier than removing some option (which is always a pain)

charlag:
I've tested a bit and I think button either needs background or bold text or both but I'd like to hear others
Otherwise seems OK to me

HellPie:
The button should be bold by default since buttons on Android are bold by default and I am not adding any style to the text whatsoever.
I will check on that, Idk about the background, I feel like it will mark it as the primary action of the content while I think the user should first read the first 500 characters and then notice the button if they want to continue, unlike the CW button which often only shows a 4 words extremely generic summary of the content and therefore should (and does) gain a predominant role in the UI of the post.

charlag:
We discussed it with connyduck, we agree that the button should look like CW button. They promised to give more feedback tomorrow.
Thanks for your efforts! 🙇‍♂️

HellPie:
Thanks @connyduck, spent quite some time on it, but it didn't turn out too hard to implement fortunately.

Not quite sure how to handle the Codacy complaint about complexity since they suggest splitting in more methods and that adds a lot of overhead for a method that must always work as fast as possible, also it wouldn't really fix the complaint from a practical point of view.

I fixed the import issue, that was quite critical imo since it's vital that if someone edits the code they won't pick up the wrong import statement since there are double APIs in place rn (bless fragmentation).

Also still waiting on @charlag to confirm all past changes are actually resolved, since they are still marked as requested on GitHub because I marked them manually rather than letting GitHub handle it through automation.

connyduck:
oh please ignore the codacy thing, I enabled it by mistake

HellPie:
ok, that import still bothered me tho so I'll keep that commit @connyduck 
idk why now Circle is failing if before it worked, code looks fine, I'd also stick a "builds on my machine" badge. I'm investigating as we speak.

connyduck:
I think its failing because its out of memory 🙄. Its fine as long as bitrise passes.

HellPie:
I'll call it a day then, right on time before I had to leave, too, ty for the review btw

charlag:
I can look tomorrow but I'm pretty sure it's okay if connyduck approved it.



16 Sep 2018, 21:34 by notifications@github.com:

PR #831, opened 2018-09-02
fix image preview rotation in ComposeActivity

### LINKED ISSUE #827: Tusky composer doesn't honor orientation metadata
My only takes pictures in landscape and stores its orientation in EXIF metadata. While it looks fine in the timeline once it is posted, it isn't respected in the composer (see screenshot)

screenshot_20180901-145041_tusky

[actual toot](https://octodon.social/@Siphonay/100650598134386494)

* * * *
- Tusky Version: 3.0
- Android Version: 8.0
- Android Device: Samsung Galaxy S9
- Mastodon instance (if applicable): any

connyduck:
closes #827 

We already had the code in Tusky, just needed to call it from the right place.
For some reason Picasso ignores orientation when used with uris as well, so I used RxJava instead.

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/ComposeActivity.java
+++ b/app/src/main/java/com/keylesspalace/tusky/ComposeActivity.java
@@ -18,6 +18,7 @@ package com.keylesspalace.tusky;
 import android.Manifest;
 import android.annotation.SuppressLint;
 import android.app.ProgressDialog;
+import android.arch.lifecycle.Lifecycle;
 import android.content.ContentResolver;
 import android.content.Context;
 import android.content.DialogInterface;
@@ -135,11 +136,19 @@ import java.util.Locale;
 import javax.inject.Inject;
 
 import at.connyduck.sparkbutton.helpers.Utils;
+import io.reactivex.Single;
+import io.reactivex.SingleObserver;
+import io.reactivex.android.schedulers.AndroidSchedulers;
+import io.reactivex.disposables.Disposable;
+import io.reactivex.schedulers.Schedulers;
 import okhttp3.MediaType;
 import okhttp3.MultipartBody;
 import retrofit2.Call;
 import retrofit2.Callback;
 import retrofit2.Response;
+
+import static com.uber.autodispose.AutoDispose.autoDisposable;
+import static com.uber.autodispose.android.lifecycle.AndroidLifecycleScopeProvider.from;
 
 public final class ComposeActivity
         extends BaseActivity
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/keylesspalace/tusky/ComposeActivity.java
```diff
@@ -1121,11 +1130,24 @@ private void makeCaptionDialog(QueuedMedia item) {
...
+                    @Override
+                    public void onSubscribe(Disposable d) {}
+
+                    @Override
+                    public void onSuccess(Bitmap bitmap) {
+                        imageView.setImageBitmap(bitmap);
+                    }
+
+                    @Override
+                    public void onError(Throwable e) { }
```
charlag: Did you use `SingleObserver` and not consumer to swallow errors? You could still use `BiConsumer` or two `Consumer`s to save a little bit of space but... idk. I just don't like Java's huge declarations but that's not important.

app/src/main/java/com/keylesspalace/tusky/util/MediaUtils.java
```diff
@@ -176,4 +180,81 @@ public static int calculateInSampleSize(BitmapFactory.Options options, int reqWi
...
+            return result;
+        } catch (OutOfMemoryError e) {
+            return null;
+        }
+    }
+
+    public static int getImageOrientation(Uri uri, ContentResolver contentResolver) {
+        InputStream inputStream;
+        try {
+            inputStream = contentResolver.openInputStream(uri);
```
charlag: We could rewrite it as try-with-resources but whatever

connyduck:
Yes I am waiting for Picasso 3 release, this 2.71828 Version doesn't have a changelog

Siphonay:
Good job friends, thank you!

app/src/main/java/com/keylesspalace/tusky/util/MediaUtils.java
```diff
@@ -176,4 +180,81 @@ public static int calculateInSampleSize(BitmapFactory.Options options, int reqWi
...
+        }
+        try {
+            Bitmap result = Bitmap.createBitmap(bitmap, 0, 0, bitmap.getWidth(),
+                    bitmap.getHeight(), matrix, true);
+            if (!bitmap.sameAs(result)) {
+                bitmap.recycle();
+            }
+            return result;
+        } catch (OutOfMemoryError e) {
+            return null;
```
igalic: would giving the user an error just cause another OOM at this point?

connyduck: No, garbage collector frees up the memory reserved for the bitmap

connyduck:
User see a generic error message when an OOM exception occurs. We could show a more specific error message.

PR #833, opened 2018-09-03
Update bitrise badge

connyduck:
test PR for new bitrise config that will test&build
build log should now be public!


charlag:
Wow, CI which doesn't crash when building us, good job!

PR #834, opened 2018-09-04
Make page margin to truly black for tablets.

qwazix:
As requested


connyduck:
looks good, thank you!

PR #842, opened 2018-09-10
update targetSdkVersion to 28

connyduck:
not sure when we will switch to Android X, this should do for now


charlag:
And CIs cannot pick it up yet, hmmmmmmmmm

PR #845, opened 2018-09-11
fix notifications loading (again)

connyduck:
apparently I still missed a case where the notification tab does not notice its at the end and keeps on loading forever in #807 🙄 
(it loaded forever when there was exactly 1 notification)
(I start hating this)

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/fragment/NotificationsFragment.java
+++ b/app/src/main/java/com/keylesspalace/tusky/fragment/NotificationsFragment.java
@@ -642,7 +642,7 @@ public class NotificationsFragment extends SFragment implements
                     adapter.removeItemAndNotify(this.notifications.size());
                 }
 
-                if (adapter.getItemCount() > 1) {
+                if (adapter.getItemCount() > 0) {
                     addItems(notifications, fromId);
                 } else {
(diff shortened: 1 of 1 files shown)


charlag:
Uhm, is this okay? Wouldn't you override this counters on every request or do I miss something?

connyduck:
Yes, thats the point - it did not stop loading because the `bottomId` was not set to null. But I will check again if everything is fine & how we do it in other places.

connyduck:
Ok, I am sure its fine. We had a similar bug that was also fixed by removing the `!= null` checks around `bottomId` -> https://github.com/tuskyapp/Tusky/commit/b579a9baea81b9c1e3b350ca1c1dc83081f6fd5e

charlag:
I still don't understand how it should work.
For example, we do loading on bottom, passed `topId` is `null`. We override the field `topId` with `null`.
After that, we pull the top and we don't have `topId`.
Same with bottom and the middle.
What do I miss?

connyduck:
Ah now I see what you mean. Dammit, it does not work, you are right.

connyduck:
New try. Explains why the problem occured only with exactly 1 toot and not everytime.

charlag:
Looks better now but I need a proper look to figure it out. If it works, then just merge it I guess

PR #852, opened 2018-09-17
Enable sharing media directly from Tusky

Tak:
Enables sharing images and videos/gifs directly from Tusky to other android applications that register for the appropriate mime types. Includes migration of relevant classes to kotlin.

Resolves #822 and #799

PR #857, opened 2018-09-19
Make Tusky a share target for video

Tak:
Extracted from https://github.com/tuskyapp/Tusky/pull/852


charlag:
I wonder why tests have failed :thinking:

charlag:
Changes look okay to me but need to check that it actually works

connyduck:
works like a charm

PR #862, opened 2018-09-22
Short emoji code on long press in statuses

charlag:
Using `LinkMovementMethod` for this.
Replaces hack with multiple ClickListeners too.
Code can be messy, tell me if something needs to be fixed.

cf5b33c7d18258d1


connyduck:
I have to be honest, I don't like this. I just tried it out, it is almost impossible to exactly touch a emoji, they are too small. Its just super weird, especially if there are also normal emojis around.
And the thing with the MovementMethod is a bit too hacky for my liking.
But I agree, we need a way to see the shortcodes somehow. 🤔

charlag:
Welp, that's true, it may be hard to hit.
MovementMethod does what it's parent does
I'll close it then, thanks for the review

Issue #738, opened 2018-07-28
[feature] Deletion confirmation


LogicalDash:
Web Mastodon shows a confirmation dialog when I ask to delete a toot. This would actually be *more* useful in Tusky, because touchscreen input is less precise. I'm more likely to bounce my thumb wrong on the menu button.

* * * *
- Tusky Version: 2.3
- Android Version: 8.0.0
- Android Device: onePlus 3t
- Mastodon instance (if applicable): cybre.space

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

charlag:
Hi! I think that's not bad idea, thanks for reporting

connyduck:
closed by https://github.com/tuskyapp/Tusky/pull/865

PR #865, opened 2018-09-30
Add deletion confirmation

pt2121:
re #738

Show a confirmation dialog when a user deletes a toot


connyduck:
Awesome, thank you!

PR #867, opened 2018-10-01
Updated it (Italian) translation

silkevicious:


connyduck:
Why does stringlate always start from some random branch 😐
Can you try editing the file from master manually?

Issue #443, opened 2017-11-08
Add the ability to pin/unpin statuses, show pinned statuses in profiles


connyduck:

charlag:
I've checked and API sends pinned toots first anyway. Good to know.

charlag:
I want to show pinned statuses pretty badly but I think we should wait for #562 to be merged because it makes some changes to the timeline.
Heck, this pinned statuses thing can add quite a lot of complexity.

wxl:
It looks like pin/unpin is now an option; what about viewing the pins?

charlag:
We somehow have support for pinned statuses thanks to #994 so I think this should be closed?
We should improve it in the future and remove that tab but it can be another issue.

PR #872, opened 2018-10-03
Add ability to pin/unpin statuses

charlag:
Quick implementation of pin/unpin.
It looks weird without displaying them but as I said before I don't want to make conflicts with #809 again.
I did not add menu item using XML because I think adding two items for one button is weir. I can change that but I would rather change other toggles to programmatic.


charlag:
Part of #443 ofc

connyduck:
Looks good and works fine! Thank you!

charlag:
Thanks! 🙇

PR #877, opened 2018-10-07
Welsh translation of strings.xml

mynameismonkey:
Translation by Jaz-Michael King and Jason Morgan - @jaz@toot.wales


charlag:
Wow, nice, thanks, we'll check that out!

connyduck:
Thank you very much!

PR #879, opened 2018-10-08
Prevent multiple regist to eventHub on same TimelineFlagment

kyori19:
https://github.com/tuskyapp/Tusky/blob/6cad8feea2f7c1abcb68ff0944c2571fa50c8627/app/src/main/java/com/keylesspalace/tusky/fragment/TimelineFragment.java#L381-L384

onActivityCreated is called every time when switching tabs.
Since subscribe is called every time when switching tab, it gets timeline unnecessarily many times after toot.

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/fragment/TimelineFragment.java
+++ b/app/src/main/java/com/keylesspalace/tusky/fragment/TimelineFragment.java
@@ -119,6 +119,7 @@ public class TimelineFragment extends SFragment implements
     public TimelineCases timelineCases;
     @Inject
     public EventHub eventHub;
+    private boolean eventRegistered = false;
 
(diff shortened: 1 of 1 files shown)


connyduck:
Indeed it does, nice find. I just wonder if there isnt a better solution to this, like moving the subscribtion to `onCreate`?

charlag:
Hi!
Thanks for the help, we could really miss that (these fragment lifecycles, oh.
I would prefer to move it to something which is executed only once but that's probably nowhere to be found, I think.

I'll try it out later this week

charlag:
@connyduck it is risky because we may need UI, we should check that manually

kyori19:
Thanks for your reply.
@connyduck I already tried to move to onCreate but app crashed because we need UI.

Tak:
Aha, I bet this addresses #880

Issue #563, opened 2018-04-08
Post containing only U+00AD SOFT HYPHEN crashes tusky


chr-1x:
after making [this post](https://cybre.space/@chr/99821978136424081) i received a few reports that people trying to scroll past it were getting tusky crashes. i don't have details but the reports are visible in my replies: https://cybre.space/@chr/99821978136424081

* * * *
- Tusky Version: unknown, probably most recent as of this issue filing
- Android Version: unknown
- Android Device: unknown
- Mastodon instance (if applicable): cybre.space

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

Zatnosk:
I left one of those reports.

It happens in Tusky 1.5.2 on Android 6.0.1 on my FairPhone 2

Zatnosk:
And to answer a question in the mastodon thread, I installed from Play Store.

Zatnosk:
No, wait, I thought I did, but F-Droid has it in it's list of installed apps on my phone. Sorry for multiposting :P

remi6397:
@Zatnosk F-droid lists also programs installed from other sources. If Play has Tusky listed on My Apps, it is probably downloaded from Play.

ciphercharly:
can confirm the issue, happening for me with Tusky 1.5.2 installed from F-droid on device running Android 6.0.1

connyduck:
I cannot reproduce this issue on any of my 3 devices and I cant find a related stacktrace in the Google Play console unfortunately. Can somebody help me out, I would really like to fix this.

ciphercharly:
@connyduck no issues at all with the f-droid version on all 3?

connyduck:
> @connyduck no issues at all with the f-droid version on all 3?

yes I just tested it

ciphercharly:
weird... what extra info could I provide to help you?

connyduck:
I would need a stacktrace. Do you know how to read the log of your device?

ciphercharly:
never done for an app, I can try if guided

ciphercharly:
done 'official' reporting after crashing, log should be attached (hated going through Google though), hope it helps

ciphercharly:
👍
out of curiosity, how does the log look like? could you paste a bit here?

connyduck:
It seems to be a bug in the Android framework and not in Tusky
```
java.lang.IndexOutOfBoundsException: 
[...]

PR #884, opened 2018-10-12
On android 6.0, filter 0x00ad (unicode soft hyphen) from statuses

Tak:
Works around #563 

Tested with known offenders
- https://cybre.space/@em/100805088534019375
- https://cybre.space/@chr/99821978136424081

The implementation isn't too elegant in reading-the-code terms, but I was pretty conscious about _every_ status going through this codepath, so I tried to keep iteration and allocation to an absolute minimum.

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/viewdata/StatusViewData.java
+++ b/app/src/main/java/com/keylesspalace/tusky/viewdata/StatusViewData.java
@@ -15,7 +15,9 @@
 
 package com.keylesspalace.tusky.viewdata;
 
+import android.os.Build;
 import android.support.annotation.Nullable;
+import android.text.SpannableStringBuilder;
 import android.text.Spanned;
 
 import com.keylesspalace.tusky.entity.Attachment;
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/keylesspalace/tusky/viewdata/StatusViewData.java
```diff
@@ -288,6 +300,32 @@ public boolean deepEquals(StatusViewData o) {
                     Objects.equals(card, concrete.card)
                     && isCollapsed == concrete.isCollapsed;
         }
+
+        static Spanned ReplaceCrashingCharacters(Spanned content) {
+            return (Spanned) ReplaceCrashingCharacters((CharSequence) content);
+        }
+
+        static CharSequence ReplaceCrashingCharacters(CharSequence content) {
```
connyduck: methods should start with lowercase character

app/src/main/java/com/keylesspalace/tusky/viewdata/StatusViewData.java
```diff
@@ -288,6 +300,32 @@ public boolean deepEquals(StatusViewData o) {
...
+            Boolean replacing = false;
+            SpannableStringBuilder builder = null;
+
+            for (int index = 0; index < content.length(); ++index) {
```
connyduck: its not guaranteed that `length()` is a simple getter, moving it out of the loop should be faster

Issue #882, opened 2018-10-11
Show Reply Count in Timeline (NOT DUPLICATE)


RadarNyan:
![image](https://user-images.githubusercontent.com/2091529/46815263-84778480-cdb5-11e8-800c-d9eaf6b28a87.png)
I need to see the reply count in the timeline, the official web UI (https://mastodon.social/web/) already does this and it's very useful, especially when "show replies" is turned off for Home timeline.

Personally I don't care about boosts count, the official web UI don't have it yet, but since there's also an option for "show boosts", it would make sense to show that count in the timeline as well.

I'm aware of #79 and I ensure you this is not a duplicate. At the end of that issue, it only accomplished showing count for Boosts and Favorites, **not Replies** - in the detailed view only, not in timeline.

* * * *
- Tusky Version: 3.1
- Android Version: 8.1
- Android Device: Mi Mix 2
- Mastodon instance (if applicable): Mastodon.social

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

charlag:
Hey, thanks for reporting.
Yeah, we probably could display it but I am personally not a fan of it, especially of all these zeroes. I would maybe add some indicators for "there is a reply", pretty much likeweb but without "empty" state.

charlag:
I'm closing it because #885 is closed

PR #885, opened 2018-10-12
Implement reply count indicator to track web UI

Tak:
Addresses #882 

Web UI:
web-detailed web-timeline

Tusky:
tusky-detailed tusky-timeline


connyduck:
Tbh I don't like this. It pollutes the the otherwise clean interface and I don't think the information is worth it. 

What about showing the number only on detail statuses? I'd be fine with that.

(also, BottomSheetActivityTest.kt is failing)

charlag:
I've grown to accept it in web. I don't think it makes too much sense to show it in detail because this indicator serves exactly the purpose of showing if status if worth opening.

Maybe we could find anything super subtle or make it optional? Don't want to deviate from web too much.
Also I don't think it's needed inside thread?

Tak:
Oh, I thought we would want to keep parity with the web ui (which also has them in threads).

I find them nice for knowing whether there's a point to expanding the thread (even more in Tusky, where it doesn't just populate a column, but replaces the full screen view), but I don't feel _super_ strongly about it.

charlag:
Oh, I've never noticed them in threads, actually. My bad.

-- 
Sincerely,
Ivan



13 Oct 2018, 07:52 by notifications@github.com:

charlag:
@connyduck are you still against this? I don't notice them in web anymore but I don't find them terribly useful either

connyduck:
> @connyduck are you still against this? 

yes (sorry @Tak). At least for main timelines. We could add it for detailed statuses, but doesn't make too much sense since one can already see all replies there.

PR #892, opened 2018-10-21
[WIP] Simple notifications streaming

charlag:
This is a moonshot to get realtime notifications to Tusky (mostly when app is open for now).
It is based on #809 for no real reason other than avoiding future conflicts.
I've tried and it works but it needs UX changes, like not sending system notifications when notifications are open and figuring out gaps behavior.


charlag:
I should probably take the latest commit and rebase it.

PR #894, opened 2018-10-22
Fix media transitions, reduce code duplication between media fragments

Tak:


app/src/main/java/com/keylesspalace/tusky/fragment/ViewMediaFragment.kt
```diff
@@ -60,14 +68,32 @@ abstract class ViewMediaFragment : BaseFragment() {
...
         toolbarVisibiltyDisposable = (activity as ViewMediaActivity).addToolbarVisibilityListener(object: ViewMediaActivity.ToolbarVisibilityListener {
             override fun onToolbarVisiblityChanged(isVisible: Boolean) {
                 onToolbarVisibilityChange(isVisible)
             }
         })
     }
 
+    protected fun updateDescriptionVisibility(visible: Boolean) {
+        // Setting visibility without animations so it looks nice when you scroll media
+        if (visible) {
```
charlag: you can use `View.visible()` extension and maybe we don't need this method then

app/src/main/java/com/keylesspalace/tusky/fragment/ViewImageFragment.kt
```diff
@@ -132,38 +130,21 @@ class ViewImageFragment : ViewMediaFragment() {
         val arguments = this.arguments!!
         val attachment = arguments.getParcelable<Attachment>(ARG_ATTACHMENT)
         val url: String?
+        var description : String? = ""
```
charlag: if it's optional, why set it to empty string?

Tak:
Yeah, missed a situation where the description could be null, fixed now

charlag:
Hi! Much better now. But I've found some problems (they were probably introduced in the previous PR)
<details>
<summary>
When video takes the whole screen, image caption is not visible. Maybe putting it on top of controls is a better idea
</summary>

screenshot_tusky_20181027-165304
</details>

<details>
<summary>
When you change between from video to picture video controls are still visible for some time
</summary>

screenshot_tusky_20181027-165320

</details>

Also it feels like video controls and action bar are not properly synchronized.

As this was introduced not in this PR, I don't consider it blocking in any way but I thought I would let you know.

charlag:
There's still a little bit of un-synchronousness between action bar and video controls but it's already much better and I think can be improved later

Issue #869, opened 2018-10-02
Poor behaviour after entering too long(?) an image caption


aziraphale:
I just entered quite a long caption for an attached image and when I tapped "OK" Tusky just popped up a toast saying "Unable to set caption." and took me back to the main Compose window, thus losing the content of that rather long caption that I'd just spent ages typing out 😕 (I did re-enter the caption dialog to confirm that it had indeed been lost, and was presented with an empty text box.)

I'm assuming that the caption that I entered was too long, but obviously without a more descriptive error message I can't know for sure!

At the very least Tusky needs to temporarily store the entered caption when it fails, and re-fill the caption dialog's text field when re-opening it.

Preferably it'd also display a more useful error message (does the Mastodon API not return a more specific error?).

And ideally the caption dialog would indicate the length of your caption and/or the remaining allowed number of characters. The caption dialog could do with a multi-line text box too, really. Having to scroll left and right to change something or even just to see what you've written gets super frustrating!

That said, Tusky is a million miles ahead of the other Mastodon Android clients that I've tried! And this is one of very few issues that I've had with it. So very well done to the dev(s) for that! <3 ❤️ 

Thanks! :)

* * * *
- Tusky Version: 3.1
- Android Version: 8.1.0
- Android Device: Nexus 6P
- Mastodon instance (if applicable): gravitas.cafe (but probably not relevant)

- [✔️] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

charlag:
Hi! Thanks for reporting.
It's totally our fault but it's also a fault of Mastodon - there's no way to get image description length nor it's documented (but set to 420 in https://github.com/tootsuite/mastodon/blob/master/app/models/media_attachment.rb#L94).
So the biggest thing we can do is to save description if it fails (or not close the dialog).

connyduck:
I think we should actually blow up the caption dialog to a whole activity. Then we have enough space to add multiline textfield and later maybe even some image manipulation tools like rotate etc.

charlag:
It makes sense in context of editing but in general I would say that people don't like changing screens and modal seems like a better fit to me here.

kitkatloyalist:
As with Firefox forms the lesson is to compose in a text editor then copy to final posting form. 

i hope my pain will spare you similar vexation

aziraphale:
> As with Firefox forms the lesson is to compose in a text editor then copy to final posting form.

[...]

PR #898, opened 2018-10-24
Media description improvements

Tak:
- Enforce media description character limit in entry UI
- Store(/restore) media descriptions when saving toots

Addresses #869


Tak:
Yeah, I don't like the hint either. Ideally, we'd use a countdown like the compose activity, but that's a lot of refactoring etc. and then we end up writing a whole new activity for editing descriptions.
(Which maybe should be done, but this seems like a decent incremental improvement until somebody gets the time/motivation.)

charlag:
Yeah, I don't think we need countdown (even though it's not so hard to write)

app/src/main/res/values/strings.xml
```diff
@@ -305,7 +305,7 @@
     <string name="compose_active_account_description">Posting with account %1$s</string>
 
     <string name="error_failed_set_caption">Failed to set caption</string>
-    <string name="hint_describe_for_visually_impaired">Describe for visually impaired</string>
+    <string name="hint_describe_for_visually_impaired">Describe for visually impaired (%d character limit)</string>
```
connyduck: you should change all other languages as well, or delete the  string in other languages

app/src/main/java/com/keylesspalace/tusky/ComposeActivity.java
```diff
@@ -429,13 +433,19 @@ public void onFailure(@NonNull Call<List<Emoji>> call, @NonNull Throwable t) {
...
             String savedJsonUrls = intent.getStringExtra(SAVED_JSON_URLS_EXTRA);
+            String savedJsonDescriptions = intent.getStringExtra(SAVED_JSON_DESCRIPTIONS_EXTRA);
             if (!TextUtils.isEmpty(savedJsonUrls)) {
-                // try to redo a list of media
                 loadedDraftMediaUris = new Gson().fromJson(savedJsonUrls,
                         new TypeToken<ArrayList<String>>() {
                         }.getType());
             }
+            if (!TextUtils.isEmpty(savedJsonDescriptions)) {
+                loadedDraftMediaDescriptions = new Gson().fromJson(savedJsonDescriptions,
```
connyduck: I don't like the `new Gson()` because of performance. I think you should be able to inject a shared Gson instance into the activity.

app/src/main/java/com/keylesspalace/tusky/ComposeActivity.java
```diff
@@ -551,10 +562,11 @@ public void afterTextChanged(Editable s) {
 
         // These can only be added after everything affected by the media queue is initialized.
         if (!ListUtils.isEmpty(loadedDraftMediaUris)) {
-            for (String uriString : loadedDraftMediaUris) {
-                Uri uri = Uri.parse(uriString);
+            for (int mediaIndex = 0; mediaIndex < loadedDraftMediaUris.size(); ++mediaIndex) {
+                Uri uri = Uri.parse(loadedDraftMediaUris.get(mediaIndex));
                 long mediaSize = getMediaSize(getContentResolver(), uri);
-                pickMedia(uri, mediaSize);
+                String description = loadedDraftMediaDescriptions == null ? null : loadedDraftMediaDescriptions.get(mediaIndex);
```
connyduck: Just had a IndexOutOfBoundsException here after trying to open a toot that was saved on master. I guess a range check would fix this.

Issue #648, opened 2018-05-10
Support the focal point feature introduced in Mastodon 2.3


connyduck:
https://github.com/tootsuite/mastodon/releases/tag/v2.3.0

- [ ] Set image previews correctly according to their focal points
- [ ] Support setting a focal point when uploading media

I think we should definitely implement the first point, not sure if we really need the second.

iknownothingaboutcomputers:
@connyduck I think both are needed. It really helps.

iknownothingaboutcomputers:
ETA?

connyduck:
its not planned currently

jfml:
I would really love to have this! Just went through the trouble of going to the desktop to repost a toot that had an image where the head was cut off in the preview +___+

Tak:
Implemented in #899 and #2620

PR #899, opened 2018-10-25
Set image previews correctly according to their focal points

jchmrt:
This PR is in relation to issue #648, specifically it aims to close the first point of this issue: "Set image previews correctly according to their focal points". This is my first time contributing to Tusky and it was some time ago since I last wrote android code, so if there are any issues with my PR please let me know :)

### Summary
Currently Tusky simply centers all image previews, not honouring the focal point information. This PR adds code which ensures that the image previews of media attachments to toots are correctly cropped to always show the focal point of the image (if it is specified). It should not in any way influence how previews of media without a focal point are shown, these continue to simply be centered.

For some general information about the focal point feature of mastodon, see [here](https://github.com/jonom/jquery-focuspoint#1-calculate-your-images-focus-point). In a nutshell, the focal point information is a tool to allow users to specify which part of the image is the focus. Client applications are then expected to try to insure that this point is visible at all times.

### Implementation
The first component of the implementation is the addition of two data classes, one for the meta-data field of the attachment and one for the focus object within the meta-data. These classes are not complete, but simply include the information that is used at this time. If at some point more of the meta-data information is needed, the class can be easily extended to include the other properties.

To achieve the correct crop on the image based on this information a few additional components were
needed:

First of all we needed a way to influence how the image is cropped into the ImageView. It turns out that the preferred way to do this is by setting the ScaleType to MATRIX and adjusting the matrix of the
image as needed. This matrix allows us to scale and transform the image in the way we need to make sure that the focal point is visible within the view. For this purpose we have the FocalPointEnforcer which can calculate and set the appropriate matrix on an ImageView as soon
as the image is loaded.

However a second problem is that we need to make sure that this matrix is updated whenever the size of the ImageView changes. The size might change for example because the orientation of the device changed from portrait to landscape or vice versas, or for a number of other reasons such as the screen being split vertically or something like that.

To be able to hook onto this event we need to create a new extended version of the ImageView class, which we call MediaPreviewImageView. This class behaves exactly the same as a normal ImageView, however if the focalPointEnforcer of this view is set, then it will call this enforcer to update the image matrix any time the size is changed.

So this PR changes all media previews in the item_status.xml and item_status_detailled.xml layout files to the new MediaPreviewImageView class. Additionally in the code for loading the images into the previews a new case is added which tests if there is a focus attribute in the meta-data. If so it makes sure to create and set the FocalPointEnforcer.

### Testing
Some rudimentary unit testing of the FocalPointEnforcer's calculations is done in the FocalPointEnforcerTest. However the rest of the code is data classes and UI code and can therefore not easily be tested meaningfully using automated tests. For that reason I did a lot of manual testing of the UI. 

A good focussed manual test to ensure that the focal points are honoured is opening [this test profile I made](https://mastodon.social/@testing12345678) in Tusky and looking at the timeline. This timeline contains 6 very tall images and 6 very wide ones. In each of these images I added a red dot and put the focal point on this dot. In the current version of Tusky looking at this timeline will show a lot of previews with just white in them. But in the version of this PR, each of the previews will show the red dot.


jchmrt:
Hmm, the CI is failing, I'm looking into it now.

charlag:
For Circle it was OOM (it's normal, I'll restart it) but bitrise doesn't fail usually.
I'll try to find energy to look into it tomorrow but more likely on the weekend. thanks for the PR!

jchmrt:
I'm not really sure what exactly the bitrise one is testing, so I'll leave it to you to look into that. If you need anything from me just let me know!

app/src/main/java/com/keylesspalace/tusky/util/FocalPointEnforcer.kt
```diff
@@ -0,0 +1,192 @@
...
+ */
+class FocalPointEnforcer(private val imageView: ImageView, private val focus: Focus) : Callback {
+    // This companion object defines the helper functions used:
+    companion object {
+        /**
+         * Create the appropriate matrix for the given parameters.
+         *
+         * How it works is using the following steps:
+         *   - First we determine if the image is to wide or to tall for the view size. If it is
+         *   to wide, we need to crop it horizontally and scale the height to fit the view
```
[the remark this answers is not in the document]
jchmrt: Fixed :heavy_check_mark:

app/src/test/java/com/keylesspalace/tusky/FocalPointEnforcerTest.kt
```diff
@@ -0,0 +1,108 @@
...
+import org.junit.Assert
+import org.junit.Test
+
+class FocalPointEnforcerTest {
+    private val eps = 0.01f
+
+    // focal[X|Y]ToCoordinate tests
+    @Test
+    fun positiveFocalXToCoordinateTest() {
+        Assert.assertEquals(FocalPointEnforcer.focalXToCoordinate(0.4f), 0.7f, eps)
```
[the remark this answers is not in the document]
jchmrt: Good point, I didn't know that was possible in Kotlin. Fixed :heavy_check_mark:

jchmrt:
Thanks for the review! You make some good points about the optimization.

> Now we allocate enforcer every time we need to adjust matrix.

Do you mean that we allocate the enforcer every time the size of the view is changed? Because I don't think this is true, as far as I know we only allocate an enforcer once for every image preview on [this line](https://github.com/tuskyapp/Tusky/pull/899/files#diff-7d674936cb4f30ba9df00ab02fc2955fR274). 

Or do you mean that we allocate a different enforcer for each image preview? This is correct, I suppose we could make use of a single enforcer and just pass it the parameters it needs each time it needs to update a matrix, if you think that would make a significant performance difference.

> One way is to do what ```PhotoViewAttacher``` does: assign itself a listener. Then we don't need custom view class.

I'm not sure how that could be done, since I don't know of any listener which gives events at the right time of resizing. When I was initially creating this feature I tried to do it using a listenerer, but none of the listener types I tried resulted in the correct behaviour regarding resizes. The listener I tried which definitely don't create signals at the right time is the ```OnGlobalLayoutListener```, this one was not called when the orientation of the screen changed, so it didn't work.

Another listener I tried was the ```OnPreDrawListener```, which worked somewhat. However this was very inefficient since it would result in extra code being run on each draw of the screen. So in the end the only way to be called only when the size of the view changed I found was to extend the class and overwrite the ```onSizeChanged``` method. I tried searching for listeners which did the same but I couldn't find any, so if you know of such a listener that would be a great solution!

> Another way it to move all this listening logic into our custom view class and only use ```FocalPointEnforcer.```
The advantage is that we don't allocate new objects.

So if I understand it correctly you mean that we could use only a single static ```FocalPointEnforcer``` and then move the code which updates the matrix and gets the size of the image and view into the view itself. That way we allocate less objects and centralize the code concerning the view into the view object itself. Is that what you mean? 

If so I think that seems like a good solution to ensure less unnecessary objects. If we combine that with reusing the same matrix within the view, I think that would solve the performance issues in an elegant way.

Again thanks a lot for the thorough review!

app/src/main/java/com/keylesspalace/tusky/entity/Focus.kt
```diff
@@ -0,0 +1,31 @@
+/* Copyright 2017 Jochem Raat <jchmrt@riseup.net>
```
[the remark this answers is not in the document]
charlag: Yeah, that works for me!

charlag:
I meant that we allocate both enforcers and `Matrix` objects. We keep save `Matrix` object as a field and reset it when necessary instead of allocating via `Matrix()` on each resize. Please, See `PhotoViewAttacher` which we use in the project for an example.

The listener I had in mind as an option is `addOnLayoutChangeListener()`.

Yeah, that's what I meant about enforcer: only leave static methods inside of it.

jchmrt:
I just did a small experiment and the ```addOnLayoutChangeListener()``` seems to work correctly, great suggestion :+1: .

So I guess we just have to choose between just having the ```FocalPointEnforcer``` attach itself as a listener using the example of the PhotoViewAttacher or making the enforcer a static utils object and moving all the code handling inside of the view itself, right? I'm fine with both options, so I'd just like to hear what you and the other contributors think the best option is. Then I'll work on rewriting the code to that model after that.

app/src/main/java/com/keylesspalace/tusky/entity/Focus.kt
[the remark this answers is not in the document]
jchmrt: I pushed another commit which moves the classes into the Attachment class. Could you take a look if the way I did this is idiomatic Kotlin?I don't have a lot of experience with Kotlin, so I'm not sure.

charlag:
I would prefer subscribing in a custom view. it's better separation of concerns and less allocations.

jchmrt:
Alright, great! I hope I'll be able to start working on that in the upcoming week, but I will be pretty busy so it might take somewhat longer.

charlag:
@jchmrt no pressure, we're pretty packed for the next release anyway

jchmrt:
I think I have incorporated all changes you suggested in this new version:

- I have moved all listening code into the custom view, now only using the separate FocalPointEnforcer class as a static util object (now called FocalPointUtil) which is used to abstract the calculations away from the custom view.
- For each view only only one Matrix is allocated, which is then reused for each resizing.

Could you take a look at the new changes and let me know what you think? Thanks :)

app/src/main/java/com/keylesspalace/tusky/view/MediaPreviewImageView.kt
```diff
@@ -15,45 +15,67 @@
...
     /**
-     * Set the focal point enforcer for this view.
+     * Set the focal point for this view.
      */
-    fun setFocalPointEnforcer(enforcer: FocalPointEnforcer) {
-        this.focalPointEnforcer = enforcer
+    fun setFocalPoint(focus: Attachment.Focus) {
+        scaleType = ScaleType.MATRIX
+        this.focus = focus
+        this.focalMatrix = Matrix()
```
charlag: I would allocate it only if it's not already there

app/src/main/java/com/keylesspalace/tusky/util/FocalPointUtil.kt
```diff
@@ -0,0 +1,161 @@
...
+     * @param viewHeight The height of the imageView
+     * @param imageWidth The width of the actual image
+     * @param imageHeight The height of the actual image
+     * @param focus The focal point to focus
+     * @param mat The matrix to use, this matrix is reset() and then updated with the new
+     * configuration. We reuse the old matrix to prevent unnecessary allocations.
+     *
+     * @return The matrix which correctly crops the image
+     */
+    fun updateFocalPointMatrix(viewWidth: Float,
```
charlag: I think returning it is a little bit redundant but maybe it doesn't hurt to be have more flexible interface

charlag: On the other hand, it's less clear that matrix is mutated if it's also returned

app/src/main/java/com/keylesspalace/tusky/view/MediaPreviewImageView.kt
```diff
@@ -15,45 +15,67 @@
...
+    }
+
+    /**
+     * Called when the size of the view changes, it calls the FocalPointUtil to update the
+     * matrix if we have a set focal point. It then reassigns the matrix to this imageView.
      */
     override fun onSizeChanged(width: Int, height: Int, oldWidth: Int, oldHeight: Int) {
-        if (drawable != null) {
-            focalPointEnforcer?.updateFocalPointMatrix()
+        if (drawable != null && focus is Attachment.Focus && focalMatrix is Matrix) {
```
charlag: I would replace `is` checks with `!= null`. `is` is more about type checking

app/src/main/java/com/keylesspalace/tusky/view/MediaPreviewImageView.kt
jchmrt: Good point, I've changed it :+1:

app/src/main/java/com/keylesspalace/tusky/util/FocalPointUtil.kt
jchmrt: I think you're right, it's clearer that it mutates the matrix if it doesn't return it. I've changed it to not return the matrix.

app/src/main/java/com/keylesspalace/tusky/view/MediaPreviewImageView.kt
jchmrt: Good point, I changed this as well :+1:

connyduck:
Setting the focus point seems to work fine and I would really love to merge this - but it is crashing when opening Image fullsize because the PhotoView library does not support Matrix scale type:
https://github.com/chrisbanes/PhotoView/blob/296b7839ea1673abd5f44656b203e08273fed9a9/photoview/src/main/java/com/github/chrisbanes/photoview/Util.java#L29

😔

jchmrt:
That's weird it wasn't crashing for me when opening the full view when I tried it earlier. I'll have a look and try and see what's changed.

jchmrt:
Hmm, it seems that for some reason with the new commits from master merged back into this branch, the scaleType of the PhotoView for the fullsize image is set to the scaleType of the MediaPreviewImageView. I don't think this should be happening at all, but I'm not sure where in the code this is being done at all. The crash can be prevented by inserting the following function into the MediaPreviewImageView class:

````kotlin
    override fun getScaleType(): ScaleType {
        return ScaleType.CENTER_CROP
    }
````

Which makes sure that the PhotoView is now simply set to CENTER_CROP for the images with a focal point as well. Using this fix the feature works as expected again and it doesn't crash when opening the full size.

However I'm not sure if this is the right solution, since I think the PhotoView should not be taking its scale type from the MediaPreviewImageView at all. Do you have any idea what might have changed to cause this to happen @connyduck?

**Comment by charlag** [comment]
My guess is that it's a transition code by Android

**Comment by jchmrt** [comment]
Ah, that might be it. When I insert a stacktrace into the modified getScaleType() function it does show the Transition class from android in there. I'm trying to look into what's changed which might cause this different behaviour, but I'm having some trouble figuring it out since I'm not that familiar with android transitions.

If anyone else with some more experience with android and/or transitions could look into it that would be great, but if not I'll try and look at it some more in the upcoming weeks but I might not get to it soon since my finals are starting this week.

**Comment by connyduck** [comment]
```
    override fun getScaleType(): ScaleType {
        return ScaleType.CENTER_CROP
    }
```

Yes I think thats our best option here. I investigated, the scale Type of the `photoView` is indeed set by transition code.

Another thing I noticed: I think the `MediaPreviewImageView` needs to be reset to `centerCrop` when the image displayed has no focus point information attached. Its used in an `RecyclerView`, one Instance may show an image with info, only to be recycled and show one without info.

connyduck:
@jchmrt can you do this small change so I can merge this PR please?

connyduck:
now the build is failing - you need to move `MediaPreviewImageView.kt` to AndroidX

Issue #835, opened 2018-09-04
Better support of Glitch-Soc


jusdepatate:
Glitch-soc add some good things:
```
What’s different?

glitch-soc adds a number of experimental features to Mastodon, such as:

    Media improvements
        Images inside the CW spoiler
        fullwidth images
        scaling options
    An app settings modal
    Collapsible toots
    Toot visibility icons
    Local-only toots
    Threaded mode
    data-* attributes on statuses for custom CSS targeting
    Advanced theming via flavours+skins
    Bookmarks
```
(from official docs)

* * * *
- Tusky Version: Tusky Nightly 3.0 1-791a87e
- Android Version: EMUI 8.0.0 (Android 8.0.0)
- Android Device: Huawei P10 Lite
- Mastodon instance (if applicable) : 
```
Who is running glitch-soc?

glitch-soc is currently in use on the following instances:

    dev.glitch.social
    toot-lab.reclaim.technology
    sdfn-01.ninjawedding.org
    ostatus.lardbucket.org
    scifi.fyi
    sprite.land
    awoo.space
    monsterpit.net
    donphan.social
    aleph.land
    im-in.space
    social.wxcafe.net
    imaginair.es
    soc.ialis.me
    crazynoisybizarre.town
```
(From official docs)

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

charlag:
hi! thanks for reporting
It's not clear from your post which features you would like us to integrate into Tusky

jusdepatate:
o/, Local only and bookmarks could be great on Teuky

ashkitten:
i'd love to see bookmarks make their way in. not sure i'd use tusky still, over the webapp, but it'd certainly make it much closer

ashkitten:
hey, just wanted to update this issue a bit with current data - as of right now according to the-federation.info, there are around 55 glitchsoc instances with an estimated 20,000 users total. i'd like to request that implementing relevant features such as bookmarks become higher priority, if possible

tastytea:
Pleroma also implements the bookmarks API and has 17879 users according to [fediverse.network](https://fediverse.network/pleroma).

ariasuni:
Bookmarks have been implemented since it’s been added to Mastodon upstream.

So the main things that Tusky could implement are local-only posts and support for content-type.

rscmbbng:
+1 for local only!

Tak:
[...]

PR #901, opened 2018-10-26
Add visibility icons to statuses

wbrawner:
Hi! :wave:

I realize this isn't a part of standard Mastodon though I'm a fan of the features requested in #835 and the data is already there so it may as well be used, right? I'm open to any suggestions for improvements here. Screenshots below:

EDIT: I just realized that devices that have modified the system font size end up with really bad alignment with the icon/text. I'll work on fixing that but I'd appreciate any comments/concerns in the meantime :) Also CI doesn't seem to be happy so I'll fix whatever I broke :P

image
image

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/adapter/StatusDetailedViewHolder.java
+++ b/app/src/main/java/com/keylesspalace/tusky/adapter/StatusDetailedViewHolder.java
@@ -197,7 +197,7 @@ class StatusDetailedViewHolder extends StatusBaseViewHolder {
                 .addOnGlobalLayoutListener(new ViewTreeObserver.OnGlobalLayoutListener() {
                     @Override
                     public void onGlobalLayout() {
-                        final int size = statusTimestampInfo.getHeight();
+                        final int size = (int) statusTimestampInfo.getTextSize();
                         visibilityDrawable.setBounds(
                                 0,
                                 0,
--- a/app/src/main/java/com/keylesspalace/tusky/adapter/StatusDetailedViewHolder.java
+++ b/app/src/main/java/com/keylesspalace/tusky/adapter/StatusDetailedViewHolder.java
@@ -3,6 +3,7 @@ package com.keylesspalace.tusky.adapter;
 import android.content.ClipData;
 import android.content.ClipboardManager;
 import android.content.Context;
+import android.content.res.Resources;
 import android.support.annotation.Nullable;
 import android.text.SpannableStringBuilder;
 import android.text.Spanned;
@@ -11,8 +12,10 @@ import android.text.method.LinkMovementMethod;
 import android.text.style.URLSpan;
 import android.view.View;
 import android.view.ViewGroup;
+import android.view.ViewTreeObserver;
 import android.widget.ImageView;
 import android.widget.LinearLayout;
+import android.widget.RelativeLayout;
 import android.widget.TextView;
 import android.widget.Toast;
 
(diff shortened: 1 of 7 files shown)


connyduck:
Im fine with having this on the detailed status, but not in timelines. There is already not enough space for everything in this one line.

charlag:
Same here

wbrawner:
Yeah, it does seem a bit cluttered. I went ahead and pulled it out of the timeline statuses. I also implemented dynamic scaling of the visibility icon based on the font size of the timestamp, so that should look a lot better now on devices where users have adjusted the system font size.

wbrawner:
I moved the logic for selecting the icon into the `setStatusVisibility()` method, since it's most likely the only place it'll be used as far as I can tell. I also had to migrate `setStatusVisibility()` into the `StatusDetailedViewHolder` class since I no longer was able to check for the existence of the `ImageView`. I used `TextView.getTextSize()` for the height of the icon, and it seems to be closer, but still off by a couple of pixels. If you have any suggestions for how to get it pixel-perfect I'd appreciate it!

charlag:
Thanks for the changes! I'll look up how to make drawable of the same size.

wbrawner:
I've been doing some research into this myself as well, and I think that `[EditText#getTextSize()](https://developer.android.com/reference/android/widget/TextView#getTextSize())` is as close as we'll get with the sizing of the drawable. From my tinkering around with it, it seems that the little bit of extra space on the top that the drawable extends past is still considered as the text height, as accents and umlauts extend up into that space on capitalized letters.

charlag:
@wbrawner extra space is not a problem, I think it's as intended.
Another option would be taking sp and calculating text size by hand.
Here are some options: https://stackoverflow.com/questions/2069810/how-to-assign-text-size-in-sp-value-using-java-code

wbrawner:
> Another option would be taking sp and calculating text size by hand

I think that might be a bit of unnecessary extra work. According to the docs for `getTextSize()`, it should already be returning the size in pixels of the text, which is what I need to give to the drawable for its size. At any rate, I'll give it a shot and see if it gives me anything different from `getTextSize()`

charlag:
It just may be easier than setting a listener etc but honestly I don't mind, it's not a timeline.

wbrawner:
Ah, I see. Yeah the listener was necessary to wait until the TextView had already been drawn, but I can see if the `sp` -> `px` conversion is any simpler without losing accuracy

charlag:
I may have a look again but I think that w need decision by @connyduck

app/src/main/java/com/keylesspalace/tusky/adapter/StatusDetailedViewHolder.java
```diff
@@ -154,6 +155,55 @@ void setupWithStatus(final StatusViewData.Concrete status, final StatusActionLis
...
+                break;
+            case DIRECT:
+                visibilityIcon = R.drawable.ic_email_24dp;
+                break;
+            default:
+                visibilityIcon = 0;
+        }
+
+        if (visibilityIcon == 0) {
+            return;
```
connyduck: you could just `default: return` above and save these three lines

wbrawner:
Fixed. Sorry for the delay, been a bit busy lately :)

connyduck:
thx!

kyori19:
I think this is mistake.
It should be `ic_lock_outline`, not `ic_lock_open`.

https://github.com/tuskyapp/Tusky/blob/1c34d21a2327e0bf7111ec0b04fcc74d37efa1b5/app/src/main/java/com/keylesspalace/tusky/adapter/StatusDetailedViewHolder.java#L175

connyduck:
Yes you are right @kyori19 
Would you mind to send a pull request?

Issue #770, opened 2018-08-18
Honor the visible_in_picker attribute of emojis


connyduck:
https://github.com/tootsuite/documentation/blob/master/Using-the-API/API.md#emoji

Only show emojis in the picker if `visible_in_picker=true`.

PR #907, opened 2018-10-30
Filter emoji where "visible_in_picker" is set to false.

Tak:
Addresses #770

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/entity/Emoji.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/entity/Emoji.kt
@@ -16,10 +16,12 @@
 package com.keylesspalace.tusky.entity
 
 import android.os.Parcelable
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/keylesspalace/tusky/entity/Emoji.kt
```diff
@@ -16,10 +16,14 @@
...
+import android.support.annotation.Nullable
+import com.google.gson.annotations.SerializedName
 import kotlinx.android.parcel.Parcelize
 
 @Parcelize
 data class Emoji(
         val shortcode: String,
-        val url: String
+        val url: String,
+        @Nullable
```
Tak: Not sure if this is needed, since the type is `Boolean?`, but android studio complained about possible NPE at the call site without it, so...

PR #909, opened 2018-11-05
Do not show "Pin" in "more" menu of direct status.

kyori19:
`status.getPinned()` generates NullPointerException when show menu of direct toot.
So I made a new method which returns only true or false.
https://github.com/accelforce/Tusky/blob/7e4798e2e37923afd90065cf186c397aa931c489/app/src/main/java/com/keylesspalace/tusky/entity/Status.kt#L105-L107
And changed not to show "Pin" option in direct toot's menu.
https://github.com/accelforce/Tusky/blob/7e4798e2e37923afd90065cf186c397aa931c489/app/src/main/java/com/keylesspalace/tusky/fragment/SFragment.java#L156-L163


connyduck:
Yeah noticed this as well recently, thank you for fixing!

PR #916, opened 2018-11-11
fix link's too large hitbox

### LINKED ISSUE #846: Links have a too large hitbox
When there's a link right at the end of the toot text, hitting anywhere to the right of it will open that link instead of the toot, as usual. That's kind of an inconvenience. "Link" includes ats and hashtags, so all blue text.

* * * *
- Tusky Version: 3.1 (beta)
- Android Version: 8.0
- Android Device: LG G6 h870

kyori19:
Fix #846 .


kyori19:
I didn't noticed to use non-width space.
I confirmed there's nothing wrong with using it.
And I also added comments to these codes.

PR #932, opened 2018-11-26
remove unnecessary compat code

connyduck:
Removes compat code that is no longer necessary since the upgrade to min api 21.
Other compat code or where I was not sure was left as it is.


charlag:
If we're not using library, I would even consider keeping compat code. It often fixes framework bugs.
In this places it seems to be safe, though.
Do we also need to migrate to AndroidX later?

connyduck:
Yes, they are all save, they just call through to the platform with a now obsolete `if SDK_INT` check.

Upgrade to AndroidX will be one of my next PRs :)

PR #942, opened 2018-12-02
Apply suggestions for ProgressImageView

charlag:
I didn't start it because of the cable problems, can do it a little bit later. If docs are correct, this should be a safe change


connyduck:
I tried it out, it works as intended. Thanks!

charlag:
Thanks!

PR #954, opened 2018-12-07
Display mentions toot hidden

### LINKED ISSUE #292: Display @ mentions outside the content warning

ariasuni:
Fix #292 

The core of the functionality is here but there is still some bugs:

- Toots aren’t hidden by default even if they should be
- On timelines (it works on notifications mentions and threaded view), toots with clickable elements don’t hide (but somehow make a hide animation)

So it definitely looks like what I do in `StatusBaseViewHolder` is overridden under some conditions.

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/adapter/StatusBaseViewHolder.java
+++ b/app/src/main/java/com/keylesspalace/tusky/adapter/StatusBaseViewHolder.java
@@ -9,6 +9,9 @@ import android.support.v7.content.res.AppCompatResources;
 import android.support.v7.widget.RecyclerView;
 import android.text.InputFilter;
 import android.text.Spanned;
+import android.text.Spannable;
+import android.text.SpannableStringBuilder;
+import android.text.style.URLSpan;
 import android.text.TextUtils;
 import android.view.View;
 import android.view.ViewGroup;
@@ -36,6 +39,7 @@ import java.text.SimpleDateFormat;
 import java.util.Date;
 import java.util.List;
 import java.util.Locale;
+import java.lang.StringBuilder;
 
 import at.connyduck.sparkbutton.SparkButton;
 import at.connyduck.sparkbutton.SparkEventListener;
@@ -50,6 +54,8 @@ abstract class StatusBaseViewHolder extends RecyclerView.ViewHolder {
     private SparkButton reblogButton;
     private SparkButton favouriteButton;
     private ImageButton moreButton;
+    private Spannable contentText;
+    private Spannable mentionsOnlyContentText;
     private boolean favourited;
     private boolean reblogged;
     private ImageView[] mediaPreviews;
@@ -80,6 +86,8 @@ abstract class StatusBaseViewHolder extends RecyclerView.ViewHolder {
         reblogButton = itemView.findViewById(R.id.status_reblog);
         favouriteButton = itemView.findViewById(R.id.status_favourite);
         moreButton = itemView.findViewById(R.id.status_more);
+        contentText = null;
+        mentionsOnlyContentText = null;
         reblogged = false;
         favourited = false;
         mediaPreviews = new ImageView[]{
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/keylesspalace/tusky/util/LinkHelper.java
```diff
@@ -117,20 +117,74 @@ public void onClick(View widget) {
...
+        boolean firstMention = true;
+        for (Status.Mention mention : mentions) {
+            String accountUsername = mention.getLocalUsername();
+            final String accountId = mention.getId();
+            ClickableSpan customSpan = new ClickableSpanNoUnderline() {
+                @Override
+                public void onClick(View widget) { listener.onViewAccount(accountId); }
+            };
+
+            customSpan = new CustomURLSpan(mention.getUrl()) {
```
connyduck: The `ClickableSpan` is created twice, only the second one is used. I believe the first one is the one we want, as it opens profiles directly and not via url.

app/src/main/java/com/keylesspalace/tusky/adapter/StatusBaseViewHolder.java
```diff
@@ -399,17 +407,28 @@ private void setSpoilerText(String spoilerText, List<Emoji> emojis,
...
         contentWarningButton.setVisibility(View.GONE);
-        content.setVisibility(View.VISIBLE);
+        this.setTextVisible(true);
+    }
+
+    private void setTextVisible(boolean visible) {
+        if (visible) {
+            content.setText(this.contentText);
+            this.content.setVisibility(View.VISIBLE);
+        } else if (this.mentionsOnlyContentText == null) {
```
connyduck: I think this method is called prior to the `mentionsOnlyContentText` being created. So the first time this will always be null. You could try creating the `mentionsOnlyContentText` here instead of making it a member variable.

app/src/main/java/com/keylesspalace/tusky/adapter/StatusBaseViewHolder.java
```diff
@@ -399,17 +407,28 @@ private void setSpoilerText(String spoilerText, List<Emoji> emojis,
...
+            this.content.setVisibility(View.GONE);
+        } else {
+            content.setText(this.mentionsOnlyContentText);
```
connyduck: I think you need to make the `content` `VISIBLE` again here.

app/src/main/java/com/keylesspalace/tusky/adapter/StatusBaseViewHolder.java
```diff
@@ -123,8 +129,10 @@ private void setUsername(String name) {
     private void setContent(Spanned content, Status.Mention[] mentions, List<Emoji> emojis,
                             StatusActionListener listener) {
         Spanned emojifiedText = CustomEmojiHelper.emojifyText(content, emojis, this.content);
-
         LinkHelper.setClickableText(this.content, emojifiedText, mentions, listener);
+        this.contentText = this.content.getText();
```
connyduck: Instead of saving the text as a member var I would just create it when needed.

app/src/main/java/com/keylesspalace/tusky/adapter/StatusBaseViewHolder.java
```diff
@@ -9,6 +9,7 @@
 import android.support.v7.widget.RecyclerView;
 import android.text.InputFilter;
 import android.text.Spanned;
+import android.text.style.URLSpan;
```
connyduck: unused import

app/src/main/java/com/keylesspalace/tusky/adapter/StatusBaseViewHolder.java
ariasuni: Well, to create `contentText` later, I need to save `content` or `emojifiedText`, `mentions` and `listener` somewhere (same for `mentionsOnlyContentText`).

app/src/main/java/com/keylesspalace/tusky/adapter/StatusBaseViewHolder.java
ariasuni: If `mentionsOnlyContentText` is initialized before this function, its content never changes so for a given toot, it can only enter either line 428 or line 430.

ariasuni:
It seems fully working. Now `mentionsOnlyContentText` is created only if the toot can be hidden. I’m not sure if [`LinkHelper.makeMentionsText()` should be called here](https://github.com/tuskyapp/Tusky/pull/954/commits/e7e673e1f62799a1c4d091f0aa767581a74bad15#diff-7d674936cb4f30ba9df00ab02fc2955fR535) or in `setSpoilerText()` but I think it’s quite good.

ariasuni:
I think the code is organized better now.

app/src/main/java/com/keylesspalace/tusky/util/LinkHelper.java
```diff
@@ -117,20 +117,63 @@ public void onClick(View widget) {
...
 
+    /**
+     * Put mentions in a piece of text and makes them clickable, associating them with callbacks to
+     * notify when they're clicked.
+     *
+     * @param view the returned text will be put in
+     * @param mentions any '@' mentions which are known to be in the content
+     * @param listener to notify about particular spans that are clicked
+     */
+    public static void setClickableMentions(
```
charlag: This method could return Spaanable or something like that but some other methods also take TextView :thinking:

connyduck:
Its not perfect, but I think its fine to merge now? @charlag

charlag:
I think it's good

PR #955, opened 2018-12-10
sort emojis alphabetically

pupper68k:
Hello!
One thing that has always annoyed me about Tusky is that when using emojis I have to search for the correct one.

Many instances have over 100 emojis enabled, and Tusky's emoji picker does not handle this well.
Most instances have sets of emojis that act as font replacements and it is annoying to have to run around and find each letter.

Here is a one line fix that solves this problem by sorting emojis in the emoji picker by shortcode.

Thanks!


connyduck:
Emojis are already sorted alphabetically here: https://github.com/tuskyapp/Tusky/blob/73670b8576b307ae869216f72260e00497906548/app/src/main/java/com/keylesspalace/tusky/ComposeActivity.java#L317

PR #956, opened 2018-12-10
Share filters with web client

Tak:
Synchronize filters with server
- Filters are loaded from server and applied to timelines according to specified context
- User can manage filters in account preferences


app/src/main/java/com/keylesspalace/tusky/fragment/TimelineFragment.java
```diff
@@ -260,15 +263,52 @@ private void setupTimelinePreferences() {
...
+    }
+
+    private static boolean filterContextMatchesKind(Kind kind, List<String> filterContext) {
+        // home, notifications, public, thread
+        switch(kind) {
+            case HOME:
+                return filterContext.contains("home");
+            case PUBLIC_FEDERATED:
+            case PUBLIC_LOCAL:
+            case USER:
```
Tak: I think `USER` and `USER_WITH_REPLIES` need to be removed, since the web UI doesn't apply filters to the account UI

app/src/main/java/com/keylesspalace/tusky/fragment/preference/AccountPreferencesFragment.kt
```diff
@@ -243,6 +296,64 @@ class AccountPreferencesFragment : PreferenceFragmentCompat(),
...
+                        existing.context.filter { c -> c != context }.toList(),
+                        existing.irreversible,
+                        existing.wholeWord,
+                        null
+                ).enqueue(object: Callback<Filter>{
+                    override fun onFailure(call: Call<Filter>, t: Throwable) {
+                        Toast.makeText(activity?.applicationContext, "Error updating filter", Toast.LENGTH_SHORT).show()
+                    }
+                    override fun onResponse(call: Call<Filter>, response: Response<Filter>) {
+                        contextFilters.remove(existing)
```
Tak: TODO: Also remove context from `Filter.context`

connyduck:
(Sidenote: I really don't like how Mastodon filters are designed, feels like they took the easy route on the server side and now its a pita to implement on a client)

Nice! 
Regarding the interface, I think the filters should be moved to their own activity so there is more space. The first thing the user sees should be a list of the filters.
Also I think we should completely remove our old client side filters to have cleaner code and not confuse the user (even if it means dropping a feature)

Tak:
@connyduck @charlag Ok, what do you think about the current UI?

image
image
image
image

charlag:
Looks okay to me

Tak:
Ok, if there are no more comments, then let's do this!

app/src/main/java/com/keylesspalace/tusky/FiltersActivity.kt
```diff
@@ -0,0 +1,179 @@
+package com.keylesspalace.tusky
+
+import android.app.AlertDialog
```
connyduck: You need to use the import from AndroidX to get the right dialog styling.

```suggestion
import androidx.appcompat.app.AlertDialog
```

app/src/main/java/com/keylesspalace/tusky/fragment/TimelineFragment.java
connyduck: yes `USER` and `USER_WITH_REPLIES` should be removed

app/src/main/res/layout/dialog_filter.xml
```diff
@@ -0,0 +1,46 @@
...
+    <EditText
+        android:id="@+id/phraseEditText"
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content"
+        android:paddingEnd="24dp"
+        android:paddingStart="24dp"
+        android:hint="@string/filter_add_description"
+        app:layout_constraintTop_toTopOf="parent"
+        />
+        <!-- Mimic AlertDialog positive/negative/neutral layout, but with our padding -->
```
connyduck: I don't think this is a good idea if not absolutely necessary, can we stay with the default layout please?

app/src/main/java/com/keylesspalace/tusky/FiltersActivity.kt
charlag: You can check out how list editing dialog is done (it's very similar dialog):
https://github.com/tuskyapp/Tusky/pull/1104/files#diff-4bf8dc8fd7cbf714a810d8653cdfae55R116
code is not pretty but it works.

app/src/main/res/layout/dialog_filter.xml
Tak: Ah, it's not necessary after the `AlertDialog` import fix

PR #957, opened 2018-12-11
Fix a lot of easy-to-fix warnings

ariasuni:
Reduce warnings produced by `./gradlew lint` from 108 to 53 (half less!).

I added annotation to ignore some warnings on big SVG since I don’t see a way to optimize them without reducing their quality.

Finally, I edited very slightly the French translation to add strings required by a change introduced in #901.


charlag:
Hey, sounds awesome! Thanks.
I would prefer to have translation fixes separately but I think it wouldn't hurt to have them in.
If I'm going to feel better tomorrow, I'll have a look!

PR #960, opened 2018-12-19
Fix wrong visibility icon

kyori19:
Icon for private status should be `ic_lock_outline`, not `ic_lock_open`.
https://github.com/tuskyapp/Tusky/pull/901#issuecomment-447646647


connyduck:
awesome, thx

PR #962, opened 2018-12-21
add the ability to see who faved or boosted a toot

### LINKED ISSUE #455: Add the ability to see who boostet or favourited a status
Should only be available on detail statuses. Probably it is best to do it like the Twitter app - show a line with boost and fav numbers where one can click for details and seperate action buttons.

connyduck:
Closes #455 

- move fav/boost count up, so it is independently clickable from the fav/boost buttons
- will be hidden when both fav and boost count are 0
- the object returned by the server when a status is liked/boosted will now be used so the count will no longer stay the same as it does now (works most of the time, no idea why servers sometimes return wrong count)
- converted `AccountListActivity` and `AccountListFragment` to Kotlin
- yes I know, the `item_status_detailed.xml` layout is a mess, I will change this later so this PR does not get too big


app/src/main/java/com/keylesspalace/tusky/AccountListActivity.kt
```diff
@@ -0,0 +1,102 @@
...
+                AccountListActivity.Type.REBLOGGED -> setTitle(R.string.title_reblogged_by)
+                AccountListActivity.Type.FAVOURITED -> setTitle(R.string.title_favourited_by)
+            }
+            setDisplayHomeAsUpEnabled(true)
+            setDisplayShowHomeEnabled(true)
+        }
+
+        supportFragmentManager
+                .beginTransaction()
+                .replace(R.id.fragment_container, AccountListFragment.newInstance(type, id))
```
charlag: Is it okay to do this? I have a feeling it can lead to duplicate fragments or replacing fragments even when not needed because they're retained automatically

app/src/main/java/com/keylesspalace/tusky/fragment/AccountListFragment.kt
```diff
@@ -0,0 +1,336 @@
...
+            adapter.addItems(accounts)
+        } else {
+            adapter.update(accounts)
+        }
+
+        bottomId = fromId
+
+        fetching = false
+
+        adapter.setBottomLoading(false)
```
charlag: done twice

app/src/main/java/com/keylesspalace/tusky/fragment/ViewThreadFragment.java
```diff
@@ -250,32 +254,16 @@ public void onFailure(@NonNull Call<Status> call, @NonNull Throwable t) {
...
-
     @Override
     public void onFavourite(final boolean favourite, final int position) {
         final Status status = statuses.get(position);
         timelineCases.favouriteWithCallback(statuses.get(position), favourite, new Callback<Status>() {
             @Override
             public void onResponse(@NonNull Call<Status> call, @NonNull Response<Status> response) {
                 if (response.isSuccessful()) {
-                    setFavForStatus(position, status, favourite);
+                    statuses.set(position, response.body());
```
charlag: There was some reason we had `setFavForStatus()` but I cannot remember now. Did we need to reuse the instance for some reason? Did we try to avoid some visual glitch? Is there a difference in behavior now?

app/src/main/java/com/keylesspalace/tusky/interfaces/StatusActionListener.java
```diff
@@ -37,4 +37,17 @@
      * @param position    The position of the status in the list.
      */
     void onContentCollapsedChange(boolean isCollapsed, int position);
+
+    /**
+     * called when the reblog count has been clicked
+     * @param position The position of the status in the list.
+     */
+    default void onShowReblogs(int position) {}
```
charlag: It's the general interface for all status containers but you can only to these things on one of them so maybe it should be a separate interface which extends the other one, idk. Maybe later.

app/src/main/java/com/keylesspalace/tusky/network/MastodonApi.java
```diff
@@ -131,16 +135,12 @@
     @GET("api/v1/statuses/{id}/reblogged_by")
     Call<List<Account>> statusRebloggedBy(
             @Path("id") String statusId,
-            @Query("max_id") String maxId,
-            @Query("since_id") String sinceId,
-            @Query("limit") Integer limit);
+            @Query("max_id") String maxId);
```
charlag: Why did you remove all these parameters?

app/src/main/java/com/keylesspalace/tusky/adapter/StatusDetailedViewHolder.java
```diff
@@ -68,6 +74,47 @@ protected void setCreatedAt(@Nullable Date createdAt) {
         }
     }
 
+    private void setReblogAndFavCount(int reblogCount,  int favCount, StatusActionListener listener) {
+
+        NumberFormat numberFormat = NumberFormat.getNumberInstance();
```
charlag: I would maybe reuse this format if possible

app/src/main/java/com/keylesspalace/tusky/AccountListActivity.kt
connyduck: We have been using the sam in Java, so I guess its fine

app/src/main/java/com/keylesspalace/tusky/fragment/ViewThreadFragment.java
connyduck: Yes there is a difference, now the fav/boost count will update. It works fine, no idea why this was never done.

app/src/main/java/com/keylesspalace/tusky/interfaces/StatusActionListener.java
connyduck: yes I know, but the `StatusDetailedViewHolder`  `StatusViewHolder` class hierachy makes it difficult to pass different listeners, I think the default interface methods are the best tradeoff

app/src/main/java/com/keylesspalace/tusky/adapter/StatusDetailedViewHolder.java
connyduck: good idea

app/src/main/java/com/keylesspalace/tusky/fragment/AccountListFragment.kt
connyduck: nice find

connyduck:
ok there is now a bug when liking posts that have a card attached 🙄

app/src/main/java/com/keylesspalace/tusky/AccountListActivity.kt
charlag: I know that we've been doing this but I think we unnecessarily reload the list on rotation now.

charlag:
Is it because we replace viewdata now?

connyduck:
yes

connyduck:
this fixes the problem, still needs some more testing I guess

connyduck:
tested everything again, looks fine, except for this issue: https://github.com/tootsuite/mastodon/issues/9638  (works fine with pleroma)

PR #963, opened 2018-12-22
Export Room schemes

charlag:
It is extracted part of the #809. I had to rebase it and now I don't have a schema version for the current database. Again.
It's really great to have scheme versions to test migrations without re-installing something 10 times. Room team also insist that it should be checked in.


connyduck:
yep its probably a good idea

charlag:
Thanks a lot

Issue #878, opened 2018-10-07
Problem when combine persian and english


herotux:
When you combine right-to-left language, such as Persian and Arabic, with the left-to-right language, like English, you face the problem of displaying the wrong words.

herotux:
For example:
برنامه tusky
May shown :
Tusky برنامه

connyduck:
Where does this occur? In text from Tusky? Or user input? Can you give me screenshots please?

herotux:
![screenshot_ -](https://user-images.githubusercontent.com/10483351/46591654-fc4d6280-cac8-11e8-9a45-ec82fcdc03e1.png)
In this pic "tusky" word maybe in right side and is in wrong in left.
If sentenses start with english direction be like this pic.sory for my bad english :rose:

charlag:
Oh, but we're not doing anything special there, it should be handled by the system. I'm not even sure we can do something but we can check.

herotux:
![screenshot_ -](https://user-images.githubusercontent.com/10483351/46604396-f4121900-cb02-11e8-8120-15a17243a71a.png)

This is not handled by the system see this sms .

charlag:
Well we're not drawing text by ourselves, it is a very, very hard thing to do. If it is some UI problem we can adapt to please, mention it but we cannot change the way text is rendered, I'm sorry.

connyduck:
I think we can control this somehow with the `textDirection` attribute https://developer.android.com/reference/android/view/View.html#attr_android:textDirection
It seems `firstStrong` is the default value, but `anyRtl` is the desired behavior?

herotux:
maybe connydock have a idea to solve this

mabkenar:
> I think we can control this somehow with the `textDirection` attribute https://developer.android.com/reference/android/view/View.html#attr_android:textDirection
> It seems `firstStrong` is the default value, but `anyRtl` is the desired behavior?

It's a pitty that Android doesn't have the smart direction algorithm that Mastodon web [has](https://github.com/tootsuite/mastodon/blob/47bca3394590b59fb818d16bc05b9ad08cdc64dd/app/helpers/stream_entries_helper.rb#L150) (and almost all credible web-apps). That would be the best method for detecting directionality (i.e., to count if RTL-strong chars are at least 30% of the total string).

mabkenar:
[...]

PR #964, opened 2018-12-23
Improve RTL support

ahangarha:
By making these changes, texts in relevant pages will appear currectly
in RTL layout.

### DIFF
--- a/app/src/main/res/layout/activity_about.xml
+++ b/app/src/main/res/layout/activity_about.xml
@@ -16,6 +16,7 @@
         <androidx.core.widget.NestedScrollView
             android:layout_width="match_parent"
             android:layout_height="wrap_content"
+            android:textDirection="anyRtl"
             android:layout_gravity="center">
 
             <LinearLayout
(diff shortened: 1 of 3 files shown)


connyduck:
But I see no difference? Rtl is already supported everywhere? Please explain what this improves, maybe with screenshots.

ahangarha:
The most please look at about page while your phone language is English and
compare it with what you see when phone language is Persian. Everything is
perfect except when a sentence starts with a Latin character. this makes
the text in the textbox appears LTR not RTL.

With my little change, this issue is resolved. If you need screenshot, I
will share one tonight.

ahangarha:
This is the screenshot of the difference between current master branch and RTL-improve branch in about page.

tusky-rtl-imprive

connyduck:
Ah I see. Thank you!

charlag:
Didn't we have this discussion at https://github.com/tuskyapp/Tusky/issues/878 and decided against anyRTL?

connyduck:
yes but #878 was about user generated content and this is about our interface (or am I missing something?)

ahangarha:
What I have done on about page, has solved the problem. But other commits
might be subject to the mentioned discussion.
I even think all my PR could be dropped and solve the problem by changing
the translation in a way it doesn't start with Latin character.

herotux:
thanks @ahangarha

PR #967, opened 2018-12-26
Rework status layout

### LINKED ISSUE #481: Long names are overlapping "X seconds ago"
Long names seem to overlap it. I suggest to replace the rest of the name with "..."

Usernames on the other hand are working fine

### LINKED ISSUE #645: Migrate from RelativeLayout to ConstraintLayout
RelativeLayout is de-facto deprated. It is a very slow layout (which measures children *twice*) and it's kinda buggy and tied to platform.
ConstraintLayout is a modern and very powerful alternative. In most cases nesting layouts is a bug with ConstraintLayout.

The most obvious case is [item_status.xml](https://github.com/tuskyapp/Tusky/blob/master/app/src/main/res/layout/item_status.xml). Right now it has *two* `RelativeLayout`s, `ConstraintLayout` and `LinearLayout` which is clearly too much.
It's kinda tricky to not break anything but I believe that we can simplify layout a lot.

charlag:
This is a long overdue migration to `ConstraintLayout` for status layouts (closes #645).
By mistake I started from detailed layout so we may have to merge at some point or split this PR.

If we want to fix #481 we need to either 1) migrate nested `RelativeLayout` to `ConstrainedLayout` (meh) or 2) Made code changes to show/hide multiple views (also meh but more efficient IMO). I'm inclined towards the option 2.


connyduck:
But we can remove the `RelativeLayout` and position everything inside the outer `ConstraintLayout`, right? Maybe with the help of a `Barrier`?

Posts with spoilers don't look quite right yet:
screenshot_1545921355

charlag:
That's the option 2 which I would like to use

thanks for checking, I missed something!

connyduck:
Yes, statuses look great now!
Im so glad we are finally getting this done, thank you so much!

I think its possible to get rid of the `Space`s as well with `layout_constraintHorizontal_chainStyle="spread_inside"`?

charlag:
I'm relying on you for finishing touched because you have a better eye on how things should look like, e.g. name is slightly off now but I'm not sure if it's okay.
My first approach was just a chain but I had to add spaces because of numbers, I can try again now when they're gone. I didn't try spread_inside yet, thanks foe the hint!

charlag:
I tried chains, looks good. The last button is a little bit closer to the side now but I'm not sure if it's desirable or not.

charlag:
Also high emoji in display name shift everything down :thinking:

app/src/main/res/layout/item_status.xml
```diff
@@ -13,156 +13,168 @@
...
-            android:layout_toStartOf="@+id/status_timestamp_info"
-            android:ellipsize="end"
-            android:maxLines="1"
-            android:textColor="?android:textColorSecondary"
-            android:textSize="?attr/status_text_medium"
-            tools:text="\@Entenhausen" />
+        android:layout_marginStart="14dp"
+        android:ellipsize="end"
+        android:maxLines="1"
+        android:paddingStart="0dp"
```
connyduck: this line does nothing and can be removed

app/src/main/res/layout/item_status.xml
```diff
@@ -13,156 +13,168 @@
...
+        android:paddingEnd="@dimen/status_display_name_padding_end"
+        android:textColor="?android:textColorPrimary"
+        android:textSize="?attr/status_text_medium"
+        android:textStyle="normal|bold"
+        app:layout_constrainedWidth="true"
+        app:layout_constraintEnd_toStartOf="@id/status_timestamp_info"
+        app:layout_constraintHorizontal_bias="0"
+        app:layout_constraintStart_toEndOf="@id/status_avatar"
+        app:layout_constraintTop_toTopOf="@id/status_avatar"
```
connyduck: to have it look like before, this needs 
```
app:layout_constraintTop_toBottomOf="@id/status_reblogged"
android:layout_marginTop="@dimen/status_avatar_padding"
```
(not sure why this has the name "status_avatar_padding", but thats what was used previously. Feel free to remove and just use 10dp)

app/src/main/res/layout/item_status.xml
```diff
@@ -13,156 +13,168 @@
...
+        tools:visibility="visible" />
 
     <com.keylesspalace.tusky.view.RoundedImageView
         android:id="@+id/status_avatar"
         android:layout_width="48dp"
         android:layout_height="48dp"
-        android:layout_below="@+id/status_reblogged"
-        android:layout_marginEnd="14dp"
         android:layout_marginTop="14dp"
+        android:layout_marginEnd="14dp"
```
connyduck: `marginEnd` not necessary when `status_display_name` has the same `marginStart`

app/src/main/res/layout/item_status_detailed.xml
```diff
@@ -19,58 +19,65 @@
...
+        android:textColor="?android:textColorPrimary"
+        android:textSize="?attr/status_text_medium"
+        android:textStyle="normal|bold"
+        app:layout_constrainedWidth="true"
+        app:layout_constraintBottom_toTopOf="@id/status_username"
+        app:layout_constraintEnd_toEndOf="parent"
+        app:layout_constraintHorizontal_bias="0"
+        app:layout_constraintRight_toRightOf="parent"
+        app:layout_constraintStart_toEndOf="@id/status_avatar"
+        app:layout_constraintTop_toTopOf="@id/status_avatar"
```
connyduck: (same as on `item_status.xml` - align it below the top view with 10dp margin instead of to the top of the avatar)

app/src/main/res/layout/item_status.xml
charlag: Yea, I forgot that margin only applied to the source of the constraint and I forgot to remove it

connyduck:
awesome ^___^

PR #976, opened 2019-01-06
fix long toots not collapsing correctly in timelines

### LINKED ISSUE #974: Problems with auto-collapsed long toots
Long toots start out uncollapsed, with a button below them that says "expand". When you tap the expand button, the toot collapses and the button says "collapse".

* * * *
- Tusky Version: 4.1-40346ff
- Android Version: 7.1.2
- Android Device: Asus z00ed

connyduck:
Closes #974 
Fixes a regression (probably from #954)

The `InputFilter` have to be set before any text or they don't work.

I also moved the code from `StatusBaseViewHolder` to `StatusViewHolder` as it is not used for detailed statuses which are always expanded.


charlag:
Ooops, I completely forgot about this!

connyduck:
Yes thats all, took me a while to figure out so I wrote a comment :D

PR #984, opened 2019-01-11
Fix unnecessary reloading of notifications

charlag:
This removes `topId` as it is not needed and just plainly uses status id
if needed. During initial loading of notifications `topId`/`bottomId` are
not set so we ended up reloading everything.

We could remove `bottomId` as well but it is used in some logic of reaching the end so I decided to not touch it.


connyduck:
Code looks fine, but what do you mean by reloading everything?

charlag:
Well it was sending unbounded notifications request and was replacing all the views as a consequence.

PR #985, opened 2019-01-11
Refactor notifications adapter

charlag:
This one indirectly depends on #984 

I mostly copied code from timeline adapter handling to use `AsyncListDiffer`.

It's low priority as it doesn't fix anything really so we can push it to one of the next releases but it seems to be okay to me.


charlag:
I'm not 100% sure that everything has merged correctly but it seems to be in the working state. There's not much to it actually, just getting it in line with `TimelineFragment` and getting rid of that weird adapter interface.
There's little to none UI code in both that it's a shame that they're not extracted to proper ViewModel classes to be easily testable. This should be a next step I think.

PR #986, opened 2019-01-11
fix mentions and more menu after a fresh login

connyduck:
When a user logs in, the `AccountEntity` object does not have a username or id when creating fragments because the network request has not returned yet -> the wrong more menu will be shown & replies will contain the own username. Fixed by deferring the loading of the username and id in `SFragment`.

Annoys me as hell, but real users probably don't log in as often as i do :)

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/fragment/SFragment.java
+++ b/app/src/main/java/com/keylesspalace/tusky/fragment/SFragment.java
@@ -19,7 +19,6 @@ import android.content.ClipData;
 import android.content.ClipboardManager;
 import android.content.Context;
 import android.content.Intent;
-import android.os.Bundle;
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
 import androidx.core.app.ActivityOptionsCompat;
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/keylesspalace/tusky/fragment/SFragment.java
```diff
@@ -146,6 +138,12 @@ protected void more(@NonNull final Status status, View view, final int position)
         final String accountUsername = status.getActionableStatus().getAccount().getUsername();
         final Spanned content = status.getActionableStatus().getContent();
         final String statusUrl = status.getActionableStatus().getUrl();
+        String loggedInAccountId = null;
```
charlag: Huh? Why it's reset here? Confusing

connyduck: its not reset its initialized

charlag: Ah, my bad, it seemed like it's the same initialization in another method.

Issue #958, opened 2018-12-15
[Feature Request] 3-dotted menu > Open the link in account...


sprite-1:
Hi! I would love for an added option in the 3-dotted menu beneath every post to have an option below the "Copy the link" called "Open the link in..." which would prompt the user to pick one of the logged in accounts from other instances to open the link in. It would really help with people who have one main account they use to interact with others. But also have an account in other instances just to see that instance's local timeline in Tusky.

Right now, it takes too many steps of:
1. Tap the 3-dotted menu
2. Tap "Copy the link"
3. Go back to main account.
4. Pull up the search
5. Paste the URL.
6. Finally be able to interact with the toot.

If this feature is implemented it would shorten the above steps into:
1. Tap the 3-dotted menu
2. Tap "Open the link in..."
3. Pick the account and interact

* * * *
- Tusky Version: 4.1

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

connyduck:
closed by #988

PR #988, opened 2019-01-15
Implement "Open status as (another account)" #958

Tak:
Adds a context action to open statuses as a different account.
if the user only has one account registered, the action is hidden.
If the user has exactly two accounts, the action is presented as "Open as @You@YourOtherInstance".
If the user has more than two accounts, the action is presented as "Open as …", and the user gets a list popup.

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/fragment/SFragment.java
+++ b/app/src/main/java/com/keylesspalace/tusky/fragment/SFragment.java
@@ -19,12 +19,16 @@ import android.content.ClipData;
 import android.content.ClipboardManager;
 import android.content.Context;
 import android.content.Intent;
+
 import android.text.Spanned;
 import android.view.Menu;
+import android.view.MenuItem;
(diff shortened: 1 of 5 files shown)


app/src/main/java/com/keylesspalace/tusky/fragment/SFragment.java
```diff
@@ -66,6 +70,7 @@
     protected abstract void onReblog(final boolean reblog, final int position);
 
     private BottomSheetActivity bottomSheetActivity;
+    private String openAsTitle;
```
connyduck: Its weird that this is a member variable and not a local one. Can you move it somehow, maybe set the title to `openAsItem` directly?

Tak: Good point - the compiler was hating because I was passing it to the lambda, but I can just get the title from `openAsItem` at that point

PR #994, opened 2019-01-17
Add pinned status tab

kyori19:
Related to #443.

I added new tab "pinned" to AccountActivity, which shows user's pinned toot.

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/AccountActivity.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/AccountActivity.kt
@@ -16,27 +16,27 @@
 package com.keylesspalace.tusky
 
 import android.animation.ArgbEvaluator
-import androidx.lifecycle.Observer
-import androidx.lifecycle.ViewModelProviders
 import android.content.Context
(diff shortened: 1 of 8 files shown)


charlag:
Wow, that's an interesting idea, I've never thought about that possibility, always thought we would need to change timeline.
The problem is that "pinned" loses some sense in this case because it's hidden by default but it's certainly better than nothing IMO.

connyduck:
Its definitely better than nothing but I still think we should change it later so the pinned toots are shown above the other toots on the first tab.

kyori19:
Actually , I had another idea to show pinned toots above normal toots.
https://github.com/accelforce/Tusky/commit/1ab0aef2c0bed9f8bffce02490ee62ec406d66f4
But this method causes many problems so I gave up this idea.

connyduck:
Can you switch the `TabLayout` in `activity_account.xml` to `app:tabMode="scrollable"`? It makes the tabs look way better on small screens

kyori19:
@connyduck OK, I changed fixed to scrollable.

PR #999, opened 2019-01-19
Never use 0 account ids to prevent caching problems

### LINKED ISSUE #990: Problem with toot cache when logging in with a new account
Steps to reproduce:
Start with a fresh installation of Tusky.
Log in with A. Notice the timeline loading.
Log in with another user B. Notice the timeline already being filled but suddenly changing. Some toots will be shown that obviously belong to the timeline of A.
The problem resolves itself when switching accounts to A and back to B.

Tusky 4.1-922f165

connyduck:
closes #990 

Note that the `(autoGenerate = true)` could now be removed from `AccountEntity.id`, but I am leaving it so we don't have to do another database migration.

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/db/AccountManager.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/db/AccountManager.kt
@@ -58,7 +58,9 @@ class AccountManager(db: AppDatabase) {
             accountDao.insertOrReplace(it)
         }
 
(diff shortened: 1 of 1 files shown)


connyduck:
Oh yeah we should definitely clear the timeline when logging out or the next login will see the old cache of the removed account, good thinking. I will send a seperate PR for that.

charlag:
As I said, there's no ID collisions so you won't see old timeline but it'll take space for two weeks.

charlag:
Ah, if you delete all of them, of course!

PR #1011, opened 2019-01-25
Add support for selecting account when sharing from outside apps

Tak:
Includes direct-share support when using API 23+

share-as
direct-share

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/adapter/AccountSelectionAdapter.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/adapter/AccountSelectionAdapter.kt
@@ -1,0 +1,58 @@
+/* Copyright 2019 Levi Bard
+ *
+ * This file is a part of Tusky.
+ *
+ * This program is free software; you can redistribute it and/or modify it under the terms of the
+ * GNU General Public License as published by the Free Software Foundation; either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * Tusky is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even
+ * the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU General
+ * Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License along with Tusky; if not,
+ * see <http://www.gnu.org/licenses>. */
+
+package com.keylesspalace.tusky.adapter
+
+import android.content.Context
+import android.text.TextUtils
+import android.view.LayoutInflater
+import android.view.View
+import android.view.ViewGroup
+import android.widget.ArrayAdapter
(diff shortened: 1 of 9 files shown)


app/src/main/java/com/keylesspalace/tusky/service/AccountChooserService.kt
```diff
@@ -0,0 +1,65 @@
...
+import com.keylesspalace.tusky.util.NotificationHelper
+import com.squareup.picasso.Picasso
+
+@TargetApi(23)
+class AccountChooserService : ChooserTargetService() {
+    lateinit var accountManager: AccountManager
+
+    override fun onCreate() {
+        super.onCreate()
+        val appDatabase = Room.databaseBuilder(applicationContext, AppDatabase::class.java, "tuskyDB")
```
charlag: Should we pull it somewhere? Is it possible?

app/src/main/java/com/keylesspalace/tusky/adapter/AccountSelectionAdapter.kt
```diff
@@ -0,0 +1,57 @@
...
+        var view = convertView
+
+        if (convertView == null) {
+            val layoutInflater = context.getSystemService(Context.LAYOUT_INFLATER_SERVICE) as LayoutInflater
+            view = layoutInflater.inflate(R.layout.item_autocomplete, parent, false)
+        }
+
+        val account = getItem(position)
+        if (account != null) {
+            val username = view!!.findViewById<TextView>(R.id.username)
```
charlag: We probably won't have enough rows to recycle so implementing holder doesn't make too much sense, right?..

app/src/main/java/com/keylesspalace/tusky/service/AccountChooserService.kt
Tak: Yeah, I wasn't sure about that...

Tak:
> Not sure how you do all of that so quick but thanks!
Fortunately all the time-consuming (from my perspective) pieces were already there :slightly_smiling_face:

app/src/main/java/com/keylesspalace/tusky/service/AccountChooserService.kt
connyduck: you can inject it like in `SendTootService`

app/src/main/java/com/keylesspalace/tusky/adapter/AccountSelectionAdapter.kt
connyduck: nah I wouldn't bother either. But I would use Kotlin Android extension instead of `findViewById`

connyduck:
Can you resolve the build error when merging with master, please? Looks like a layout file got renamed.

Tak:
Done!

PR #1018, opened 2019-01-30
Fix empty state for account media, fix tablets

### LINKED ISSUE #1014: Crash on opening Pinned section in profile
When I open the app, go to my profile and tap on "pinned" section (I have 0 pinned toots), app just restarts without any prompts. It doesn't look like crash, just restarts. If I go to my profile and tap on "pinned" section again, it crashes with default Android dialog (About app / Kill app / Send feedback)

* * * *
- Tusky Version: Tusky Nightly 4.1-c0c73f5
- Android Version: 9.0
- Android Device: Oneplus 5

charlag:
Fix #1014


connyduck:
👍 
but I think you added the db file by mistake

PR #1020, opened 2019-02-01
Attach subject when composing with shared content from other apps as optional function

lenchan139:
In Current, if I share some content from some apps(e.g. YouTube), ComposeActivity catch the url/text (Intent.EXTRA_TEXT) only. For other SNS application is fine since they will generate preview when composing. But for Mastodon, the preview only generate / show in limit case. So, there may attach subject(Intent.EXTRA_SUBJECT) with text/url(Intent.EXTRA_TEXT) would be better.

Since most Mastodon instances have characters limits, this function is optional. User can enable in preference. (default off)

screenshot 2019-02-01 at 12 08 08

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/ComposeActivity.java
+++ b/app/src/main/java/com/keylesspalace/tusky/ComposeActivity.java
@@ -620,13 +620,12 @@ public final class ComposeActivity
                         String subject = intent.getStringExtra(Intent.EXTRA_SUBJECT);
                         String text = intent.getStringExtra(Intent.EXTRA_TEXT);
                         String shareBody = null;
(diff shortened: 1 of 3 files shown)


connyduck:
Nobody is going to understand what that option is doing. I had to look at the code to understand it.
We can change it, but for everybody.

connyduck:
Yes I understand what this Pull request is doing, but I don't think it we should offer an option because it is too confusing for the user. Things like sharing to Tusky should work out of the box without the user having to think about it.

charlag:
Can we just always add it?

lenchan139:
Ok,
I removed option and make subject would be added if possible :)

PR #1026, opened 2019-02-02
Fix incorrectly incrementing IDs before sending to server.

### LINKED ISSUE #1022: Tusky not loading new toots
Tusky is not loading new toots when pulling to refresh for one of my accounts.

the reason is it is querying the api with a wrong parameter 
`api/v1/timelines/home?since_id=10151730742138560z&limit=32` (note the `z`)
looks like a regression from #1016
will try to come up with steps to reproduce

- Tusky Version: current master

charlag:
Fix #1022 
It *seems* to be working but I've made so many stupid errors before and while fixing it now that I'm not sure about anything anymore. I want to write some tests for that but I'm not sure when I'll do it. I would appreciate if someone else tried it out before merging.

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/db/TimelineDao.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/db/TimelineDao.kt
@@ -59,7 +59,11 @@ LIMIT :limit""")
     }
 
     @Query("""DELETE FROM TimelineStatusEntity WHERE authorServerId = null
-AND timelineUserId = :acccount AND serverId > :sinceId AND serverId < :maxId""")
+AND timelineUserId = :acccount AND
+(LENGTH(serverId) < LENGTH(:maxId) OR LENGTH(serverId) == LENGTH(:maxId) AND serverId < :maxId)
+AND
+(LENGTH(serverId) > LENGTH(:sinceId) OR LENGTH(serverId) == LENGTH(:sinceId) AND serverId > :sinceId)
+""")
     abstract fun removeAllPlaceholdersBetween(acccount: Long, maxId: String, sinceId: String)
 
     @Query("""UPDATE TimelineStatusEntity SET favourited = :favourited
--- a/app/src/test/java/android/text/FakeSpannableString.kt
+++ b/app/src/test/java/android/text/FakeSpannableString.kt
@@ -1,0 +1,50 @@
+package android.text
+
+// Used for stubbing Android implementation without slow & buggy Robolectric things
+@Suppress("unused")
+class SpannableString(private val text: CharSequence) : Spannable {
+
+    override fun setSpan(what: Any?, start: Int, end: Int, flags: Int) {
+        throw NotImplementedError()
+    }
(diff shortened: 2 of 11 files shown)


charlag:
Started with tests, setup is usually the longest part, should be able to cover other logic tomorrow.

charlag:
I think I'm done with it for now. So glad I wrote tests, found so many sneaky things, even in the dumbest 2 line implementation of `String#inc()`.

app/src/main/java/com/keylesspalace/tusky/db/TimelineDao.kt
```diff
@@ -59,7 +59,11 @@ LIMIT :limit""")
     }
 
     @Query("""DELETE FROM TimelineStatusEntity WHERE authorServerId = null
-AND timelineUserId = :acccount AND serverId > :sinceId AND serverId < :maxId""")
+AND timelineUserId = :acccount AND
+(LENGTH(serverId) < LENGTH(:maxId) OR LENGTH(serverId) == LENGTH(:maxId) AND serverId < :maxId)
+AND
+(LENGTH(serverId) > LENGTH(:sinceId) OR LENGTH(serverId) == LENGTH(:sinceId) AND serverId > :sinceId)
+""")
```
Tak: Maybe use parentheses to make the boolean operation precedence explicit?

app/src/test/java/android/text/FakeSpannableString.kt
```diff
@@ -0,0 +1,50 @@
+package android.text
+
+// Used for stubbing Android implementation without slow & buggy Robolectric things
+@Suppress("unused")
+class SpannableString(private val text: CharSequence) : Spannable {
```
Tak: Maybe this should be merged with the one in `SpanUtilsTest`?

app/src/test/java/com/keylesspalace/tusky/StringUtilsTest.kt
```diff
@@ -20,6 +22,15 @@ class StringUtilsTest {
         notLessList.forEach { (l, r) -> assertFalse("not $l < $r", l.isLessThan(r)) }
     }
 
+    @Test
+    fun inc() {
+        listOf(
+                "122" to "123",
+                "12A" to "12B",
+                "1" to "2"
```
Tak: No test case for something ending with `z`?

Tak: or `9`?

charlag: `inc()` is very dumb and I just let it increase any character, really. Even if it's z it should become something bigger.

connyduck:
It fixes the bug and the code looks fine for me, so I think we should merge it? Got a lot of complaints about the timeline not loading on nightly.

charlag:
I hit the bug like five times. Now I'm on this branch without any problems so far.

PR #1036, opened 2019-02-12
call db operation on background thread in TabPreferenceActivity

connyduck:


charlag:
LGTM
Did you also check conversations ViewModel? It looked to me that it has the same problem.

PR #1037, opened 2019-02-12
delete conversation cache on logout

connyduck:


app/src/main/java/com/keylesspalace/tusky/components/conversation/ConversationsRepository.kt
```diff
@@ -93,9 +95,16 @@ class ConversationsRepository @Inject constructor(val mastodonApi: MastodonApi,
         )
     }
 
+    fun deleteCacheForAccount(accountId: Long) {
+        Single.fromCallable {
+            db.conversationDao().deleteForAccount(accountId)
+        }.observeOn(Schedulers.io())
```
charlag: You probably want to `subscribeOn()` and not `observeOn()`

PR #1040, opened 2019-02-13
call db functions on background thread in ConversationsViewModel

connyduck:
really nice that they added `Completable` and `Single` support directly to Room in 2.1.0

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -112,9 +112,9 @@ dependencies {
     // architecture components
     implementation 'androidx.lifecycle:lifecycle-extensions:2.0.0'
     //room
-    implementation 'androidx.room:room-runtime:2.1.0-alpha04'
-    kapt 'androidx.room:room-compiler:2.1.0-alpha04'
-    implementation 'androidx.room:room-rxjava2:2.1.0-alpha04'
+    implementation 'androidx.room:room-runtime:2.0.0'
+    kapt 'androidx.room:room-compiler:2.0.0'
+    implementation 'androidx.room:room-rxjava2:2.0.0'
     implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$kotlin_version"
     testImplementation 'junit:junit:4.12'
     implementation "com.google.dagger:dagger:$daggerVersion"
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/keylesspalace/tusky/components/conversation/ConversationsViewModel.kt
```diff
@@ -67,6 +71,8 @@ class ConversationsViewModel  @Inject constructor(
                     lastStatus = conversation.lastStatus.copy(expanded = expanded)
             )
             database.conversationDao().insert(newConversation)
+                    .subscribe()
```
charlag: I don't think it makes much sense to add them to disposables in all 3 cases. We don't want to cancel these actions when the screen is closed, right? We still want to update DB or something.

app/build.gradle
```diff
@@ -112,9 +112,9 @@ dependencies {
     // architecture components
     implementation 'androidx.lifecycle:lifecycle-extensions:2.0.0'
     //room
-    implementation 'androidx.room:room-runtime:2.0.0'
-    kapt 'androidx.room:room-compiler:2.0.0'
-    implementation 'android.arch.persistence.room:rxjava2:1.1.1'
+    implementation 'androidx.room:room-runtime:2.1.0-alpha04'
```
charlag: I think that you can return `Single` from the inception:
https://medium.com/androiddevelopers/room-rxjava-acb0cd4f3757
(post from June 207)
So no need for alpha, really

app/src/main/java/com/keylesspalace/tusky/db/ConversationsDao.kt
```diff
@@ -18,17 +18,18 @@ package com.keylesspalace.tusky.db
...
+import io.reactivex.Completable
 
 @Dao
 interface ConversationsDao {
     @Insert(onConflict = OnConflictStrategy.REPLACE)
     fun insert(conversations: List<ConversationEntity>)
 
     @Insert(onConflict = OnConflictStrategy.REPLACE)
-    fun insert(conversation: ConversationEntity)
+    fun insert(conversation: ConversationEntity): Completable
```
charlag: If we don't want to cancel them (I think we don't) we could as well leave them as blocking ones.

app/build.gradle
connyduck: nope, in the article you linked: 

> Later edit 2: starting with Room 2.1.0-alpha01, DAO methods annotated with @Insert, @Delete or @Update support Rx return types Completable, Single<T> and Maybe<T>.

charlag: Ah, sorry, I misunderstood, this are mutation functions! They've been always blocking and now you can subscribe on another thread, that's really great.

Issue #938, opened 2018-11-29
Tusky crashes upon proximity of a post


tleydxdy:
when ever [this](https://pl.smuglo.li/notice/8043569) post is shown, tusky crashes. 

* * * *
- Tusky Version:4.0 fdroid
- Android Version:7.1.2
- Android Device:sony xperia z5
- Mastodon instance (if applicable):gnu/smug

- [✓] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

tleydxdy:
seems like it was caused by the huge picture.
logcat shows

11-29 12:55:18.846 E/AndroidRuntime(27720): java.lang.RuntimeException: Canvas: trying to draw too large(108000000bytes) bitmap.

tleydxdy:
the picture is 6000x4500 pixels big

connyduck:
Looks like pleroma (or some instances?) don't resize images for preview, so Tusky loads the fullsize images. Thats mostly a pleroma problem, but we still should not crash

stom79:
I got the same issue with Pleroma: https://gitlab.com/tom79/mastalab/issues/521

I now use this helper function (can be improved/customized)
```Java
public static Bitmap compressImageIfNeeded(Bitmap bmToCompress){

        int size = bmToCompress.getByteCount();
        double resizeby =  33554432; //4Mo
        double resize = ((double)size)
        if( resize > 1 ) {
            Bitmap newBitmap = Bitmap.createScaledBitmap(bmToCompress, (int) (bmToCompress.getWidth() / resize),
                    (int) (bmToCompress.getHeight() / resize), true);
            ByteArrayOutputStream bos = new ByteArrayOutputStream();
            newBitmap.compress(Bitmap.CompressFormat.JPEG, 80, bos);
            return newBitmap;
        }
        return bmToCompress;
    }
```

connyduck:
should be fixed by #1043

PR #1043, opened 2019-02-15
Fix large image attachments not showing

MightyPork:
this is a fix for the following error on oneplus one with lineageos, viewing large photos uploaded to Pleroma

```
W OpenGLRenderer: Bitmap too large to be uploaded into a texture (4160x3120, max=4096x4096)
W OpenGLRenderer: Shape too large to be rendered into a texture (4172x3132, max=4096x4096)
W OpenGLRenderer: Bitmap too large to be uploaded into a texture (4160x3120, max=4096x4096)
```

there are still some errors in the log, but the images show OK.
- `E/BufferItemConsumer: [unnamed-4641-1] Failed to release buffer: Unknown error -1 (1)`
- `W/Adreno-EGL: <qeglDrvAPI_eglGetConfigAttrib:607>: EGL_BAD_ATTRIBUTE`

this *might* break true animated gifs (i.e. not mastodon video gifs). I'm not at all sure if they worked before, but they certainly don't now.

the size is 1920 because 1920x1080 and also it seemed to work a little better than 2048. but that may be a coincidence. At any rate 4096 caused horrible lag in the timeline. 2048 was mostly OK too

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/fragment/AccountMediaFragment.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/fragment/AccountMediaFragment.kt
@@ -302,8 +302,8 @@ class AccountMediaFragment : BaseFragment(), Injectable {
             holder.imageView.setBackgroundColor(Color.HSVToColor(itemBgBaseHSV))
             val item = items[position]
 
-            val maxW = context!!.resources.getInteger(R.integer.media_max_width)
-            val maxH = context!!.resources.getInteger(R.integer.media_max_height)
+            val maxW = holder.imageView.context.resources.getInteger(R.integer.media_max_width)
+            val maxH = holder.imageView.context.resources.getInteger(R.integer.media_max_height)
 
             Picasso.with(holder.imageView.context)
                     .load(item.attachment.previewUrl)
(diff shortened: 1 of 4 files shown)


connyduck:
Related issue https://github.com/tuskyapp/Tusky/issues/938

app/src/main/java/com/keylesspalace/tusky/fragment/AccountMediaFragment.kt
```diff
@@ -301,8 +301,15 @@ class AccountMediaFragment : BaseFragment(), Injectable {
             itemBgBaseHSV[2] = random.nextFloat() * (1f - 0.3f) + 0.3f
             holder.imageView.setBackgroundColor(Color.HSVToColor(itemBgBaseHSV))
             val item = items[position]
+
+            val maxW = context!!.resources.getInteger(R.integer.media_max_width)
+            val maxH = context!!.resources.getInteger(R.integer.media_max_height)
```
connyduck: Please never `context!!` this can cause crashes. You can get a non-null context from a view, so here `holder.imageView.context`

charlag:
Calculate exactly to what? To the view? In timelines we crop previews anyway, we could multiply that of course...

MightyPork:
context!! is now gone, i tried and it still works too

PR #1044, opened 2019-02-15
Russian language updated:

MasterGroosha:
• Added strings for tabs customization.
• Added strings for Likes/Boosts counters.
• Added strings for external open/share.
• Some other fixes and additions.

String "maximum of 5 tabs reached" is untranslatable atm, by the way.


connyduck:
> String "maximum of 5 tabs reached" is untranslatable atm, by the way.

Its now translateable. Thx for noticing that.

PR #1047, opened 2019-02-15
fix statuses collapsing when faving/boosting in ViewThreadFragment

connyduck:


charlag:
Sorry for being slow but seems to be fine

PR #1053, opened 2019-02-17
fix crash when opening a thread in the browser

connyduck:
Steps to reproduce: see a reblog in the timeline, open it. Click "open in browser" in the menu -> crash


connyduck:
actually, `actionableStatus.getActionableId()` is redundant

PR #1056, opened 2019-02-17
fix close button behavior in ComposeActivity

connyduck:
If I click the close button, I definitely want the activity to close, and not hide the bottom sheets.


charlag:
Hm, so if you press it on top it won't ask if you want to save? I think I tap there accidentally sometimes... also different behavior is confusing (we don't have a button to save draft e.g. and if you always use top button you'll never discover it).

connyduck:
No it will always ask if you want to save. It just does not hide the bottom sheets before doing so.

charlag:
Ah, okay then, LGTM!

Issue #1003, opened 2019-01-21
[Enhancement] Reduce the number of swipes a TalkBack user needs for a toot


MarcoZehe:
## Problem

When the TalkBack screen reader is on, a blind user can swipe left or right to walk sequentially through the items on the screen.

When reading toots or other items, right now a TalkBack user needs several swipes to walk through a toot. The standard items are:

1. Profile name, and CW text if applicable.
2. Profile button.
3. Tot text, or Show More button if hidden behind CW.
4. If 3 was a CW enhancer, toot text.
5. Reply Button.
6. Boost button.
7. Favorite Button.
8. More Options button.

If there is one image, between items 4 and 5, two more are inserted, plus one more for each aditional image.

That makes a minimum of 7 or 8, and a maximum of 10 swipes for _one_ toot, with at least half of the items being repetitive information like the buttons etc. This makes browsing toots very inefficient for a blind user.

Other clients, for example for Twitter, have adopted a model where each item only consists of one swipable item, and the actual actions are then provided as Accessibility.AccessibilityAction items. That reduces the number of swipes per item to 1, and the actions can be taken if needed or wanted, or one can just continue with the next toot.

## Proposed solution

To do this in Tusky, the following would need to happen:

1. All items except one would need to be marked [Accessibility.AccessibilityNodeInfo.SetImportantForAccessibility](https://developer.android.com/reference/android/view/accessibility/AccessibilityNodeInfo.html#setImportantForAccessibility(boolean)) false, so TalkBack no longer swipes to them.
2. Set the contentDescription to a string of text containing all relevant information from most to least important. They should be separated by a comma, so the speech synthesizer inserts a slight pause between the items, but not too much of a pause, like it would do if a period was used instead. Alternatively, a semicolon ; could also be tried. A proposed order would be:
   1. Profile screen name, but not the Mastodon username.
   2. The toot text, or CW if present, plus a hint that this is a CW.
   3. If this is a boost, something like "Boosted by xyz".
   4. The date this was composed.
   5. The full Mastodon username.
   6. If there are images attached and these have descriptions, add them one by one to the end.
3. Add [AccessibilityNodeInfo.AccessibilityAction](https://developer.android.com/reference/android/view/accessibility/AccessibilityNodeInfo.AccessibilityAction.html) classes for all the supported actions. Those actions would also be in the order "most likely" to "least likely", as they are presented top to bottom in the local context menu of TalkBack. The proposed order is:
[...]

PR #1059, opened 2019-02-17
Timeline a11y

charlag:
(After Tusky 5)
Apply suggestions from #1003.
Does only minimal changes, there needs to be more done for threads and notifications (also no way to refresh timeline now).


app/src/main/java/com/keylesspalace/tusky/util/ListStatusAccessibilityDelegate.kt
```diff
@@ -0,0 +1,135 @@
...
+import androidx.core.view.AccessibilityDelegateCompat
+import androidx.core.view.accessibility.AccessibilityNodeInfoCompat
+import androidx.core.view.accessibility.AccessibilityNodeInfoCompat.AccessibilityActionCompat
+import androidx.recyclerview.widget.RecyclerView
+import androidx.recyclerview.widget.RecyclerViewAccessibilityDelegate
+import com.keylesspalace.tusky.R
+import com.keylesspalace.tusky.interfaces.StatusActionListener
+import com.keylesspalace.tusky.viewdata.StatusViewData
+
+// Not using lambdas because there's boxing of int then
```
connyduck: did not know that, thx

app/src/main/java/com/keylesspalace/tusky/adapter/StatusViewHolder.java
```diff
@@ -92,11 +95,88 @@ protected void setupWithStatus(StatusViewData.Concrete status, final StatusActio
             }
 
             rebloggedBar.setOnClickListener(v -> listener.onOpenReblog(getAdapterPosition()));
+        }
+        setContentDescription(status);
+        // Workaround for RecyclerView 1.0.0 / androidx.core 1.0.0
+        // RecyclerView tries to set AccessibilityDelegateCompat to null
+        // but ViewCompat code replaces is with the default one. RecyclerView never
+        // fetches another one from its delegate because it checks that it's set so we remove it
+        // and let RecyclerView ask for a new delegate.
```
connyduck: Will this be fixed in a upcoming release of AndroidX?

app/src/main/java/com/keylesspalace/tusky/adapter/StatusViewHolder.java
```diff
@@ -92,11 +95,88 @@ protected void setupWithStatus(StatusViewData.Concrete status, final StatusActio
...
+            itemView.setContentDescription(
+                    itemView.getContext().getString(R.string.load_more_placeholder_text));
+        } else {
+            setStatusDescriptionForStatus(status);
         }
 
     }
 
+    private void setStatusDescriptionForStatus(@NonNull StatusViewData.Concrete status) {
+        CharSequence postedAgo = DateUtils.getRelativeTimeSpanString(
```
connyduck: Shouldn't this respect the "use absolute time" setting?

app/src/main/java/com/keylesspalace/tusky/adapter/StatusViewHolder.java
```diff
@@ -92,11 +95,88 @@ protected void setupWithStatus(StatusViewData.Concrete status, final StatusActio
...
+                status.getCreatedAt().getTime(),
+                System.currentTimeMillis(),
+                DateUtils.SECOND_IN_MILLIS,
+                DateUtils.FORMAT_ABBREV_RELATIVE
+        );
+        Context context = itemView.getContext();
+
+
+        String description = itemView.getContext().getString(R.string.status_description,
```
connyduck:
```suggestion
        String description = context.getString(R.string.status_description,
```

app/src/main/java/com/keylesspalace/tusky/util/ListStatusAccessibilityDelegate.kt
charlag: In this particular case. Kotlin exports it's lambda interface as generic method to Java. Generics are always boxed (that's why there are `IntStream`  and such).

app/src/main/java/com/keylesspalace/tusky/adapter/StatusViewHolder.java
charlag: I didn't find any info anywhere. Didn't report it tbh.

app/src/main/java/com/keylesspalace/tusky/adapter/StatusViewHolder.java
charlag: Yep, we probably should.

charlag:
I did not extend it to the detailed status because I was unable to reproduce some native TextView functionality. When you focus TextView, there's an accessibility action to select a specific link. I could reproduce it with mentions but I'm unsure about links, maybe if I could scan for spannables.
Anyway, I could try to reproduce it and then we can decide.

charlag:
I keep description card intentionally, otherwise detailed status behaves like a regular one now.

charlag:
Hey
I've added/fixed some things and it can be reviewed at any point.
I'm not sure if we need to make a build to let people try it out or we can just get feedback from nightly later.

connyduck:
I would merge it after the next bugfix release if thats ok? In the meantime here would be the link for a build from this branch: https://app.bitrise.io/artifact/12488468/p/debe1dde81121cdf1d4514238f094c91

charlag:
@connyduck sure, as you like. That's for the link, I'll share it today!

Issue #1061, opened 2019-02-19
When sharing files to tusky, it doesn't return control after posting


vaartis:
Before Tusky 5 tusky returned control back to the app the file was shared from, but as of Tusky 5 beta 1 (from google play), it just opens the main timeline instead.

Steps to reproduce:
- Open any aplication that can share files with others
- Select tusky
- Send the post

Expected:
The control is transfered back to the application sharing the file

What actually happens:
Tusky opens the main timeline. An interesting detail is that in the "switch application" menu the icon is of the original application and pressing "back" while on the main timeline returns the control to that application.

* * * *
- Tusky Version: 5.0 beta 1
- Android Version: 8.0.0
- Android Device: OnePlus 5T
- Mastodon instance (if applicable):

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

connyduck:
closed by #1062

PR #1062, opened 2019-02-19
Fixed sharing files doesn't return control after posting

mihudevcom:
Issue #1061


charlag:
I didn't try it out but seems fine to me

connyduck:
I would have tried to move the sharing directly to `ComposeActivity`, but this is an even easier solution. Thx!

PR #1065, opened 2019-02-19
fix crash in AccountListFragment

connyduck:
Steps to reproduce: Go on a slow internet connection. Open and close one of the `AccountListFragment`s very fast. If the network request returns after the Fragment was closed, it crashed.

```
java.lang.NullPointerException: 
  at com.keylesspalace.tusky.util.ViewExtensionsKt.show (ViewExtensionsKt.java:6)
  at com.keylesspalace.tusky.fragment.AccountListFragment.onFetchAccountsFailure (AccountListFragment.java:327)
```

I moved the api requests to `Single`s, because if they get cancelled the callback is not invoked so it does not crash anymore. With the `callList` the error callback was invoked when cancelled.

This is the only crash I could find in Google Play for Tusky 5.0 beta 1.

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/fragment/AccountListFragment.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/fragment/AccountListFragment.kt
@@ -20,6 +20,7 @@ import android.util.Log
 import android.view.LayoutInflater
 import android.view.View
 import android.view.ViewGroup
+import androidx.lifecycle.Lifecycle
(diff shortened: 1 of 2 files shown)


charlag:
I believe we set up retrofit call adapter to use it's own connection pool automatically so no need to subscribe on IO?
I'm unsure where we subscribe (hard to read from mobile) but otherwise seems fine. Can look again tomorrow if needed.

connyduck:
yes you are right, thx!

Issue #1021, opened 2019-02-01
Hotkey to refresh views


NoraCodes:
I would really like to be able to refresh e.g. the home timeline or notifications view using the F5 key on my hardware keyboard.

* * * *
- Tusky Version: 5.1 from F-Droid
- Android Version: 6.0.1
- Android Device: Pinebook 11.4" 1080p
- Mastodon instance (if applicable):

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

PR #1068, opened 2019-02-20
Hotkey to refresh views

mihudevcom:
Fixes issue #1021

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/MainActivity.java
+++ b/app/src/main/java/com/keylesspalace/tusky/MainActivity.java
@@ -285,7 +285,9 @@ public final class MainActivity extends BottomSheetActivity implements ActionBut
         composeIntent.setAction(intent.getAction());
         composeIntent.setType(intent.getType());
         composeIntent.putExtras(intent);
+        composeIntent.setFlags(Intent.FLAG_ACTIVITY_NEW_TASK | Intent.FLAG_ACTIVITY_CLEAR_TASK);
         startActivity(composeIntent);
+        finish();
     }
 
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/keylesspalace/tusky/components/conversation/ConversationsFragment.kt
```diff
@@ -186,4 +193,11 @@ class ConversationsFragment : SFragment(), StatusActionListener, Injectable {
     companion object {
         fun newInstance() = ConversationsFragment()
     }
+
+    private val refreshReceiver = object : BroadcastReceiver() {
+        override fun onReceive(context: Context, intent: Intent) {
+            //refresh
+            initSwipeToRefresh()
```
connyduck: That does not look right, I think you just have to call `viewModel.refresh()` here

app/src/main/java/com/keylesspalace/tusky/MainActivity.java
```diff
@@ -593,4 +596,20 @@ public FloatingActionButton getActionButton() {
         return fragmentInjector;
     }
 
+    @Override
+    public boolean dispatchKeyEvent(KeyEvent event) {
```
connyduck: There is already a `onKeyDown` method in `MainActivity` where we handle other shortcuts. Not sure where the right place is for this, but it should be one single place to not clutter the code.

charlag:
I agree, hotkey should refresh only the current fragment and there should be interface to do that perhaps (it's only needed for timelnes and notifications or am I wrong? Could be even hardcoded if it's only two places)

PR #1071, opened 2019-02-22
fix crash on Android 5

connyduck:
I guess there are more sophisticated ways to fix this, but this needs to get out asap

cc @Tak


charlag:
Do we know why we cannot inject there?

connyduck:
The Injector will have a reference to the class and that causes `NoClassDefFoundError` even if its never injected on that platform.

See here for more info/other solutions: https://github.com/google/dagger/issues/1064

PR #1079, opened 2019-02-25
l10n: update Japanese translation

musashino205:
Synchronized values-ja/strings.xml with values/strings.xml and added
translations.

Signed-off-by: INAGAKI Hiroshi <musashino.open@gmail.com>


connyduck:
Thank you very much!

Issue #945, opened 2018-12-03
[Discussion] should we include Conscrypt for TLS 1.3 support?


connyduck:
Previously I was against including Conscrypt just to fix TLS compatibility of old Android versions, because it increases apk size significally (I think it was ~5Mb).
But adding it to support TLS 1.3 is something I would really like to discuss. On Google Play we have Android App Bundles, so at least for users who load Tusky from there the size increase won't be that bad. There are already quite a few Mastodon instances that support TLS 1.3.

A nice sideeffect would be that the strange elliptic curve bug on Android 7 will be goine as well.

👍 = yes TLS 1.3 
👎  = I prefer super small apk

charlag:
I think these days 5mb is *nothing*. If it's 15 or 30, then we should think about it maybe

mimi89999:
I think that it is a good idea to include Conscrypt. Other applications like Conversations already include it.

tateisu:
which instances support TLS1.3 ?

connyduck:
mamot.fr mstdn.jp are the bigger ones I know.

stom79:
Just my two cents about conscrypt. Take care of the releases. Currently 1.3 is stable 1.4 and 1.4.1 created a lot of troubles with Mastalab.

PR #1083, opened 2019-02-26
upgrade OkHttp, add Conscrypt

connyduck:
this disables TLS 1 and 1.1 and enables TLS 1.3

closes https://github.com/tuskyapp/Tusky/issues/945


app/build.gradle
```diff
@@ -94,8 +94,9 @@ dependencies {
     implementation 'com.squareup.retrofit2:converter-gson:2.5.0'
     implementation 'com.squareup.retrofit2:adapter-rxjava2:2.5.0'
     implementation 'com.squareup.picasso:picasso:2.5.2'
-    implementation 'com.squareup.okhttp3:okhttp:3.12.0'
-    implementation 'com.squareup.okhttp3:logging-interceptor:3.12.0'
+    implementation 'com.squareup.okhttp3:okhttp:3.13.1'
+    implementation 'com.squareup.okhttp3:logging-interceptor:3.13.1'
+    implementation "org.conscrypt:conscrypt-android:2.0.0"
```
charlag: Did you measure the size change already?

app/src/main/java/com/keylesspalace/tusky/util/OkHttpUtils.java
```diff
@@ -121,47 +81,6 @@ private static Interceptor getUserAgentInterceptor() {
         };
     }
 
-    /**
```
charlag: We don't need this anymore because we just bundle the provider, right?

connyduck:
> This can be a breaking thing but I support it. Just have concerns about the size and how many instances may break.

instances that don't support TLS 1.2  x_X   must be the worst admin ever

app/src/main/java/com/keylesspalace/tusky/util/OkHttpUtils.java
connyduck: yes

app/src/main/java/com/keylesspalace/tusky/TuskyApplication.java
```diff
@@ -62,6 +66,8 @@
     public void onCreate() {
         super.onCreate();
 
+        Security.insertProviderAt(Conscrypt.newProvider(), 1);
```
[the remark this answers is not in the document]
connyduck: from the doc of this method:

> Adds a new provider, at a specified position. The position is the preference order in which providers are searched for requested algorithms.  The position is 1-based, that is, 1 is most preferred, followed by 2, and so on.

charlag:
Well, true, maybe it would even motivate some people. I think it's good to go then.

app/build.gradle
connyduck: Size of Tusky 5.1 apk: 3.749kB
size apk with Conscrypt:  6.986kB (download size will be much smaller for Google Play users because of App Bundles)

charlag: While it's 2x increase, I don't think many people will notice because it's still quite small.

connyduck:
I tried it out with [this interceptor](https://gist.github.com/handstandsam/d7d7956a05698b23cad14dd404350e23) and TLS 1.3 works on all Android versions except 5. No idea why it does not work there, but there is also no reason to investigate further, TLS 1.2 is fine.

PR #1085, opened 2019-02-27
fix crash in account activity

connyduck:
A try to fix this mysterious crash I keep seeing in Google Play crash reports. 
https://gist.github.com/connyduck/bdf1c765fbdd526e590c4ef390948f23

Its only on Android 8 and the approach is to just not override this method anymore.


connyduck:
Good news: This fixed the crash, I don't see it in logs anymore!

PR #1089, opened 2019-02-28
Add autocompletion for custom emoji

### LINKED ISSUE #398: Support for the custom emoji
Mastodon custom emoji seems to be completed by now and it makes sense to implement them.
There are actually two parts: showing them in timeline and writing them.
I think we should start with showing them in timeline. I think it can be done with `ImageSpan`.
 I may try to do this if I have time but I just want to track it by now. If anyone's going to implement it - please, post it here.

- [x] Show custom emoji in posts
- [x] Add ability to use custom emoji
- [ ] Autocomplete custom emoji

autumnontape:
Closes #398.

Maybe it'd be a good idea to add more tests?

I don't have much Android experience, so I might not be doing everything in the best way.

autocompletion of "hello :masto", showing ":masto_ball:" and ":mastodon:" as suggestions, with high-resolution previews of the emoji in question

(Don't worry about the low resolution of the ":mastodon:" emoji, that's just what it looks like on that instance.)

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/ComposeActivity.java
+++ b/app/src/main/java/com/keylesspalace/tusky/ComposeActivity.java
@@ -315,16 +315,13 @@ public final class ComposeActivity
             mastodonApi.getCustomEmojis().enqueue(new Callback<List<Emoji>>() {
                 @Override
                 public void onResponse(@NonNull Call<List<Emoji>> call, @NonNull Response<List<Emoji>> response) {
-                    emojiList = response.body();
+                    List<Emoji> emojiList = response.body();
                     if(emojiList == null) {
                         emojiList = Collections.emptyList();
                     }
                     Collections.sort(emojiList, (a, b) ->
                         a.getShortcode().toLowerCase(Locale.ROOT).compareTo(
                             b.getShortcode().toLowerCase(Locale.ROOT)));
-
-                    emojiListRetrievalLatch.countDown();
-
                     setEmojiList(emojiList);
                     cacheInstanceMetadata(activeAccount);
                 }
@@ -332,10 +329,7 @@ public final class ComposeActivity
                 @Override
                 public void onFailure(@NonNull Call<List<Emoji>> call, @NonNull Throwable t) {
                     Log.w(TAG, "error loading custom emojis", t);
-
                     loadCachedInstanceMetadata(activeAccount);
-
-                    emojiListRetrievalLatch.countDown();
                 }
             });
         } else {
@@ -1621,13 +1615,16 @@ public final class ComposeActivity
         if(instanceEntity != null) {
             Integer max = instanceEntity.getMaximumTootCharacters();
             maximumTootCharacters = (max == null ? STATUS_CHARACTER_LIMIT : max);
-            emojiList = instanceEntity.getEmojiList();
-            setEmojiList(emojiList);
+            setEmojiList(instanceEntity.getEmojiList());
             updateVisibleCharactersLeft();
         }
     }
 
     private void setEmojiList(@Nullable List<Emoji> emojiList) {
+        this.emojiList = emojiList;
+
+        emojiListRetrievalLatch.countDown();
+
         if (emojiList != null) {
             emojiView.setAdapter(new EmojiAdapter(emojiList, ComposeActivity.this));
             enableButton(emojiButton, true, emojiList.size() > 0);
--- a/app/src/main/java/com/keylesspalace/tusky/ComposeActivity.java
+++ b/app/src/main/java/com/keylesspalace/tusky/ComposeActivity.java
@@ -1596,6 +1596,11 @@ public final class ComposeActivity
                             }
                         }
 
+                        if (!results.isEmpty() && !resultsInside.isEmpty()) {
+                            // both lists have results. include a separator between them.
+                            results.add(new ComposeAutoCompleteAdapter.ResultSeparator());
+                        }
+
                         results.addAll(resultsInside);
                         return results;
                     } else {
--- a/app/src/main/java/com/keylesspalace/tusky/ComposeActivity.java
+++ b/app/src/main/java/com/keylesspalace/tusky/ComposeActivity.java
@@ -66,7 +66,7 @@ import com.google.android.material.snackbar.Snackbar;
 import com.google.gson.Gson;
 import com.google.gson.reflect.TypeToken;
 import com.keylesspalace.tusky.adapter.EmojiAdapter;
-import com.keylesspalace.tusky.adapter.MentionTagAutoCompleteAdapter;
+import com.keylesspalace.tusky.adapter.ComposeAutoCompleteAdapter;
 import com.keylesspalace.tusky.adapter.OnEmojiSelectedListener;
 import com.keylesspalace.tusky.db.AccountEntity;
 import com.keylesspalace.tusky.db.AppDatabase;
@@ -84,7 +84,7 @@ import com.keylesspalace.tusky.service.SendTootService;
 import com.keylesspalace.tusky.util.CountUpDownLatch;
 import com.keylesspalace.tusky.util.DownsizeImageTask;
 import com.keylesspalace.tusky.util.ListUtils;
-import com.keylesspalace.tusky.util.MentionTagTokenizer;
+import com.keylesspalace.tusky.util.ComposeTokenizer;
 import com.keylesspalace.tusky.util.SaveTootHelper;
 import com.keylesspalace.tusky.util.SpanUtilsKt;
 import com.keylesspalace.tusky.util.StringUtils;
@@ -158,7 +158,7 @@ import static com.uber.autodispose.android.lifecycle.AndroidLifecycleScopeProvid
 public final class ComposeActivity
         extends BaseActivity
         implements ComposeOptionsListener,
-        MentionTagAutoCompleteAdapter.AutocompletionProvider,
+        ComposeAutoCompleteAdapter.AutocompletionProvider,
         OnEmojiSelectedListener,
         Injectable, InputConnectionCompat.OnCommitContentListener {
 
@@ -521,8 +521,8 @@ public final class ComposeActivity
         });
 
         textEditor.setAdapter(
-                new MentionTagAutoCompleteAdapter(this));
-        textEditor.setTokenizer(new MentionTagTokenizer());
+                new ComposeAutoCompleteAdapter(this));
+        textEditor.setTokenizer(new ComposeTokenizer());
 
         // Add any mentions to the text field when a reply is first composed.
         if (mentionedUsernames != null) {
@@ -1547,7 +1547,7 @@ public final class ComposeActivity
     }
 
     @Override
-    public List<MentionTagAutoCompleteAdapter.AutocompleteResult> search(String token) {
+    public List<ComposeAutoCompleteAdapter.AutocompleteResult> search(String token) {
         try {
             switch (token.charAt(0)) {
                 case '@':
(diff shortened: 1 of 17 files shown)


autumnontape:
Heads up, I'm going to do a force push in a few minutes to get it down to two commits.

autumnontape:
It may be better to move the `emojiListRetrievalLatch.countDown()` calls into `setEmojiList()` and either remove the `emojiList` argument to `setEmojiList()` or have `setEmojiList()` set the `emojiList` field itself rather than relying on the field having been set already. There doesn't seem to be any reason not to make any of these changes.

autumnontape:
I went ahead and made the last change I mentioned in the former comment. I think this makes the code less prone to future error.

app/src/main/java/com/keylesspalace/tusky/ComposeActivity.java
```diff
@@ -1569,6 +1571,28 @@ private void saveDraftAndFinish() {
                         Log.e(TAG, String.format("Autocomplete search for %s failed.", token));
                         return Collections.emptyList();
                     }
+                case ':':
+                    try {
+                        emojiListRetrievalLatch.await();
```
charlag: I'm not sure we need to wait for latch. Autocomplete should be fast and if it doesn't work then maybe it makes sense to just skip. Did you notice the change when you used it?

app/src/main/res/layout/item_autocomplete_emoji.xml
```diff
@@ -0,0 +1,33 @@
+<?xml version="1.0" encoding="utf-8"?>
+<RelativeLayout xmlns:android="http://schemas.android.com/apk/res/android"
```
charlag: Please change `RelativeLayout` to be something else please (`LinearLayout` or `ConstraintLayout`). Ask for help if needed.

app/src/main/java/com/keylesspalace/tusky/ComposeActivity.java
autumnontape: The latch is there so that if you begin to type an emoji code and then stop, the result will always be consistent. So you'll never have to type an additional character and then delete it to see the results you want, if you're using autocomplete as a search.

The case where the `await` doesn't return immediately is out of the ordinary, and I have a fast connection, so I don't notice it in practice. I did test the feature with a throttled connection, and it works as expected.

I could remove the latch without affecting most potential users. I just figure it's a pretty small difference in code for some nice consistency.

app/src/main/java/com/keylesspalace/tusky/adapter/MentionTagAutoCompleteAdapter.java
```diff
@@ -44,6 +45,7 @@
         implements Filterable {
     private static final int ACCOUNT_VIEW_TYPE = 0;
     private static final int HASHTAG_VIEW_TYPE = 1;
+    private static final int EMOJI_VIEW_TYPE = 2;
```
Tak: Maybe it makes sense to convert this to a Real Enum™ at this point?

autumnontape: Sure, I'll do that real quick

Tak:
Maybe it makes sense to show the completion results without the colons?
:+1: overall

app/src/main/java/com/keylesspalace/tusky/adapter/MentionTagAutoCompleteAdapter.java
autumnontape: Oh, `BaseAdapter` mandates that a lot of the methods that deal with these constants use ints. It probably wouldn't be worthwhile to use an enum, then.

charlag:
I think colons are okay too

app/src/main/java/com/keylesspalace/tusky/ComposeActivity.java
charlag: Yeah, maybe it's for the best there

autumnontape:
The emoji shortcodes feel naked to me without colons :stuck_out_tongue:

autumnontape:
All right, I converted the RelativeLayout to a LinearLayout as @charlag asked me to. Android Studio is telling me now that I could replace the whole thing with just a TextView with compound drawables, but doing that doesn't seem worth it, since it doesn't look like it'd be straightforward to hook a compound drawable into Picasso.

autumnontape:
Hmm, hang on, I accidentally decreased the padding between the emoji preview and the shortcode. I liked it better the way it was before, so I'm going to fix that and force-push again.

charlag:
Huh, it can actually be a TextView, that's right. It's not hard to hook it to Picasso either, we have already some places where we use `into(new Target())` (maybe in Kotlin), you can check it

autumnontape:
Okay, the LinearLayout is pixel-for-pixel identical to the old RelativeLayout now.

autumnontape:
@charlag I'll check it out.

charlag:
I don't mind either, just saying

autumnontape:
It looks like I'd have to add code to set the size of the image and code to make it work with Picasso, if I used a compound drawable here. It doesn't really seem like a step forward.

Any more feedback?

charlag:
I wouldn't bother tbh, let's wait until @connyduck looks at this
Great job btw!

autumnontape:
Thanks!

autumnontape:
Hmm, maybe `MentionTagAutoCompleteAdapter` should be renamed to `ComposeAutoCompleteAdapter` or something.

charlag:
That's a good idea

autumnontape:
All right, I've renamed `MentionTagAutoCompleteAdapter`, `MentionTagTokenizer`, and `MentionTagTokenizerTest` to start with "Compose" instead.

app/src/main/java/com/keylesspalace/tusky/ComposeActivity.java
```diff
@@ -1557,18 +1559,40 @@ private void saveDraftAndFinish() {
...
+                    } catch (InterruptedException e) {
+                        Log.e(TAG, String.format("Autocomplete search for %s was interrupted.", token));
+                        return Collections.emptyList();
+                    }
+                    if (emojiList != null) {
+                        final String incomplete = token.substring(1);
+
+                        List<Emoji> results = CollectionsKt.filter(
+                                emojiList,
+                                emoji -> emoji.getShortcode().startsWith(incomplete)
```
connyduck: Why only `startsWith`? Wouldn't it be better to return all emojis that contain the string, with emojis that start with it first?

app/src/main/res/layout/item_autocomplete_emoji.xml
```diff
@@ -0,0 +1,33 @@
...
+    <TextView
+        android:id="@+id/shortcode"
+        android:layout_width="wrap_content"
+        android:layout_height="wrap_content"
+        android:layout_marginStart="8dp"
+        android:ellipsize="end"
+        android:maxLines="1"
+        android:textColor="?android:textColorPrimary"
+        android:textSize="?attr/status_text_large"
+        android:textStyle="normal|bold"
```
connyduck: Bold? It looks very different than the autocomplete of a hashtag, I think thats weird. But I actually like the bold, maybe add a dedicated autocomplete layout for hashtags that is bold as well?

app/src/main/java/com/keylesspalace/tusky/adapter/ComposeAutoCompleteAdapter.java
```diff
@@ -159,6 +163,31 @@ public View getView(int position, @Nullable View convertView, @NonNull ViewGroup
...
+                if (view.getTag() == null) {
+                    view.setTag(new EmojiViewHolder(view));
+                }
+                emojiViewHolder = (EmojiViewHolder) view.getTag();
+
+                EmojiResult emojiResult = ((EmojiResult) getItem(position));
+                if (emojiResult != null) {
+                    Emoji emoji = emojiResult.emoji;
+                    String format = context.getString(R.string.emoji_shortcode_format);
+                    String formattedShortcode = String.format(format, emoji.getShortcode());
```
connyduck: `context.getString` takes format arguments, you could save a line of code if you want

app/src/main/java/com/keylesspalace/tusky/ComposeActivity.java
autumnontape: Wow, that's a great idea! The second list will be a lot longer than the first in many cases, so maybe it'd be a good idea to put a horizontal rule or something between them?

I should probably also make the matching case-insensitive.

app/src/main/res/layout/item_autocomplete_emoji.xml
autumnontape: All right, I'll try it.

autumnontape: I wasn't sure how it'd be, but I like the bold hashtag autocomplete a lot. Now all three kinds of autocomplete bold their results. :+1:

autumnontape:
All right, I've addressed all of the issues @connyduck brought up. I'm going to see if I can figure out how to put a horizontal rule in the autocomplete results now.

autumnontape:
Hmm... The default divider is exactly the same color as the autocomplete results' background in the dark theme.

autumnontape:
All right, I had to add some new resources to get the line to show up in the dark theme, but I added a divider. Is it good? Should I change/revert it?

divider

autumnontape:
Oh, I might as well admit -- I have no idea what this `//noinspection ConstantConditions` comment I've been copy-pasting along with the layout inflating code in ComposeAutoCompleteAdapter.java actually does? So maybe some or all of the ones I added shouldn't be there.

charlag:
@autumnontape it's a marker for the IDE to not warn you about certain things, like it think that the variable is `@Nullable` but you know that it's not and don't want to check. Try to remove some and see if IDE complains.
You did a stellar work there :star:

autumnontape:
@charlag I removed all the "noinspection" comments in `ComposeAutoCompleteAdapter.getView()`, including the one that was already there, and ran a full build, but nothing seems to be complaining about it. So I guess either the original comment doesn't need to be there anymore, or whatever tool it was talking to isn't being run on my end/I'm not seeing its output.

autumnontape:
I am using Android Studio and not seeing anything in-editor either, FYI.

autumnontape:
Also FYI, I'm guessing my next push will be the last one, so it'll be another force-push to clean things up just a little.

autumnontape:
So all that's left for me is to possibly remove the `noinspection` comments since they don't seem to be doing anything. I have the "constant conditions and exceptions" inspection enabled, which I think is the one those comments are bypassing.

I just need to know if removing them is okay for you all or if I'm possibly missing something.

charlag:
Yes, please, I checked and it doesn't have a warning, feel free to remove them.

I had another look and it seems that shortcode format doesn't need to me translatable? Do I miss something here?
Otherwise I think it's good to go, asking @connyduck to merge it

autumnontape:
It wasn't marked `translatable="false"` because I copy-pasted it from this:

```xml
<string name="status_username_format">\@%s</string>
```

Which seemingly has identical "translations" across all the translation files.

In any case, I've now marked it as not translatable and removed the unnecessary `noinspection` comments, including the one that was there before, so I believe it's ready to merge :tada:

connyduck:
Awesome, thx!

autumnontape:
@connyduck Did you forget to merge this?

connyduck:
no, I did not want it in the 5.2 release

PR #1091, opened 2019-03-01
Fix crash on Android 9 when replying from notification

connyduck:
`android.util.AndroidRuntimeException: Calling startActivity() from outside of an Activity  context requires the FLAG_ACTIVITY_NEW_TASK flag.`


connyduck:
Its a [change in behavior on Android 9](https://developer.android.com/about/versions/pie/android-9.0-changes-all#fant-required) and I guess previously not enough people were using Android 9 to draw my attention to this crash

PR #1094, opened 2019-03-03
Fix visibility of answer from DM column, from private to direct

ariasuni:
#1012 hardcoded visibility level when answering from a toot displayed in the DM column to `private` but direct messages have the visibility level `direct`, so this PR corrects the value.

Also, since people can post by mistake to their followers instead of directly, I think you should consider releasing a minor version of Tusky containing this fix.


connyduck:
Thx! Tusky 5.2 will be released soon.

PR #1108, opened 2019-03-06
Add support for language switching

### LINKED ISSUE #644: [i18n] Language unavailable on Android
I'm opening this issue because for some language it's impossible to have the application translated as the application switches to the system language.
The problem is that for example, Android is not available in Occitan, so the application won't show in this language.
Could we add a menu in the setting to set the language one wants?

ariasuni:
Fix #644 

I mostly followed [this tutorial](https://proandroiddev.com/change-language-programmatically-at-runtime-on-android-5e6bc15c758), adapting it to Tusky.

* Use default system language or any language added to Tusky (language list is static and needs to be updated when a new translation is added, but it is the simpler way)
* Language change immediately, similarly to theme change but with a bit more code here and there to correctly update locale.
* Works with locales that are more than two letters (e.g. `pt-rBR` or `zh-rTW`).

Possible changes, based on your feedback:

* I hardcoded the current locales in `donottranslate.xml`, because it is the simplest yet clean solution. See [here](https://stackoverflow.com/questions/11611065/get-the-applications-resources-languages) for other ways to do it that you might find better.
* The code from the tutorial suggest that the class `LocaleManager` should be provided to different classes through dependency injection, but I do not know if it’s relevant here or how to do it.

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/BaseActivity.java
+++ b/app/src/main/java/com/keylesspalace/tusky/BaseActivity.java
@@ -59,8 +59,6 @@ public abstract class BaseActivity extends AppCompatActivity implements Injectab
 
     protected static final int PERMISSIONS_REQUEST_WRITE_EXTERNAL_STORAGE = 1;
 
-    private final String TAG = "BaseActivity";
-
     @Override
     protected void onCreate(@Nullable Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
--- a/app/src/main/java/com/keylesspalace/tusky/BaseActivity.java
+++ b/app/src/main/java/com/keylesspalace/tusky/BaseActivity.java
@@ -102,7 +102,6 @@ public abstract class BaseActivity extends AppCompatActivity implements Injectab
     @Override
     protected void attachBaseContext(Context base) {
         super.attachBaseContext(TuskyApplication.localeManager.setLocale(base));
-        Log.d(TAG, "attachBaseContext");
     }
 
     protected boolean requiresLogin() {
--- a/app/src/main/java/com/keylesspalace/tusky/BaseActivity.java
+++ b/app/src/main/java/com/keylesspalace/tusky/BaseActivity.java
@@ -16,6 +16,7 @@
 package com.keylesspalace.tusky;
 
 import android.app.ActivityManager;
+import android.content.Context;
 import android.content.Intent;
 import android.content.SharedPreferences;
 import android.graphics.Bitmap;
@@ -58,6 +59,8 @@ public abstract class BaseActivity extends AppCompatActivity implements Injectab
 
     protected static final int PERMISSIONS_REQUEST_WRITE_EXTERNAL_STORAGE = 1;
 
+    private final String TAG = "BaseActivity";
+
     @Override
     protected void onCreate(@Nullable Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
@@ -94,6 +97,12 @@ public abstract class BaseActivity extends AppCompatActivity implements Injectab
         }
 
         callList = new ArrayList<>();
+    }
+
+    @Override
+    protected void attachBaseContext(Context base) {
+        super.attachBaseContext(TuskyApplication.localeManager.setLocale(base));
+        Log.d(TAG, "attachBaseContext");
     }
 
     protected boolean requiresLogin() {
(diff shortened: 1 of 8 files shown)


app/src/main/java/com/keylesspalace/tusky/BaseActivity.java
```diff
@@ -96,6 +99,12 @@ protected void onCreate(@Nullable Bundle savedInstanceState) {
         callList = new ArrayList<>();
     }
 
+    @Override
+    protected void attachBaseContext(Context base) {
+        super.attachBaseContext(TuskyApplication.localeManager.setLocale(base));
+        Log.d(TAG, "attachBaseContext");
```
connyduck: I would remove these Logs, I don't think they are needed anymore

app/src/main/res/xml/preferences.xml
```diff
@@ -39,6 +39,14 @@
             android:key="absoluteTimeView"
             android:title="@string/pref_title_absolute_time" />
 
+        <ListPreference
+            android:defaultValue="default"
+            android:entries="@array/language_entries"
+            android:entryValues="@array/language_values"
+            android:key="language"
+            android:summary="%s"
+            android:title="@string/pref_title_language_settings" />
```
connyduck: Move the setting up. At least above the two SwitchPreferences, but I would even put it above the text size.
Also, please give it this icon: http://materialdesignicons.com/icon/translate

app/src/main/java/com/keylesspalace/tusky/util/LocaleManager.kt
```diff
@@ -0,0 +1,47 @@
...
+
+import com.keylesspalace.tusky.util.getNonNullString
+
+
+class LocaleManager(context: Context) {
+
+    private var prefs: SharedPreferences = PreferenceManager.getDefaultSharedPreferences(context)
+
+    fun setLocale(context: Context): Context {
+        val language = prefs.getNonNullString("language", "")
```
connyduck: why is the default `""`? I think it should be `"default"`.

app/src/main/res/values/donottranslate.xml
```diff
@@ -30,4 +30,65 @@
...
+        <item>ko</item>
+        <item>zh-TW</item>
+        <item>zh-SG</item>
+        <item>zh-MO</item>
+        <item>zh-CN</item>
+        <item>zh-HK</item>
+        <item>ja</item>
+    </string-array>
+
+    <string name="pref_title_language_settings">Language</string>
```
connyduck: This string is translateable and should be in `strings.xml`

app/src/main/java/com/keylesspalace/tusky/PreferencesActivity.kt
```diff
@@ -124,31 +124,38 @@ class PreferencesActivity : BaseActivity(), SharedPreferences.OnSharedPreference
...
             "statusTextSize" -> {
                 restartActivitiesOnExit = true
             }
             "absoluteTimeView" -> {
                 restartActivitiesOnExit = true
             }
+            "language" -> {
+                val language = sharedPreferences.getNonNullString("language", "en")
+                Log.d("activeLanguage", language)
+                sharedPreferences.edit().putString("language", language).apply()
```
connyduck: You can delete these three lines, they are doing nothing.

charlag:
I would say that we should inject LocaleManager to not suffer with tests but we can do it later if it becomes a problem.

app/src/main/java/com/keylesspalace/tusky/util/LocaleManager.kt
ariasuni: Well there is already a default value in the XML file but that costs nothing to add `default` here too.

app/src/main/java/com/keylesspalace/tusky/TuskyApplication.java
```diff
@@ -103,6 +111,19 @@ protected void initSecurityProvider() {
...
+    protected void attachBaseContext(Context base) {
+        localeManager = new LocaleManager(base);
+        super.attachBaseContext(localeManager.setLocale(base));
+    }
+
+    @Override
+    public void onConfigurationChanged(Configuration newConfig) {
+        super.onConfigurationChanged(newConfig);
+        localeManager.setLocale(this);
+        Log.d(TAG, "onConfigurationChanged: " + newConfig.locale.toString());
```
connyduck: log

app/src/main/java/com/keylesspalace/tusky/BaseActivity.java
```diff
@@ -58,6 +59,8 @@
 
     protected static final int PERMISSIONS_REQUEST_WRITE_EXTERNAL_STORAGE = 1;
 
+    private final String TAG = "BaseActivity";
```
connyduck: unused

Issue #452, opened 2017-11-13
[Feature Request]: Recalculate toots relative timestamp on pull-to-refresh


MasterGroosha:
When you pull-to-refresh any timeline, relative timestamp mark (as seen on screenshot below) isn't recalculated relatively to the current time. It does update on switching back and forth between different timelines though.

![timestamp in toot view](https://user-images.githubusercontent.com/2780044/32713651-440cc4f6-c85b-11e7-935a-10708090e2c4.png)

Using Tusky 1.3.1.

charlag:
We could do better than refreshing on pull, I think we should automatically change it with time.
Something like this may work:
https://stackoverflow.com/a/14512130/1923879
I'm not sure if broadcastReceiver is cheap or not and if it worths putting in a holder. If it's not we'd better notify adapter.

MasterGroosha:
So time will be recalculated in real-time for each new item from ListView appearing on screen? Won't it cause any kind of freezes on low-end devices?

charlag:
This thing does this only every minute.
It will only recalculate it for views currently visible (and probably in the nearby timeline as well).
I've also found some libraries
https://github.com/damien5314/TimeSinceTextView

https://github.com/curioustechizen/android-ago

connyduck:
Yes, we could do better (and we should) but I think a quck fix that calls `notifyDataSetChanged();` on every refresh would also be ok for now.

charlag:
This leaves us with other problems. First, it may be quite expensive but it shouldn't be too bad. Second, we may loose things like position in the timeline (but since we're in the top on refresh it shouldn't be that bad). This can and should be solved by utilizing stable IDs. The thing is that we need mapping from ```String``` IDs to ```long```s. Maybe we could keep a map or something. Pinterest released a framework for RecyclerView recently which does some of that bookkeeping, maybe we should check how they do it.
I would love to just do ```notfiyDatasetHasChanged()``` because it would let us get rid of some ugly logic (and move loading from the Fragments)

charlag:
It's implemented with #1130 / #1113 I believe

PR #1113, opened 2019-03-07
Refresh message timestamp every minute

pandasoft0:
tuskyapp/Tusky#452

This pull request doesn't add the "update a timestamp" on refresh, but it adds "update timestamp" every minute.
The main idea is to send the notifyItemRangeChanged to the adapter with a payload.
Viewholder checks if payload exists and if so - it updates only the fields specified at the payload (creation timestamp for now). If payloads are empty - adapter works just like before.

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/adapter/StatusBaseViewHolder.java
+++ b/app/src/main/java/com/keylesspalace/tusky/adapter/StatusBaseViewHolder.java
@@ -44,7 +44,9 @@ import at.connyduck.sparkbutton.SparkEventListener;
 import kotlin.collections.CollectionsKt;
 
 public abstract class StatusBaseViewHolder extends RecyclerView.ViewHolder {
-
+    public static class Key{
+        public static final String KEY_CREATED = "created";
+    }
     private TextView displayName;
     private TextView username;
     private ImageButton replyButton;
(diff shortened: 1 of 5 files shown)


charlag:
Hey
Sorry for not reacting to it so far. I briefly looked at the code and I have some design questions. Proposed solution looks like a hack to me tbh. Contents didn't change but we want to update views. Adapter shouldn't be concerned about that (more, we use diffing mechanisms and I wouldn't go around them).
I would rather start a timer in each ViewHolder on bind and stop it on rebind/unmount. Because there are only so many `ViewHolder` (around 10-20 I would say) it shouldn't be a big problem.
Maybe even better than that, there may be some global timer to which they could subscribe/unsubscribe to update simultaneously when the signal is sent.
I would love to hear from others, e.g. @connyduck @Tak

Tak:
I agree with @charlag's analysis

connyduck:
I think starting the timers should be in the Fragments because Fragments have lifecycles and the timers need to be stopped. But also we should not be going around the differ. What about just sending the same list to the differ and the differ dispatches the payload?

pandasoft0:
@connyduck I've updated the logic. Now, interval observable calls the updateAdapter method every minute. Diff callback always retruns the "false" on areContentTheSame, but at the getPayload it compare old item with new and if they are equal - returns only payload with update timestamp. If items are different - returns null. It means the full refresh the view. I added some comments to to the code to explain this logic.

charlag:
I don't ask to change anything yet, but my opinion is that it still breaks some of the diffing. It will force adapter to update all contents and generally do more work than necessarily. I think adapter might create some timer and pass it to `ViewHolder`s so that they can redraw or something like that.

pandasoft0:
@charlag this is a very simple solution. It works over your code and doesn't change the logic of the adapter or the differ. Could you provide an example when updated code refresh all content instead of change timestamps for the same content? I don't see a sense to pass any timer to the adapter or view holders. It may cause a memory or resource leaks. If you think, that provide solution is bad - just close the pull request and i will not create it again, but i will be very grateful if you send me a link to your solution when you fix this issue.

charlag:
@pandasoft0 hey, I'm sorry if I sound too harsh. I just introduced many bugs in timelines and I'm very cautious. I'm very grateful that you're trying to solve this. Please don't take it personally. I'm not a maintainer and I wouldn't close anything even then.
Such an example would be boost/fav of a status which would change all viewHolders instead of just one. Same goes for adding/removing items. We disabled update animation of individuals items, we may enable it again to see what's changing. Maybe there are more cases but I cannot remember any. It's not critical of course but kinda defeats the point of the differ.

pandasoft0:
@charlag My solution uses the old ** deepEquals** method. But i moved call of this function from the **areContentsTheSame** to the **getChangePayload** method. If content of items are the same - it updates the timestamp only, if content of items differ - it returns null as a payload and viewholder refreshes all content.

charlag:
Ah. my apologies! I somehow completely missed that it's part of the `ItemCallback`. It's pretty elegant actually.
I think we can dismiss my abstract concerns.
Sorry again.

PR #1120, opened 2019-03-09
Improve Dependency Injection

connyduck:
I also tried to inject `LocaleManager`, but unfortunately we need it in `attachBaseContext` and thats called prior to injection. 😔


charlag:
We can try to inject at the earlier stage (in Injector) but I'm not sure it's possible. I can look at it later.

connyduck:
Yes would be cool if you can figure it out. For now its fine as long as it works.

Issue #966, opened 2018-12-26
Add a media download button to cards on timelime


cun-tee:
Add a media download button to cards on timelime. The current implementation is clunky and inefficient.

C.

* * * *
- Tusky Version: 4.1
- Android Version: oreo
- Android Device: samsung
- Mastodon instance (if applicable):

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

charlag:
Hi!
Would you mind describing your use case and why current implementation doesn't work for you? Thanks.

cun-tee:
Yes when downloading artwork from blogs have to find the art  open it, then click again a few times to get the download button. I'm sure others have this issue and just slows down the process. 4/5 times on each item to get to the download link versus 1 button press on the item download button on the timeline.

C.

cun-tee:
When scrolling through a lost of artwork, I have to 1) click once to open it, then 2) click it again as the buttons have disappeared, then 3) press the download button, then 4) press the back button, then 5) scroll down and repeat.

other apps have the download button on the timeline (see Tumbviewer on play store) so I have to 1) press download and 2) scroll down. Which is 60% less steps and over time is a lot faster and efficient.

C.

cun-tee:
Tumbviewer here https://play.google.com/store/apps/details?id=com.nutrition.humblr

Tak:
This one can be closed, right?

Tak:
Implemented in #1121

PR #1121, opened 2019-03-11
Add option to download all media for a toot with one tap

Tak:
...plus one to open the menu :wink: 
- Refactor / generalize external storage permission request
- Add extended status menu item for downloading all media
Addresses #966

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/BaseActivity.java
+++ b/app/src/main/java/com/keylesspalace/tusky/BaseActivity.java
@@ -19,11 +19,13 @@ import android.app.ActivityManager;
 import android.content.Context;
 import android.content.Intent;
 import android.content.SharedPreferences;
+import android.content.pm.PackageManager;
 import android.graphics.Bitmap;
 import android.graphics.BitmapFactory;
 import android.graphics.Color;
 import android.graphics.PorterDuff;
 import android.graphics.drawable.Drawable;
+import android.os.Build;
 import android.os.Bundle;
 import android.preference.PreferenceManager;
 import android.util.Log;
@@ -37,9 +39,11 @@ import com.keylesspalace.tusky.db.AccountEntity;
 import com.keylesspalace.tusky.db.AccountManager;
 import com.keylesspalace.tusky.di.Injectable;
 import com.keylesspalace.tusky.interfaces.AccountSelectionListener;
+import com.keylesspalace.tusky.interfaces.PermissionRequester;
 import com.keylesspalace.tusky.util.ThemeUtils;
 
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.List;
 
 import javax.inject.Inject;
@@ -48,6 +52,8 @@ import androidx.annotation.Nullable;
 import androidx.annotation.StringRes;
 import androidx.appcompat.app.AlertDialog;
 import androidx.appcompat.app.AppCompatActivity;
+import androidx.core.app.ActivityCompat;
+import androidx.core.content.ContextCompat;
 import retrofit2.Call;
 
 public abstract class BaseActivity extends AppCompatActivity implements Injectable {
@@ -59,7 +65,9 @@ public abstract class BaseActivity extends AppCompatActivity implements Injectab
     @Inject
     public AccountManager accountManager;
 
-    protected static final int PERMISSIONS_REQUEST_WRITE_EXTERNAL_STORAGE = 1;
+    protected static final int BUILD_VERSION_ANY = -1;
+    private static final int REQUESTER_NONE = Integer.MAX_VALUE;
+    private HashMap<Integer, PermissionRequester> requesters;
 
     @Override
     protected void onCreate(@Nullable Bundle savedInstanceState) {
@@ -98,6 +106,7 @@ public abstract class BaseActivity extends AppCompatActivity implements Injectab
         }
 
         callList = new ArrayList<>();
+        requesters = new HashMap<>();
     }
 
     @Override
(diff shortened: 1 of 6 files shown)


charlag:
Sounds cool, thanks! I'll try to have a look once I have enough energy

app/src/main/java/com/keylesspalace/tusky/BaseActivity.java
```diff
@@ -222,4 +235,37 @@ public void showAccountChooserDialog(CharSequence dialogTitle, boolean showActiv
...
+            }
+            if (permissionsToRequest.isEmpty()) {
+                return true;
+            }
+
+            int newKey = requester == null ? REQUESTER_NONE : requesters.size();
+            String[] permissionsCopy = new String[permissionsToRequest.size()];
+            permissionsToRequest.toArray(permissionsCopy);
+            ActivityCompat.requestPermissions(this, permissionsCopy, newKey);
+            if (newKey != REQUESTER_NONE) {
```
charlag: I would put request key there before requesting the permission. I had such bugs before when permissions are given right away or similar.

app/src/main/java/com/keylesspalace/tusky/BaseActivity.java
```diff
@@ -222,4 +235,37 @@ public void showAccountChooserDialog(CharSequence dialogTitle, boolean showActiv
...
+                requesters.put(newKey, requester);
+            }
+            return false;
+        }
+
+        // This version doesn't require the permissions we want
```
charlag: Would it be simpler if we just called callback?

app/src/main/java/com/keylesspalace/tusky/BaseActivity.java
```diff
@@ -222,4 +235,37 @@ public void showAccountChooserDialog(CharSequence dialogTitle, boolean showActiv
...
     }
+
+    @Override
+    public void onRequestPermissionsResult(int requestCode, @NonNull String[] permissions, @NonNull int[] grantResults) {
+        if (requestCode != REQUESTER_NONE && requesters.containsKey(requestCode)) {
+            requesters.remove(requestCode).onRequestPermissionsResult(permissions, grantResults);
+        }
+    }
+
+    public boolean requestPermissions(String[] permissions, int minimumBuildVersion, @Nullable PermissionRequester requester) {
```
charlag: I would rather accept lambda or returned `Single` here. Can be such a fluent API:
```kotlin
requestPermission(Manifest.permission.WRITE_EXTERNAL_STORAGE) { givenPermissions ->
   if (givenPermissions.isNotEmpty()) {
       downloadFile()
   |
}
```

app/src/main/java/com/keylesspalace/tusky/BaseActivity.java
```diff
@@ -222,4 +235,37 @@ public void showAccountChooserDialog(CharSequence dialogTitle, boolean showActiv
...
+        if (minimumBuildVersion == BUILD_VERSION_ANY || Build.VERSION.SDK_INT >= minimumBuildVersion) {
```
charlag: Do we even need this check? `checkSelfPermission` should just return true?.

app/src/main/java/com/keylesspalace/tusky/BaseActivity.java
Tak: This way, it can directly chain the callback from `ActivityCompat.requestPermissions`

Tak: ...but I did update it so that the callsite as lambdaable as java allows...

app/src/main/java/com/keylesspalace/tusky/BaseActivity.java
Tak: I don't know? I assumed the original check was there for a reason. :smiley:

app/src/main/java/com/keylesspalace/tusky/fragment/SFragment.java
```diff
@@ -342,4 +356,31 @@ private void showOpenAsDialog(String statusUrl, CharSequence dialogTitle) {
...
+            DownloadManager.Request request = new DownloadManager.Request(uri);
+            request.setDestinationInExternalPublicDir(Environment.DIRECTORY_DOWNLOADS, filename);
+            downloadManager.enqueue(request);
+        }
+    }
+
+    private void requestDownloadAllMedia(Status status) {
+        pendingDownloadStatus = status;
+        String[] permissions = new String[]{ Manifest.permission.WRITE_EXTERNAL_STORAGE };
+        ((BaseActivity)getActivity()).requestPermissions(permissions, Build.VERSION_CODES.M, (permissions1, grantResults) -> {
```
charlag: if you don't use lambda arg, I would rename it to `__` but that's not important obv

app/src/main/java/com/keylesspalace/tusky/BaseActivity.java
charlag: Yeah, maybe leave it there

app/src/main/java/com/keylesspalace/tusky/fragment/SFragment.java
Tak: Aha, so the interface being in kotlin is why I couldn't lambdify the kotlin call

app/src/main/java/com/keylesspalace/tusky/ViewMediaActivity.kt
```diff
@@ -188,36 +186,26 @@ class ViewMediaActivity : BaseActivity(), ViewImageFragment.PhotoActionsListener
...
-                    downloadMedia()
-                } else {
-                    showErrorDialog(toolbar, R.string.error_media_download_permission, R.string.action_retry) { downloadMedia() }
-                }
-            }
-        }
+    private fun downloadMedia() {
+        val url = attachments!![viewPager.currentItem].attachment.url
+        val filename = Uri.parse(url).lastPathSegment
+        val toastText = String.format(resources.getString(R.string.download_image), filename)
```
charlag: why not `getString()` with argument?

charlag:
I think it's ready to merge, with or without that last change
@connyduck

connyduck:
thx!

Issue #1102, opened 2019-03-04
Add support for polls


connyduck:
https://github.com/tootsuite/mastodon/pull/10111

I have not yet looked at it and I really hope its not a breaking change that will crash Tusky. 
But we definitely need to add support for displaying polls and voting very soon. 
Creating polls can wait some time, but we of course we will also add that feature.

connyduck:
I will do this in three parts:
1) polls in timelines & voting
2) polls in notifications
3) creating polls

1) is about 60% finished

rinpatch:
Would be nice if poll creation used `poll_limits` object from `/api/v1/instance` used by glitch-soc and (soon) pleroma instead of depending on hardcoded values

PR #1123, opened 2019-03-13
Fix crash with unknown notification type

kyori19:
Fix part of #1102 .

If there's "Your poll has ended." notification in user's notification list, app crashes.
This is a fatal problem because once app started to crash, we can't even switch account.
So I fixed not to crash app for now.

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/adapter/NotificationsAdapter.java
+++ b/app/src/main/java/com/keylesspalace/tusky/adapter/NotificationsAdapter.java
@@ -136,6 +136,7 @@ public class NotificationsAdapter extends RecyclerView.Adapter {
                     (NotificationViewData.Concrete) notification;
             Notification.Type type = concreteNotificaton.getType();
             switch (type) {
+                default:
                 case MENTION: {
                     StatusViewHolder holder = (StatusViewHolder) viewHolder;
                     StatusViewData.Concrete status = concreteNotificaton.getStatusViewData();
--- a/app/src/main/java/com/keylesspalace/tusky/viewdata/NotificationViewData.java
+++ b/app/src/main/java/com/keylesspalace/tusky/viewdata/NotificationViewData.java
@@ -59,7 +59,7 @@ public abstract class NotificationViewData {
         }
 
         public Notification.Type getType() {
-            return type;
+            return type == null ? Notification.Type.UNKNOWN : type;
         }
 
         public String getId() {
--- a/app/src/main/java/com/keylesspalace/tusky/viewdata/NotificationViewData.java
+++ b/app/src/main/java/com/keylesspalace/tusky/viewdata/NotificationViewData.java
@@ -59,7 +59,7 @@ public abstract class NotificationViewData {
         }
 
         public Notification.Type getType() {
-            return type == null ? Notification.Type.UNKNOWN : type;
+            return type;
         }
 
         public String getId() {
--- a/app/src/main/java/com/keylesspalace/tusky/entity/Notification.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/entity/Notification.kt
@@ -16,7 +16,6 @@
 package com.keylesspalace.tusky.entity
 
 import com.google.gson.annotations.JsonAdapter
-import com.google.gson.annotations.SerializedName
 import com.keylesspalace.tusky.json.NotificationTypeAdapter
 
 data class Notification(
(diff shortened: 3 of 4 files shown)


app/src/main/java/com/keylesspalace/tusky/adapter/NotificationsAdapter.java
```diff
@@ -136,6 +136,7 @@ public void onBindViewHolder(@NonNull RecyclerView.ViewHolder viewHolder, int po
                     (NotificationViewData.Concrete) notification;
             Notification.Type type = concreteNotificaton.getType();
             switch (type) {
+                default:
```
charlag: Huh? I don't get it, default is a mention?

app/src/main/java/com/keylesspalace/tusky/viewdata/NotificationViewData.java
```diff
@@ -59,7 +59,7 @@ public Concrete(Notification.Type type, String id, Account account,
         }
 
         public Notification.Type getType() {
-            return type;
+            return type == null ? Notification.Type.UNKNOWN : type;
```
charlag: New notification does not have a type at all? Why it's `null`?

charlag:
I'm still grumpy about Mastodon just breaking clients. If we're sure that it will stay the same in stable (very likely) I propose to merge this before the stable release.

app/src/main/java/com/keylesspalace/tusky/adapter/NotificationsAdapter.java
kyori19: When type is null, switch statement returns NullPointerException.
This seems to be a Java's specification.

app/src/main/java/com/keylesspalace/tusky/viewdata/NotificationViewData.java
kyori19: https://github.com/tuskyapp/Tusky/blob/08c1bbd25340269fdb173bb61ca4416a9d931eef/app/src/main/java/com/keylesspalace/tusky/entity/Notification.kt#L26

New notification type `poll` is not defined in Notification.Type so type will be null.

app/src/main/java/com/keylesspalace/tusky/adapter/NotificationsAdapter.java
charlag: Indeed it is and indeed it does throw. I just don't understand how "default" changes that:
https://ideone.com/I82Z5B

app/src/main/java/com/keylesspalace/tusky/viewdata/NotificationViewData.java
charlag: Ah, I see. Maybe more correct would be to alter parser code so that it returns `UNKNOWN` if it doesn't match any.
Right now it makes a really confusing API in the data layer (we can have both `null` and `UNKNOWN`).
We could mark type field as `@JsonAdapter` and write our adapter (I can help with that). It could take 10 more minutes but it will be future-proof IMO.

connyduck: I think its even possible without a `JsonAdapter`, have a look how we handle `Visibility` in `Status.kt`

kyori19: When I passed instance original visibility `unleakable`, app crashed with same error.
So I think we need `JsonAdapter`...

```
2019-03-15 16:50:24.068 2655-2655/com.keylesspalace.tusky E/AndroidRuntime: FATAL EXCEPTION: main
    Process: com.keylesspalace.tusky, PID: 2655
    java.lang.NullPointerException: Attempt to invoke virtual method 'int com.keylesspalace.tusky.entity.Status$Visibility.ordinal()' on a null object reference
```

kyori19: Added `NotificationTypeAdapter` and anotated `Notification.Type` by `JsonAdapter`.
Please check whether it works correctly.

charlag:
Looks fine. We don't need `@SerializedName` now I guess.

kyori19:
@charlag It's true. Removed `@SerializedName` from `Notifiaction.Type`.

connyduck:
Poll notifications now look like I mentioned myself. But better than crashing I guess. And it hopefully wont stay this way for long.

PR #1127, opened 2019-03-16
ListsActivity: create IconicsDrawable with right color instead of tinting it later

connyduck:
Thats what I meant, sorry for not explaining it better @charlag 
It should be slightly more efficient


connyduck:
but it probably does not matter 🙃

charlag:
Idk, seems fine, let's ship it soon!

PR #1129, opened 2019-03-16
fix ProGuard issue with NotificationTypeAdapter

connyduck:
Proguard strips the class away so release build are crashing

There are at least three ways to fix this:

- register the adapter with the Gson instance like we do with `SpannedTypeAdapter`
- add an explicit keep rule for ProGuard
- move the adapter to the `Notification.kt` file so its kept by the rule that keeps the `Notification` class like we do with `MediaTypeDeserializer`

No idea whats better, I went with the last option.

cc @kyori19


charlag:
Damn, ProGuard. We mention the class explicitly. It could as well stay there :roll_eyes: 
I think this option is okay, we mention some parsing related things there anyway.

kyori19:
I didn't notice it... Sorry.
I think last option is better.

PR #1130, opened 2019-03-17
Update Notification Adapter to support status created timestamp refresh

pandasoft0:
Issue #452 
Refresh items timestamp at the Notification fragment with 1 minute interval.
Implemented the same logic as at the pull request #1113

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/adapter/NotificationsAdapter.java
+++ b/app/src/main/java/com/keylesspalace/tusky/adapter/NotificationsAdapter.java
@@ -124,12 +124,24 @@ public class NotificationsAdapter extends RecyclerView.Adapter {
 
     @Override
     public void onBindViewHolder(@NonNull RecyclerView.ViewHolder viewHolder, int position) {
+        bindViewHolder(viewHolder,position,null);
+    }
+
+    @Override
+    public void onBindViewHolder(@NonNull RecyclerView.ViewHolder viewHolder, int position, @NonNull List payloads) {
+        bindViewHolder(viewHolder,position,payloads);
+    }
+
+    private void bindViewHolder(@NonNull RecyclerView.ViewHolder viewHolder, int position, @Nullable List payloads){
+        Object payloadForHolder = payloads!=null&&!payloads.isEmpty()?payloads.get(0):null;
         if (position < this.dataSource.getItemCount()) {
             NotificationViewData notification = dataSource.getItemAt(position);
             if (notification instanceof NotificationViewData.Placeholder) {
-                NotificationViewData.Placeholder placeholder = ((NotificationViewData.Placeholder) notification);
-                PlaceholderViewHolder holder = (PlaceholderViewHolder) viewHolder;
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/keylesspalace/tusky/adapter/NotificationsAdapter.java
```diff
@@ -124,12 +124,24 @@ public NotificationsAdapter(AdapterDataSource<NotificationViewData> dataSource,
...
+
+    private void bindViewHolder(@NonNull RecyclerView.ViewHolder viewHolder, int position, @Nullable List payloads){
+        Object payloadForHolder = payloads!=null&&!payloads.isEmpty()?payloads.get(0):null;
         if (position < this.dataSource.getItemCount()) {
             NotificationViewData notification = dataSource.getItemAt(position);
             if (notification instanceof NotificationViewData.Placeholder) {
-                NotificationViewData.Placeholder placeholder = ((NotificationViewData.Placeholder) notification);
-                PlaceholderViewHolder holder = (PlaceholderViewHolder) viewHolder;
-                holder.setup(statusListener, placeholder.isLoading());
+                if (payloadForHolder==null) {
```
charlag: Why is it important to check payload for placeholder?

pandasoft0: @charlag If payloadForHolder==null - need to update a whole ViewHolder. If payloadsForHolder!=null and it is a List - enum all items of this list and update only required fields. It may contain only a single field for update now.
`
public static class Key{
        public static final String KEY_CREATED = "created";
}
`

But may be easy extended at the future.

kyori19:
I found that wrong small notification icon shows with boost notification after this PR merged.( I used `git bisect` to identify this commit)
I'm looking for a problem but I haven't found it yet.

cc: @pandasoft0

pandasoft0:
@kyori19 Could you please provide a screenshot? Please, mark the wrong icon at the screenshot.

kyori19:
@pandasoft0 Correct icon is mastodon default icon(First one). Tusky shows wrong icon(Next one).
 
Steps to reproduce:
1. Run app with notification tab.
2. Receive boost or favorite notification.
3. Refresh notification tab.
4. Wrong icon appears.

connyduck:
Can reproduce the problem. Also pretty sure it is caused by this PR because reverting it fixes the problem.

pandasoft0:
@kyori19 @connyduck Could you please say me on what site you tested? I tried to reproduce the issue on 2 different sites and tested on 2 different devices but i never saw the issue. Maybe the reason is the both sites have a default icon for empty avatars and app never show the default app icon via setImageResource but loads the icon with Picasso?

connyduck:
I don't think the instance matters. I tested this with three accounts. Send Account A about 10 likes from Account B. Open the notification column of A in Tusky. Send one like from Account C to account A, refresh the notification column -> the new notification from C shows the profile picture of B.

pandasoft0:
Thanks, i will check it today once more

PR #1138, opened 2019-03-20
Added "Warning" icon instead of "CW" text.

MasterGroosha:
> **Disclaimer**: this needs to be tested, since I haven't touched Android development (with all its things) for years.

In short: I don't like "CW" text in toot compose screen, since it _has_ to be translated and there's no space for 3+ characters. I went to https://ionicons.com, grabbed "Warning" icon (AFAIK they are under MIT Licence, so I can do that) and placed instead of "CW" text.

Screenshot:
image

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/ComposeActivity.java
+++ b/app/src/main/java/com/keylesspalace/tusky/ComposeActivity.java
@@ -29,6 +29,7 @@ import android.content.res.Resources;
 import android.graphics.Bitmap;
 import android.graphics.PorterDuff;
 import android.graphics.drawable.Drawable;
+import android.media.Image;
 import android.net.Uri;
 import android.os.AsyncTask;
 import android.os.Build;
@@ -202,7 +203,7 @@ public final class ComposeActivity
     private TootButton tootButton;
     private ImageButton pickButton;
     private ImageButton visibilityButton;
-    private Button contentWarningButton;
+    private ImageButton contentWarningButton;
     private ImageButton emojiButton;
     private ImageButton hideMediaToggle;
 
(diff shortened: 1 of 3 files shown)


MasterGroosha:
I guess I should also edit copyrights or something like that. It would be nice if someone points me what exactly should I do. I'm not good in such things :(

charlag:
I have a mixed feeling about it. How does Mastodon Web UI does that?

MasterGroosha:
@charlag Mastodon Web UI shows "CW"

charlag:
Pinafore and Hyperspace use the same icon so maybe it's fine?

MasterGroosha:
Should I close the PR then?

charlag:
Opposite, I meant that Pinafore and Hyperspace use the same icon you proposed so I'm more for it (even though I think that the icon is kinda ugly)

connyduck:
I'm not sure. I prefer the old CW, but you are right its bad that its not translateable 🤔

jalcine:
All good to me; that's a universal 'warning' symbol

Tak:
Makes sense to me

connyduck:
What about this icon? https://material.io/tools/icons/?icon=error&style=baseline (proposed by @MasterGroosha on Mastodon)
It fits better with the others

charlag:
While it is nice becuse it's rounded it has semantics of an error. "Post an error!". I would even prefer triangle to that.
Would be even better if we could go from warning to "spoiler" semantics somehow.

charlag:
"mode_comment"? "short_text"? "arrow_drop_down"? "unfold_more"? "priority_high", "sms_failed", "new_releases"?

connyduck:
sms_failed seems really nice 🤔 https://material.io/tools/icons/?icon=sms_failed&style=baseline

MasterGroosha:
Updated the icon to "sms_failed". This is how it looks now with both dark and light theme:

изображение
изображение

PR #1142, opened 2019-03-21
[WIP] Start handling expired authentication.

charlag:
Based on
https://pleroma.site/notice/9gz5HsnovdYyoLeXmC
and
https://pleroma.site/notice/9gzKeoEXa4pic5XcQK

The plan is to help Mastodon migrate to normal authentication model where tokens actually expire. We still need to do something in case of invalid login because currently we don't handle it in any way other than not doing anything.

So I started with recording `refresh_token`, `expires_in` and saving `clientId` and `clientSecret` to the DB.

There's no rush for this but I think it falls together quite nicely.


a1batross:
That would be nice for OAuth compliance but in fact, Mastodon still doesn't support refreshing tolen and Pleroma supports refreshing but it's useless as tokens never expire. And there is no plans to make them expirable as far as I know.

connyduck:
Pixelfed definitely expires tokens, I think after 2 weeks. There are quite some users using Tusky+Pixelfed, so token refreshing would be very nice to have.

a1batross:
Ah, I don't know about Pixelfed. Sure indeed.

PR #1146, opened 2019-03-23
Add ability to scroll to top by tab click at the Account activity

pandasoft0:
Issue: tuskyapp#1078

TimellineFragment and NotificationFragment did a findViewById with a tab_layout id. But the Account activity uses a tab layout with another id.

I've added an interface and add this interface implementation to the Main and Account activities. Allows do not depend on tab layout id.


app/src/main/java/com/keylesspalace/tusky/fragment/TimelineFragment.java
```diff
@@ -432,24 +432,26 @@ private void showNothing() {
...
 
-        if (jumpToTopAllowed()) {
-            TabLayout layout = requireActivity().findViewById(R.id.tab_layout);
-            if (layout != null) {
-                onTabSelectedListener = new TabLayout.OnTabSelectedListener() {
-                    @Override
-                    public void onTabSelected(TabLayout.Tab tab) {
-                    }
+        if (requireActivity() instanceof TabbedActivity) {
+            if (jumpToTopAllowed()) {
```
connyduck: `jumpToTopAllowed()` should no longer be needed - if the Activity is a `TabbedActivity` we can always jump to the top. Also it is outdated - hashtag tabs can now be included in the main tabs.

pandasoft0: 1. Remove jumpToTopAllowed;
2. Move add/remove tab listener to the setMenuVisibility method;
3. Added a listener at the conversation fragment.

pandasoft0:
@connyduck I updated the code, but i don't know how to re request your review

app/src/main/java/com/keylesspalace/tusky/components/conversation/ConversationsFragment.kt
```diff
@@ -172,6 +179,53 @@ class ConversationsFragment : SFragment(), StatusActionListener, Injectable {
...
+        }
+    }
+
+    private fun addTabListener() {
+        val activity = activity
+        if (activity is TabbedActivity) {
+            // MainActivity's layout is guaranteed to be inflated until onCreate returns.
+            val layout = (activity as TabbedActivity).getTabLayout()
+            if (layout != null) {
+                onTabSelectedListener = object : TabLayout.OnTabSelectedListener {
```
connyduck: What about reusing the `OnTabSelectedListener` instead of creating a new instance everytime?

app/src/main/java/com/keylesspalace/tusky/components/conversation/ConversationsFragment.kt
```diff
@@ -172,6 +179,53 @@ class ConversationsFragment : SFragment(), StatusActionListener, Injectable {
...
     }
 
+    private fun jumpToTop() {
+        if (isAdded) {
+            layoutManager?.scrollToPosition(0)
+            recyclerView.stopScroll()
+        }
+    }
+
+    override fun setMenuVisibility(menuVisible: Boolean) {
```
connyduck: I don't think this method is meant for this usecase. What about letting each fragment know its position so it cn check itself when it should jump to the top?

pandasoft0:
About setMenuVisibility. I use this method in a several projects and it works as expected. Also, i am not sure that the fragment must know about it position at the ViewPager. To be honesty, i don't like the current implementation when fragment add a listener to the view from the parent activity. I would prefer the solution when fragment implement the interface:
```
enum class ScrollDirection{
  Top
}
interface ScrollableFragment{
  fun scrollTo(direction:ScrollDirection=ScrollDirection.Top)
}
```
At the ViewPager adapter saves fragments at `instantiateItem` to the map (and remove them at the `destroyItem`)

Add a tab click listener at the main activity. On tab reselect: get the current fragment from the adapter, checks is fragment implements the `ScrollableFragment` and call the `scrollTo` if so.
What do you think about it?

pandasoft0:
@connyduck Could you please answer on my questions?

connyduck:
> Add a tab click listener at the main activity. On tab reselect: get the current fragment from the adapter, checks is fragment implements the ScrollableFragment and call the scrollTo if so.
What do you think about it?

Seems like a good idea. There already is a `OnTabSelectedListener` in MainActivity you can extend.

pandasoft0:
I've update the code

PR #1147, opened 2019-03-24
add shortcut to ComposeActivity

### LINKED ISSUE #1031: Make a shortcut to compose a new toot
It would be great if there was a long-press shortcut to go straight to the Compose Toot activity, so you could skip the timeline.

connyduck:
now with pretty icons

closes #1031


app/src/blue/res/xml-v25/shortcuts.xml
```diff
@@ -0,0 +1,20 @@
...
+    <shortcut
+        android:icon="@mipmap/ic_shortcut_compose"
+        android:shortcutId="com.keylesspalace.tusky.Compose"
+        android:shortcutLongLabel="@string/compose_shortcut_long_label"
+        android:shortcutShortLabel="@string/compose_shortcut_short_label">
+
+        <intent
+            android:action="android.intent.action.VIEW"
+            android:targetClass="com.keylesspalace.tusky.SplashActivity"
+            android:targetPackage="com.keylesspalace.tusky"/>
```
connyduck: super annoying: Had to duplicate the whole file because its not possible to use `@string` or a placeholder for `android:targetPackage` 🙄

app/src/green/res/xml-v25/shortcuts.xml
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0" encoding="utf-8"?>
+<shortcuts xmlns:android="http://schemas.android.com/apk/res/android">
+    <shortcut
+        android:icon="@mipmap/ic_shortcut_compose"
+        android:shortcutId="com.keylesspalace.tusky.Compose"
+        android:shortcutLongLabel="@string/compose_shortcut_long_label"
+        android:shortcutShortLabel="@string/compose_shortcut_short_label">
+
+        <intent
```
charlag: I don't understand why there are two `Intent`s here, I'm really not familiar with shortcuts.

connyduck: They open both so you see the splash screen for a short time and when you close the compose screen you are back at the main screen. Not sure if thats the best behavior, maybe we should just open `ComposeActivity` only 🤔

connyduck:
I tried removing the `SplashScreen` intent, but then the screen turns white before the `ComposeActivity` is shown and thats weird

PR #1150, opened 2019-03-25
fix crash when filter request returns an error

connyduck:
happens, for example, when the instance is on a mastodon version that does not support filters yet

cc @Tak


Tak:
Looks good to me

charlag:
I've basically fixed that in #1142 but there's some problem on Pleroma side so I'm not sure if it's finished.
I guess that's better than nothing for now.

charlag:
(i think I did lol, I guess it's fine, too hard with words today, sorry!)

PR #1157, opened 2019-03-26
remove CustomTabsHelper & ResourcesUtils

connyduck:
They are just not needed xD
Also sneaked in a `setShowTitle(true)` which should make the Chrome custom tabs more comfortable.


app/src/main/java/com/keylesspalace/tusky/util/ResourcesUtils.java
```diff
@@ -1,29 +0,0 @@
...
-package com.keylesspalace.tusky.util;
-
-import android.content.Context;
-import androidx.annotation.AnyRes;
-
-/**
- * Created by remi on 1/14/18.
- */
-
-public class ResourcesUtils {
```
charlag: OMG what's that, never seen this

connyduck:
We were both checking if the user has an app installed that can handle the intent and catching the `ActivityNotFoundException`. Only the latter is needed.

charlag:
Oh... okay, I would think that trying to resolve activity should be enough but it required some nontrivial checks for some reason, it's really easy to catch
LGTM

PR #1158, opened 2019-03-26
Update Deps + Cleanup

bernd289:
Update Deps
Update Grandle Wrapper

Fix some warnings/compile errors


app/src/main/java/com/keylesspalace/tusky/fragment/preference/AccountPreferencesFragment.kt
```diff
@@ -57,19 +58,19 @@ class AccountPreferencesFragment : PreferenceFragmentCompat(),
     @Inject
     lateinit var eventHub: EventHub
 
-    private lateinit var notificationPreference: Preference
-    private lateinit var tabPreference: Preference
-    private lateinit var mutedUsersPreference: Preference
-    private lateinit var blockedUsersPreference: Preference
+    private var notificationPreference: Preference? = null
```
charlag: I wouldn't do that tbh. They should be there, we shouldn't check for them in general.

app/src/main/java/com/keylesspalace/tusky/fragment/preference/PreferencesFragment.kt
```diff
@@ -33,17 +34,17 @@ class PreferencesFragment : PreferenceFragmentCompat() {
 
         addPreferencesFromResource(R.xml.preferences)
 
-        val themePreference: Preference = findPreference("appTheme")
-        themePreference.icon = IconicsDrawable(themePreference.context, GoogleMaterial.Icon.gmd_palette).sizePx(iconSize).color(ThemeUtils.getColor(themePreference.context, R.attr.toolbar_icon_tint))
+        val themePreference: Preference? = findPreference("appTheme")
+        themePreference?.icon = IconicsDrawable(themePreference?.context, GoogleMaterial.Icon.gmd_palette).sizePx(iconSize).color(ThemeUtils.getColor(themePreference?.context as Context, R.attr.toolbar_icon_tint))
```
charlag: Again, if it's not there we silently ignore the error, not good

app/src/main/java/com/keylesspalace/tusky/fragment/preference/AccountPreferencesFragment.kt
connyduck: Yes, I agree with @charlag, it should crash when the preference is not found. Since it is used a lot and to keep the code clean I would add an extension function `requirePreference` that returns a non-null preference or throws a exception.

connyduck:
`requirePreference` is a perfect candidate for a extension function on `PreferenceFragmentCompat`: https://kotlinlang.org/docs/reference/extensions.html

charlag:
Seems fine now!

PR #1164, opened 2019-03-29
make a copy of file before upload on ComposeActivity => solve upload failed bug

lenchan139:
Thanks @charlag , I solved it by make a copy to cache before upload :)

~~Hi, I bought Samsung S10e currently and try to use Tusky on it. But there have a problem if I upload photos that taken by stock camera. To fix it, the simplest way is always to compress photo before uploading to mastodon instance.~~

### DIFF
--- a/app/src/main/res/layout/activity_view_media.xml
+++ b/app/src/main/res/layout/activity_view_media.xml
@@ -1,28 +1,30 @@
 <?xml version="1.0" encoding="utf-8"?>
 <FrameLayout xmlns:android="http://schemas.android.com/apk/res/android"
-    android:orientation="vertical"
+    xmlns:app="http://schemas.android.com/apk/res-auto"
     android:layout_width="match_parent"
     android:layout_height="match_parent"
-    android:background="@android:color/black">
+    android:background="@android:color/black"
+    android:orientation="vertical">
 
     <com.keylesspalace.tusky.view.ImageViewPager
+        android:id="@+id/viewPager"
         android:layout_width="match_parent"
-        android:layout_height="match_parent"
-        android:id="@+id/viewPager" />
+        android:layout_height="match_parent" />
 
     <androidx.appcompat.widget.Toolbar
         android:id="@+id/toolbar"
         android:layout_width="match_parent"
         android:layout_height="?attr/actionBarSize"
         android:background="@color/toolbar_view_media"
-        android:theme="@style/AppTheme.Account.AppBarLayout"/>
+        android:theme="@style/ViewMediaActivity.AppBarLayout"
+        app:titleTextColor="@color/white" />
 
     <ProgressBar
         android:id="@+id/progressBarShare"
         style="?android:attr/progressBarStyleLarge"
         android:layout_width="wrap_content"
         android:layout_height="wrap_content"
-        android:visibility="gone"
-        android:layout_gravity="center"/>
+        android:layout_gravity="center"
+        android:visibility="gone" />
 
 </FrameLayout>
--- a/app/src/main/res/values/styles.xml
+++ b/app/src/main/res/values/styles.xml
@@ -30,10 +30,8 @@
         <item name="status_text_large">22sp</item>
     </style>
 
-    <style name="AppTheme.Account.AppBarLayout" parent="ThemeOverlay.AppCompat">
-        <item name="android:textColorPrimary">@color/text_color_primary_dark</item>
-        <item name="android:textColorSecondary">@color/text_color_secondary_dark</item>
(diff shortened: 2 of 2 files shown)


charlag:
Hey
Can you please explain what is the problem with uploads? I think we would prefer to make solution which works automatically rather than adding an option

lenchan139:
Hi @charlag ,
Yeah, You are right. :(
###
If I upload a photo and it was taken by the Samsung stock camera, there have some photo will cause the error from OkHttp and it says:
`java.net.ProtocolException: expected 4021 bytes but received 4182`

But, if I upload a re-save version(no matter upload to telegram and download it, use Snapseed re-save it) with that photo and upload it, it will upload success.

charlag:
Hm, thanks.
Doe sit happen with every app which provides media?

lenchan139:
No.
I can upload same photo on another Mastodon client called Fedilab and It happens only Tusky. :(

charlag:
then we better find the error :thinking: 
we can try to copy it into our own storage first and then upload, not sure if it would help

lenchan139:
Oh, many thank @charlag !!!
It works now without error if I make a copy to cache and upload the cached copy.

charlag:
Hey, thank you, I encountered similar thing in other apps, we'll try to review it in reasonable time 😅

app/src/main/java/com/keylesspalace/tusky/ComposeActivity.java
```diff
@@ -1431,13 +1432,36 @@ protected void onActivityResult(int requestCode, int resultCode, Intent intent)
     }
 
 
-    private void pickMedia(Uri uri, long mediaSize, String description) {
+    private void pickMedia(Uri inUri, long mediaSize, String description) {
+        Uri uri = inUri;
+        ContentResolver contentResolver = getContentResolver();
+        String mimeType = contentResolver.getType(uri);
+        String surfix = MimeTypeMap.getSingleton().getExtensionFromMimeType(mimeType);
```
Tak: Why not use the filename suffix from the uri?

app/src/main/java/com/keylesspalace/tusky/ComposeActivity.java
```diff
@@ -1431,13 +1432,36 @@ protected void onActivityResult(int requestCode, int resultCode, Intent intent)
...
+        try {
+            InputStream tempInput;
+            FileOutputStream out;
```
Tak: When do these get closed?

app/src/main/java/com/keylesspalace/tusky/ComposeActivity.java
```diff
@@ -1431,13 +1432,36 @@ protected void onActivityResult(int requestCode, int resultCode, Intent intent)
...
+            }
+            uri = FileProvider.getUriForFile(this,
+                    BuildConfig.APPLICATION_ID+".fileprovider",
+                    file);
+            Log.v("checker","a");
+        }catch(IOException e){
+            e.printStackTrace();
+            Log.w(TAG, e);
+            uri = inUri;
+        }
```
Tak: Does `mediaSize` need to be updated?

charlag:
@lenchan139 hi! Do you plan to continue working on this?

lenchan139:
@charlag Hi, It is done now. :)

app/src/main/java/com/keylesspalace/tusky/ComposeActivity.java
```diff
@@ -1431,13 +1432,42 @@ protected void onActivityResult(int requestCode, int resultCode, Intent intent)
...
+            }
+            uri = FileProvider.getUriForFile(this,
+                    BuildConfig.APPLICATION_ID+".fileprovider",
+                    file);
+
+            mediaSize = getMediaSize(getContentResolver(), uri);
+            tempInput.close();
+            out.close();
+        }catch(IOException e){
+            e.printStackTrace();
```
charlag: I think we should try to close streams on error?

app/src/main/java/com/keylesspalace/tusky/ComposeActivity.java
```diff
@@ -1431,13 +1432,42 @@ protected void onActivityResult(int requestCode, int resultCode, Intent intent)
...
+            Log.w(TAG, e);
```
connyduck: `e.printStackTrace();` is deprecated, `Log.w(TAG, e);` alone should be fine.

connyduck:
there is a `closeQuietly` method in IOUtils you can use to make this way shorter

lenchan139:
repoen at #1294

PR #1166, opened 2019-03-29
Fix wrong notification avatar being displayed sometimes

connyduck:
bug report: https://github.com/tuskyapp/Tusky/pull/1130#issuecomment-477482909 thx @kyori19 

Im not sure it is actually caused by #1130, it probably just made the bug appear more often.

Removing the `fit()` fixes the problem. The weird thing was that only the notification icon was wrong - so it had to have something to do with Picasso. If it was a recycling issue more stuff would have been wrong.


charlag:
Something about reusing something, yes. I'm afraid that it will lead to some visual bugs but let's see, it's better than wrong ones (should update Picasso finally)

Issue #1082, opened 2019-02-26
Migrate from Picasso to Glide


connyduck:
So I noticed that some images are way more blurry than they should be and I tried to play around with Picasso to make them sharper to no avail (both Picasso 2.5.2 and 2.71828). I tried Glide and the result was much better. See the attached images for comparision.
My suggestion is we migrate our image loading to Glide because A) better quality and B) Glide also supports gifs (#961 #108)

my Avatar, loaded with Picasso:
![picasso](https://user-images.githubusercontent.com/10157047/53441388-0ab00280-3a07-11e9-96f7-b63b1bf4bef5.png)

my avatar, loaded with Glide:
![glide](https://user-images.githubusercontent.com/10157047/53441387-0ab00280-3a07-11e9-93b1-de337aa2c2fd.png)

charlag:
It may be a good idea, but let's first check some things, because Glide is... big.
Do we use custom transformation to make avatars rounded? If so, the problem is likely there
Do we use `fit()` or something like that? It can use downsampling with all the consequences.
I checked both and it seems like...we just tell to download it

```java
            Picasso.with(avatar.getContext())
                    .load(url)
                    .placeholder(R.drawable.avatar_default)
                    .into(avatar);
```

Maybe we *should* try calling `fit()` there. It can be Android resizing images in a weird way.
I checked and I don't experience it so only you can check it.

Glide makes sense for other reasons (you can save configurations and use them afaik) but I would check this first anyway.

connyduck:
Ok yes, its better with `fit()`, but not on a level with Glide
![picasso-fit](https://user-images.githubusercontent.com/10157047/53442686-28cb3200-3a0a-11e9-88bd-3f0ed240673d.png)

How can we add gif support if we stay with Picasso?

charlag:
Ouch, that's still looking bad and I have no idea why, really
I'm not sure about gifs. I think Glide implements custom Drawable for that (and so does Fresco but for other reasons) so we could use something similar.
Probably it's just easier to migrate to Glide. It's made by Google Photos team I think and it is actively developed.

charlag:
>Round Pictures: CircleImageView/CircularImageView/RoundedImageView are known to have issues with TransitionDrawable (.crossFade() with .thumbnail() or .placeholder()) and animated GIFs, use a BitmapTransformation (.circleCrop() will be available in v4) or .dontAnimate() to fix the issue.

connyduck:
#1175

PR #1169, opened 2019-03-30
[WIP] Update Picasso library

pandasoft0:
Update Picasso library to latest.
Should solve the issue #1130 with a small avatar

@connyduck Could you please check? Updated library+OkHttpDownloader should solve the issue with avatar.


pandasoft0:
I revert the fit() on small avatar back. It works without issue at my devices after update library.

connyduck:
I would prefer to switch to Glide tbh. See https://github.com/tuskyapp/Tusky/issues/1082

pandasoft0:
I suggest to merge these changes to the main branch because it fixes the issue with wrong images. I checked the issue. App pass the correct links to the Picasso library (old version), but it shows the wrong images. I didn't check the reason. New version always shows the correct images. Also, it has embedded OkHttp downloader instead of 3rd party library. Maybe it was a reason of the issues with wrong images.
I worked with Glide library too. I will replace Picasso with Glide at this or next week, but this PR maybe used as a temporary solution.

PR #1173, opened 2019-04-02
Add bot indicator overlay to avatars

Tak:
I saw this suggestion fly past on the federated timeline and already lost it. :grinning: 

bot-dark-background bot-light-background

The best "bot" icon I found is the android logo, but I guess that's not ok for trademark reasons...

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/PreferencesActivity.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/PreferencesActivity.kt
@@ -143,6 +143,9 @@ class PreferencesActivity : BaseActivity(), SharedPreferences.OnSharedPreference
             "absoluteTimeView" -> {
                 restartActivitiesOnExit = true
             }
+            "showBotOverlay" -> {
+                restartActivitiesOnExit = true
+            }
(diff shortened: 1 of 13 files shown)


connyduck:
Im not sure if I like this. Why show the bot status and not, for example, the locked status of an account?
(but thats the icon I would be using http://materialdesignicons.com/icon/robot )

Tak:
Oh, I didn't find that one - it's much better.

The benefits of marking bots (from my point of view) are:
- I know whether there's a point in trying to interact with the account at all
- I know whether it's a markov chain generator or if I just don't understand what the person is saying (ebooks bots)
- I know whether it's always posting promotional links or if there's a chance it will post other things too (especially helpful for languages I don't/barely understand)

We could also add an overlay for locked accounts, but you already have to go to the account view to follow people, so it doesn't seem too helpful...

connyduck:
Ok if you change the icon and add an option to turn this off I will merge it

connyduck:
Ok, can I merge this? (its still marked as WIP)

Tak:
Oops, yeah

Issue #968, opened 2018-12-27
Tusky freeze when opening picture if your network is not responding


a1batross:
I was in metro, when tried to open random picture from wxw.moe and using cellular network, where Android may report LTE is active, but in reality network doesn't work.

Tusky was freezed in home activity, so it even don't reported that image is loading currently and image viewer activity don't created yet.

However, Android doesn't recognized this as ANR.

* * * *
- Tusky Version: 4.1
- Android Version: 8.1
- Android Device: Xiaomi Redmi Note 4
- Mastodon instance (if applicable): niu.moe(local), wxw.moe

- Yes, I searched or browsed the repo’s other issues to ensure this is not a duplicate.

Dostoi:
I'm having an issue similar to this one. I noticed the tusky wait for all available thumbnail to be loaded before opening and loading the picture you're trying to load.
A good way to test it is to use orbot on a low network and go to someone media page and try to open a picture before all thumbnail are loaded.

averageunicorn:
An addition: As I experienced it, Tusky doesn't freeze completely, but you have to press back before you can do anything again.

charlag:
Happens to me on a slow network when I 
1) try to open it
2) Don't wait until it loads and press back
3) Try to open it again. Visually nothing changes but it's unresponsive until you press back

My theory is that it's happening because our image loading library (Picasso) says that it has a cached version so we try to play an animation after loading because it should be super fast but it actually isn't. Maybe it will go away when we update Picasso.

Another option is that it's some state/timing issue.

connyduck:
fixed by #1175

PR #1175, opened 2019-04-03
Migrate to Glide

pandasoft0:
Replace Picasso library with Glide library. Issue: #1082


pandasoft0:
Updated the code to use a Glide instead of Picasso.
Not sure about:
https://github.com/roma-apps/roma-android/blob/upstream/1082/migrate_to_glide/app/src/main/java/com/keylesspalace/tusky/util/CustomEmojiHelper.java
https://github.com/roma-apps/roma-android/blob/upstream/1082/migrate_to_glide/app/src/main/java/com/keylesspalace/tusky/fragment/ViewImageFragment.kt#L184-L212
https://github.com/roma-apps/roma-android/blob/upstream/1082/migrate_to_glide/app/src/main/java/com/keylesspalace/tusky/view/MediaPreviewImageView.kt#L101-L108

app/src/main/java/com/keylesspalace/tusky/util/CustomEmojiHelper.java
```diff
@@ -118,20 +123,19 @@ public void draw(@NonNull Canvas canvas, CharSequence text, int start, int end,
...
-            if(view != null) {
-                imageDrawable = new BitmapDrawable(view.getContext().getResources(), bitmap);
-                view.invalidate();
-            }
-        }
-
-        @Override
-        public void onBitmapFailed(Drawable errorDrawable) {}
+        Target<Bitmap> getTarget(){
+            return new SimpleTarget<Bitmap>() {
```
connyduck: `SimpleTarget` is deprecated. I think we need `CustomTarget` here

app/src/main/java/com/keylesspalace/tusky/adapter/StatusBaseViewHolder.java
```diff
@@ -323,10 +325,10 @@ protected void setMediaPreviews(final List<Attachment> attachments, boolean sens
             mediaPreviews[i].setVisibility(View.VISIBLE);
 
             if (TextUtils.isEmpty(previewUrl)) {
-                Picasso.with(context)
+                Glide.with(mediaPreviews[i])
                         .load(mediaPreviewUnloadedId)
-                        .resize(maxW, maxH)
-                        .onlyScaleDown()
+                        .override(maxW,maxH)
+                        .downsample(DownsampleStrategy.CENTER_INSIDE)
```
connyduck: This downsampling was used to prevent a crash when loading huge images. It looks like Glide is already resizing to the size of the ImageView, no need to help it

app/src/main/java/com/keylesspalace/tusky/view/MediaPreviewImageView.kt
```diff
@@ -94,18 +98,16 @@ defStyleAttr: Int = 0
...
-    override fun onSuccess() {
-        onSizeChanged(width, height, width, height)
+    override fun onLoadFailed(e: GlideException?, model: Any?, target: Target<Drawable>?, isFirstResource: Boolean): Boolean {
+        return false
     }
 
-    // We do not handle the error here, instead it will be handled higher up the call chain.
-    override fun onError() {
+    override fun onResourceReady(resource: Drawable?, model: Any?, target: Target<Drawable>?, dataSource: DataSource?, isFirstResource: Boolean): Boolean {
+        onSizeChanged(width, height, width, height)
```
connyduck: looks fine

app/src/main/java/com/keylesspalace/tusky/adapter/StatusBaseViewHolder.java
pandasoft0: I've removed downsampling

pandasoft0:
@connyduck 
1. I fixed the crash
2. But i see an ugly side effect and i have no ides how to solve it.
How to reproduce:
1. Upload a large image (3-4Mb)
2. Click on the image at the TimelineFragment while it doesn't loaded for preview yet.
a) at the Picasso version of the app it stucks until image load and starts the transition after
b) i updated this behaviour at the Glide version. It starts the transition immediately but show the progress bar until image load at the image preview screen. When you click "back" it start the exit transition, but shows the black placeholder at the timeline fragment. I think, Glide has not cached preview for this size and loads the image again, but i am not sure. Do you have any suggestions how to remove this side effect? I think, custom cache signature may help (https://bumptech.github.io/glide/doc/caching.html#cache-invalidation) What do you think?

connyduck:
The code in `ViewImageFragment` is such a mess. 
Here is how I refactored it and it seems to work and fixes #968:

- delete the whole `if (arguments!!.getBoolean(ViewMediaFragment.ARG_START_POSTPONED_TRANSITION)) {` block including the `else`.
- call `loadImageFromNetwork` instead 
- remove `.skipMemoryCache(true)` and  `.diskCacheStrategy(DiskCacheStrategy.NONE)` (**just let Glide decide whether to load the image from network or cache** - this seems to be the main problem here)
- remove `Handler().post {` (why is this even here)
- inline the `loadImageFromNetwork`
- do `photoActionsListener.onBringUp()` after the Glide call and remove it from `finishLoadingSuccessfully` (which could also be inlined)

- probably the whole `ARG_START_POSTPONED_TRANSITION` can also be removed from `ViewVideoFragment` but that requires some more investigation.

pandasoft0:
> remove Handler().post { (why is this even here)
I have added this line to fix the crash. Glide crashes when app starts the new image load request to the same target from another callback.
I will test the app at my devices in few hours

connyduck:
Ahh yes makes sense. So it should be no longer needed.

charlag:
Please don't do Single.create(). fromCallaback should be used instead.
Original idea with disk loading was to know if transition should be instant or not. I cannot look at the code properly as I still lack proper connection.

connyduck:
Yes this was the idea, but it does not work as #968 shows. So the better option imho is to just transition always and show the ProgressBar if the image is not yet loaded.

pandasoft0:
> Please don't do Single.create(). fromCallaback should be used instead.

Done

pandasoft0:
> Here is how I refactored

Did you commit your changes or it's just your thoughts? If thoughts - i will update the code today

connyduck:
I tried it out but I discarded the changes afterwards. Yes, please update the code and resolve the merge conflicts. I will then test it again to see if its ready to merge.

app/src/main/java/com/keylesspalace/tusky/fragment/ViewImageFragment.kt
```diff
@@ -74,53 +85,16 @@ class ViewImageFragment : ViewMediaFragment() {
...
-                            }
-                        }
-                    })
-        } else {
-            // if we're not initial page, don't bother.
-            loadImageFromNetwork(url, photoView)
-        }
+        loadImageFromNetwork(url, photoView)
 
+        //Add a delay to prevent screen blink on load image from the cache
```
pandasoft0: @connyduck I was need to do this trick to prevent blinking. See video below

pandasoft0:
Without a trick
https://www.dropbox.com/s/rim21jxczqtbt5h/without_trick.mp4?dl=0

With a trick
https://www.dropbox.com/s/llzpak0jq8ud2y1/with_trick.mp4?dl=0

pandasoft0:
My thoughts.
When image cached at the memory - Glide fast enough to load it to the image view before transition, but if the image is cached at the local storage - app starts the transition before image load, it doesn't see the image and show activity without transition.

connyduck:
I think its not a good idea to wait with showing the Activity. This feels like lag. I rather have the activity open immediately than the animation.
Can we do something like this:
Check if the large image is in the memory cache, if yes, show it. If not, check if the preview image is in the memory cache, if yes, show it (maybe below the ProgressBar) and load the large image in the background. If the preview is not in memory cache, show the ProgressBar and load the large image in the background.

pandasoft0:
I am not sure, we can check is image cached or not. I see a single way only - as it was done before - load image with cache=memory only and load from Network on fail. We can postpone transition and load image from any cache. If fail - show progress bar and load from internet...
Maybe, we can use custom cache implementation (I didn't check the ability) but it looks too complicated for this task.

connyduck:
My first preference is still just one single Glide call (basically everything you have now without the `Completable.timer`). Its simple and I think having no transition when loading from disk cache is ok for now.

> I am not sure, we can check is image cached or not. I see a single way only - as it was done before - load image with cache=memory only and load from Network on fail.

Yes you check if the image is cached by trying to load it from cache only. If you want to have an animation when loading from disk cache you have to load from any cache and not memory only. If you get everything to work this way its great, but we can also add it later.

pandasoft0:
Please, check this new solution. It uses the 2 Glide requests but in the "Glide" documented way:
https://bumptech.github.io/glide/doc/options.html#starting-a-new-request-on-failure

https://github.com/roma-apps/roma-android/blob/ecc6169be793355833aecd4a7510f9608763dd1a/app/src/main/java/com/keylesspalace/tusky/fragment/ViewImageFragment.kt#L135-L149

PR #1176, opened 2019-04-04
Update Deps

bernd289:


app/build.gradle
```diff
@@ -122,26 +122,26 @@ dependencies {
     implementation 'androidx.room:room-runtime:2.0.0'
     kapt 'androidx.room:room-compiler:2.0.0'
     implementation 'androidx.room:room-rxjava2:2.0.0'
-    implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$kotlin_version"
+    implementation "org.jetbrains.kotlin:kotlin-stdlib:$kotlin_version"
```
connyduck: I think `kotlin-stdlib-jdk7` is the right dependency

bernd289: oops. yes. fixed :)

PR #1186, opened 2019-04-10
confirm delete notifications

bernd289:


app/src/main/java/com/keylesspalace/tusky/fragment/NotificationsFragment.java
```diff
@@ -252,6 +254,25 @@ public View onCreateView(@NonNull LayoutInflater inflater, @Nullable ViewGroup c
...
+        dialog.setPositiveButton(R.string.yes, new DialogInterface.OnClickListener() {
+            public void onClick(DialogInterface dialog, int which) {
+                clearNotifications();
+            }
+        });
+
+        dialog.setNeutralButton(R.string.no, new DialogInterface.OnClickListener() {
+            public void onClick(DialogInterface dialog, int which) {
+                //Do nothing.
+            }
```
connyduck: if the callback does nothing, you can remove it and use `null` instead

app/src/main/res/values/strings.xml
```diff
@@ -472,4 +472,8 @@
     <string name="compose_shortcut_short_label">Compose</string>
     <string name="pref_title_bot_overlay">Show indicator for bots</string>
 
+    <string name="notification_clear_text">Are you sure you want to permanently clear all your notifications?</string>
+    <string name="yes">Yes</string>
+    <string name="no">No</string>
```
connyduck: there is `android.R.string.yes` and `android.R.string.no`, no need to define it ourselves

bernd289:
@connyduck changes applied :)

Issue #1155, opened 2019-03-26
Accessibility: Give an inserted photo a better contentDescription


MarcoZehe:
When composing a toot with an image, like when sharing from Google Photos or inserting via the Add Image facility, the inserted image gets a contentDescription of "Delete", at least in German. This is confusing because what actually happens when you double-tap this with TalkBack is that it opens a popup where you can either add a description or delete the image. So calling this simply "Delete" is inappropriate.

Suggestion: Give this a meaningful contentDescription like "Actions for Image ...", where ... can be either a number, like 1 or 2 etc., or if a description has already been provided, append that so the user already knows which of the images they are about to manipulate.

* * * *
- Tusky Version: both 5 and 6 Nightly.
- Android Version: 9
- Android Device: Google Pixel 3
- Mastodon instance (if applicable): All.

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

charlag:
Thanks for reporting! We should definitely do that!

connyduck:
https://github.com/tuskyapp/Tusky/pull/1188

PR #1188, opened 2019-04-12
Add correct Content Description for the preview images at the Compose screen

pandasoft0:
Fix the #1155 issue.
Add description to the preview image:
1. "Actions for the image %description%" if image has the caption
2. "Actions for the image %idx+1%" if image has not caption, where idx is the index of the image at the queue list.

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/ComposeActivity.java
+++ b/app/src/main/java/com/keylesspalace/tusky/ComposeActivity.java
@@ -133,6 +133,25 @@ import java.util.concurrent.CountDownLatch;
 
 import javax.inject.Inject;
 
+import androidx.annotation.ColorInt;
+import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
+import androidx.annotation.Px;
+import androidx.annotation.StringRes;
+import androidx.appcompat.app.ActionBar;
+import androidx.appcompat.app.AlertDialog;
+import androidx.appcompat.content.res.AppCompatResources;
+import androidx.appcompat.widget.Toolbar;
+import androidx.core.app.ActivityCompat;
+import androidx.core.content.ContextCompat;
+import androidx.core.content.FileProvider;
+import androidx.core.view.inputmethod.InputConnectionCompat;
+import androidx.core.view.inputmethod.InputContentInfoCompat;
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/keylesspalace/tusky/ComposeActivity.java
```diff
@@ -1158,6 +1158,33 @@ private void addMediaToQueue(@Nullable String id, QueuedMedia.Type type, Bitmap
...
+
+    private void updateContentDescription(QueuedMedia item) {
+        if (item.preview != null) {
+            String imageId;
+            if (!TextUtils.isEmpty(item.description)) {
+                imageId = item.description;
+            } else {
+                int idx = getImageIdx(item);
+                if (idx < 0)
+                    imageId = getString(R.string.compose_preview_image_description_image_unknown);
```
connyduck: Is this a case that can happen? When?

pandasoft0: If item has the `description` - app set this description as `content description`
If item has not an item description - app set its `index+1` as description, but if item not found at the queue - it set to `unknown`. Generally, it should never happening, but, i prefer to handle all possible cases even they are not possible now, but it may possible in the future after some behaviour change

connyduck: I would just set un-set the description in that case, no need for an additional string resource. Keep in mind it needs to be translated into 28 languages.

pandasoft0: Ah, got it. Thanks, i will update code now

app/src/main/res/values/strings.xml
```diff
@@ -466,6 +466,7 @@
     <string name="pref_title_bot_overlay">Show indicator for bots</string>
 
     <string name="notification_clear_text">Are you sure you want to permanently clear all your notifications?</string>
+    <string name="compose_preview_image_description">Actions for the image %s</string>
```
connyduck:
```suggestion
    <string name="compose_preview_image_description">Actions for image %s</string>
```

ralexey: Done

PR #1193, opened 2019-04-13
Added check if text in compose toot field has only whitespaces

### LINKED ISSUE #1192: Improve “empty” text send behaviour
When you try to send empty post, there’s native error “Field must not be empty”. However, when you try to send whitespaces without any other characters, Tusky tries to create a post, resulting in unclickable (!) error “Status cannot be sent. A copy saved in Drafts”.

There are different approaches, one is simply stripping all whitespaces in the beginning of the text and in the end of it, however IMO it’s not good. If there’s some text in status, let user put as many whitespaces as he wishes.
Probably the best solution is to check whether there are any other characters. If no, show the error.

* * * *
- Tusky Version: 6.1 latest nightly
- Android Version: 9.0

MasterGroosha:
I'd like to fix #1192 myself. 
I'm not sure though whether my approach is the best one. 

There's a dedicated `calculateTextLength()` function, maybe I should change it to:
```java
    int calculateTextLength() {
        String composeText = textEditor.getText().toString();  // NEW
        if (composeText.trim().length() <= 0)                  // NEW
            return 0;                                          // NEW
        
        int offset = 0;
        URLSpan[] urlSpans = textEditor.getUrls();
        if (urlSpans != null) {
            for (URLSpan span : urlSpans) {
                offset += Math.max(0, span.getURL().length() - MAXIMUM_URL_LENGTH);
            }
        }
        int length = textEditor.length() - offset;
        if (statusHideText) {
            length += contentWarningEditor.length();
        }
        return length;
    }
```

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/ComposeActivity.java
+++ b/app/src/main/java/com/keylesspalace/tusky/ComposeActivity.java
@@ -1016,7 +1016,7 @@ public final class ComposeActivity
             spoilerText = contentWarningEditor.getText().toString();
         }
         int characterCount = calculateTextLength();
-        if (characterCount <= 0 && mediaQueued.size() == 0) {
+        if ((characterCount <= 0 || contentText.trim().length() <= 0) && mediaQueued.size() == 0) {
             textEditor.setError(getString(R.string.error_empty));
             enableButtons();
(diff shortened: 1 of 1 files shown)


connyduck:
I think your solution is fine!

PR #1198, opened 2019-04-18
upgrade build tools and enable r8 full mode

connyduck:
~250kB smaller build

### DIFF
--- a/app/src/main/res/layout/fragment_view_video.xml
+++ b/app/src/main/res/layout/fragment_view_video.xml
@@ -12,7 +12,6 @@
         android:id="@+id/mediaDescription"
         android:layout_width="match_parent"
         android:layout_height="wrap_content"
-        app:layout_constraintTop_toBottomOf="@+id/toolbar"
         app:layout_constraintBottom_toTopOf="@+id/videoPlayer"
         android:background="#60000000"
         android:lineSpacingMultiplier="1.1"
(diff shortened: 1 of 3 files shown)


app/src/main/res/layout/fragment_view_video.xml
```diff
@@ -12,7 +12,6 @@
         android:id="@+id/mediaDescription"
         android:layout_width="match_parent"
         android:layout_height="wrap_content"
-        app:layout_constraintTop_toBottomOf="@+id/toolbar"
```
connyduck: lint was blocking release builds because of this warning

charlag: Is it okay? Will it overlap?

connyduck:
As of Build Tools 3.4, R8 is the default shrinker, nothing else to do. `android.enableR8.fullMode=true` enables even more optimizations, but removes compatibility with ProGuard.

app/src/main/res/layout/fragment_view_video.xml
connyduck: Oh, looks like this was never actually working. I'll fix it.

charlag:
Do we need to add some rules for it then? We use Retrofit so we have some dynamicity.

connyduck:
Yes, R8 uses ProGuard rules, we leave them in the project. Or do you mean something else?

charlag:
Nah, I was just concerned about this moment. Good go go then I guess!

PR #1202, opened 2019-04-20
fix some lint warnings

connyduck:


app/src/main/java/com/keylesspalace/tusky/view/RoundedImageView.java
```diff
@@ -194,12 +194,10 @@ public ColorFilter getColorFilter() {
     }
 
     private void applyColorFilter() {
-        if (mBitmapPaint != null) {
-            mBitmapPaint.setColorFilter(mColorFilter);
-        }
+        mBitmapPaint.setColorFilter(mColorFilter);
```
charlag: Didn't we.. add it for some reason? Can't remember

connyduck: It is a final variable thats always initialized, I really can't see any reason for the null check

charlag: Ah sure

PR #1203, opened 2019-04-21
fix issue with foreground service on Oreo+

connyduck:
According to Google Play logs Tusky sometimes crashes with a `IllegalStateException` when sending a toot, this should fix it.

https://stackoverflow.com/questions/46445265/android-8-0-java-lang-illegalstateexception-not-allowed-to-start-service-inten


connyduck:
Yes it should, but seems there is a rare condition where it isn't. It crashed only 3 times in the last week.

PR #1204, opened 2019-04-21
fix crash when status has more than 4 images

connyduck:
and accessibility service is active

Pleroma allows attaching more than 4 images -> `IndexOutOfBoundsException`

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/util/ListStatusAccessibilityDelegate.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/util/ListStatusAccessibilityDelegate.kt
@@ -16,8 +16,10 @@ import androidx.recyclerview.widget.RecyclerView
 import androidx.recyclerview.widget.RecyclerViewAccessibilityDelegate
 import com.keylesspalace.tusky.R
 import com.keylesspalace.tusky.adapter.StatusBaseViewHolder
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/keylesspalace/tusky/util/ListStatusAccessibilityDelegate.kt
```diff
@@ -60,7 +62,8 @@ class ListStatusAccessibilityDelegate(
                         R.id.action_open_media_2,
                         R.id.action_open_media_3,
                         R.id.action_open_media_4)
-                for (i in 0 until status.attachments.size) {
+                val attachmentCount = min(status.attachments.size, MAX_MEDIA_ATTACHMENTS)
```
charlag: So we will add actions only for first 4 attachments right? Seems legit.

charlag:
And sorry for introducing another crash :disappointed:

app/src/main/java/com/keylesspalace/tusky/util/ListStatusAccessibilityDelegate.kt
connyduck: Yes we also show only 4 images. Maybe we should extend the number sometime.

PR #1212, opened 2019-04-22
add support for multiple choice polls

connyduck:
4 additional views to every status layout just for this :/ 
I did not find a good way to reuse the RadioButtons though

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/adapter/StatusBaseViewHolder.java
+++ b/app/src/main/java/com/keylesspalace/tusky/adapter/StatusBaseViewHolder.java
@@ -8,6 +8,7 @@ import android.text.TextUtils;
 import android.view.View;
 import android.view.ViewGroup;
 import android.widget.Button;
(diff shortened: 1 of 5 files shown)


app/src/main/java/com/keylesspalace/tusky/adapter/StatusBaseViewHolder.java
```diff
@@ -839,6 +806,94 @@ protected void setupPoll(Poll poll, List<Emoji> emojis, StatusActionListener lis
...
 
+
+        }
+    }
+
+    private void setupPollResult(Poll poll, List<Emoji> emojis) {
+        List<PollOption> options = poll.getOptions();
+
+        for(int i = 0; i < Status.MAX_POLL_OPTIONS; i++) {
+            if(i < options.size()) {
```
charlag: Here and in all other places, can we get a crash if the server supports more options for some reason?

PR #1225, opened 2019-04-26
fix deleted status reappearing in the timeline

connyduck:
Steps to reproduce: Have a deleted status as the second status in you timeline. Switch accounts (or restart Tusky), notice how it disappears when the timeline is reloaded. Switch again, the toot appears and disappears again.

This PR fixes the problem by deleting the whole range of loaded status instead of only relying on `onConflict = REPLACE`
Not 100% sure this is the correct approach, but it seems to work. Will do more testing tomorrow.

this is part 1/x of fixes to the timeline glitches

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -133,6 +133,7 @@ dependencies {
         exclude group: 'com.android.support', module: 'support-annotations'
     })
     androidTestImplementation 'android.arch.persistence.room:testing:1.1.1'
+    androidTestImplementation 'androidx.test.ext:junit:1.1.0'
     testImplementation 'androidx.test.ext:junit:1.1.0'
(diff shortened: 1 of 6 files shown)


charlag:
Looks good bug I would propose to remove and insert in one transaction. That's probably not terribly important tho.
Also I think that test is a must here, maybe very simple one with two statuses outside the range and one inside.

connyduck:
Whats the best way to get this into a transaction? I tried adding a method to TimelineDao, but then it looses its testability :/

charlag:
Maybe it makes sense to add it to that method which inserts into timeline and test them together? I don't know. Maybe it's also okay to leave it like that, that's not going to reduce consistency.

connyduck:
I will merge it so we can see how it works for users, if there are problems we can change it again.

PR #1229, opened 2019-05-01
Poll notifications

connyduck:


connyduck:
device-2019-05-01-221137

app/src/main/java/com/keylesspalace/tusky/util/NotificationHelper.java
```diff
@@ -569,22 +588,43 @@ private static String titleForType(Context context, Notification notification, B
             case REBLOG:
                 return String.format(context.getString(R.string.notification_reblog_format),
                         accountName);
+            case POLL:
+                if(notification.getStatus().getAccount().getId().equals(account.getAccountId())) {
```
charlag: I don't even know if we need this distinction or generic "the poll has ended" would be enough but if it's there cool then

app/src/main/java/com/keylesspalace/tusky/db/AppDatabase.java
```diff
@@ -279,4 +279,11 @@ public void migrate(@NonNull SupportSQLiteDatabase database) {
         }
     };
 
+    public static final Migration MIGRATION_15_16 = new Migration(15, 16) {
```
charlag: We should update `MigrationTest` sometime...

app/src/main/java/com/keylesspalace/tusky/adapter/StatusViewHolder.java
```diff
@@ -81,30 +82,39 @@ protected void setupWithStatus(StatusViewData.Concrete status, final StatusActio
 
                 String rebloggedByDisplayName = status.getRebloggedByUsername();
                 if (rebloggedByDisplayName == null) {
-                    hideRebloggedByDisplayName();
+                    hideStatusInfo();
                 } else {
                     setRebloggedByDisplayName(rebloggedByDisplayName);
+                    statusInfo.setOnClickListener(v -> listener.onOpenReblog(getAdapterPosition()));
```
charlag: Listener is set here but not replaced for poll?

connyduck: Agreed, its ugly, but its save because its are never used on the same ViewHolder. Will refactor once I have more time.

PR #1233, opened 2019-05-02
Update Gradle wrapper to 5.4.1

friederbluemle:


connyduck:
Last time we did this, FDroid hat problems building Tusky so I would rather not update yet and wait until the Android plugin requires it.

friederbluemle:
@connyduck - Thanks for checking. What were the build problems you were seeing?
The Android Gradle plugin `3.4.0` (which is currently used) _already_ comes with Gradle wrapper `5.4` by default.

connyduck:
Their automated build system could not handle it for some reason, they had to do some manual action and the Tusky release was delayed several days. I will merge this after the next release, ok?

friederbluemle:
@connyduck - Sure, sounds good, thanks!
I'll keep this PR rebased. :+1:

PR #1235, opened 2019-05-02
Update Deps

bernd289:
Alpha Versions: Update to Gradle promoted library version


connyduck:
Sorry, I would rather not update to alpha versions if we do not need the additional features really bad.

PR #1238, opened 2019-05-03
Poll fixes

connyduck:
Some small bugs with polls. Might add more fixes as I discover problems.


app/src/main/java/com/keylesspalace/tusky/adapter/StatusBaseViewHolder.java
```diff
@@ -864,17 +869,23 @@ private void setupPollVoting(Poll poll, List<Emoji> emojis, StatusActionListener
 
             pollButton.setOnClickListener(v -> {
 
-                List<Integer> pollResult = new ArrayList<>(options.size());
-                for(int i = 0; i < options.size(); i++) {
-                    if(pollCheckboxOptions[i].isChecked()) {
-                        pollResult.add(i);
+                int position = getAdapterPosition();
+
+                if (position != RecyclerView.NO_POSITION) {
```
connyduck: we have this check in the other listeners as well and it seems to prevent some crashes

PR #1240, opened 2019-05-05
fix timespan plurals

connyduck:
Sorry, now translators have to translate them again :/

please check if this is what you need @MasterGroosha


MasterGroosha:
Looks good to me, I've already tested new plurals and they work fine.

2 things bother me however:
1) For now, I had to remove "left" word from `poll_info_time_relative` string and move it to `timespan_X`, since that word has different cases with different plurals. This is fine now, but may cause potential problems in future is timespans will be reused somewhere. Can't say for other languages though.
2) When I see "1 minute left" I personally think as "0-59 seconds left". However, Tusky shows "1 minute left" for "61-119 seconds" and after that shows "59 seconds left", "58 seconds left" and so on. There might be cultural differences and I don't know what is the right way, just noticed.

connyduck:
I will change the name of the string and the helper function to make sure it is not used for anything else and we should be fine.

PR #1249, opened 2019-05-06
Fix bot icon indicator again

connyduck:
There was another issue with recycling the indicator, maybe the Glide request did not get cancelled when doing `setImageResource`, not 100% sure. The bigger issue was, that the bot flag did not get saved to db and the indicator sometimes was not shown at all.


connyduck:
Here is a screenshot of the bug, Testaccount has Gargs profilepicture instead of the bot indicator 
Screenshot_20190506-113901

PR #1256, opened 2019-05-11
fix images with focal point not shown in timeline

### LINKED ISSUE #1255: [Bug] Some images disappear
It seems that Tusky changes the image library in 7.0 release and some images just disappear after loaded. Most of them are PNGs from pawoo.net.

PS: If you touch the eye button at the left-top corner twice it will show again, but then if you swipe up and down it will disappear again... 😂 

* * * *
- Tusky Version: 7.0
- Android Version: 9.0
- Mastodon instance (if applicable): pawoo.net

connyduck:
Apparently Glide calls the `RequestListener` before it sets the image (Picasso did it the other way), so the first calculation of the matrix gets the wrong size of the drawable and nothing is displayed. This should fix the problem.

Closes #1255

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/view/MediaPreviewImageView.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/view/MediaPreviewImageView.kt
@@ -88,7 +88,6 @@ defStyleAttr: Int = 0
      * Overridden setScaleType method which only accepts the new type if we don't have a focal
      * point set.
      *
-     *
      */
     override fun setScaleType(type: ScaleType) {
         if (focus != null) {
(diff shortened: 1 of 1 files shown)


charlag:
As far as I understand you extracted a method because we cannot can super `onSizeChanged()`?

LGTM in any case

-- 
Sincerely,
Ivan



11 May 2019, 13:27 by notifications@github.com:

connyduck:
I extracted the method so the Drawable received from `onResourceReady` can be used, the one from `getDrawable` seems to be not updated at this point.

pandasoft0:
I created a draft PR #1257. I used a bit another solution - create the Matrix transformation for glide. It has an issue with transformation from a thumbnail to a full image view.

Please, note, this is a draft commit: need to update matrix transformation.

It might help you. I will close it later.

PR #1257, opened 2019-05-11
[TEMP] Implement Glide matrix transformation

pandasoft0:
Implement Matrix transformation for thumbnails with focal point


pandasoft0:
#1255

connyduck:
Pros of this implementation: We don't need a custom view, Glide seems to be able to cache it
Con of this implementation: Additional bitmap creation which is a big perf hit

pandasoft0:
Yes, it was just an idea ;)

PR #1263, opened 2019-05-16
upgrade deps

connyduck:

PR #1268, opened 2019-05-18
fix upgrade of material lib

connyduck:
In #1263 I removed the `colorSurface` because it made the main toolbar look ugly. Turns out it is needed for the color of drop downs. 

here is the code change that causes the problem: https://github.com/material-components/material-components-android/commit/6d4ffa29e3b9476352273b42dbdd16077eb88dc9

So my workaround is using a shape drawable instead of a color drawable. Once the materialcomponents are stable I will do a proper refactoring.

### DIFF
--- a/app/src/main/res/drawable/background_main_toolbar.xml
+++ b/app/src/main/res/drawable/background_main_toolbar.xml
@@ -1,0 +1,5 @@
+<?xml version="1.0" encoding="utf-8"?>
+<shape xmlns:android="http://schemas.android.com/apk/res/android"
+    android:shape="rectangle">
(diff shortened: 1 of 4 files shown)


connyduck:
check this line https://github.com/material-components/material-components-android/commit/6d4ffa29e3b9476352273b42dbdd16077eb88dc9#diff-0f63a43f16e30832772a3e5ff0c4d483R214

when its a ColorDrawable they manipulate it

charlag:
Okay then. As long as square shape works, should be fine. Thanks for pointing me to the place.

Issue #1221, opened 2019-04-26
Headings of categories in settings shouldn't be tappable 


DragoCubed:

connyduck:
https://github.com/tuskyapp/Tusky/pull/1239

PR #1273, opened 2019-05-20
downgrade alpha libraries

connyduck:
The two `-alpha` dependencies transitively upgrade the appcompat library which breaks our in-app language switching. (Its not working when theme is set to dark.) This took me hours to figure out 😠 and I hope we can get rid of this alpha stuff soon and forever. For now I am downgrading again. Unfortunately I cannot go back to a stable version as we are already using a lot of the new features. Yes, this probably introduces #1221 again, but that one is not as bad as the other problems.


charlag:
Oh, I'm sorry, I didn't have time yet to look at it yet, but now I see that it's a tint chanhe. Hopefully they'll release something stable soon.

Issue #810, opened 2018-08-26
Feature Request: Add @ and # characters to compose window


bannisdale:
As # and @ are characters, people usually don't use a lot outside of social networks, it's sometimes kinda cumbersome to use them with your keyboard of choice on a mobile device. Given that these are very essential characters for the Fediverse, though, it might be helpful to have them at hand directly in the compose window.

In my opinion, there is a lot of unused space in the upper bar that would be perfect. My personal proposition would be to add them (and maybe other characters, too?) in the upper-right corner, left of the profile picture of the posting account.

mal0ki:
So I just checked my Twitter app of choice, since it's the only app that's remotely similar, and lo and behold they do have # and @ just above the compose window next to the character limit. 

Interesting. I've never noticed before, and never used it. 

It's a good idea.

charlag:
It may be a nice idea but I we barely made enough space for all the options when @connyduck was re-designing it last time. Would be nice if some layout options were proposed by the champion ;)

charlag:
Hey @Tak is this a part of some PR or it's just in your fork? How does it work on small screens? I guess we could just add it if there's enough space but oh boy we have more functions than Twitter with CWs and sensitive images and custom emojis.

bannisdale:
@charlag I haven't examined the Tak-Commit but given that we have 6 areas in the lower bar already (Attachment, Visibility, CW, Emojis, Sign Count, 'TOOT!' button) and just 2 areas in the upper bar (close ... ... ... avatar), I would assume that adding '#' and '@' next to the avatar (lefthand side, of course) shouldn't result in space problems. :)

Tak:
Oh, I was experimenting in my fork - I didn't realize that referencing the issue in the commit on my fork would create a reference here. :smiley: 
My first attempt was to add them in the lower bar, and there's not enough room on smaller devices.
I also considered adding them in the emoji picker, but that's not terribly intuitive, and makes the emoji picker code less nice.
Maybe the upper bar could be a decent option...

charlag:
I think that we should ask Connyduck about design stuff, I'm quite terrible at such UI things :D

bannisdale:
@connyduck Any opinion on that? :)

bannisdale:
So ... Uhm ... Would anyone mind to ask the duck? :D

connyduck:
[...]

PR #1277, opened 2019-05-23
Add @ and # helper buttons to compose activity

Tak:
Addresses #810

(Styling help is welcome)

buttons

### DIFF
--- a/app/src/main/res/layout/activity_compose.xml
+++ b/app/src/main/res/layout/activity_compose.xml
@@ -21,6 +21,32 @@
             android:padding="8dp"
             tools:ignore="ContentDescription" />
         <!--content description will be set in code -->
+
+        <androidx.appcompat.widget.AppCompatButton
+            android:id="@+id/atButton"
+            style="?attr/image_button_style"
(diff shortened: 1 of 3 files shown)


app/src/main/res/layout/activity_compose.xml
```diff
@@ -21,6 +21,34 @@
...
+            android:layout_width="?attr/actionBarSize"
+            android:layout_height="?attr/actionBarSize"
+            android:layout_gravity="end"
+            android:contentDescription="\@"
+            android:padding="8dp"
+            android:layout_marginHorizontal="4dp"
+            android:text="\@"
+            android:textColor="?android:textColorTertiary"
+            android:textSize="?attr/status_text_large"
+            />
```
connyduck: make both these Buttons `androidx.appcompat.widget.AppCompatButton`, width 40dp, textStyle bold.
`layout_marginHorizontal` only works on Api 26+

`contentDescription` should not be necessary when `text` is the same. Please extract the string resource in a string to make the warning go away (you can put it in donottranslate.xml)

PR #1279, opened 2019-05-24
Animate gif avatars

connyduck:


charlag:
I started reviewing it, had some concerns about transformations (it will be much more expensive than before) but I need to update couple of things locally first.

app/src/main/java/com/keylesspalace/tusky/adapter/TimelineAdapter.java
```diff
@@ -111,13 +119,29 @@ public void setMediaPreviewEnabled(boolean enabled) {
...
 
     public boolean getMediaPreviewEnabled() {
         return mediaPreviewEnabled;
     }
 
+    public void setShowBotOverlay(boolean showBotOverlay) {
+        this.showBotOverlay = showBotOverlay;
+    }
+
+    public boolean getShowBotOverlay() {
```
charlag: These getters are never used

app/src/main/java/com/keylesspalace/tusky/fragment/NotificationsFragment.java
```diff
@@ -740,13 +744,12 @@ public void onPreferenceChanged(String key) {
                 hideFab = PreferenceManager.getDefaultSharedPreferences(getContext()).getBoolean("fabHide", false);
                 break;
             }
-            case "mediaPreviewEnabled": {
+            case "absoluteTimeView": {
```
charlag: Not clear why this has changed? Should be both?..

app/src/main/java/com/keylesspalace/tusky/util/ImageLoadingHelper.kt
```diff
@@ -0,0 +1,43 @@
...
+
+    if(url.isNullOrBlank()) {
+        Glide.with(imageView)
+                .load(R.drawable.avatar_default)
+                .into(imageView)
+    } else {
+        if (animate) {
+            Glide.with(imageView)
+                    .load(url)
+                    .transform(
```
charlag: This results in two additional bitmap allocations and two expensive transformations but it seems to be the only way to do this without writing our own transformation...

app/src/main/java/com/keylesspalace/tusky/util/ImageLoadingHelper.kt
```diff
@@ -0,0 +1,43 @@
...
+import androidx.annotation.Px
+import com.bumptech.glide.Glide
+import com.bumptech.glide.load.resource.bitmap.FitCenter
+import com.bumptech.glide.load.resource.bitmap.RoundedCorners
+import com.keylesspalace.tusky.R
+
+
+private val fitCenterTransformation = FitCenter()
+
+fun loadAvatar(url: String?, imageView: ImageView, @Px radius: Int, animate: Boolean) {
```
charlag: I would propose two changes to make this helper more useful and to cut down boilerplate:
- Let it decide on itself if it should animate or not. I don't think there's a case where it shouldn't decide that? It has a context and can figure it out. I don't know if there's any performance difference from fetching the pref each time, maybe it could be an object instead?
- Let it fetch the size in pixels by itself. It can either accept resource or some enum.

That's largely an ergonomic/DRY proposal, it's really not nice to fetch all these sizes and prefs each time you want to display an avatar.

app/src/main/java/com/keylesspalace/tusky/ComposeActivity.java
```diff
@@ -303,14 +286,20 @@ public void onCreate(Bundle savedInstanceState) {
...
-                        .placeholder(R.drawable.avatar_default)
-                        .into(composeAvatar);
-            }
+
+            int[] actionBarSizeAttr = new int[] { R.attr.actionBarSize };
+            TypedArray a = obtainStyledAttributes(null, actionBarSizeAttr);
+            int avatarSize = a.getDimensionPixelSize(0, 1);
+            a.recycle();
+
+            boolean animateAvatars = preferences.getBoolean("animateGifAvatars", false);
```
charlag: We need some helper which can fetch preference with correct key and correct type. At least we need constants for keys.

charlag: That's a different thing obv.

connyduck: yes we need that, but not in this PR

app/src/main/java/com/keylesspalace/tusky/util/ImageLoadingHelper.kt
connyduck: yes 😔
Its worse than our current solution, but its done on an background thread and cached, so I think we can go with it.

app/src/main/java/com/keylesspalace/tusky/fragment/NotificationsFragment.java
connyduck: nope this was a mistake, thx for finding it

app/src/main/java/com/keylesspalace/tusky/adapter/TimelineAdapter.java
connyduck: thx

app/src/main/java/com/keylesspalace/tusky/util/ImageLoadingHelper.kt
connyduck: Imho, pixel size and pref should be loaded only once when the adapter is created for perf reasons, but yes its not DRY at all. Is it ok for you if I leave it like that for now?

charlag: Absolutely, it's just a proposal. For perf reasons I proposed an an object which can be created once per holds or so (I think both dimensions and preys are cached)

PR #1281, opened 2019-05-25
remove some unneeded code from BaseActivity

connyduck:


app/src/main/java/com/keylesspalace/tusky/BaseActivity.java
```diff
@@ -92,11 +88,6 @@ protected void onCreate(@Nullable Bundle savedInstanceState) {
 
         setTaskDescription(new ActivityManager.TaskDescription(appName, appIcon, recentsBackgroundColor));
 
-        long accountId = getIntent().getLongExtra("account", -1);
-        if (accountId != -1) {
-            accountManager.setActiveAccount(accountId);
-        }
-
```
connyduck: `"account"` is not used anywhere else in the app. The real account switching code is in `MainActivity`

app/src/main/java/com/keylesspalace/tusky/BaseActivity.java
```diff
@@ -165,24 +156,6 @@ protected void redirectIfNotLoggedIn() {
...
-        }
-        for (int i = 0; i < menu.size(); i++) {
-            Drawable icon = menu.getItem(i).getIcon();
-            if (icon != null) {
-                icon.setColorFilter(color, PorterDuff.Mode.SRC_ATOP);
-            }
-        }
-        return super.onCreateOptionsMenu(menu);
-    }
-
```
connyduck: this was once needed, but with the new Material themes it apparently is no longer

PR #1282, opened 2019-05-25
improve notification fragment layout

connyduck:
- Remove two layouts to make it more efficient
- The button bar is slightly smaller now. A lot of users requested that, but I would not make it too small so the buttons are still easy to touch.

cc @pandasoft0

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/di/NetworkModule.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/di/NetworkModule.kt
@@ -77,7 +77,7 @@ class NetworkModule {
                 .apply {
                     addInterceptor(InstanceSwitchAuthInterceptor(accountManager))
                     if (BuildConfig.DEBUG) {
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/keylesspalace/tusky/di/NetworkModule.kt
```diff
@@ -77,7 +77,7 @@ class NetworkModule {
                 .apply {
                     addInterceptor(InstanceSwitchAuthInterceptor(accountManager))
                     if (BuildConfig.DEBUG) {
-                        addInterceptor(HttpLoggingInterceptor().setLevel(HttpLoggingInterceptor.Level.BODY))
+                        addInterceptor(HttpLoggingInterceptor().setLevel(HttpLoggingInterceptor.Level.BASIC))
```
connyduck: Ah yes I keep changing that around depending on what I want to debug. BASIC is best for most usecases I guess.

connyduck:
> Seems fine?

yeah its quite an obvious change

> Toolbar was higher, right?

yes

> also, do you plan to change sw640 version as well?

Yes of course, why do I always forget it >_<

PR #1283, opened 2019-05-25
improve performance when clearing notifications

connyduck:


app/src/main/java/com/keylesspalace/tusky/util/NotificationHelper.java
```diff
@@ -465,13 +468,18 @@ public static void disablePullNotifications() {
 
     public static void clearNotificationsForActiveAccount(@NonNull Context context, @NonNull AccountManager accountManager) {
         AccountEntity account = accountManager.getActiveAccount();
-        if (account != null) {
-            account.setActiveNotifications("[]");
-            accountManager.saveAccount(account);
```
connyduck: Database access! On the main thread! And in 90% of cases it's redundant!

connyduck:
This should make the perf problems slightly less worse @koyuawsmbrtn

app/src/main/java/com/keylesspalace/tusky/util/NotificationHelper.java
charlag: Could it be caught by `android.os.StrictMode`? Could you try it out please? Event `StrictMode.enableDefaults()` should be enough.

charlag: THIS is what causing lags for me! I thought it's some debugging in my branch but it's *this*. Good job finding it!

koyuawsmbrtn:
Nice! Sign me up.

connyduck:
No its not caught by `StrictMode.enableDefaults()`. Probably because SQLite only writes it to a cache?

charlag:
Weird, I rememeber Room catching it and there was a trick to allow it while debugging.

charlag:
Yup, unconditionally for some reason.
https://github.com/tuskyapp/Tusky/blob/master/app/src/main/java/com/keylesspalace/tusky/TuskyApplication.java#L72

connyduck:
We should get rid of this `allowMainThreadQueries()`. I guess there are more main thread queries that are not that easy to remove though.

PR #1286, opened 2019-05-26
fix text color in dropdown of light theme

connyduck:
before
device-2019-05-26-100309

after
device-2019-05-26-100139

### DIFF
--- a/app/src/main/res/values/styles.xml
+++ b/app/src/main/res/values/styles.xml
@@ -30,7 +30,7 @@
         <item name="status_text_large">22sp</item>
     </style>
 
-    <style name="AppTheme.Account.AppBarLayout" parent="ThemeOverlay.AppCompat">
+    <style name="ViewMediaActivity.AppBarLayout" parent="ThemeOverlay.AppCompat">
         <item name="android:colorControlNormal">@color/white</item>
     </style>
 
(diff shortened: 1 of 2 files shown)


app/src/main/res/values/styles.xml
```diff
@@ -31,9 +31,7 @@
     </style>
 
     <style name="AppTheme.Account.AppBarLayout" parent="ThemeOverlay.AppCompat">
-        <item name="android:textColorPrimary">@color/text_color_primary_dark</item>
-        <item name="android:textColorSecondary">@color/text_color_secondary_dark</item>
-        <item name="android:textColorTertiary">@color/text_color_tertiary_dark</item>
+        <item name="android:colorControlNormal">@color/white</item>
```
charlag: White is okay for both themes? I'm confused tbh

connyduck: Yes this only affects the ViewMediaActivity hat is completely black regardless of theme.

charlag: Okay, maybe we should document that so that we don't have surprises later

connyduck: true, lets rename the theme to make clear where its used

Issue #1287, opened 2019-05-26
Accessibility problems with polls


charlag:
- Polls are not announced by the screen reader
- No way to vote for polls with a screen reader

It shouldn't be too hard to add it but the problem is that translations must be updated accordingly. Not sure how to solve this yet, any ideas @connyduck ?

charlag:
Maybe I can smash poll description inside the content, not sure yet

charlag:
After testing I see that non-finished polls can actually be acted upon but finished ones are not announced at all

PR #1288, opened 2019-05-26
Minimal screen reader support for polls

charlag:
This is a minimal fix for #1287. Notifications are still not announced correctly and I didn't add any special actions for polls so one may have to scroll through them manually. I think they're rare enough that it's not a big problem yet.
I checked `description_status` in all languages and it's either left as-is (even in Asian languages) or not translated at all. So I've made it non-translatable.


connyduck:
Lgtm, definitely better than nothing. What do you mean, not announced correctly?

Making `description_status` non-translateable is a good idea, it was confusing translators a lot.

charlag:
Well poll notifications are announced like statuses, there's no "your poll has ended/poll you voted in has ended", there's no easy way to propogate this info currently. If you have ideas, I would be glad.

connyduck:
Ah, yes, I see 🤔 
No idea. I will merge this since its still an improvement to the current status?

charlag:
Yeah, I think it's better than not announcing polls in any case

PR #1289, opened 2019-05-26
[WIP] Streaming, again

charlag:
Cherry picked #892 which is beyond salvation by now.
Surprisingly it really works well, I think adding some refresh/gap thing and some disconnect/connect behavior could be enough to ship it. Maybe a toggle as well, although web doesn't have it.

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/fragment/TimelineFragment.java
+++ b/app/src/main/java/com/keylesspalace/tusky/fragment/TimelineFragment.java
@@ -55,6 +55,7 @@ import com.keylesspalace.tusky.appstore.FavoriteEvent;
 import com.keylesspalace.tusky.appstore.MuteEvent;
 import com.keylesspalace.tusky.appstore.NewHomeTimelineStatusEvent;
 import com.keylesspalace.tusky.appstore.PreferenceChangedEvent;
+import com.keylesspalace.tusky.appstore.ProflileStreamingReconnectedEvent;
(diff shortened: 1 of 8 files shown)


connyduck:
Ready for review?

charlag:
I think nor yet, need to do something with timeline on resume

charlag:
I've rebased it and added things to load top events on resume. Also I've tried to sync them a little. There's much more of what can be done, I would feedback from anyone on this.

app/src/main/java/com/keylesspalace/tusky/fragment/TimelineFragment.java
```diff
@@ -976,33 +992,48 @@ private void sendFetchTimelineRequest(@Nullable String maxId, @Nullable String s
...
-            Callback<List<Status>> callback = new Callback<List<Status>>() {
-                @Override
-                public void onResponse(@NonNull Call<List<Status>> call, @NonNull Response<List<Status>> response) {
-                    if (response.isSuccessful()) {
-                        onFetchTimelineSuccess(liftStatusList(response.body()), fetchEnd, pos);
-                    } else {
-                        onFetchTimelineFailure(new Exception(response.message()), fetchEnd, pos);
-                    }
-                }
+            Single<?> result = Single
```
charlag: Maybe I shouldn't do this and should instead convert all of the timeline requests into `Single`s.

charlag:
Updated it a little bit because I've found some ridiculous bugs.

PR #1302, opened 2019-06-01
cleanup entity classes and ViewThreadFragment

connyduck:
At first, I just wanted to cleanup the entitiy classes

- `var` -> `val` where possible
- use data classes
- remove unused attributes
- fix nullability

But then I found that since Mastodon 2.6 every status has the Card already embedded so there is no need for that extra request in `ViewThreadFragment`.

Also fixes the problem that avatars where never animated in `ViewThreadFragment`

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/entity/Status.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/entity/Status.kt
@@ -31,8 +31,8 @@ data class Status(
         val emojis: List<Emoji>,
         @SerializedName("reblogs_count") val reblogsCount: Int,
         @SerializedName("favourites_count") val favouritesCount: Int,
-        var reblogged: Boolean = false,
-        var favourited: Boolean = false,
+        var reblogged: Boolean,
+        var favourited: Boolean,
         var sensitive: Boolean,
         @SerializedName("spoiler_text") val spoilerText: String,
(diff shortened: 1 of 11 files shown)


app/src/main/java/com/keylesspalace/tusky/entity/Status.kt
```diff
@@ -120,45 +121,17 @@ data class Status(
     }
 
 
-    class Mention {
-        var id: String? = null
+    data class Mention (
```
charlag: Why it was not even a data class before? Was there some problem with embedding somewhere?

charlag:
I checked it, seems to be fine

charlag:
I noticed one more thing about cards. Pleroma seems to return 0/0 for width/height of the image, at least for some of them. We hide the image in this case, even though we don't use the size.

https://github.com/tuskyapp/Tusky/blob/master/app/src/main/java/com/keylesspalace/tusky/adapter/StatusDetailedViewHolder.java#L159

app/src/main/java/com/keylesspalace/tusky/entity/Status.kt
connyduck: probably a mistake

connyduck:
Thx! Will fix the cards with Pleroma in a seperate PR.

Issue #1217, opened 2019-04-25
Using sensitive content button once force the toot to be marked as sensitive


rapenne-s:
When tooting, if using the "Sensitive content" button twice, without filling the warning text, the toot will still be posted with sensitive parameter set to yes.

* * * *
- Tusky Version: 6.1
- Android Version: 8.1.0
- Android Device: something
- Mastodon instance (if applicable): bsd.network

- [ ] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

connyduck:
Thats really weird. I can not reproduce the issue, can you give me a more detailed description or a video of the bug please?

connyduck:
Ahh I see, it only makes a difference when adding an image to the status. Will fix!

connyduck:
fixed #1309

PR #1309, opened 2019-06-06
Fix issue with wrong sensitive status

pandasoft0:
Save current sensitivity state before force switch it to true and restore when switch content sensitivity to false
tuskyapp/Tusky#1217

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/ComposeActivity.java
+++ b/app/src/main/java/com/keylesspalace/tusky/ComposeActivity.java
@@ -992,7 +992,7 @@ public final class ComposeActivity
         }
 
         Intent sendIntent = SendTootService.sendTootIntent(this, content, spoilerText,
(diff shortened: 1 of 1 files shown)


pandasoft0:
I will add this check on "send status". But i suggest to leave my changes also.
Use case:
1. User attaches the image
2. Set the "sensitive content". (it auto set the "sensitive media")
3. Unset the "sensitive content". Sensitive media still be set.

My changes. App stores the "sensitive media" state and restore previous state it if user unset the "sensitive content".

But if previous solution was ok - i will revert changes back and add check only to send status method

pandasoft0:
Done

connyduck:
I do not think that this is a problem. Mastodon web behaves the same way? Once the "sensitive" flag is set, you can only unset it with the dedicated button.

pandasoft0:
Got it.
I left last commit only.

Issue #1143, opened 2019-03-22
[Feature Request] Domain Blocks


aka-rin:
To block annoying messages in federated public timeline...

API reference: https://docs.joinmastodon.org/api/rest/domain-blocks/

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

Px3XjG6P:
If this the same as blocking/muting an entire mastodone instance, then it's exactly what I'm looking for.

connyduck:
#1311

PR #1311, opened 2019-06-07
Implement instance mutes

Tak:
Adds a menu item to the account activity, similar to the web UI, and an activity in account preferences similar to the muted users activity.

Addresses #1143

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/AccountActivity.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/AccountActivity.kt
@@ -656,7 +656,7 @@ class AccountActivity : BottomSheetActivity(), ActionButtonActivity, HasAndroidI
     private fun showMuteDomainWarningDialog(instance: String) {
         AlertDialog.Builder(this)
                 .setMessage(getString(R.string.mute_domain_warning, instance))
-                .setPositiveButton(android.R.string.ok) { _, _ -> viewModel.muteDomain(instance) }
+                .setPositiveButton(getString(R.string.mute_domain_warning_dialog_ok)) { _, _ -> viewModel.muteDomain(instance) }
                 .setNegativeButton(android.R.string.cancel, null)
                 .show()
     }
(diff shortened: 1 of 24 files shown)


connyduck:
😍

app/src/main/res/values/strings.xml
```diff
@@ -92,6 +93,7 @@
     <string name="action_view_favourites">Favourites</string>
     <string name="action_view_mutes">Muted users</string>
     <string name="action_view_blocks">Blocked users</string>
+    <string name="action_view_domain_mutes">Muted instances</string>
```
connyduck: the wording used in mastodon web is `Hidden domains`. I think we should follow that.

app/src/main/java/com/keylesspalace/tusky/AccountActivity.kt
```diff
@@ -642,6 +655,14 @@ class AccountActivity : BottomSheetActivity(), ActionButtonActivity, HasSupportF
                 .show()
     }
 
+    private fun showMuteDomainWarningDialog(instance: String) {
+        AlertDialog.Builder(this)
+                .setMessage(getString(R.string.mute_domain_warning, instance))
+                .setPositiveButton(android.R.string.ok) { _, _ -> viewModel.muteDomain(instance) }
```
connyduck: I think the button should say "hide entire domain" like in Mastodon web.

app/src/main/res/values/strings.xml
```diff
@@ -179,6 +182,7 @@
     <string name="dialog_unfollow_warning">Unfollow this account?</string>
     <string name="dialog_delete_toot_warning">Delete this toot?</string>
     <string name="dialog_redraft_toot_warning">Delete and re-draft this toot?</string>
+    <string name="mute_domain_warning">Are you really, really sure you want to block all of %s? In most cases, a few targeted blocks or mutes are sufficient and preferable.</string>
```
connyduck: How about

Are you sure you want to block the entire <domain>? You will not see content from that domain in any public timelines or your notifications. Your followers from that domain will be removed.

Tak:
Ok, I think I've addressed everything except the migration to the new coding style.
(I'll still do that, but you can merge like this and I'll make a follow-up PR, if you prefer)

connyduck:
Nice! Now, when I go to the instance mutes page while being in Airplane mode, I see the ProgressBar forever instead of the offline elephant friend.

Tak:
Hm, I can't repro that in the emulator or on my device. Any thoughts about what's missing?

connyduck:
Tried it again, now it works. Code also looks good. Maybe it was my mistake. 
Thx again for the awesome PR!

PR #1312, opened 2019-06-08
fix R8 rules to avoid crash when downloading Emoji Font

connyduck:
Seems R8 is optimizing away too much and the AsyncTask was missing methods. This also explains why it worked in a debug build.
cc @C1710

### DIFF
--- a/app/proguard-rules.pro
+++ b/app/proguard-rules.pro
@@ -94,9 +94,10 @@
 
 -dontwarn com.google.errorprone.annotations.*
 
-# work around a bug in proguard
-# see https://sourceforge.net/p/proguard/bugs/729/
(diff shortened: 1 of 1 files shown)


app/proguard-rules.pro
```diff
@@ -94,9 +94,10 @@
 
 -dontwarn com.google.errorprone.annotations.*
 
-# work around a bug in proguard
-# see https://sourceforge.net/p/proguard/bugs/729/
--keepnames public interface com.uber.autodispose.lifecycle.CorrespondingEventsFunction { *; }
```
connyduck: This old rule can be removed as apparently R8 does not have this problem like ProGuard did

Issue #1306, opened 2019-06-05
Setting to allow the disabling of the Clear|Filter buttons from the notification column


BNolet:
Issue raised by [Maddie|@MadestMadness@meow.social](https://meow.social/@MadestMadness)

Currently the two buttons are persistent until you scroll down, limiting the already scarce screen real-estate. I'd like to be able to disable this and possibly also disable it for a set period of time. Disabling at all is the most important. 

* * * *
- Tusky Version: 7.0
- Android Version: 8.0.1
- Android Device: Moto Z^2 Play
- Mastodon instance (if applicable):

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

connyduck:
fyi in Tusky 8 the buttons will be smaller

BNolet:
That's fantastic, but could it be possible to be able to turn them off completely?

connyduck:
yes of course

connyduck:
#1314

PR #1314, opened 2019-06-08
Add setting to show/hide notification filter bar

pandasoft0:
Add setting to show/hide notification filter bar
tuskyapp/Tusky#1306

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/fragment/NotificationsFragment.java
+++ b/app/src/main/java/com/keylesspalace/tusky/fragment/NotificationsFragment.java
@@ -284,8 +284,6 @@ public class NotificationsFragment extends SFragment implements
             //Clear behaviour to hide app bar
             params.setBehavior(null);
         }
-
-        swipeRefreshLayout.requestLayout();
     }
 
     private void confirmClearNotifications() {
--- a/app/src/main/java/com/keylesspalace/tusky/fragment/NotificationsFragment.java
+++ b/app/src/main/java/com/keylesspalace/tusky/fragment/NotificationsFragment.java
@@ -611,18 +611,7 @@ public class NotificationsFragment extends SFragment implements
     private void clearNotifications() {
         //Cancel all ongoing requests
         swipeRefreshLayout.setRefreshing(false);
-        for (Call callItem : callList) {
-            callItem.cancel();
-        }
-        callList.clear();
-        bottomLoading = false;
-        topLoading = false;
-
-        //Disable load more
-        bottomId = null;
-
-        //Clear exists notifications
-        notifications.clear();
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/keylesspalace/tusky/fragment/NotificationsFragment.java
```diff
@@ -255,10 +265,30 @@ public View onCreateView(@NonNull LayoutInflater inflater, @Nullable ViewGroup c
...
+            //Set content behaviour to hide filter on scroll
+            params.setBehavior(new AppBarLayout.ScrollingViewBehavior());
+        } else {
+            appBarOptions.setExpanded(false, false);
+            appBarOptions.setVisibility(View.GONE);
+            //Clear behaviour to hide app bar
+            params.setBehavior(null);
+        }
+
+        swipeRefreshLayout.requestLayout();
```
connyduck: this should not be necessary

ralexey: This is a workaround. If i hide the app bar without this workaround - it left the empty place instead. 
The same issue there: https://stackoverflow.com/questions/32704775/appbarlayout-take-space-after-setvisibilityview-gone/35126429

pandasoft0: This is a workaround. If i hide the app bar without this workaround - it leave the empty place instead of remove the bar.
https://stackoverflow.com/questions/32704775/appbarlayout-take-space-after-setvisibilityview-gone/35126429

pandasoft0:
It reloads notifications on show/hide the bar. Because some notifications type may be disabled via filter and were not loaded. So, when you switch the setting - need to reload them.

app/src/main/java/com/keylesspalace/tusky/fragment/NotificationsFragment.java
connyduck: Yes, the setting of the scrolling behavior is necessary, but the `requestLayout` should not be.

connyduck:
Yes, I understand why you reload, but it seems to be buggy. 
Here is a video so you see what I mean. https://connyduck.at/device-2019-06-10-200420.webm

pandasoft0:
Hm... I see, i will check.

app/src/main/java/com/keylesspalace/tusky/fragment/NotificationsFragment.java
pandasoft0: Ah. Sorry. Done.

pandasoft0:
Check new version, please

PR #1315, opened 2019-06-08
Update Dagger & other deps

connyduck:
Dagger added `HasAndroidInjector` which replaces the other `Has*Injector`s. Nice.

I did deliberatly not update the androidx alpha deps as they require code changes and I want to wait until they are stable.


connyduck:
I think that `Any` is need for compatibility with the old Injectors

PR #1296, opened 2019-05-30
improve usability: add shortcuts for compose/send new toot [#505]

unstabler:
https://twingyeo.kr/@cheesekun/102184993516052526

sorry for my bad english > <);

PR #1317, opened 2019-06-10
Fix back button not working

kyori19:
Since #1296 merged, back button is not working in `ComposeActivity`.
This PR fixes it.


connyduck:
Awesome, thx!

PR #1323, opened 2019-06-15
update Room to 2.1.0

connyduck:


charlag:
I would check tho if it executes is lazily or not, because if not then all this `subscribeWith()` doesn't make sense

connyduck:
I checked the sample app, looks like its necessary to subscribe.

PR #1326, opened 2019-06-18
Tusky is no longer a fully compliant Mastodon client

ealgase:
Given the fact that Tusky now blocks Gab (#1303), it is not a fully compliant client according to https://docs.joinmastodon.org/api/guidelines/.  

In particular:
image
Since a user can no longer log in to any Mastodon server, it is no longer a fully compliant client, so the README should reflect that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@
 
 ![](/fastlane/metadata/android/en-US/images/icon.png)
 
-Tusky is a beautiful Android client for [Mastodon](https://github.com/tootsuite/mastodon). Mastodon is a GNU social-compatible federated social network. That means not one entity controls the whole network, rather, like e-mail, volunteers and organisations operate their own independent servers, users from which can all interact with each other seamlessly.
+Tusky is a beautiful Android client for some [Mastodon](https://github.com/tootsuite/mastodon) instances. Mastodon is a GNU social-compatible federated social network. That means not one entity controls the whole network, rather, like e-mail, volunteers and organisations operate their own independent servers, users from which can all interact with each other seamlessly. While Tusky is not fully compliant, it supports most Mastodon features on most Mastodon instances.
 
 [<img src="/assets/fdroid_badge.png" alt="Get it on F-Droid" height="80" />](https://f-droid.org/repository/browse/?fdid=com.keylesspalace.tusky)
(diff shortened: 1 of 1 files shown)


freespeech4ever:
Now instead of having a few more users on your app, connecting to an instance you don't care about, your code will be forked and all of your effort goes towards a Gab branded app.  Thanks for your support!

charlag:
Ah, yes, we are totally afraid to not have nazi support who failed to write their own code for the server too.
>Guidelines

It just says that we should use OAuth and not hardcode one instance (like Librem One).

There's no "compliance suice" or whatever you came up with.
Pathetic.

ealgase:
> The user must be able to login to any Mastodon server from the app.

> any Mastodon server

Am I missing something?

charlag:
Yes, you're missing

> Guidelines

ealgase:
> Yes, you're missing
> 
> > Guidelines

Yes, the app is no longer compliant with the Guidelines, so it should be labelled as such.

Laurelai:
They are guidelines, not rules and compliance is relative, you dont have to follow them all strictly to be in compliance with them. Just mostly and adapted to your situation. Thats why they are guidelines.

charlag:
I shouldn't even mention that the person who wrote these guidelines is the same who proposed blocking one nasty hole.

ealgase:
image

He asked them to consider it.  An app developer who looked at their actual users finds that most of them would prefer Gab to not be blocked (although whether they do what their users want is another question).

Laurelai:
Tusky can code their app however they want to. I keep hearing threats to fork the app so why dont you all just shut up and actually do it. If you think your forked gab allowing app will be more popular go do it.

ealgase:
> Tusky can code their app however they want to. I keep hearing threats to fork the app so why dont you all just shut up and actually do it. If you think your forked gab allowing app will be more popular go do it.

https://github.com/antiantifeature/Tusky

Laurelai:
Then why are you wasting your time here then?

ealgase:
> Then why are you wasting your time here then?

Because the app has a misleading statement and it should be corrected.

Alternatively, mentioning incompatibility with Gab would also work.

Laurelai:
Nazis dont count and its expected by default to not cater to them.

ealgase:
> Nazis dont count and its expected by default to not cater to them.

Nazis are users, too.

Laurelai:
Nazis arent even fit to use the same oxygen as the rest of us, we had a whole world war over if its ok to be a nazi and the answer is no.

ealgase:
Then put that in the README, that Nazis are not welcome to use the software, and certain Nazi instances are blocked.

charlag:
Okay, that's enough.
Maybe Nazis are users of *your* apps but they're not welcome here.
We may or may not add it when (if) there will be an instance which is preemptively blocked now.
It doesn't make us more or less compliant to any imaginary standard of yours.

Issue #1081, opened 2019-02-26
Improve our search results


connyduck:
The Mastodon search can now do a lot more things than previously: https://github.com/tootsuite/mastodon/pull/10091

We should definitely take advantage of that, and, for example, present our search results in three (posts, users, hashtags) infinitely scrolling tabs. Or present the results as we do now, but with a "more" button for each section of the results.

Of course we need to wait until the majority of instances run this feature.

pandasoft0:
I checked the search APIv2.
1. What is prefer solution:
a) show first 5 items per category (statuses, accounts, hashtags) with "more" button below each category. On More click replace this summary fragment with category fragment and show infinite scroll list with items of selected category only;
b) show 3 tabs at top, and show infinite scroll lists per tab.
JFYI: in APIv2 we can't select what items to load. We can specify limit and offset, but anyway - it will load items for all 3 categories: accounts, statuses, hashtags.

2. Should i use the exists fragment and adapter to show statuses or create a new adapter in the same way i did for report's statuses? Should i implement all features of exists statuses adapter at the new?

3. Should i update exists activity+fragment code or create a new activity with fragments according to selected solution at p.1?

connyduck:
I prefer solution B. Only showing results of one type after "more" clicked is a bad user experience.

I think the 3 tabs and the search bar should belong to the activity, and the tab contents are three different fragments. They should all share one ViewModel probably.
The ViewHolders we have for statuses and accounts should definitely be reused, other than that you can decide.

ralexey:
@connyduck I am trying to use v2/search at mastodon.social but result looks strange.
So, first request. I want to get first 10 items per category:
```
https://mastodon.social/api/v2/search?q=test&resolve=true&limit=10&offset=0&following=false 
{
  "accounts": [
    {
      "id": "963"
    },
    {
      "id": "179860"
    },
    {
      "id": "1184"
    },
    {
      "id": "179862"
    },
    {
      "id": "820018"
    },
    {
      "id": "248304"
    },
    {
      "id": "725124"
    },
    {
      "id": "298091"
    },
    {
      "id": "217924"
    },
    {
      "id": "61339"
    }
  ],
  "statuses": [
    {
      "id": "102115228601373686"
    },
    {
      "id": "102146124881012866"
    },
    {
      "id": "102143335650922023"
    },
    {
      "id": "102246304503430384"
    },
    {
      "id": "102100944210578451"
    },
    {
      "id": "101836442594147121"
[...]

PR #1327, opened 2019-06-18
Improve search results

pandasoft0:
Issue #1081


app/src/main/java/com/keylesspalace/tusky/components/search/fragments/SearchStatusesFragment.kt
```diff
@@ -0,0 +1,147 @@
...
+    override fun onVoteInPoll(position: Int, choices: MutableList<Int>) {
+        //TODO("not implemented") //To change body of created functions use File | Settings | File Templates.
+    }
+
+    override fun removeItem(position: Int) {
+        //TODO("not implemented") //To change body of created functions use File | Settings | File Templates.
+    }
+
+    override fun onReblog(reblog: Boolean, position: Int) {
+        //TODO("not implemented") //To change body of created functions use File | Settings | File Templates.
```
connyduck: Don't forget to implement these methods

app/src/main/java/com/keylesspalace/tusky/components/search/SearchViewModel.kt
```diff
@@ -0,0 +1,66 @@
...
+import com.keylesspalace.tusky.viewdata.StatusViewData
+import io.reactivex.disposables.CompositeDisposable
+import javax.inject.Inject
+
+class SearchViewModel @Inject constructor(private val mastodonApi: MastodonApi) : ViewModel() {
+    private val disposables = CompositeDisposable()
+
+    private val statusesRepository = SearchRepository<Pair<Status, StatusViewData.Concrete>>(mastodonApi)
+    private val accountsRepository = SearchRepository<Account>(mastodonApi)
+    private val hastagsRepository = SearchRepository<HashTag>(mastodonApi)
```
connyduck: typo

app/src/main/java/com/keylesspalace/tusky/components/search/SearchViewModel.kt
```diff
@@ -0,0 +1,66 @@
...
+import com.keylesspalace.tusky.entity.Status
+import com.keylesspalace.tusky.network.MastodonApi
+import com.keylesspalace.tusky.util.Listing
+import com.keylesspalace.tusky.util.NetworkState
+import com.keylesspalace.tusky.util.ViewDataUtils
+import com.keylesspalace.tusky.viewdata.StatusViewData
+import io.reactivex.disposables.CompositeDisposable
+import javax.inject.Inject
+
+class SearchViewModel @Inject constructor(private val mastodonApi: MastodonApi) : ViewModel() {
```
connyduck:
```suggestion
class SearchViewModel @Inject constructor(private mastodonApi: MastodonApi) : ViewModel() {
```

app/src/main/java/com/keylesspalace/tusky/interfaces/AnchorActivity.kt
```diff
@@ -0,0 +1,7 @@
+package com.keylesspalace.tusky.interfaces
+
+import android.view.View
+
+interface AnchorActivity {
```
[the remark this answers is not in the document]
pandasoft0: It worked not correct because activity has a scrollable toolbar and fragments have CoordinatorLayout as root. When app searches the layout for snackbar it founds the fragment's CoordinatorLayout. But root CoordinatorLayout hides the bottom of the fragment (with Snackbar) when toolbar shown. I've removed child CoordinatorLayout, so, Snackbar always uses the Activity's CoordinatorLayout

app/src/main/java/com/keylesspalace/tusky/entity/HashTag.kt
```diff
@@ -0,0 +1,15 @@
+package com.keylesspalace.tusky.entity
+
+import com.google.gson.annotations.SerializedName
+
+data class HashTag(
+
+        @field:SerializedName("name")
+        val name: String,
```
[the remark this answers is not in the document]
pandasoft0: We must not forget to add this file to proguard rules in this case. I prefer to set the SerializedName and do not worry about proguard.

app/src/main/java/com/keylesspalace/tusky/entity/HashTag.kt
```diff
@@ -0,0 +1,15 @@
...
+
+        @field:SerializedName("url")
+        val url: String,
+
+        @field:SerializedName("history")
+        val history: List<History?>? = null
```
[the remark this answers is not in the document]
pandasoft0: done

app/src/main/java/com/keylesspalace/tusky/di/NetworkModule.kt
```diff
@@ -77,7 +77,7 @@ class NetworkModule {
                 .apply {
                     addInterceptor(InstanceSwitchAuthInterceptor(accountManager))
                     if (BuildConfig.DEBUG) {
-                        addInterceptor(HttpLoggingInterceptor().setLevel(HttpLoggingInterceptor.Level.BASIC))
+                        addInterceptor(HttpLoggingInterceptor().setLevel(HttpLoggingInterceptor.Level.BODY))
```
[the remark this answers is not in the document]
pandasoft0: Me too ;)
Done.

app/src/main/java/com/keylesspalace/tusky/components/search/SearchViewModel.kt
pandasoft0: Done

app/src/main/java/com/keylesspalace/tusky/components/search/SearchViewModel.kt
pandasoft0: It was left for further updates, but it's not useful now.
Updated

app/src/main/res/layout/activity_search2.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="utf-8"?>
+<androidx.constraintlayout.widget.ConstraintLayout xmlns:android="http://schemas.android.com/apk/res/android"
```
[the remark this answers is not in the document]
pandasoft0: Removed

pandasoft0:
- Tabs style updated
- I left the SFragment because all 3 fragments has the similar setup and functionality. Yes, i thought about something like a 
`SearchBasicFragment<T>: Fragment(), Injectable, LinkListener{}`
for a Hashtag and Account fragments, and SFragment for SearchStatusFragment, but most code will be the same and from another side - SFragment is lightweight.
As another solution - i think we may not to use the SFragment at all, but copy-paste some methods from SFragment to a SearchStatusFragment: SearchBascicFragment<...>. Please, let me know if this is a better solution and i will implement it. It's not a problem.

pandasoft0:
About error handling.
Search fragment may show the error not only on initial load but during a scroll also.
Also, it calls 3 different search request for every page.
Yes, we may show the elephant for initial load error, but i prefer to show it per fragment. And refresh content only for that fragment where user clicked the refresh button. So, app may return success on first requests, but error on load next data request for statuses and we don't need to refresh all pages because data on other pages are valid. Also, we must not repeat all 3 requests even if initial request failed for all 3 pages. If user wants to find a hashtag and search failed, he will click repeat on hashtag page and app will do a single request but will not to do 2 other unneeded requests...
But if you want to repeat all failed requests - not a problem for me. I will do.

app/src/main/java/com/keylesspalace/tusky/components/search/fragments/SearchStatusesFragment.kt
pandasoft0: Implemented. I decided to do a trick.
ViewModel stores loaded statuses at the separate list.
When user update the status - app updates the status at this list and invalidate DataSource. DataSource uses this precached list as initial load data. When user changes the search phrase, app clear this list and app uses the network request to load data.
Not sure that this is the best solution, but it allow to use standard paging libraries.

app/src/main/java/com/keylesspalace/tusky/components/search/adapter/SearchAccountsAdapter.kt
```diff
@@ -0,0 +1,68 @@
...
+                setupWithAccount(item)
+                setupLinkListener(linkListener)
+            }
+        }
+
+    }
+
+    public override fun getItem(position: Int): Account? {
+        return super.getItem(position)
+    }
```
connyduck: this override is unnecessary, it just calls super

app/src/main/java/com/keylesspalace/tusky/components/search/adapter/SearchHashtagsAdapter.kt
```diff
@@ -0,0 +1,63 @@
...
+            (holder as? HashtagViewHolder)?.apply {
+                setup(item.name, linkListener)
+            }
+        }
+
+    }
+
+    public override fun getItem(position: Int): HashTag? {
+        return super.getItem(position)
+    }
```
connyduck: unnecessary override

app/src/main/java/com/keylesspalace/tusky/components/search/adapter/SearchStatusesAdapter.kt
```diff
@@ -0,0 +1,67 @@
...
+        getItem(position)?.let { item ->
+            (holder as? StatusViewHolder)?.setupWithStatus(item.second, statusListener,
+                    mediaPreviewEnabled, showBotOverlay, animateAvatar)
+        }
+
+    }
+
+    public override fun getItem(position: Int): Pair<Status, StatusViewData.Concrete>? {
+        return super.getItem(position)
+    }
```
connyduck: unnecessary overrride

app/src/main/java/com/keylesspalace/tusky/entity/HashTag.kt
```diff
@@ -0,0 +1,8 @@
+package com.keylesspalace.tusky.entity
+
+import com.google.gson.annotations.SerializedName
+
+data class HashTag(
+        @field:SerializedName("name")
```
connyduck: `SerializedName` does nothing if the property has the same name anyway.

app/src/main/java/com/keylesspalace/tusky/components/search/fragments/SearchFragment.kt
```diff
@@ -0,0 +1,128 @@
...
+
+    private fun initAdapter() {
+        searchRecyclerView.addItemDecoration(DividerItemDecoration(searchRecyclerView.context, DividerItemDecoration.VERTICAL))
+        searchRecyclerView.layoutManager = LinearLayoutManager(searchRecyclerView.context)
+        adapter = createAdapter()
+        searchRecyclerView.adapter = adapter
+
+    }
+
+    protected fun showNoData(isEmpty: Boolean) {
```
connyduck: can be private

app/src/main/java/com/keylesspalace/tusky/components/search/fragments/SearchFragment.kt
```diff
@@ -0,0 +1,128 @@
...
+    private fun subscribeObservables() {
+        data.observe(viewLifecycleOwner, Observer {
+            adapter.submitList(it)
+        })
+
+        networkStateRefresh.observe(viewLifecycleOwner, Observer {
+            if (it == NetworkState.LOADING) {
+                searchProgressBar.show()
+            } else
+                searchProgressBar.hide()
```
connyduck: there is a extension function `visible` so you can do these checks shorter
`searchProgressBar.visible(it == NetworkState.LOADING)`

app/src/main/java/com/keylesspalace/tusky/components/search/fragments/SearchStatusesFragment.kt
```diff
@@ -0,0 +1,429 @@
...
+        mentionedUsernames.remove(loggedInUsername)
+        val intent = ComposeActivity.IntentBuilder()
+                .inReplyToId(inReplyToId)
+                .replyVisibility(replyVisibility)
+                .contentWarning(contentWarning)
+                .mentionedUsernames(mentionedUsernames)
+                .replyingStatusAuthor(actionableStatus.account.localUsername)
+                .replyingStatusContent(actionableStatus.content.toString())
+                .build(context)
+        activity!!.startActivity(intent)
```
connyduck:
```suggestion
        requireActivity.startActivity(intent)
```

app/src/main/java/com/keylesspalace/tusky/components/search/fragments/SearchStatusesFragment.kt
```diff
@@ -0,0 +1,429 @@
...
+        val accountId = status.actionableStatus.account.id
+        val accountUsername = status.actionableStatus.account.username
+        val content = status.actionableStatus.content
+        val statusUrl = status.actionableStatus.url
+        val accounts = viewModel.getAllAccountsOrderedByActive()
+        var openAsTitle: String? = null
+
+        val loggedInAccountId = viewModel.activeAccount?.accountId
+
+        val popup = PopupMenu(context!!, view)
```
connyduck: `view.context` is always non-null

app/src/main/java/com/keylesspalace/tusky/components/search/fragments/SearchStatusesFragment.kt
```diff
@@ -0,0 +1,429 @@
...
+                    .setNegativeButton(android.R.string.cancel, null)
+                    .show()
+        }
+    }
+
+    private fun showConfirmEditDialog(id: String, position: Int, status: Status) {
+        if (activity == null) {
+            return
+        }
+        AlertDialog.Builder(activity!!)
```
connyduck: `activity?.let{` makes this shorter and you don't have to use `!!`

app/src/main/java/com/keylesspalace/tusky/components/search/adapter/SearchStatusesAdapter.kt
noln: This one is required in order to change visibility of the method from `protected` to `public` , used at: 

- [SearchStatusesFragment.kt:78](https://github.com/roma-apps/roma-android/blob/04d36ff1cc58451522ca13bb80c0f27f417839a8/app/src/main/java/com/keylesspalace/tusky/components/search/fragments/SearchStatusesFragment.kt#L78), 

- [SearchStatusesFragment.kt:84](https://github.com/roma-apps/roma-android/blob/04d36ff1cc58451522ca13bb80c0f27f417839a8/app/src/main/java/com/keylesspalace/tusky/components/search/fragments/SearchStatusesFragment.kt#L84), 

- [SearchStatusesFragment.kt:90](https://github.com/roma-apps/roma-android/blob/04d36ff1cc58451522ca13bb80c0f27f417839a8/app/src/main/java/com/keylesspalace/tusky/components/search/fragments/SearchStatusesFragment.kt#L90), 

- [SearchStatusesFragment.kt:96](https://github.com/roma-apps/roma-android/blob/04d36ff1cc58451522ca13bb80c0f27f417839a8/app/src/main/java/com/keylesspalace/tusky/components/search/fragments/SearchStatusesFragment.kt#L96).

noln:
@connyduck could you provide the video showing the weird retry behaviour that you mention above please? I've not been able to replicate it.

connyduck:
Do you have a new GitHub account?

connyduck:
Here is a video https://connyduck.at/device-2019-07-07-143455.mp4
Notice how only the first tab shows data after going back into WiFi and clicking retry.

noln:
@connyduck Nope, different person. I'm leading on Roma for Android now, so aiming to tidy up and close out any remnant PRs from the fork before starting any new ones.

noln:
@connyduck 

- Themes and colours are now showing as requested.
- That retry bug is fixed too.
- I've added pull-to-refresh for each tab too.

One thing to note is that searching hashtags results in infinite paging, full of the same result, which looks like a bug. However that's actually what the V2 API returns from server-side when the `offset` 
 parameter is changed, so the client-side code is implemented correctly. Therefore the issue will go away when the server returns the correct result.

I think that satisfies all of the prior comments, but would welcome further review and feedback.

noln:
Good point, hadn't thought of that! A few of approaches come to mind:

1. Show a couple of tappable TextViews that tell the user that there are other results on the other tabs, which takes the user straight there on tap.
2. Snackbar to do the same as above.
3. Update the tab titles to show the number of results returned ("99+" for large numbers of results), so the user can see that there are results loaded on the other tabs.

connyduck:
I merge this now so more people can try it out, lets see. 

> Update the tab titles to show the number of results returned

Would be nice, but I think we don't get this info from the api?

Thx again

connyduck:
Oh no, found another bug: The account list does not stop loading and shows the same account over and over again:

device-2019-07-19-201534

@noln

noln:
@connyduck 😵

Looking at that, it seems that it's instance-specific so perhaps a server-side issue, as it doesn't occur on the (admittedly Pleroma) instances I've tried with the head of `master`.

Could you raise an issue against me for that, with details of which instance version you're using please?

screenshot-1563560589989

connyduck:
Ok, i tried debugging this and it looks like it is a bug in Mastodon and only happens when searching for ones own account 🤣 
Like, I can send in offset=100 and will still get back my own account

connyduck:
https://github.com/tootsuite/mastodon/issues/11365

PR #1331, opened 2019-06-20
Fix build warnings

connyduck:
These warnings were occuring on every build, this PR fixes them

`w: /Users/connyduck/Documents/AndroidApps/Tusky/app/src/main/java/com/keylesspalace/tusky/FiltersActivity.kt: (178, 24): Parameter 'v' is never used`

`warn: multiple substitutions specified in non-positional format; did you mean to add the formatted="false" attribute?`

```
/Users/connyduck/Documents/AndroidApps/Tusky/app/src/main/java/com/keylesspalace/tusky/adapter/StatusBaseViewHolder.java:800: warning: non-varargs call of varargs method with inexact argument type for last parameter;
            return context.getString(R.string.description_poll, args);
                                                                ^
  cast to Object for a varargs call
  cast to Object[] for a non-varargs call and to suppress this warning
```

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/adapter/StatusBaseViewHolder.java
+++ b/app/src/main/java/com/keylesspalace/tusky/adapter/StatusBaseViewHolder.java
@@ -783,7 +783,7 @@ public abstract class StatusBaseViewHolder extends RecyclerView.ViewHolder {
         if (poll == null) {
             return "";
         } else {
-            CharSequence[] args = new CharSequence[5];
+            Object[] args = new CharSequence[5];
             List<PollOption> options = poll.getOptions();
             for (int i = 0; i < args.length; i++) {
                 if (i < options.size()) {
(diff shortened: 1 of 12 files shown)


app/src/main/res/values-ar/strings.xml
```diff
@@ -438,7 +438,6 @@
 <string name="compose_preview_image_description">إجراءات على الصورة %s</string>
 
     <string name="caption_notoemoji">حزمة الإيموجي الحالية لـ غوغل</string>
-    <string name="description_poll">استطلاع رأي بالخيارات: %s, %s, %s, %s; %s</string>
```
connyduck: unfortunately I do not know how these have to look correctly in Arabic 
cc @BoFFire

BoFFire: I'm checking this string. Thanks.

app/src/main/java/com/keylesspalace/tusky/adapter/StatusBaseViewHolder.java
```diff
@@ -783,7 +783,7 @@ private CharSequence getPollDescription(Context context,
         if (poll == null) {
             return "";
         } else {
-            CharSequence[] args = new CharSequence[5];
+            Object[] args = new CharSequence[5];
```
charlag: Huh?

connyduck: Makes the warning go away. But not pretty at all, do you have a better idea?

PR #1333, opened 2019-06-21
Additional Gab Domains

wiresharklain:
I would add these domains into the defence code. Looking other the mother PR, I have no idea if subdomains are automatically included, so I added the development instance to be on the safe side.

I would like to state that gab.io is unconfirmed at this point, but I will check.

You're also welcome to tell me if I'm being overkill or anything! Just thinking of ideas!

### DIFF
--- a/app/src/main/res/values/donottranslate.xml
+++ b/app/src/main/res/values/donottranslate.xml
@@ -107,6 +107,8 @@
     <string-array name="rick_roll_domains" translatable="false">
         <item>gab.com</item>
         <item>gab.ai</item>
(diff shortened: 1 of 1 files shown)


connyduck:
Thx. Yes its overkill, subdomains are handled by the code.

wiresharklain:
Got it. Also, gab.io seems to be unrelated, as it seems to be registed to a group called "Gaia AG".

Without any further evidence to suppport .io as one. I will close the PR. It can be added later, if it turns out to be the case.

PR #1343, opened 2019-06-23
fix vanishing toots bug when refreshing

### LINKED ISSUE #1228: Refresh issue when existing refresh running
When an existing background refresh is running (eg: just opened app, timed background refresh), and the user pulls down to refresh, you get issues with toots disappearing from the Home timeline. Doesn't always happen though, but the below should allow someone to reproduce somewhat reliably.

1. Start Tusky and wait for an update to finish, so it should have all the latest toots.
2. Quit Tusky (force kill if needed so it's not running in the background).
3. Log into the web version of the Mastodon instance and wait till 2-3 new toots have appeared in the Home timeline.
4. Start Tusky and within 2-3 seconds of the UI loading, pull down to refresh.
5. Compare this list of toots with the Mastodon web version. If they are different, forcing a change of the theme seems to restore the missing toots (eg: change from Black to Dark to Black again).

Notes:
- I suspect a workaround for this would be to not allow multiple refreshes to fire if a refresh is already happening (eg: some sort of lock around the refresh function). A status icon that shows when a refresh is happening would probably help as well (at least with further diagnosis).
- The issue can happen if you manage to pull down refresh as a background timed refresh has just started as well (not just at startup).

* * * *
- Tusky Version: 6.1
- Android Version: 8.0
- Android Device: Samsung SM-G930F
- Mastodon instance (if applicable): social.chinwag.org

- [*] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

connyduck:
closes #1228

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/fragment/TimelineFragment.java
+++ b/app/src/main/java/com/keylesspalace/tusky/fragment/TimelineFragment.java
@@ -284,6 +284,7 @@ public class TimelineFragment extends SFragment implements
                     }
 
                     this.updateCurrent();
+                    this.loadAbove();
                 });
     }
 
@@ -292,7 +293,7 @@ public class TimelineFragment extends SFragment implements
         if (this.statuses.isEmpty()) {
             topId = null;
         } else {
-            topId = CollectionsKt.first(statuses, Either::isRight).asRight().getId();
+            topId = CollectionsKt.first(this.statuses, Either::isRight).asRight().getId();
         }
         this.timelineRepo.getStatuses(topId, null, null, LOAD_AT_ONCE,
                 TimelineRequestMode.NETWORK)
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/keylesspalace/tusky/fragment/TimelineFragment.java
```diff
@@ -305,22 +306,32 @@ private void updateCurrent() {
                             if (!statuses.isEmpty()) {
                                 filterStatuses(statuses);
 
-                                // Working around a bug when Mastodon API doesn't return the first
-                                // status because of string "id < maxId". Hacking with ID doesn't
-                                // help.
-                                if (!this.statuses.isEmpty()) {
-                                    Either<Placeholder, Status> firstOld = this.statuses.get(0);
-                                    this.statuses.clear();
```
connyduck: So this line basically was the problem: it removes all statuses. But there might have been a concurrent top load going on in the meantime that inserted newer stauses. I fixed it by clearing only statuses that are older than the `topId` at the moment the load was started.

app/src/main/java/com/keylesspalace/tusky/fragment/TimelineFragment.java
```diff
@@ -284,6 +284,7 @@ private void tryCache() {
                     }
 
                     this.updateCurrent();
+                    this.loadAbove();
```
connyduck: Now that concurrent refreshes and top loads are possible, lets to them in parallel.

app/src/main/java/com/keylesspalace/tusky/fragment/TimelineFragment.java
```diff
@@ -1171,7 +1183,7 @@ private void updateStatuses(List<Either<Placeholder, Status>> newStatuses, boole
 
     private void removeConsecutivePlaceholders() {
         for (int i = 0; i < statuses.size() - 1; i++) {
-            if (!statuses.get(i).isRight() && !statuses.get(i + 1).isRight()) {
+            if (statuses.get(i).isLeft() && statuses.get(i + 1).isLeft()) {
```
connyduck: using `isLeft` is better than `!isRight`

app/src/main/java/com/keylesspalace/tusky/fragment/TimelineFragment.java
Tak: Is the mastodon bug mentioned in the comment still addressed now?

connyduck: I am pretty sure it is, but I will check again

charlag: We should consider using new API parameters finally but not in this PR ofc

app/src/main/java/com/keylesspalace/tusky/fragment/TimelineFragment.java
```diff
@@ -305,22 +306,32 @@ private void updateCurrent() {
...
-                                    this.statuses.add(firstOld);
-                                } else {
-                                    this.statuses.clear();
+
+                                if (!this.statuses.isEmpty() && topId != null) {
+                                    // clear old cached statuses
+                                    Iterator<Either<Placeholder, Status>> iterator = statuses.iterator();
+                                    while (iterator.hasNext()) {
+                                        Either<Placeholder, Status> item = iterator.next();
+                                        if(item.isRight()) {
```
charlag: You could shorten it if you put id into the variable and left only one comparison

charlag: Could also find the first newer one and delete everything older but not sure if there's any performance/readability benefit

app/src/main/java/com/keylesspalace/tusky/fragment/TimelineFragment.java
mal0ki: :rofl:

connyduck:
Tested it again, everything seems to work correctly

PR #1284, opened 2019-05-25
Show media descriptions in timeline when previews are disabled.

charlag:
3 more text views? Yes. Is it worth it? I think absolutely.
End goal (and original branch) is to automate it to be on/off depending on network state but I re-did just UI part for now because it's the most benefit (and because I couldn't rebase my older one lol).

Example: https://birb.site/@charlag/102158167894189708

PR #1344, opened 2019-06-23
Use cached preview as thumbnail in ViewImageFragment, fix #1267

### LINKED ISSUE #1267: Progressively enhance images on viewing
Mastodon offers image thumbnails that Tusky already loads when displaying posts with images on the TL.
On a slow connection loading the full image might take some time, so when touching the image to take a closer look, one might have to wait a few seconds in order to see the image again. All this while the thumbnail is already available.

I would propose, when viewing an image in full, to display the thumbnail instead of a loading screen, and then replace it with the full resolution image once loaded.
A similar thing could be done with video thumbnails.

* * * *
- Tusky Version: 7.0

charlag:
It's not really finished yet but I want to ask for feedback.
Basically when the full image is not in cache, it tried out preview from cache and then loads the complete image.
There are couple of things:
1. Maybe progress bar on top would work better then. We could even show progress but we need to play a little with OkHttp: https://learnpainless.com/android/show-progress-while-loading-images-with-glide
2. I've tried to do the same with video but... let's say it wasn't really successful

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/ViewMediaActivity.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/ViewMediaActivity.kt
@@ -30,7 +30,6 @@ import android.net.Uri
 import android.os.Bundle
 import android.os.Environment
 import android.transition.Transition
-import android.transition.TransitionListenerAdapter
 import android.util.Log
 import android.view.Menu
 import android.view.MenuItem
@@ -62,6 +61,8 @@ import java.io.FileOutputStream
 import java.io.IOException
 import java.util.*
 
+typealias ToolbarVisibilityListener = (isVisible: Boolean) -> Unit
+
 class ViewMediaActivity : BaseActivity(), ViewImageFragment.PhotoActionsListener {
     companion object {
         private const val EXTRA_ATTACHMENTS = "attachments"
@@ -84,23 +85,16 @@ class ViewMediaActivity : BaseActivity(), ViewImageFragment.PhotoActionsListener
         }
     }
 
+    var isToolbarVisible = true
+        private set
+
     private var attachments: ArrayList<AttachmentViewData>? = null
-
-    private var toolbarVisible = true
-    private val toolbarVisibilityListeners = ArrayList<ToolbarVisibilityListener>()
-
-    interface ToolbarVisibilityListener {
-        fun onToolbarVisiblityChanged(isVisible: Boolean)
-    }
+    private val toolbarVisibilityListeners = mutableListOf<ToolbarVisibilityListener>()
 
     fun addToolbarVisibilityListener(listener: ToolbarVisibilityListener): Function0<Boolean> {
         this.toolbarVisibilityListeners.add(listener)
-        listener.onToolbarVisiblityChanged(toolbarVisible)
+        listener(isToolbarVisible)
         return { toolbarVisibilityListeners.remove(listener) }
-    }
-
-    fun isToolbarVisible(): Boolean {
-        return toolbarVisible
     }
 
     override fun onCreate(savedInstanceState: Bundle?) {
@@ -113,7 +107,10 @@ class ViewMediaActivity : BaseActivity(), ViewImageFragment.PhotoActionsListener
         attachments = intent.getParcelableArrayListExtra(EXTRA_ATTACHMENTS)
         val initialPosition = intent.getIntExtra(EXTRA_ATTACHMENT_INDEX, 0)
 
-        val adapter = if (attachments != null) {
+        // Adapter is actually of existential type PageAdapter & SharedElementsTransitionListener
+        // but it cannot be expressed and if I don't specify type explicitly compilation fails
+        // (probably a bug in compiler)
+        val adapter: PagerAdapter = if (attachments != null) {
             val realAttachs = attachments!!.map(AttachmentViewData::attachment)
             // Setup the view pager.
             ImagePagerAdapter(supportFragmentManager, realAttachs, initialPosition)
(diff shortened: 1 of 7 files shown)


charlag:
Yeah, I noticed animation!
Thanks for checking.
I did not notice that overload, at least not in the IDE

connyduck:
Wait, looks like we broke animation somewhere else, its not on master. o_0
I will check, this is a release blocker imho.

charlag:
Okay, this was also my thought :D

connyduck:
`git bisect` says the bug was introduced here: https://github.com/tuskyapp/Tusky/pull/1284

Its probably because a `null` view is passed to `listener.onViewMedia`. Do you want to fix it or should I?

charlag:
If you could fix it, it would be great, I'm not sure when I'll have time to work on this yet

charlag:
@connyduck I managed to get it working but it still fails sometimes - maybe previews are evicted from cache, idk.
I would be glad if someone tested this.

app/src/main/java/com/keylesspalace/tusky/ViewMediaActivity.kt
```diff
@@ -125,7 +125,7 @@ class ViewMediaActivity : BaseActivity(), ViewImageFragment.PhotoActionsListener
             AvatarImagePagerAdapter(supportFragmentManager, avatarUrl)
         }
 
-        viewPager.adapter = adapter
+        viewPager.adapter = adapter as PagerAdapter
```
connyduck: no cast needed

app/src/main/java/com/keylesspalace/tusky/pager/ImagePagerAdapter.kt
```diff
@@ -1,20 +1,26 @@
...
 class ImagePagerAdapter(
         fragmentManager: FragmentManager,
         private val attachments: List<Attachment>,
         private val initialPosition: Int
-) : FragmentStatePagerAdapter(fragmentManager) {
+) : FragmentStatePagerAdapter(fragmentManager), SharedElementTransitionListener {
+
+    private var primaryItem: ViewMediaFragment? = null
+
+    override fun setPrimaryItem(container: ViewGroup, position: Int, `object`: Any) {
```
connyduck: Can you use a different variable name so the backticks are not needed please?

app/src/main/java/com/keylesspalace/tusky/fragment/ViewImageFragment.kt
```diff
@@ -131,49 +132,74 @@ class ViewImageFragment : ViewMediaFragment() {
...
+            if (isCacheRequest && !isThumnailRequest) photoActionsListener.onBringUp()
+            // Hide progress bar only on fail request from internet
+            if (!isCacheRequest) progressBar?.hide()
             return false
         }
 
-        override fun onResourceReady(resource: Drawable?, model: Any?, target: Target<Drawable>?, dataSource: DataSource?, isFirstResource: Boolean): Boolean {
-            progressBar?.hide() //Always hide the progress bar on success
+        override fun onResourceReady(resource: Drawable?, model: Any?, target: Target<Drawable>?,
+                                     dataSource: DataSource?, isFirstResource: Boolean): Boolean {
```
connyduck: I think the params in this override can all be non-null. Makes the subsequent code more readable. Glide does not use `@NonNull` annotations in Java so Kotlin can't pick it up, but they always document nullable stuff.

app/src/main/java/com/keylesspalace/tusky/fragment/ViewImageFragment.kt
```diff
@@ -131,49 +132,74 @@ class ViewImageFragment : ViewMediaFragment() {
...
      */
-    private inner class ImageRequestListener(private val isCacheRequest: Boolean) : RequestListener<Drawable> {
-        override fun onLoadFailed(e: GlideException?, model: Any?, target: Target<Drawable>?, isFirstResource: Boolean): Boolean {
-            if (isCacheRequest) //Complete the transition on failed image from cache
-                completeTransition()
-            else
-                progressBar?.hide() //Hide progress bar only on fail request from internet
+    private inner class ImageRequestListener(
+            private val isCacheRequest: Boolean,
+            private val isThumnailRequest: Boolean) : RequestListener<Drawable> {
```
connyduck: Can't you use `isFirstResource` instead of this custom boolean? Shouldn't it be the same?

charlag: If there is a thumbnail. There might be none.

charlag:
I"ve changed some small parts but there are not so many things which seem very Java-ish to me, it's a regular, ugly Android UI code

app/src/main/java/com/keylesspalace/tusky/ViewMediaActivity.kt
charlag: Interestingly enough, the first time I wrote it it was fine (it resolves to existential - union of two types which is not representable in Kotlin because it's not representable in Java I guess) but when I was editing code before sending I got errors. Now I don't get them again. Weird, maybe something didn't conform to something.

Tak:
> Do you mind cleaning up the legacy part of this code a little? Some parts look like they were created by the Java->Kotlin converter.

Those are probably the parts I converted 😀

charlag:
Pushed again because of the compilation errors. Sorry!

connyduck:
Tested again, looks good. Thanks <3

charlag:
Thank you! 💚

PR #1347, opened 2019-06-24
Add exited.eu to list of gab domains

Tak:
Screenshot from 2019-06-24 08-48-17

### DIFF
--- a/app/src/main/res/values/donottranslate.xml
+++ b/app/src/main/res/values/donottranslate.xml
@@ -107,6 +107,7 @@
     <string-array name="rick_roll_domains" translatable="false">
         <item>gab.com</item>
         <item>gab.ai</item>
+        <item>exited.eu</item>
     </string-array>
 
     <string name="rick_roll_url">https://www.youtube.com/watch?v=dQw4w9WgXcQ</string>


m4sk1n:
Can you please get something funnier than just a rickroll for my domain? 🤔

charlag:
How does a brick sou... *cough* I mean, wait and see, we prepare something.

Laurelai:
> How does a brick sou... _cough_ I mean, wait and see, we prepare something.

[If you need some ideas let me know](https://i.imgur.com/7zEOPFy.jpg)

connyduck:
It is an edgelord instance with only 2 users, thats quite a difference to gab with potentially hundred thousand users. Do we now block every domain running gab? I would be in for that, but it probably means a lot of additional work for us.

charlag:
That's a good point, I think we can't tell yet how big instances will be. I'm okay with setting some threshold.

m4sk1n:
Please, take a look at rules of my instance, look for example that I have actually made a post with block recommendation for a homophobic instance (and reported them to get.lgbt), I just wanted to make a ‘safe-space’ Gab instance for people who wanted to test gab-social as a software and the way you, Fedi people handle it is, well, hard to find a word to describe…

rnickson:
if this gets merged, I'll either fork this app or switch to Fedilab

Laurelai:
If this gets merged ill make it the officially reccomended app for my instance.

m4sk1n:
@Laurelai (you probably blocked me), what’s wrong with my instance? Why it needs to be blocked?

BluRaf:
I'd recommend finding better ways to spending your time instead of taking such obvious bait.

PS: Death and harm threats are punishable by law.

m4sk1n:
@BluRaf, it (the logo) was intended to be a bait for a few minutes I wanted to bait gab users (the way some mastodon users reacted was what I expected to see from gabbers), then I realized it’s actually a good idea to have instance like this, so it was actually a bait just for a few minutes ;)

Laurelai:
I really doubt with all of the support you have given gab in public you were trying to bait their users.

m4sk1n:
So, I have given a lot of support to Mastodon (translating front-end+back-end, joinmastodon, documentation and their blog), does it mean I’m disallowed to bait you, as a Mastodon user, @Laurelai? 🤔

Laurelai:
We all know what side you are on, dont try to gaslight us.

m4sk1n:
So, I’m the only one who don’t know :/ Could you tell me what side I’m on?

charlag:
@Laurelai @m4sk1n could y'all stop please? I don't like baiting and I don't believe in the good intent here but it's totally not fun to get all these notifications and I don't think that's the place.
Thanks.

m4sk1n:
Ok, this discussion has absolutely no point, if anyone wants to call me a nazi without distracting others here, dm me on matrix: @m4sk1n:matrix.org, discord: mkljczk#7643 or even on a f@#king facebook messenger: @mkljczk

mal0ki:
No matter how funny it is, I don't think PRs for rickrolling/blocking more instances should be made without discussions between the team first.
As we know gab is already known to cause harm, that's the biggest reason for us to add it in the first place.

We will also not have baiting and attacks of each other in the comments of our issues.

ealgase:
Should be noted, cyzed.com is another Gab Social instance

05nelsonm:
Any news on someone forking this censored app to give users the freedom to do as they wish? Currently running an older version that does not include this PR, but would like to stay up to date.

Thanks.

Issue #1349, opened 2019-06-24
Filters do not work when looking at profiles or conversations


aCuteLittleBox:
when viewing a conversation/thread, or while viewing someone's profile and looking through their posts, Tusky does not apply any filters. This is a dangerous oversight for those who rely on filters for their own safety and it is important it is correctly quickly

* * * *
- Tusky Version: 7.1
- Android Version: LineageOS 15.1 (Android 8.1)
- Android Device: Samsung Galaxy S7
- Mastodon instance (if applicable):

- [✅]I searched or browsed the repo’s other issues to ensure this is not a duplicate.

connyduck:
🤔 indeed, thx for noticing
We are implementing the Mastodon filter Api, and it specifies 4 values for a filter context `home`, `notifications`, `public`, `thread`. So if Tusky is not filtering on threads thats our mistake, profiles on the other hand seem to be not meant to be filtered by the Mastodon api.

connyduck:
OK we are now filtering correctly in threads. For the filtering in profiles we need a change in the Mastodon api.

PR #1351, opened 2019-06-25
Apply conversation filters to threads. Addresses #1349

Tak:
- Extract most of the filtering logic into `SFragment`
- Provide specialized overrides in `TimelineFragment` and `ViewThreadFragment`


connyduck:
Awesome! I just wonder if we maybe should cache the filters somehow so they are not reloaded every time an thread is viewed (does not have to be in this PR)

Tak:
Makes sense, we just need to decide whether/how to invalidate them in that case

connyduck:
What about invalidating them when they are changed in the app, and reload when they app is started or accounts are switched? Should be fine for most usecases, users that change filters in another client hopefully think about restarting the app.

Tak:
Ok, now the filters are cached globally unless modified locally

connyduck:
Nice! Now you just need to clear the filter cache when accounts are changed in the app.

Tak:
🤦‍♂

Tak:
Ok, account switching should be covered now :smiley:

connyduck:
awesome, thx!

PR #1352, opened 2019-06-29
fix LoginActivity created twice on unauthenticated appstart

connyduck:
The LoginActivity was started twice when opening the app while not being logged in because both the BaseActivity and the SplashActivity were redirecting. I only noticed because I pressed back to close the app and there was another login.


connyduck:
Wait a second, actually a better fix would be to have SplashActivity not inheriting from BaseActivity because that initialization stuff in BaseActivity.onCreate definitely does not need to run for SplashActivity

Issue #1346, opened 2019-06-24
American English translation


toastal:
I'd like to see an American English translation support -- as the word 'favourite' irrationally grinds my gears. Can I contribute this? It seems the translation system is homebrewed? Can it support this kind of localization to en_US when it seems there's just language codes and not region codes?

* * * *
- Tusky Version: 7.1
- Android Version: 9
- Android Device: OnePlus 5
- Mastodon instance (if applicable): cybre.space

- [ x ] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

akarshanbiswas:
Here: https://weblate.tusky.app/

toastal:
@akarshanbiswas I can't create a new language though to translate to.

PR #1358, opened 2019-07-01
add support for american/british english translations

connyduck:
closes https://github.com/tuskyapp/Tusky/issues/1346

I first tried having british as the default and overriding with `en-rUS`, but that did not work, my device always showed American, no matter the settings, so i now did it the other way round

### DIFF
--- a/app/src/main/res/values-en-rGB/strings.xml
+++ b/app/src/main/res/values-en-rGB/strings.xml
@@ -1,0 +1,23 @@
+<resources>
+
+    <string name="title_favourites">Favourites</string>
+    <string name="notification_favourite_format">%s favourited your toot</string>
+    <string name="action_favourite">Favourite</string>
+    <string name="action_unfavourite">Remove favourite</string>
+    <string name="action_view_favourites">Favourites</string>
+    <string name="action_open_faved_by">Show favourites</string>
+    <string name="pref_title_notification_filter_favourites">my posts are favourited</string>
+    <string name="notification_favourite_name">Favourites</string>
+    <string name="notification_favourite_description">Notifications when your toots get marked as favourite</string>
+
+    <string name="description_status_favourited">Favourited</string>
+
+    <string name="error_authorization_unknown">An unidentified authorisation error occurred.</string>
+    <string name="error_authorization_denied">Authorisation was denied.</string>
+
+    <plurals name="favs">
+        <item quantity="one">&lt;b>%1$s&lt;/b> Favourite</item>
+        <item quantity="other">&lt;b>%1$s&lt;/b> Favourites</item>
+    </plurals>
+
+</resources>
--- a/app/src/main/res/values-en-rGB/strings.xml
+++ b/app/src/main/res/values-en-rGB/strings.xml
@@ -12,6 +12,8 @@
 
     <string name="description_status_favourited">Favourited</string>
 
+    <string name="title_favourited_by">Favourited by</string>
+
     <string name="error_authorization_unknown">An unidentified authorisation error occurred.</string>
     <string name="error_authorization_denied">Authorisation was denied.</string>
 
--- a/app/src/main/res/values/donottranslate.xml
+++ b/app/src/main/res/values/donottranslate.xml
@@ -36,12 +36,12 @@
 
     <string-array name="language_entries">
         <item>@string/system_default</item>
-        <item>British English</item>
         <item>Català</item>
         <item>Čeština</item>
         <item>Cymraeg</item>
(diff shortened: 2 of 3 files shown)


app/src/main/res/values/donottranslate.xml
```diff
@@ -36,6 +36,7 @@
 
     <string-array name="language_entries">
         <item>@string/system_default</item>
+        <item>British English</item>
```
connyduck: Should we call the other English American English?

app/src/main/res/values-en-rGB/strings.xml
```diff
@@ -0,0 +1,23 @@
...
+    <string name="action_view_favourites">Favourites</string>
+    <string name="action_open_faved_by">Show favourites</string>
+    <string name="pref_title_notification_filter_favourites">my posts are favourited</string>
+    <string name="notification_favourite_name">Favourites</string>
+    <string name="notification_favourite_description">Notifications when your toots get marked as favourite</string>
+
+    <string name="description_status_favourited">Favourited</string>
+
+    <string name="error_authorization_unknown">An unidentified authorisation error occurred.</string>
+    <string name="error_authorization_denied">Authorisation was denied.</string>
```
connyduck: My dictionary says in BE both authorisation and authorization are ok, but not if one of them is preferable.

connyduck:
Did I miss some strings that are different in BE/AE?

app/src/main/res/values/donottranslate.xml
mlc: I'd go with
* English (UK)
* English (US)
so that they show up under `E` where you expect to find English, but this is a matter of opinion, not correctness

connyduck: Thats a good idea, thx!

app/src/main/res/values-en-rGB/strings.xml
ryliejamesthomas: 'authorisation' is preferred in BE :) 's' over 'z' in these kinds of situations, generally.

app/src/main/res/values-en-rGB/strings.xml
```diff
@@ -0,0 +1,23 @@
...
+
+    <plurals name="favs">
```
ryliejamesthomas: Would usually be 'faves'.

ryliejamesthomas: ..not sure if this is just an internal thing that doesn' actually matter though

app/src/main/res/values-en-rGB/strings.xml
Tak:
```suggestion
    <string name="error_authorization_denied">Authorisation was denied.</string>
    <string name="title_licenses">Licences</string>
    
    <string name="about_tusky_license">Tusky is free and open-source software.
        It is licensed under the GNU General Public License Version 3.
You can view the licence here: https://www.gnu.org/licenses/gpl-3.0.en.html</string>
    <string name="report_description_remote_instance">The account is from another server. Send an anonymised copy of the report there as well?</string>
```

app/src/main/res/values-en-rGB/strings.xml
```diff
@@ -0,0 +1,23 @@
...
+
+    <string name="title_favourites">Favourites</string>
+    <string name="notification_favourite_format">%s favourited your toot</string>
+    <string name="action_favourite">Favourite</string>
+    <string name="action_unfavourite">Remove favourite</string>
+    <string name="action_view_favourites">Favourites</string>
+    <string name="action_open_faved_by">Show favourites</string>
+    <string name="pref_title_notification_filter_favourites">my posts are favourited</string>
+    <string name="notification_favourite_name">Favourites</string>
+    <string name="notification_favourite_description">Notifications when your toots get marked as favourite</string>
```
Tak:
```suggestion
    <string name="notification_favourite_description">Notifications when your toots get marked as favourite</string>
    <string name="title_favourited_by">Favourited by</string>
```

app/src/main/res/values-en-rGB/strings.xml
Tak: (I didn't touch `GNU General Public License` nor `Apache License` since they're the proper names of the licenses)

app/src/main/res/values-en-rGB/strings.xml
connyduck: its just an internal thing, no user will see it

connyduck:
Thank you all for your help!

PR #1361, opened 2019-07-01
Update en-GB strings.xml

committomemory:
Just one additional change - anonymize (US) to anonymise (UK)


connyduck:
awesome, thx!

PR #1369, opened 2019-07-05
[WIP] Inline reply/boost/favourite counters

noln:
Adds counters next to reply, boost and favourite icons on posts.

As8KAjyjEy


app/src/main/java/com/keylesspalace/tusky/adapter/StatusBaseViewHolder.java
```diff
@@ -674,6 +685,18 @@ protected void setupWithStatus(StatusViewData.Concrete status,
...
 
+    private void setupCounters(StatusViewData.Concrete status) {
+        if (countFavorites!=null)
+            countFavorites.setText(status.getFavouritesCount()>0?String.format(Locale.getDefault(),"%d",status.getFavouritesCount()):"");
+
+        if (countReposts!=null)
+            countReposts.setText(status.getReblogsCount()>0?String.format(Locale.getDefault(),"%d",status.getReblogsCount()):"");
+
+        if (countReplies!=null)
+            countReplies.setText(status.getRepliesCount()>0?String.format(Locale.getDefault(),"%d",status.getRepliesCount()):"");
```
Tak: The web UI does "0", "1", "1+" for replies - do we want to do the same?
What's the plan for dealing with many-digit counts on these?

app/src/main/java/com/keylesspalace/tusky/di/NetworkModule.kt
```diff
@@ -77,7 +77,7 @@ class NetworkModule {
                 .apply {
                     addInterceptor(InstanceSwitchAuthInterceptor(accountManager))
                     if (BuildConfig.DEBUG) {
-                        addInterceptor(HttpLoggingInterceptor().setLevel(HttpLoggingInterceptor.Level.BASIC))
+                        addInterceptor(HttpLoggingInterceptor().setLevel(HttpLoggingInterceptor.Level.BODY))
```
Tak: Is this part of the counter functionality?

noln: Is not, good spot, I'll remove it. Apologies, taken this branch on from a previous dev and it was mixed in with some other features. Thought I'd untangled it all, apparently missed the odd shred.

app/src/main/java/com/keylesspalace/tusky/adapter/StatusBaseViewHolder.java
noln: This is how it works in the Roma fork so carried over the functionality, but open to suggestions.

noln:
@Tak it's always-on in this implementation, but could be switchable in the view layer if that's something that @connyduck is firm on. I've WIP'ed the PR pending feedback.

JuanjoSalvador:
That's pretty cool!

charlag:
I think we already had such a PR before?
#885
For the context

app/src/main/java/com/keylesspalace/tusky/fragment/ViewThreadFragment.java
```diff
@@ -612,32 +628,63 @@ private void handleStatusComposedEvent(StatusComposedEvent event) {
 
         if (eventStatus.getInReplyToId().equals(thisThreadsStatusId)) {
             insertStatus(eventStatus, statuses.size());
+            addReplyCount(0, 1);
```
connyduck: This looks like a bug? The detailed status is not guaranteed to be the first one in the list

app/src/main/java/com/keylesspalace/tusky/adapter/StatusBaseViewHolder.java
connyduck: We use `numberFormat.format` elsewhere, i prefer that
I actually like showing the nuber more than  "1" "1+" .

PR #1374, opened 2019-07-08
fix timeline reloading and favs/boosts/polls showing up at wrong pos

connyduck:


app/src/main/java/com/keylesspalace/tusky/db/TimelineDao.kt
```diff
@@ -74,12 +74,12 @@ AND
     abstract fun removeAllPlaceholdersBetween(account: Long, maxId: String, sinceId: String)
 
     @Query("""UPDATE TimelineStatusEntity SET favourited = :favourited
-WHERE timelineUserId = :accountId AND (serverId = :statusId OR reblogServerId - :statusId)""")
+WHERE timelineUserId = :accountId AND (serverId = :statusId OR reblogServerId = :statusId)""")
```
connyduck: what was the minus doing here :no_mouth:

app/src/main/java/com/keylesspalace/tusky/fragment/TimelineFragment.java
```diff
@@ -301,20 +301,21 @@ private void updateCurrent() {
                             if (!statuses.isEmpty()) {
                                 filterStatuses(statuses);
 
-
                                 if (!this.statuses.isEmpty() && topId != null) {
                                     // clear old cached statuses
-                                    Iterator<Either<Placeholder, Status>> iterator = statuses.iterator();
+                                    Iterator<Either<Placeholder, Status>> iterator = this.statuses.iterator();
```
connyduck: yeah... this code never actually worked because it ran on the wrong list 🙄

connyduck:
So here is what I think happened: The bug in the sql statements was there all the time, but nobody noticed, because the timeline reloading was working mostly correct. When I fixed the racecondition for timeline reloading in #1343, I actually broke timeline reloading, revealing the database bug. So this PR now fixes timeline reloading and the database.
Will do some extensive testing before I merge this. This is critical enough to warrant a v8.1 release, I will do it asap.

app/src/main/java/com/keylesspalace/tusky/fragment/TimelineFragment.java
Nutomic: That looks like very confusing variable naming. Using mStatuses or something like that for class members would really help avoid this ;)

connyduck:
Ok found a problem: When reloading the status, we request older than x. In the response we do not get back x, so we don't know if it may have been deleted meanwhile and only replace old statuses beginning from x-1. Maybe we should request older than x+1, but that would probably be buggy in case somebody has two subsequent statuses in the timeline.
Its not a big problem so I am inclined to leave it for now.

app/src/main/java/com/keylesspalace/tusky/fragment/TimelineFragment.java
charlag: No Hungarian notation will be in the code I write 😜

app/src/main/java/com/keylesspalace/tusky/db/TimelineDao.kt
charlag: :man_facepalming: 
How on Earth did it work. How did it pass tests. Ughouhg.

app/src/main/java/com/keylesspalace/tusky/fragment/TimelineFragment.java
Nutomic: Then at least don't use the same name as a class variable and a local variable :p

charlag: I would rather rename local `statuses` into something `newStatuses` or `loadedStatuses`

Tak:
The changes make sense to me - although like @charlag , I don't have a good overview of all the pieces involved

connyduck:
> Ok found a problem: When reloading the status, we request older than x. In the response we do not get back x, so we don't know if it may have been deleted meanwhile and only replace old statuses beginning from x-1. Maybe we should request older than x+1, but that would probably be buggy in case somebody has two subsequent statuses in the timeline.
Its not a big problem so I am inclined to leave it for now.

ok apart from that everything looks good

PR #1375, opened 2019-07-09
fix card view outline being off a few pixels

connyduck:
before
abefore

after
after


connyduck:
The `cardView.setClipToOutline(true);` was somewhere behind a condition and was not executed in all cases.

Issue #1381, opened 2019-07-10
Unexpected behavior when connecting to instance


centromere:
When attempting to connect to the `gab.com` instance, the app sends me to YouTube and plays [this video](https://youtu.be/dQw4w9WgXcQ). I think the source of the bug is here:

https://github.com/tuskyapp/Tusky/blob/39a59f4adeaf9cada325e62d2a21365baa4a02df/app/src/main/res/values/donottranslate.xml#L113

You probably want to remove those. I'll submit a pull request.

* * * *
- Tusky Version: 8.0
- Android Device: Nexus 6p
- Mastodon instance (if applicable): gab.com

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

Laurelai:
Its not a bug. Its a feature.

centromere:
I'm not sure I understand. How does this behavior help me connect to gab.com?

Laurelai:
It doesnt. It prevents it. Which is the point. Try and keep up.

starrychloe:
What's the point of a decentralized platform if censorship is imposed by a central authority? Why even bother to stop using Facebook at all? Facebook already blocks hate speech, so why not just continue to use Facebook? Even if the purpose of a broken app is to block 'hate speech', not all people on Gab are haters. That's collectivist thinking (and collective punishment). The KKK are Christians. Does that make the Pope a Nazi?

Who gets to decide what constitutes 'hate speech'? 
https://www.youtube.com/watch?v=NmuzrHwMkMU

Intentionally crippling an app begs for the code base to be forked.

Sixthhokage1:
> What's the point of a decentralized platform if censorship is imposed by a central authority?
Third-party client apps are not part of the platform itself. Devs maintain freedom of association, and in the case of Tusky they have made absolutely clear that they do not want their app to be used to connect to Gab. Gab forked from a platform that 1) doesn't have an official app to act as a "central authority" and 2) has a pretty decent mobile web UI. If the fork fucked up the web UI such that you *need* a client then that's on Gab's devs. Y'all trying to force Tusky to support Gab are an entitled lot.

starrychloe:
Oh it seems this project has already been forked and deployed.

https://play.google.com/store/apps/details?id=com.gabai.gabby

> This app is an exact copy of a Mastodon app called Tusky

[...]

PR #1382, opened 2019-07-10
Fix bug which causes gab.com not to work

centromere:
See also: https://github.com/tuskyapp/Tusky/issues/1381


connyduck:
I'm afraid, but this is not a bug but a well known and highly requested easter egg.

PR #1384, opened 2019-07-14
Transfer SmartLengthInputFilter license to Tusky

HellPie:
This PR remove the license header in `SmartLengthInputFilter`, which originally attributed the file's ownership to me (Diego Rossi) under the Apache 2.0 license. The removal of the license header transfers the ownership and license to the ones specified in Tusky's `LICENSE` file.

The `SmartLengthInputFilter` has also been completely rewritten in Kotlin, without using any sort of automated tooling or copy-paste. It's still the same code for the most part but I rewrote it to ensure those few spots where Kotlin has extension methods make use of them.


app/src/main/java/com/keylesspalace/tusky/util/SmartLengthInputFilter.kt
```diff
@@ -0,0 +1,117 @@
...
+		/**
+		 * Default for maximum status length on Mastodon and default collapsing length on Pleroma.
+		 */
+		const val LENGTH_DEFAULT = 500
+
+		/**
+		 * Stores a reusable singleton instance of a {@link SmartLengthInputFilter} already configured
+		 * to the default maximum length of {@value #LENGTH_DEFAULT}.
+		 */
+		val INSTANCE = SmartLengthInputFilter(LENGTH_DEFAULT)
```
charlag: Could you make it `@JvmStatic`?

charlag: There's also `@JvmField` just fyi

HellPie:
wouldn't that limit the customization assuming maybe in the future stuff changes and say ideally notifications could only show 200 characters but timeline should still show 500, this way using a class it becomes possible to have two instances, but with object doesn't that make it a non-instantiable singleton?

connyduck:
Yes thats a valid point, but I don't think its going to happen soon and when it does it would be easy to change. 
I would prefer SmartLengthInputFilter being an object, but I will also merge it like it is now if you want.

HellPie:
If you don't plan to add any setting or tweak notifications then sure I can convert it to an object, I'll do it on thursday tho since I'm gonna be working tomorrow as well. It can always be moved back to a class whenever the need will arise.

HellPie:
I just realized that you asked me to make the `hasBadRatio` function toplevel, but wouldn't that become a bit too "generic" and come up without class name in other Kotlin code? There are some parts that reference it for example in `StatusExtension`.

connyduck:
Maybe call the function differently? `isOverCollapsibleLength` or something like that?

HellPie:
Don't forget in Java it will make the function name quite long and the class name is already quite verbose. Maybe `canCollapseStatus`? Maybe `should` rather than `can`?

app/src/main/java/com/keylesspalace/tusky/util/SmartLengthInputFilter.kt
```diff
@@ -0,0 +1,96 @@
...
+
+		// Enable trimming at the end of the closest word if possible
+		if (source[keep].isLetterOrDigit()) {
+			var boundary: Int
+
+			// Android N+ offer a clone of the ICU APIs in Java for better internationalization and
+			// unicode support. Using the ICU version of BreakIterator grants better support for
+			// those without having to add the ICU4J library at a minimum Api trade-off.
+			if (android.os.Build.VERSION.SDK_INT >= android.os.Build.VERSION_CODES.N) {
+				val iterator = android.icu.text.BreakIterator.getWordInstance()
```
connyduck: I'd rather use imports than fully qualified names

app/src/main/java/com/keylesspalace/tusky/util/SmartLengthInputFilter.kt
```diff
@@ -0,0 +1,96 @@
+package com.keylesspalace.tusky.util
```
connyduck: license header is missing
```suggestion
/* Copyright 2019 Tusky contributors
 *
 * This file is a part of Tusky.
 *
 * This program is free software; you can redistribute it and/or modify it under the terms of the
 * GNU General Public License as published by the Free Software Foundation; either version 3 of the
 * License, or (at your option) any later version.
 *
 * Tusky is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even
 * the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU General
 * Public License for more details.
 *
 * You should have received a copy of the GNU General Public License along with Tusky; if not,
 * see <http://www.gnu.org/licenses>. */

package com.keylesspalace.tusky.util
```

app/src/main/java/com/keylesspalace/tusky/util/SmartLengthInputFilter.kt
HellPie: this was also here before that, but it's mostly because otherwise it would clash with the one in the else block, it would still require one of the two to be fully qualified which is why, to avoid confusion, I kept both.

PR #1388, opened 2019-07-15
make avatar images in drawer rounded squares

connyduck:
So we finally have our trademark rounded squares with radius 1/8 of size everywhere.

This is a little bit hacky, but definitely easier than forking the MaterialDrawer library.
I added some comments, so I hope it is clear how it works.

Screenshot_20190715-113625

### DIFF
--- a/app/src/main/res/drawable/materialdrawer_shape_small.xml
+++ b/app/src/main/res/drawable/materialdrawer_shape_small.xml
@@ -1,0 +1,5 @@
+<shape xmlns:android="http://schemas.android.com/apk/res/android"
+    android:shape="rectangle" >
+    <solid android:color="#000"/>
+    <corners android:radius="5dp"/>
+</shape>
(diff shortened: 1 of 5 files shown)


app/src/main/res/drawable/materialdrawer_shape_small.xml
```diff
@@ -0,0 +1,5 @@
+<shape xmlns:android="http://schemas.android.com/apk/res/android"
+    android:shape="rectangle" >
+    <solid android:color="#000"/>
+    <corners android:radius="5dp"/>
```
connyduck: unfortunately its not possible to use % here, so i had to create two files, one for the selected profile, one for the others

app/src/main/res/layout/material_drawer_header.xml
```diff
@@ -0,0 +1,134 @@
+<!-- this replaces the default material_drawer_header.xml from the MaterialDrawer library -->
```
charlag: Are there any changes to this besides changing image classes?

connyduck: also added our placeholder as default

PR #1389, opened 2019-07-15
upgrade dependencies

connyduck:
Deliberately not upgrading the alpha AndroidX libs, as that requires some work and I would like to wait until they are stable for this.

The changes in code are due to OkHttp moving to Kotlin in 4.0.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -86,7 +86,7 @@ project.tasks.withType(org.jetbrains.kotlin.gradle.tasks.KotlinCompile).all {
     }
 }
 
-ext.daggerVersion = '2.23.1'
+ext.daggerVersion = '2.23.2'
 ext.retrofitVersion = '2.6.0'
 
 // if libraries are changed here, they should also be changed in LicenseActivity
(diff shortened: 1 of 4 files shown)


connyduck:
No idea why Bitrise did not run build on this branch automatically. I triggered it manually, build passed.

app/build.gradle
```diff
@@ -106,8 +106,8 @@ dependencies {
     implementation "com.squareup.retrofit2:retrofit:$retrofitVersion"
     implementation "com.squareup.retrofit2:converter-gson:$retrofitVersion"
     implementation "com.squareup.retrofit2:adapter-rxjava2:$retrofitVersion"
-    implementation 'com.squareup.okhttp3:okhttp:3.14.2'
-    implementation 'com.squareup.okhttp3:logging-interceptor:3.14.2'
+    implementation 'com.squareup.okhttp3:okhttp:4.0.1'
```
charlag: There are proguard issues:
https://github.com/square/okhttp/issues/5167
do we use R8 already?

connyduck: Yes we use R8. I will check if everything works, thx

PR #1392, opened 2019-07-15
fix avatars and display names in AccountsInListFragment

### LINKED ISSUE #1280: Issue with user who have no displayName at list editing
Related to #1104, cc: @charlag 
Users who don't have displayName shown as if they don't have any name.
https://github.com/tuskyapp/Tusky/blob/520e0d6e7a9d478eff09f3afc32d7bffdacef752/app/src/main/java/com/keylesspalace/tusky/AccountsInListFragment.kt#L210
We should use `account.name` instead of `displayName` here.

* * * *
- Tusky Version: master 419ae80
- Android Version: 8.0.0
- Android Device: G8441
- Mastodon instance (if applicable): all

connyduck:
closes #1280

Also makes the avatars round again. Looks like I forgot to call the new method here when introducing the gif animations.


connyduck:
Wait. The names are also not emojified correctly.

PR #1393, opened 2019-07-15
update timestamps on pull-to-refresh 

connyduck:
update timestamps on pull-to-refresh even when no new statuses are fetched


connyduck:
Yes we update them every minute. But it always find it very weird when I pull to refresh and the timestamps stay at x seconds.
In threads it already works, because we are not using a differ there.

PR #1399, opened 2019-07-17
Prevent NPE when displayName/userFullName is null

### LINKED ISSUE #1398: Crash when displaying notification from a user who has been deleted from instance.
1. Have two accounts on different instances.
2. One user was deleted from the instance by admin.
3. The other user, who was in conversation with the deleted user, on a different instance (pleroma.site) crashes on login in Tusky.

ezgif com-video-to-gif

* * * *
- Tusky Version: 8.1
- Android Version: 9
- Android Device: Pixel 2 XL, Pixel 3 XL
- ~~Mastodon~~ Pleroma instance: networked.space (user deleted from), pleroma.site (user still exists, crashes on login when trying to display notification from the former)

noln:
- Fixes #1398.
- Corrects edge case that takes some effort to replicate, but is replicable on multiple devices.
- Use `.isNullOrEmpty()` instead of `isEmpty()` in `Account` class.
- Replace null display name of deleted users with placeholder text, defaults to `unknown`.

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/adapter/StatusBaseViewHolder.java
+++ b/app/src/main/java/com/keylesspalace/tusky/adapter/StatusBaseViewHolder.java
@@ -151,8 +151,14 @@ public abstract class StatusBaseViewHolder extends RecyclerView.ViewHolder {
 
     protected abstract int getMediaPreviewHeight(Context context);
 
-    protected void setDisplayName(String name, List<Emoji> customEmojis) {
-        CharSequence emojifiedName = CustomEmojiHelper.emojifyString(name, customEmojis, displayName);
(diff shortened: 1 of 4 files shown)


connyduck:
I see whats the problem, I am just not sure if this is the right fix/if we should fix it at all.
To indicate the nullability correctly in Kotlin, both `displayName` and `localUsername` in `Account` should be `String?`, but that implies a whole lot of other changes up to the database I would rather not make. 

There are 2 crash reports of this in the Google Play console, so its a very rare problem.

noln:
Yeah I chose not to change the base Account fields because of the required DB changes, instead made the small change to catch the edge cases in Roma, thought I'd share upstream.

Oh and if those crashes are in Canada, they're both me! 😆

PR #1402, opened 2019-07-20
Fix crash on empty status messages

charlag:


connyduck:
`LENGTH_DEFAULT` should be a Double so we are not doing Integer division. Now the left hand side of the comparison can only give zero, one, ... and never a fraction like 0.8

charlag:
I don't think it should be double, it should be casted when needed I think

charlag:
Updated with cast to float

connyduck:
thx <3

Issue #763, opened 2018-08-16
Following notification selects wrong user account


rixx:
Two of my user accounts both have notifications. One set of notifications I swipe away (individually, if that matters), then I tap the other set of notifications. Tusky then presents me (reliably) with the wrong account's notification tab instead of the one I just tabbed.

Both accounts were on the same instance, in case that matters.

* * * *
- Tusky Version: 2.2
- Android Device: Moto G 5
- Mastodon instance (if applicable): chaos.social

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

PR #1408, opened 2019-07-22
fix notifications opening the wrong account

connyduck:
closes https://github.com/tuskyapp/Tusky/issues/763


app/src/main/java/com/keylesspalace/tusky/util/NotificationHelper.java
```diff
@@ -268,7 +268,7 @@ public static void make(final Context context, Notification body, AccountEntity
         summaryStackBuilder.addParentStack(MainActivity.class);
         summaryStackBuilder.addNextIntent(summaryResultIntent);
 
-        PendingIntent summaryResultPendingIntent = summaryStackBuilder.getPendingIntent(notificationId,
+        PendingIntent summaryResultPendingIntent = summaryStackBuilder.getPendingIntent((int) (notificationId + account.getId() * 10000),
```
connyduck: Looks like these ids need to be unique or the behavior of previously created Intents is changed 🙄

Issue #1329, opened 2019-06-20
Use max_bio_chars API endpoint


clarfon:
glitch-soc and co have been offering a `max_toot_chars` value via `/api/v1/instance`, and florence-masto has started offering the `max_bio_chars` value there too. Although it seems that this API won't be final, it makes sense to have code in Tusky that uses this value here so that there's at least some code that handles bio length limits on a per-instance basis. I imagine that once this is in place, it'll be easier to update if the API moves to a different location.

connyduck:
Do you know any server wherer I can try this out?

clarfon:
I know that bobadon.co.uk is running this version so maybe @TrechNex can help you out? This is assuming he's running it with the fix for the API installed.

connyduck:
https://github.com/tuskyapp/Tusky/commit/ce01e6de22422fea40188a9f837b5b21cab109ac

TrechNex:
bobadon.co.uk runs whatever the latest tagged release is (0.1.2 when I last checked).

PR #1414, opened 2019-07-26
Add support for instance property `max_bio_chars`

Tak:
Addresses #1329

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/ComposeActivity.java
+++ b/app/src/main/java/com/keylesspalace/tusky/ComposeActivity.java
@@ -313,22 +313,9 @@ public final class ComposeActivity
                     getString(R.string.compose_active_account_description,
                             activeAccount.getFullName()));
 
-            mastodonApi.getInstance().enqueue(new Callback<Instance>() {
-                @Override
-                public void onResponse(@NonNull Call<Instance> call, @NonNull Response<Instance> response) {
-                    if (response.isSuccessful() && response.body().getMaxTootChars() != null) {
-                        maximumTootCharacters = response.body().getMaxTootChars();
-                        updateVisibleCharactersLeft();
-                        cacheInstanceMetadata(activeAccount);
-                    }
-                }
-
-                @Override
-                public void onFailure(@NonNull Call<Instance> call, @NonNull Throwable t) {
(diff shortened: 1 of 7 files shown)


app/src/test/java/com/keylesspalace/tusky/ComposeActivityTest.kt
```diff
@@ -109,28 +111,14 @@ class ComposeActivityTest {
...
-            override fun execute(): Response<Instance> {
-                throw Error("not implemented")
-            }
-            override fun request(): Request {
-                throw Error("not implemented")
-            }
-
-            override fun enqueue(callback: Callback<Instance>?) {
-                instanceResponseCallback?.invoke(this, callback)
+        `when`(apiMock.instance).thenReturn(object: Single<Instance>() {
```
charlag: Couldn't you just do `Single.just()` (or however it's called) instead of subclassing `Single`?

app/src/main/java/com/keylesspalace/tusky/ComposeActivity.java
```diff
@@ -1851,6 +1838,19 @@ static boolean canHandleMimeType(@Nullable String mimeType) {
                 (mimeType.startsWith("image/") || mimeType.startsWith("video/") || mimeType.equals("text/plain")));
     }
 
+    private void onFetchInstanceSuccess(Instance instance) {
+        if (instance != null && instance.getMaxTootChars() != null) {
```
charlag: Are you sure you need to check for `null`?

app/src/test/java/com/keylesspalace/tusky/ComposeActivityTest.kt
Tak: Maybe? I'll take a look

app/src/main/java/com/keylesspalace/tusky/ComposeActivity.java
Tak: No, I think I might be overcautious since I introduced bugs multiple times with `Call`/`Callback`

charlag: But here it's Single and it would be null only if server intentionally would return it

PR #1418, opened 2019-07-27
cleanup code in StatusViewHolders

connyduck:
Goal: make the code easier to read and remove unnecessary stuff

Yes I removed a lot of null checks, but they are all unnecessary with a well-behaving server. In case this introduces a crash we should fix it somewhere else, input validation should not be done when binding views.

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/adapter/StatusBaseViewHolder.java
+++ b/app/src/main/java/com/keylesspalace/tusky/adapter/StatusBaseViewHolder.java
@@ -4,8 +4,8 @@ import android.content.Context;
 import android.graphics.drawable.Drawable;
 import android.text.Spanned;
 import android.text.TextUtils;
+import android.text.format.DateUtils;
 import android.view.View;
-import android.view.ViewGroup;
 import android.widget.Button;
 import android.widget.ImageButton;
 import android.widget.ImageView;
@@ -29,7 +29,7 @@ import com.keylesspalace.tusky.entity.Emoji;
 import com.keylesspalace.tusky.entity.Status;
 import com.keylesspalace.tusky.interfaces.StatusActionListener;
 import com.keylesspalace.tusky.util.CustomEmojiHelper;
-import com.keylesspalace.tusky.util.DateUtils;
+import com.keylesspalace.tusky.util.TimestampUtils;
 import com.keylesspalace.tusky.util.HtmlUtils;
 import com.keylesspalace.tusky.util.ImageLoadingHelper;
 import com.keylesspalace.tusky.util.LinkHelper;
@@ -63,8 +63,6 @@ public abstract class StatusBaseViewHolder extends RecyclerView.ViewHolder {
     private SparkButton reblogButton;
     private SparkButton favouriteButton;
     private ImageButton moreButton;
-    private boolean favourited;
-    private boolean reblogged;
     protected MediaPreviewImageView[] mediaPreviews;
     private ImageView[] mediaOverlays;
     private TextView sensitiveMediaWarning;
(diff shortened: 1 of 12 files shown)


app/src/main/java/com/keylesspalace/tusky/adapter/StatusViewHolder.java
```diff
@@ -92,15 +88,12 @@ void setPollInfo(final boolean ownPoll) {
     }
 
     void hideStatusInfo() {
-        if (statusInfo == null) {
-            return;
-        }
```
connyduck: the view needs to be here, crashing is fine when it isn't

app/src/main/java/com/keylesspalace/tusky/util/LinkHelper.java
```diff
@@ -129,7 +129,6 @@ public void onClick(View widget) {
         }
 
         view.setText(builder);
-        view.setLinksClickable(true);
```
connyduck: this line just does nothing, the default is true anyway

app/src/main/res/layout/item_conversation.xml
```diff
@@ -329,7 +329,7 @@
             app:layout_constraintTop_toTopOf="parent" />
 
         <TextView
-            android:id="@+id/status_media_label_1"
+            android:id="@+id/status_media_label_0"
```
connyduck: we start with 0 everywhere else as well

app/src/main/java/com/keylesspalace/tusky/adapter/StatusBaseViewHolder.java
```diff
@@ -358,7 +326,6 @@ private void setRebloggingEnabled(boolean enabled, Status.Visibility visibility)
     }
 
     protected void setFavourited(boolean favourited) {
-        this.favourited = favourited;
```
connyduck: redundant, the view stores the state for us

app/src/main/java/com/keylesspalace/tusky/adapter/StatusBaseViewHolder.java
```diff
@@ -423,18 +390,20 @@ protected void setMediaPreviews(final List<Attachment> attachments, boolean sens
 
             setAttachmentClickListener(mediaPreviews[i], listener, i, attachments.get(i), true);
 
+            final int mediaPreviewHeight = getMediaPreviewHeight(context);
```
connyduck: lets call this method only once instead of 4 times (its not that expensive, but still)

app/src/main/java/com/keylesspalace/tusky/adapter/StatusBaseViewHolder.java
```diff
@@ -674,17 +640,6 @@ protected void setupWithStatus(StatusViewData.Concrete status,
         }
     }
 
-
-    private void setContentDescription(@Nullable StatusViewData.Concrete status) {
-        if (status == null) {
```
connyduck: impossible

app/src/main/java/com/keylesspalace/tusky/adapter/StatusBaseViewHolder.java
```diff
@@ -706,21 +661,19 @@ private void setDescriptionForStatus(@NonNull StatusViewData.Concrete status) {
...
         String rebloggedUsername = status.getRebloggedByUsername();
         if (rebloggedUsername != null) {
-            reblogDescriontion = context
+            return context
                     .getString(R.string.status_boosted_format, rebloggedUsername);
         } else {
-            reblogDescriontion = "";
+            return "";
         }
-        return reblogDescriontion;
```
connyduck: nice typo

charlag: I'm good at them lol

PR #1428, opened 2019-07-31
fix size of visibility icon on detailed toot page

connyduck:
Regression from https://github.com/tuskyapp/Tusky/pull/1418

I was trying something out and did not mean to commit it 🙄 

On the plus side, this seems to be the only problem introduced in #1418


connyduck:
They are set in the code above.

PR #1439, opened 2019-08-03
Load sensitive media previews for only when shown

charlag:
Couldn't find an issue for that but I remember that we've discussed this.
Lazily load sensitive media: don't load it until shown.


charlag:
Oooops. I'll have a look. Does it happen consistently with this post?

connyduck:
Yes, but it is also somehow related to view recycling. On app start its fine, then I scroll down and up again, then its gone, then its there again, and so on.

charlag:
Is this media sensitive?

connyduck:
yes

charlag:
Fixed! Should work now. That was pretty :man_facepalming:

PR #1445, opened 2019-08-06
Enable reporting users without statuses 

connyduck:
There is a new wave of spambots that don't post but have links in their bio.

Also, delete the `statusContent` param of `ReportActivity` as it is used nowhere.


connyduck:
Possible improvement for later: Skip the first page of the report activity if there are no statuses to report.

app/src/main/java/com/keylesspalace/tusky/components/report/adapter/StatusesDataSource.kt
```diff
@@ -72,14 +72,18 @@ class StatusesDataSource(private val accountId: String,
...
         initialLoad.postValue(NetworkState.LOADING)
-        mastodonApi.statusObservable(params.requestedInitialKey).zipWith(
-                mastodonApi.accountStatusesObservable(accountId, params.requestedInitialKey, null, params.requestedLoadSize - 1, true),
-                BiFunction { status: Status, list: List<Status> ->
-                    val ret = ArrayList<Status>()
-                    ret.add(status)
-                    ret.addAll(list)
-                    return@BiFunction ret
-                })
+        if (params.requestedInitialKey == null) {
```
charlag: What is this key and why this check is needed now?

connyduck: Its the id of the status that is being reported. Null means user did not select report on a toot but on a profile.

PR #1448, opened 2019-08-11
Fix image viewer & transitions

charlag:
Follow up for #1344 
After few hours of fiddling with it, it seems to finally work now. Images transition, don't jump around when touched, avatars open and so on.
At least one of the problems was that we've been waiting for the transition when we shouldn't (for 3rd and 4th images when you open the 1st one e.g.). Another problem is that we really, really need `attacher.update()` for some reason. Third problem is that after loading Glide resets the image and it could happen when we already set thumbnail...
So, a lot of races. Hopefully it's more reliable now.
I've also noticed that exit transitions don't work for images other than the one you've opened. We could look into that in the future.

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/fragment/ViewImageFragment.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/fragment/ViewImageFragment.kt
@@ -17,6 +17,7 @@ package com.keylesspalace.tusky.fragment
 
 import android.animation.Animator
 import android.animation.AnimatorListenerAdapter
+import android.annotation.SuppressLint
 import android.content.Context
 import android.graphics.drawable.Drawable
 import android.os.Bundle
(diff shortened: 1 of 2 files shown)


connyduck:
Thx. Im not sure I understand exactly whats going on, but it definitely works better than before.

Just one thing I noticed: When I scroll in the timeline, then go into airplane mode and click a preview the Image viewer stays black. Shouldn't it display the preview image that is already cached?

charlag:
Hm, it should actually. I will check it.

charlag:
Okay, I think I fixed that (I cannot properly test because I can only debug via wifi so I tried to break the urls).
I've also found and fixed another bug when full image was not displayed after scrolling to the last page and back

connyduck:
I think it looks good now?

charlag:
Let's see if it works lol

Issue #1370, opened 2019-07-05
Theme autoswitch doesn't work with main timelines


Tak:
If Tusky is alive when the "automatic at sunset" time comes, the timelines don't switch theme (although other activities do, so the app switches between light and dark as the user navigates between different activities).

* * * *
- Tusky Version: Nightly/`master`
- Android Version: 9
- Android Device: Pixel 2 XL
- Mastodon instance (if applicable):

- :ballot_box_with_check: I searched or browsed the repo’s other issues to ensure this is not a duplicate.

connyduck:
I think this was fixed by https://github.com/tuskyapp/Tusky/pull/1455

Issue #1260, opened 2019-05-15
“Use System Design” doesn't work on Android Q


cassidyjames:
Android Q officially introduced a system-wide dark theme. However, it doesn't appear the “Use System Design” theme preference in Tusky respects this and switches the theme base on this setting.

connyduck:
fixed by https://github.com/tuskyapp/Tusky/pull/1455

cassidyjames:
I'm still seeing this issue in Android 10; I'm not sure if it was a regression or wasn't actually fixed. I can provide more details if needed!

connyduck:
Works for me. Please provide more details, I'm especially interested in the exact Tusky and Android versions.

cassidyjames:
Pixel 3, Android 10 (January security update), Tusky 9.1 (Play Store beta). Will attach a screen recording shortly.

cassidyjames:
Ah I see now there hasn't been a release for several months, and this commit was after the latest release. I'll see if I can manually build an APK or something.

charlag:
1. You are on beta branch, maybe beta was not updated and only main channel was
2. There's a nightly build in play store

And you've found it, good

cassidyjames:
Yep, sorry about the confusion! I can confirm that this is fixed for me on the nightly green version. Gonna keep using that unless I hit any major issues. :smile:

PR #1455, opened 2019-08-19
upgrade appcompat, fix language switching & nightmode

connyduck:
I did not want to do this quite yet, but turns out `androidx.preference:preference:1.1.0-rc01` depends on `androidx.appcompat:appcompat:1.1.0-rc01`, so we already upgraded to it implicitely in https://github.com/tuskyapp/Tusky/pull/1421. This PR upgrades explicitly and fixes the bugs that got introduced with it.

The biggest change in appcompat 1.1.0 is that `AppCompatDelegate.setDefaultNightMode` restarts activities. Also `AppCompatDelegate.setDefaultNightMode` only needs to be set once, not for every activity (this may have been the case in older versions too, not sure).

closes https://github.com/tuskyapp/Tusky/issues/1260
closes https://github.com/tuskyapp/Tusky/issues/1370

Nice sideeffect: The `ThemeUtils` inject hack that was in place to fix the tests can be removed. Also looks like the hacks we had to make `MODE_NIGHT_FOLLOW_SYSTEM` work are now fixed in appcompat.

Known problems: In-app Language switching does not work on Android Q. On current master it does not work on any API level because the appcompat upgrade broke it.


connyduck:
Yes, its definitely better than before, but I still want to try to get it working even on Q

connyduck:
Meh, seems to affect more versions than Q.
The problem is that the `AppCompatDelegate.setDefaultNightMode` overrides whatever Locale we set in `attachBaseContext` and breaks the in-app language switching D:
Patching the context again afterwards seems to do nothing D:

connyduck:
This may be a bug in the new appcompat version. Lets see if I get an answer to this: https://issuetracker.google.com/issues/139935426

connyduck:
Wow, looks like this is indeed a bug in AndroidX and it will be fixed! I will merge this and hopefully the fix is out soon.

PR #1461, opened 2019-08-24
use status source for delete and redraft

connyduck:
ref https://github.com/tootsuite/mastodon/pull/10669

fixes bugs with delete&redraft where mentions are lost


connyduck:
Nope its not working with pleroma. Need to fallback to the old code

connyduck:
And of course it needs error handling, otherwise it crashes. Thx!

PR #1465, opened 2019-08-29
fix error and content overlapping in NotificationsFragment

connyduck:
Screenshot_20190805-163248

This happens when two requests (the initial one and another from swipe to refresh) are active + one succeeds and one fails. Fixed by disabling swipe to refresh while loading.


connyduck:
There are two possibilities where it can fail: 
- Initial request -> does not need to be enabled, error screen is shown
- Subsequent request -> should already be enabled (otherwise the request can't even be started)

charlag:
Ah yes we don't disable it while ti's refreshing. Should be fine then.

Issue #1334, opened 2019-06-21
Implement new attachment type audio


connyduck:
In the next release mastodon will add audio attachment support. https://github.com/tootsuite/mastodon/pull/11123

Note that it is not yet decided which format the audio will have, see https://github.com/tootsuite/mastodon/issues/11131

koyuawsmbrtn:
Suggestion: Allowing attachments to be audio files **and** being able to record voice notes to send them around

connyduck:
-> https://github.com/tuskyapp/Tusky/issues/1337

novadeviator:
audio support is now enabled in mastodon. it's mp3 format. this issue is not clear in terms of what is supposed to be implemented - playing or/and uploading audio. #1337 is for uploading, but playback of audio does not work in Tusky currently if i'm not mistaken.

connyduck:
This issue is about displaying the attachments correctly in timelines & playing audio.

PR #1466, opened 2019-09-02
Add audio support for timelines

connyduck:
Straightforward, minimal implementation to get audio attachments to work with Tusky. I don't have time for something more elaborate right now, sorry.
Closes https://github.com/tuskyapp/Tusky/issues/1334

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/adapter/StatusBaseViewHolder.java
+++ b/app/src/main/java/com/keylesspalace/tusky/adapter/StatusBaseViewHolder.java
@@ -334,8 +334,7 @@ public abstract class StatusBaseViewHolder extends RecyclerView.ViewHolder {
         favouriteButton.setChecked(favourited);
     }
 
-    private void loadImage(MediaPreviewImageView imageView, String previewUrl, String description,
-                           MetaData meta) {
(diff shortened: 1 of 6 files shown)


app/src/main/java/com/keylesspalace/tusky/adapter/StatusBaseViewHolder.java
```diff
@@ -505,11 +506,14 @@ private void setAttachmentClickListener(View view, StatusActionListener listener
...
+        if(attachment.getMeta().getDuration() != null && attachment.getMeta().getDuration() > 0) {
+            duration = formatDuration(attachment.getMeta().getDuration());
+        }
         if (TextUtils.isEmpty(attachment.getDescription())) {
-            return context
-                    .getString(R.string.description_status_media_no_description_placeholder);
+            return duration + " " + context.getString(R.string.description_status_media_no_description_placeholder);
         } else {
-            return attachment.getDescription();
+            return duration + " " + attachment.getDescription();
```
Tak: We'll get an extra leading space here if the duration is empty, not sure how much it matters

connyduck: good catch, I will add a `trim()`

PR #1467, opened 2019-09-02
revert appcompat upgrade in order to fix in-app-language switching

connyduck:
Temporary fix until we can solve the problems with https://github.com/tuskyapp/Tusky/pull/1455


connyduck:
Yes, this is a fix for Tusky 9, for Tusky 10 we are definitely upgrading to the new appcompat, and if we can't fix the language switcher we kick it out.

PR #1470, opened 2019-09-05
only show Progress Dialog if necessary when sending toot

connyduck:


app/src/main/java/com/keylesspalace/tusky/util/CountUpDownLatch.java
```diff
@@ -44,4 +44,8 @@ public synchronized void await() throws InterruptedException {
             wait();
         }
     }
+
+    public synchronized boolean isEmpty() {
```
connyduck: not sure how to call this method 🤔

charlag: Wait since when we have our own countdownlatch, it's not implemented correctly because it access the same variable from different synchronized methods

connyduck: The variable should be synchronized, not the methods, right? I don't think it matters though since all calls probably happen on the main thread anyway.

charlag: synchronized block/method tells "only one method can go in this piece of code right now" but it doesn't help if you do it from multiple methods. You have to either restructure it or (much better alternative) just use AtomicInteger.
Relatively non-painful change would be:
 - all methods which access it go through one internal synchronized method which both changes the variable and returns it's value (kinda like `AtomicInteger#getAndSet`)
 - the number itself must be volatile then

we don't have to fix it now of course. Method name is fine I think.

PR #1480, opened 2019-09-14
Fix crash when status date is null

connyduck:
I have no idea where null `createdAt` is coming from, but I see it in the crashlogs of the beta. So lets revert the change.


app/src/main/java/com/keylesspalace/tusky/adapter/StatusBaseViewHolder.java
```diff
@@ -252,37 +252,48 @@ private void setAvatar(String url,
 
     }
 
-    protected void setCreatedAt(@NonNull Date createdAt) {
+    protected void setCreatedAt(Date createdAt) {
```
charlag: Can be `@Nullable` then?

PR #1490, opened 2019-09-20
upgrade autodispose

connyduck:
They renamed the extension function from `autoDisposable` to `autoDispose`


charlag:
Should we disable hiding to save on allocations? https://github.com/uber/AutoDispose/pull/376
I think we don't need to protect things from ourselves

Their new `withScope()` is :fire: 
Feels like some Ruby magic except there's no magic.

connyduck:
So I should set `AutoDisposePlugins#setHideProxies()` to false somewhere in our Application class and it will improve performance?

charlag:
That's how I understand it. The difference is probably marginal but it may add up with many streams

connyduck:
nice, thx for the tip

PR #1495, opened 2019-09-23
Fix screen reader not announcing post content in some cases

charlag:
Some posts (especially cross-posted ones) are marked as sensitive but have no CW. User cannot expand them and hear the full content.


connyduck:
Thx for finding this. As far as I understand it, `sensitive` only means the media should be hidden and has nothing to do with CWs.

PR #1507, opened 2019-09-30
fix requestLayout() improperly called warning in AccountActivity

connyduck:
Noticed this warning in logcat:
`requestLayout() improperly called by com.google.android.material.appbar.CollapsingToolbarLayout{15ddc71f V.ED.... ......ID 0,0-540,568 #7f0900e4 app:id/collapsingToolbar} during layout: running second layout pass`

this fixes it


connyduck:
It was called during initial layout which is quite bad.

Issue #1413, opened 2019-07-26
Use more than colour to distinguish boost icon


clarfon:
Right now, the boost icon changes colour from grey to blue whenever a post is boosted. Right now, my phone is set to greyscale mode at night to help push me to sleep, and it's basically impossible to tell if I pressed the button in this mode.

It would be nice if a checkmark or something similar were shown on the icon to help distinguish boosts with more than just colour. This is also a general rule of thumb according to WCAG as well.

connyduck:
Thats a really good idea, thx. How about something like this:

![new_boost](https://user-images.githubusercontent.com/10157047/62010949-f4eb6880-b171-11e9-904d-8fc5b0087e8b.png)

PR #1513, opened 2019-10-02
change active boost icon

connyduck:
closes https://github.com/tuskyapp/Tusky/issues/1413

And I also changed the naming of the icons since it was inconsistent.


connyduck:
yes

PR #1532, opened 2019-10-11
add build customization options

connyduck:
Admins of Mastodon instances ask me if I can provide help with customizing Tusky for their instance. 
So here is a way to customize Tusky without having to touch any code. The app still has to be built and signed though.
Please tell me if that is what you had in mind and if it is helpful.


connyduck:
One thing that is maybe weird is that the button on the about page will always say "Tusky's account" even when it will lead to another one. Don't have an idea how to resolve that nicely yet. 🤔

mynameismonkey:
Absolutely will get the job done, hugely appreciated, we might make a few additional downstream changes but they would be Welsh-language specific so no need to think about those. Do you see this going into master or a separate themeable branch?

connyduck:
When I merge it it would go to develop first and then to master with the next release.

PR #1535, opened 2019-10-13
upgrade dependencies

connyduck:


app/build.gradle
```diff
@@ -107,16 +107,16 @@ dependencies {
...
+    implementation 'com.squareup.okhttp3:logging-interceptor:4.2.2'
     implementation 'org.conscrypt:conscrypt-android:2.2.1'
     implementation 'com.github.connyduck:sparkbutton:2.0.1'
     implementation 'com.github.chrisbanes:PhotoView:2.3.0'
     implementation 'com.mikepenz:google-material-typeface:3.0.1.3.original@aar'
     implementation('com.theartofdev.edmodo:android-image-cropper:2.8.0') {
         exclude group: 'com.android.support'
     }
-    implementation 'com.evernote:android-job:1.2.6'
+    implementation 'com.evernote:android-job:1.4.2'
```
connyduck: This now uses Workmanager under the hood. I wonder A) if this changes the reliability of notifications B) if we could also implement Workmanager directly

app/build.gradle
```diff
@@ -125,18 +125,18 @@ dependencies {
     // architecture components
     implementation 'androidx.lifecycle:lifecycle-extensions:2.1.0'
     //room
-    implementation 'androidx.room:room-runtime:2.1.0'
+    implementation 'androidx.room:room-runtime:2.2.0'
```
charlag: Should we try incremental annotation processing?
https://developer.android.com/jetpack/androidx/releases/room#2.2.0

connyduck: It can make the build faster, right? I don't have any problems with the build time but yes, lets try it.

PR #1537, opened 2019-10-15
Upgrade api level to Android 10

### LINKED ISSUE #1365: Upgrade to targetSdkVersion 29 / Android Q
ToDo:
- [x] upgrade AndroidX libraries 
- [x] implement the new sharing api (there is a [AndroidX library](https://developer.android.com/reference/androidx/core/content/pm/ShortcutInfoCompat?hl=en) for this )
- [x] check if we need to change something regarding the new behavior of the external storage permission
- [x] [PreferenceManager](https://developer.android.com/reference/android/preference/PreferenceManager) is deprecated -> migrate to AndroidX
- [x] fix https://github.com/tuskyapp/Tusky/issues/1260
- [x] [turn on hyphenation again](https://android-developers.googleblog.com/2019/07/whats-new-for-text-in-android-q.html)
- [x] test on foldable device/emulator
- [x] test with edge-to-edge display

There is no need to rush this, as new Android versions tend to roll out quite slowly. I would wait until we can test on a stable Android Q release.

connyduck:
closes #1365 

I implemented the new sharing api. As a sideeffect there is now no more "compose" launcher shortcut but one shortcut for each account the user is logged in. I also improved our settings tile, it will now no longer open ComposeActivity with the current account, but ask which account to use if multiple are present.

Also I did not fix the deprecated stuff in ComposeActivity to not create merge conflicts. @charlag


connyduck:
I will have to do some more testing before this is ready to merge, but feel free to comment on the code.

charlag:
I'm confused, should I review it or it's not ready yet?

app/src/main/java/com/keylesspalace/tusky/util/ShareShortcutHelper.kt
```diff
@@ -0,0 +1,96 @@
...
+
+            val outBmp = Bitmap.createBitmap(outerSize, outerSize, Bitmap.Config.ARGB_8888)
+
+            val canvas = Canvas(outBmp)
+            canvas.drawBitmap(bmp, (outerSize - innerSize).toFloat() / 2f, (outerSize - innerSize).toFloat() / 2f, null)
+
+            IconCompat.createWithAdaptiveBitmap(outBmp)
+        }
+
+        val person = Person.Builder()
```
charlag: `Person.Builder`
:robot:

app/src/main/java/com/keylesspalace/tusky/util/ShareShortcutHelper.kt
```diff
@@ -0,0 +1,96 @@
...
+            val outerSize = context.resources.getDimensionPixelSize(R.dimen.adaptive_bitmap_outer_size)
+
+            val bmp = Glide.with(context)
+                    .asBitmap()
+                    .load(account.profilePictureUrl)
+                    .error(R.drawable.avatar_default)
+                    .submit(innerSize, innerSize)
+                    .get()
+
+            val outBmp = Bitmap.createBitmap(outerSize, outerSize, Bitmap.Config.ARGB_8888)
```
charlag: got me couple of seconds to figure this out, would be nice to have a comment I think?

app/src/main/java/com/keylesspalace/tusky/util/ShareShortcutHelper.kt
```diff
@@ -0,0 +1,96 @@
...
+        val person = Person.Builder()
+                .setIcon(icon)
+                .setName(account.displayName)
+                .setKey(account.identifier)
+                .build()
+
+        // This intent will be sent when the user clicks on one of the launcher shortcuts. Intent from share sheet will be different
+        val intent = Intent(context, MainActivity::class.java).apply {
+            action = Intent.ACTION_SEND
+            type = "text/plain"
```
charlag: Do you need a type?

app/src/main/java/com/keylesspalace/tusky/util/ShareShortcutHelper.kt
```diff
@@ -0,0 +1,96 @@
...
+                .setShortLabel(account.displayName)
+                .setPerson(person)
+                .setLongLived(true)
+                .setIcon(icon)
+                .build()
+
+        ShortcutManagerCompat.addDynamicShortcuts(context, listOf(shortcutInfo))
+
+    }
+            .subscribeOn(Schedulers.computation())
```
charlag: Not very good to block `computation()` for network request. Maybe would be better to split it into two parts.

Tak:
Don't we lose direct share without `AccountService`?

app/src/main/java/com/keylesspalace/tusky/util/ShareShortcutHelper.kt
connyduck: Well the MainActivity expects it, I thought its easier to add it here than to change it in MainActivity

connyduck:
> I'm confused, should I review it or it's not ready yet?

Its not yet tested thoroughly, but code hopefully doesn't change too much anymore, so yes thx

> Don't we lose direct share without AccountService?

No, the AndroidX library takes care of it. When you add shortcuts via the `ShortcutManagerCompat` it will work with all api levels that support sharing.

Tak:
Are you still getting direct share suggestions for Tusky with this branch?
I'm not (although mine have been weird since I upgraded to 10…)

connyduck:
Yes, but not on all devices. I am investigating why not.

connyduck:
Ok looks like its an issue on the emulator, see https://stackoverflow.com/questions/57396667/directshare-api-with-android-q
It works on all my physical devices.

PR #1539, opened 2019-10-16
You've been banned

koyuawsmbrtn:
Not sure, but this is far better than Rick Roll imho

### DIFF
--- a/app/src/main/res/values/donottranslate.xml
+++ b/app/src/main/res/values/donottranslate.xml
@@ -115,7 +115,7 @@
         <item>gab.ai</item>
     </string-array>
 
-    <string name="rick_roll_url">https://www.youtube.com/watch?v=dQw4w9WgXcQ</string>
+    <string name="rick_roll_url">https://www.youtube.com/watch?v=FXPKJUE86d0</string>
 
 
     <string-array name="poll_duration_names">


JuanjoSalvador:
At least, is funniest.

resynth1943:
Seeing as the developers would rather redirect these people to a YouTube video 
('funny memez lolz!!!'), instead of an educational resource, I'm willing to bet they 
don't actually care about making a change.

Plus, 'banned' seems like quite a harsh word, so it'll probably attract negative feedback.

dusnm:
> Seeing as the developers would rather redirect these people to a YouTube video
> ('funny memez lolz!!! roflroflrofl'), instead of an educational resource, I'm willing to bet they
> don't actually care about making a change.
> 
> Plus, 'banned' seems like quite a harsh word, so it'll probably attract negative feedback.

It shouldn't be their job to educate nazis and trolls.

app/src/main/res/values/donottranslate.xml
```diff
@@ -115,7 +115,7 @@
         <item>gab.ai</item>
     </string-array>
 
-    <string name="rick_roll_url">https://www.youtube.com/watch?v=dQw4w9WgXcQ</string>
+    <string name="rick_roll_url">https://www.youtube.com/watch?v=FXPKJUE86d0</string>
```
JacksonChen666, line 118: seems to be changing the url without changing the key, might be a bad idea as this string is pretty much being misrepresented and a more appropriate name could be `banned_url` or something

DraconicNEO:
> > Seeing as the developers would rather redirect these people to a YouTube video
> > ('funny memez lolz!!! roflroflrofl'), instead of an educational resource, I'm willing to bet they
> > don't actually care about making a change.
> > Plus, 'banned' seems like quite a harsh word, so it'll probably attract negative feedback.
> 
> It shouldn't be their job to educate nazis and trolls.

I agree, plus if these servers really wanted to they could bypass this easily, either by plagiarizing, or they could set up different servers for logins to bypass these artificial restrictions. If they went the second route the devs would need to add new entries for every new login server. 
Honestly seems like an uphill battle, they've only blocked 3 servers out of all the terrible ones that exist.

PR #1544, opened 2019-10-28
Migrate to ViewPager2

connyduck:
This is yet untested, I bet I introduced some bugs, also I can probably clean up the code a bit.

I like ViewPager2, it seems faster than the old one. Also got rid of a few lines of code.


app/src/main/java/com/keylesspalace/tusky/fragment/AccountMediaFragment.kt
```diff
@@ -235,14 +234,6 @@ class AccountMediaFragment : BaseFragment(), RefreshableFragment, Injectable {
             topProgressBar?.show()
     }
 
-    // That's sort of an optimization to only load media once user has opened the tab
-    // Attention: can be called before *any* lifecycle method!
-    override fun setUserVisibleHint(isVisibleToUser: Boolean) {
-        super.setUserVisibleHint(isVisibleToUser)
-        this.isVisibleToUser = isVisibleToUser
-        if (isVisibleToUser && isAdded) doInitialLoadingIfNeeded()
-    }
```
connyduck: This does not work anymore with ViewPager2. I don't think its a problem as ViewPager2 seems to only start the fragment once its visible.

app/src/main/java/com/keylesspalace/tusky/pager/AccountPagerAdapter.kt
```diff
@@ -0,0 +1,66 @@
...
+
+import androidx.viewpager2.adapter.FragmentStateAdapter
+import java.lang.ref.WeakReference
+
+class AccountPagerAdapter(
+        activity: FragmentActivity,
+        private val accountId: String
+) : FragmentStateAdapter(activity) {
+
+    private val fragments = SparseArray<WeakReference<Fragment>>(TAB_COUNT)
```
connyduck: This is a hack I needed to use multiple times because there seems to be no out-of-the-box way to get the active fragments from a `FragmentStateAdapter`. Seems to work fine.

connyduck:
Ok, I did some testing. Everything looks good, except the close transition is not played everytime when closing the ViewMediaFragment after swiping to another image. I don't know how to fix this, but its also not a big issue.

app/src/main/java/com/keylesspalace/tusky/components/search/SearchActivity.kt
```diff
@@ -57,8 +58,12 @@ class SearchActivity : BottomSheetActivity(), SearchView.OnQueryTextListener, Ha
     }
 
     private fun setupPages() {
-        pages.adapter = SearchPagerAdapter(this, supportFragmentManager)
-        tabs.setupWithViewPager(pages)
+        pages.adapter = SearchPagerAdapter(this)
+
+        TabLayoutMediator(tabs, pages) {
```
charlag: odd formatting here

app/src/main/java/com/keylesspalace/tusky/pager/AccountPagerAdapter.kt
charlag: Why `SparseArray` with only 4 tabs?

app/src/main/java/com/keylesspalace/tusky/pager/AccountPagerAdapter.kt
```diff
@@ -0,0 +1,66 @@
...
+
+    fun getFragment(position: Int): Fragment? {
+        return fragments[position]?.get()
+    }
+
+    fun refreshContent() {
+        for (i in 0 until TAB_COUNT) {
+            val fragment = getFragment(i)
+            if (fragment != null && fragment is RefreshableFragment) {
+                (fragment as RefreshableFragment).refreshContent()
```
charlag: why is cast needed here?

app/src/main/java/com/keylesspalace/tusky/ViewMediaActivity.kt
```diff
@@ -153,7 +154,7 @@ class ViewMediaActivity : BaseActivity(), ViewImageFragment.PhotoActionsListener
         window.statusBarColor = Color.BLACK
         window.sharedElementEnterTransition.addListener(object : NoopTransitionListener {
             override fun onTransitionEnd(transition: Transition) {
-                (adapter as SharedElementTransitionListener).onTransitionEnd()
+                adapter.onTransitionEnd(viewPager.currentItem)
```
charlag: I'm not 100% sure that it will be current item but should be fine for common cases.

app/src/main/java/com/keylesspalace/tusky/pager/AccountPagerAdapter.kt
connyduck: It isn't, thx

connyduck: Why not `SparseArray`? (It was one in the old implementation too)

charlag: `SparseArray` is an alternative for hashmap with `long`/`int` keys, not for arrays/lists. It is used to not allocate object for the numeric key and to not allocate space for all the keys if they are sparse. It's not constant time and it uses more memory than list in common cases, it just uses less memory than numeric key map.

connyduck: indeed 🧐
thx, will change

Issue #1550, opened 2019-11-04
Crash on clicking hashtags in Persian Locale


danialbehzadi:
Here is related catalog: https://paste.ubuntu.ir/qojx

The problem is wrong l10n of `#%s` string.

I fixed it here: https://github.com/tuskyapp/Tusky/commit/36ec3e6e2b4073f6bdf9ce32d1854084480a261d

Now we just need a release to address it. Please release a minor bug fix version.

* * * *
- Tusky Version: 9.1 (F-Droid)
- Android Version: 9 (LineageOS 16)
- Android Device: Xiaomi Pocophone F1

- [X] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

ahangarha:
I also faced with this issue and along with @danialbehzadi we discovered it happens only when the app is running in Persian.

It seems his solution fixes this bug.

ahangarha:
This bug has occurred because of two mistake:
1. adding `#%s` in strings to be translated
1. misunderstanding of the string by the translator (me) as if it was referring to second!

I am happy it is detected and resolved.

Please remove this string from the list of translatable strings.

haghighi-ahmad:
+1
Same problem on **Tusky 9.1** (Android 9)

PR #1551, opened 2019-11-05
set title_tag as non-translatable

danialbehzadi:
This should prevent crashes like https://github.com/tuskyapp/Tusky/issues/1550 in future.


danialbehzadi:
done!

PR #1560, opened 2019-11-16
Bookmarks

connyduck:
straightforward - most code copied from favourites


app/build.gradle
```diff
@@ -105,6 +105,7 @@ dependencies {
 
     implementation "androidx.core:core-ktx:1.2.0-beta01"
     implementation "androidx.appcompat:appcompat:1.1.0"
+    implementation "androidx.fragment:fragment-ktx:1.1.0"
```
connyduck: added this dep so I could use the `commit` extension function in StatusListActivity

app/src/main/java/com/keylesspalace/tusky/fragment/NotificationsFragment.java
```diff
@@ -443,6 +443,21 @@ public void onFavourite(final boolean favourite, final int position) {
...
+    @Override
+    public void onBookmark(final boolean bookmark, final int position) {
+        final Notification notification = notifications.get(position).asRight();
+        final Status status = notification.getStatus();
+
+        timelineCases.bookmark(status, bookmark)
+                .observeOn(AndroidSchedulers.mainThread())
+                .as(autoDisposable(from(this)))
+                .subscribe(
+                        (newStatus) -> setFavouriteForStatus(position, status, bookmark),
```
charlag: `setFavouriteForStatus`? :thinking:

app/src/main/java/com/keylesspalace/tusky/fragment/ViewThreadFragment.java
```diff
@@ -621,6 +639,28 @@ private void handleReblogEvent(ReblogEvent event) {
         adapter.setItem(posAndStatus.first, newViewData, true);
     }
 
+    private void handleBookmarkEvent(BookmarkEvent event) {
+        Pair<Integer, Status> posAndStatus = findStatusAndPos(event.getStatusId());
+        if (posAndStatus == null) return;
+
+        boolean bookmark = event.getBookmark();
+        posAndStatus.second.setBookmarked(bookmark);
```
charlag: Is it okay that both status and reblog are bookmarked? Should one of them be bookmarked?

app/src/main/java/com/keylesspalace/tusky/MainActivity.java
```diff
@@ -415,7 +417,10 @@ public void cancel(ImageView imageView) {
                             Intent intent = new Intent(MainActivity.this, EditProfileActivity.class);
```
charlag: Maybe not in this PR but this if-else chain *screams* "make me switch statement"

app/src/main/java/com/keylesspalace/tusky/fragment/NotificationsFragment.java
connyduck: ah I knew there was a c&p mistake somewhere

app/src/main/java/com/keylesspalace/tusky/fragment/ViewThreadFragment.java
connyduck: We do it like this for favourites since ever and I don't want to debug why exactly :D

app/src/main/java/com/keylesspalace/tusky/MainActivity.java
connyduck: yes true!

connyduck:
> My biggest concern is loading of pages. I know that for favorites it doesn't work properly now because we must use response headers and I changed it to use first/last status IDs instead. Favorites are ordered differently. Because of that we cannot load beyond the first page sometimes. I don't know if this is the case for Favorites but it would be nice to fix both at once.

I tested it and it seems fine?

PR #1572, opened 2019-11-25
Small changes to the media player

### LINKED ISSUE #620: Going back from a video requires two taps on Back button
So there seems to be a small, yet annoying bug:  
When I click on a video, it opens up this video player and shows the video.  
But in order to go back to the timeline, I need to click on the Back button (the one in the Navigation Bar) twice.  
It looks like the first 'back' hides the video controls as it immediately brings me back when they're gone.

* * * *
- Tusky Version: 1.7 + current `master`
- Android Version: 7.1.1
- Android Device: Moto Z2 Play
- Mastodon instance (if applicable): mastodon.xyz

viviicat:
I've never done this before. Please let me know if there's anything I should be doing differently with this PR! I've also never coded in Kotlin so if some of the code is off or too verbose, I'm all ears.

These changes fix a few issues I encountered, and also resolve #620.

I was noticing that the media controls were hiding at a slightly different time as the top toolbar, and that this hiding was happening at unexpected times. If the video took more than 3 seconds to load, the toolbar would hide before the video started playing, and tapping the video repeatedly would cause the controls to hide very quickly and become 'desynced' with the toolbar visibility.

I also noticed that toolbar and video description fading in wasn't working because the visibility was hidden during the fade-in.

I also noticed there were issues with playing audio--the controls never appeared initially, and if the user tapped to show them, they would disappear unnecessarily after 3 seconds. The controls now always stay up while audio is playing.

And finally, when pausing video, the controls would still disappear after 3 seconds, which seems unintuitive. The controls now stay up while the video is paused.


charlag:
Hi!
Thanks for the changes, sorry for not looking at them yet, I will try to so it soon.

app/src/main/java/com/keylesspalace/tusky/view/ExposedPlayPauseVideoView.kt
```diff
@@ -0,0 +1,33 @@
+package com.keylesspalace.tusky.view
+
+import android.content.Context
+import android.util.AttributeSet
+import android.widget.VideoView
+
+class ExposedPlayPauseVideoView : VideoView {
+    private var mListener: PlayPauseListener? = null
```
connyduck: no m prefix please

app/src/main/java/com/keylesspalace/tusky/view/ExposedPlayPauseVideoView.kt
```diff
@@ -0,0 +1,33 @@
...
+
+    constructor(ctx: Context) : super(ctx)
+    constructor(ctx: Context, attrs: AttributeSet) : super(ctx, attrs)
+    constructor(ctx: Context, attrs: AttributeSet, defStyleAttr: Int) : super(ctx, attrs, defStyleAttr)
+    constructor(ctx: Context, attrs: AttributeSet, defStyleAttr: Int, defStyleRes: Int) : super(ctx, attrs, defStyleAttr, defStyleRes)
```
connyduck: use one overloaded constructor for this (like here https://github.com/tuskyapp/Tusky/blob/develop/app/src/main/java/com/keylesspalace/tusky/view/PollPreviewView.kt#L26-L30 )

viviicat:
Thanks for the review. I'm on vacation this week so I don't

viviicat:
Dammit I hit the wrong button

viviicat:
Thanks for the review! I'm on vacation all this week so I don't have access to my computer, but I can definitely address the feedback when I return.

I did notice one tiny inconsistency with my changes which was if you tap the screen twice when playing *audio* to hide/show the controls, they still autohide.

charlag:
hey @viviicat, are you still interested in working on this?

viviicat:
Yeah--sorry about the delay. I'm going to take a look at this tonight.

connyduck:
Thx! ^_^

PR #1576, opened 2019-12-02
fix formatting of %s favs %s boosts strings in some languages

connyduck:
I don't know why Weblate sometimes gets it right and sometimes it does not
¯\\\_(ツ)\_/¯

cc @MasterGroosha


MasterGroosha:
Thank you!

PR #1581, opened 2019-12-08
Use blurhash as image preview and as sensitive media cover, close #1571

### LINKED ISSUE #1571: Implement Blurhash
Implementing blurhash would save some frustration about images being loaded on slow connections as the user would receive some sort of picture, but knows that it is loading the right one.

* * * *
- Tusky Version: any
- Android Version: any
- Android Device: any

charlag:
It's 90% finished I'd say but there are couple of things:
1. Sometimes when you click button to hide media, blurhash is not applied. I did not figure it out yet
2. We probably need a setting to not use blurhash

Screenshot_20191208-182851_Tusky

Feel free to review/guess what's wrong there!

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/PreferencesActivity.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/PreferencesActivity.kt
@@ -56,7 +56,7 @@ class PreferencesActivity : BaseActivity(), SharedPreferences.OnSharedPreference
             setDisplayShowHomeEnabled(true)
         }
 
-        val fragment: Fragment = when(intent.getIntExtra(EXTRA_PREFERENCE_TYPE, 0)) {
+        val fragment: Fragment = when (intent.getIntExtra(EXTRA_PREFERENCE_TYPE, 0)) {
             GENERAL_PREFERENCES -> {
                 setTitle(R.string.action_view_preferences)
                 PreferencesFragment.newInstance()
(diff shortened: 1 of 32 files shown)


charlag:
@connyduck thanks a lot for the review!
I'm not sure about the pool, maybe? GC is much better these days that I can't answer definitely. They are also really tiny, like 32x32.

Nice idea about the message and good job spotting other places! I will try to finish it this week.

charlag:
Does this look better? (it's vpad 8, hpad 12dp)
I have to duplicate the color with alpha for that I think but it's not a big issue

Screenshot_20191221-012211_Tusky

charlag:
I think I fixed some things but there are some bugs, with focal point I believe

connyduck:
> Does this look better?

yes 😍 

Can confirm, there is a bug with focal point. Looks like its not reset when the blurhash is shown.

Also, the play overlay should not be shown while the image is hidden.

Screenshot_1576936883

charlag:
I fixed two things (I'll squash them afterwards).
Do you think that setting should be device-wise or per-account? I think it doesn't make much sense to bind it to an account...

connyduck:
I would make a device wide setting (default blurhash on).

charlag:
I think this is it now.
I would like to refactor all these booleans in one argument, something like mask, like proposed here:
https://www.luu.io/posts/dont-use-booleans/
what do you think?

charlag:
So I went ahead and tried to consolidate these options, I think it looks better. Still too much copypaste and quite some plumbing but at least we won't have to change everything for a new option now.
I'm not exactly sure in which package to put `StatusDisplayOptions` and I'm open to suggestions.
We can also make it a separate PR but I think it's a low-risk change.

charlag:
Oh, I think I should've squashed some commits before merging but that's fine, thanks for the review and test!
I started refactoring settings somewhere on 22nd to get rid of strings scattered across the project and maybe improve settings code readability but I didn't continue yet.

PR #1583, opened 2019-12-13
Make polls respect content warning visibility setting

Tak:
poll-collapsed
poll-expanded

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/adapter/StatusBaseViewHolder.java
+++ b/app/src/main/java/com/keylesspalace/tusky/adapter/StatusBaseViewHolder.java
@@ -180,7 +180,6 @@ public abstract class StatusBaseViewHolder extends RecyclerView.ViewHolder {
                                         @NonNull List<Emoji> emojis,
                                         @Nullable PollViewData poll,
                                         final StatusActionListener listener) {
-        setupPoll(poll, emojis, listener);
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/keylesspalace/tusky/adapter/StatusBaseViewHolder.java
```diff
@@ -178,11 +178,13 @@ protected void setSpoilerAndContent(boolean expanded,
                                         @Nullable String spoilerText,
                                         @Nullable Status.Mention[] mentions,
                                         @NonNull List<Emoji> emojis,
+                                        @Nullable PollViewData poll,
                                         final StatusActionListener listener) {
+        setupPoll(poll, emojis, listener);
```
connyduck: This is called even when the poll is currently not visible, right? Why not move it somewhere where its only executed when the poll is actually visible, e.g. into `setTextVisible`?

Tak: I just kept the old code flow as much as possible - I can make this change if you prefer

connyduck: Should be a quick change, so yes please do it

Tak: Updated

Issue #1586, opened 2019-12-15
Toolbar back button doesn’t work in Scheduled toots page


ariasuni:
The back button in the top toolbar doesn’t work in the scheduled toots page. Back button of my phone works correctly.

* * * *
- Tusky Version: 9.1-4573b320
- Android Version: 9
- Android Device: Oneplus 5

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

PR #1591, opened 2019-12-20
ScheduledTootActivity: fix toolbar back button #1586

a1batross:
Added forgotten onOptionsItemSelected, so toolbar back button does work again.

However, I would rather suggest move onOptionsItemSelected to BaseActivity.java and remove copypasted code from every activity that has toolbar with back button but it's not intuitive and theoretically may break unexpectedly.


connyduck:
thx

PR #1592, opened 2019-12-20
WIP: First draft of OCR image captions

Tak:
This is a quick hack-in of autosuggesting image captions using Tesseract OCR (specifically [tess-two](https://github.com/alexcohn/tess-two))

~~Right now, it's hardcoded to english, and requires that you have [training data](https://github.com/tesseract-ocr/tessdata/blob/3.04.00/eng.traineddata) stored on the device at `/storage/emulated/0/Android/data/com.keylesspalace.tusky/files/OCR/tessdata/`~~

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -170,7 +170,7 @@ dependencies {
     implementation "com.theartofdev.edmodo:android-image-cropper:2.8.0"
 
     implementation "de.c1710:filemojicompat:1.0.17"
-    implementation "com.rmtheis:tess-two:9.0.0"
+    implementation "com.rmtheis:tess-two:9.1.0"
 
     testImplementation "androidx.test.ext:junit:1.1.2"
     testImplementation "org.robolectric:robolectric:4.4"
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -170,6 +170,7 @@ dependencies {
     implementation "com.theartofdev.edmodo:android-image-cropper:2.8.0"
 
     implementation "de.c1710:filemojicompat:1.0.17"
+    implementation "com.rmtheis:tess-two:9.0.0"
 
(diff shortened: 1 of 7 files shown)


charlag:
Cool!
Questions:
can we download training data at runtime?
we can let a user pick a language too
I think there should be just a button to initiate all of that. I like how it's in Pinafore/vanilla web client. How do they detect language btw?

a1batross:
How much APK increase by Tesseract libraries? It's important to include libraries at least for x86, armeabi-v7a and aarch64.

Tak:
Yes, the data just needs to be there before we (re)initialize our tesseract instance.
The training data sets are ~5-50MB in size (English is 20, most European languages are 5-15, simplified Chinese is 40).
I was thinking we could use a similar UI to the one we have for downloading emoji packs.
You have to manually specify the set of languages you want it to support at initialization time. I didn't look into how it selects between specified languages - I presumed it just picks the best match, but 🤷‍♂ 

The size of my debug apk was 23MB (I didn't put training data into the apk), but I'm not sure what architectures are included in that by default.

charlag:
Thanks for details. It all sounds good expect for the APK size IMO - 23 mb is *ooof*
It would be nice if there was non-native build of that because we don't always need top performance, like even TesseractJS has acceptable performance for most use-cases.

a1batross:
The thing is TesseractJS is emscripten port of original C++ library. You can't compile C++ code into JVM or DEX bytecode yet. :(

>I'm not sure what architectures are included in that by default.

You can open APK with any archiver and look what's inside of `lib` folder. 

Anyway, it's very cool feature.

Tak:
This one's not going anywhere, may as well close it

PR #1595, opened 2019-12-20
upgrade androidx.browser to 1.2.0

connyduck:
They added a new method to set the Navigation bar color of the custom tab, so lets use that.


app/src/main/java/com/keylesspalace/tusky/LoginActivity.kt
```diff
@@ -345,9 +346,16 @@ class LoginActivity : BaseActivity(), Injectable {
         private fun openInCustomTab(uri: Uri, context: Context): Boolean {
 
             val toolbarColor = ThemeUtils.getColor(context, R.attr.custom_tab_toolbar)
-            val customTabsIntent = CustomTabsIntent.Builder()
+            val customTabsIntentBuilder = CustomTabsIntent.Builder()
                     .setToolbarColor(toolbarColor)
-                    .build()
+
+            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O_MR1) {
```
connyduck: We are only coloring the nav bar on Api 27+, so lets do the same here

app/build.gradle
```diff
@@ -107,7 +107,8 @@ dependencies {
     implementation "androidx.core:core-ktx:1.2.0-beta01"
     implementation "androidx.appcompat:appcompat:1.1.0"
     implementation "androidx.fragment:fragment-ktx:1.1.0"
-    implementation "androidx.browser:browser:1.0.0"
+    implementation "androidx.browser:browser:1.2.0"
+    implementation "androidx.swiperefreshlayout:swiperefreshlayout:1.0.0"
```
connyduck: looks like `browser:1.0.0` had a dependency on `swiperefreshlayout` that we are now missing

charlag: Is it like... peer dependency?

PR #1596, opened 2019-12-20
upgrade dependencies

connyduck:
these should work wih no code changes


charlag:
Just remembered that F-Droid doesn't have Gradle 6.0.1 yet (at least it's not deployed) https://gitlab.com/fdroid/admin/issues/128
I will ask them just in case

PR #1597, opened 2019-12-20
Implement identity proof api

connyduck:
I don't know if `MediatorLiveData` is the way to go here... It causes more adapter updates than necessary, but works fine.
Other possibility would be to move this to Rx and zip the two network requests, have only one `LiveData` with a `Pair<Account, List<IdentityProofs>`.

Also, instead of `Either<IdentityProof, Field>` I could map the IdentityProofs to Fields 🤔 

Thoughts?


connyduck:
oh combineOptionalLiveData is nice, thx

PR #1598, opened 2019-12-21
fix imagebutton color

connyduck:
Regression from #1541 

device-2019-12-21-114043


connyduck:
I ran `git bisect` and its definitely from compose refactoring, but I don't know why we changed it there either

PR #1601, opened 2019-12-21
delete unused stuff

connyduck:
Leftovers from #1541 and #1537


connyduck:
We don't have a general compose shortcut anymore, but one for every logged in account.

PR #1612, opened 2019-12-31
fix poll button being visible below voted poll

connyduck:
cc @Tak

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/adapter/StatusBaseViewHolder.java
+++ b/app/src/main/java/com/keylesspalace/tusky/adapter/StatusBaseViewHolder.java
@@ -215,8 +215,11 @@ public abstract class StatusBaseViewHolder extends RecyclerView.ViewHolder {
             LinkHelper.setClickableText(this.content, emojifiedText, mentions, listener);
             if (poll != null) {
                 setupPoll(poll, emojis, listener);
+            } else {
+                hidePoll();
             }
         } else {
+            hidePoll();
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/keylesspalace/tusky/adapter/StatusBaseViewHolder.java
```diff
@@ -215,23 +215,24 @@ private void setTextVisible(boolean expanded,
...
         } else {
+            hidePoll();
             LinkHelper.setClickableMentions(this.content, mentions, listener);
         }
         if (TextUtils.isEmpty(this.content.getText())) {
             this.content.setVisibility(View.GONE);
         } else {
             this.content.setVisibility(View.VISIBLE);
         }
-        setPollVisible(poll != null && expanded);
```
charlag: I'm sorry but I look at this for a few minutes already and I don't see the difference with the old logic

connyduck: The problem with the old logic is that `setPollVisible(true)` was called after the poll was setup, and `setPollVisible(true)` makes the button visible, no matter what.
Now the method is called `hidePoll` and is never called together with `setUpPoll`.

charlag: I see now, thanks, LGTM

charlag: (maybe this explanation should be in commit message :wink: )

PR #1613, opened 2020-01-01
use RxAwareViewModel for more ViewModels

connyduck:


app/src/main/java/com/keylesspalace/tusky/components/report/ReportViewModel.kt
```diff
@@ -132,62 +123,61 @@ class ReportViewModel @Inject constructor(
     }
 
     fun toggleMute() {
-        val single: Single<Relationship> = if (muteStateMutable.value?.data == true) {
+        muteStateMutable.value = Loading()
+
+        if (muteStateMutable.value?.data == true) {
```
charlag: Here and in `toggleBlock` you assign the value and check it right away. Before it was checking first.

connyduck: Looks like I was too fast, thx

PR #1614, opened 2020-01-01
Fix missing cases of using StatusDisplayOptions

charlag:
Sorry for the delay!

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/adapter/StatusBaseViewHolder.java
+++ b/app/src/main/java/com/keylesspalace/tusky/adapter/StatusBaseViewHolder.java
@@ -87,7 +87,6 @@ public abstract class StatusBaseViewHolder extends RecyclerView.ViewHolder {
 
     private PollAdapter pollAdapter;
 
-    private boolean useAbsoluteTime;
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/keylesspalace/tusky/adapter/StatusBaseViewHolder.java
```diff
@@ -198,23 +197,21 @@ protected void setSpoilerAndContent(boolean expanded,
...
     }
 
-    private void setTextVisible(boolean expanded,
-                                Spanned content,
-                                Status.Mention[] mentions,
-                                List<Emoji> emojis,
-                                @Nullable PollViewData poll,
+    private void setTextVisible(boolean expanded, Spanned content, Status.Mention[] mentions,
+                                List<Emoji> emojis, @Nullable PollViewData poll,
+                                StatusDisplayOptions statusDisplayOptions,
```
connyduck: I would prefer one arg per line

charlag: Oops, that's what Intellij did when I moved arguments order I guess

PR #1615, opened 2020-01-01
Settings refactor

charlag:
This is a teaser for my proposal: to not use XMLs for settings but use Kotlin's expressiveness to add a tiny DSL.
My main motivation is that we use damn strings around the project to get/update preference. I think later we should add a higher-level well-typed prefs abstraction but this is kinda different story and it would still have duplication with XML if we left it.
Another benefit is that we don't need to do clumsy "find the pref and update it" dance.

If you are wondering if there is a project which does that already than yes, there is one but it replaces androidx.preference completely. It's a much bigger library with kinda different purpose and the DSL is like 30 lines or so.

Let me know what you think about it!

cc @Tak


connyduck:
🤔 fancy

What I think would be really cool as well, is a `Preference` class that you can inject where you need it and it gives you all prefs in a typesafe way. Right now we do `preferences.getString("statusTextSize", "medium")` etc which is really bad

charlag:
Yea, absolutely I want this but it's kinda orthogonal to the pref screen itself

connyduck:
Yes, we should do it in a separate PR.
Do you want to make more changes here or can I make review it?

charlag:
I could go on and convert more stuff (i think there was something else) if it's okay with you

charlag:
I believe it's ready for review now. It's mostly mechanical changes but I was able to simplify some things a little bit.
I didn't try notifications screen yet but maybe you will test it anyway, I can do it otherwise.

app/src/main/java/com/keylesspalace/tusky/fragment/preference/PreferencesFragment.kt
```diff
@@ -20,58 +20,182 @@ import androidx.preference.Preference
...
+import com.keylesspalace.tusky.settings.*
 import com.keylesspalace.tusky.util.ThemeUtils
 import com.keylesspalace.tusky.util.getNonNullString
 import com.mikepenz.iconics.IconicsDrawable
 import com.mikepenz.iconics.typeface.library.googlematerial.GoogleMaterial
 import com.mikepenz.iconics.utils.colorInt
-import com.mikepenz.iconics.utils.sizeRes
+import com.mikepenz.iconics.utils.sizePx
 
 fun PreferenceFragmentCompat.requirePreference(key: String): Preference {
```
connyduck, line 31: this is now unused

charlag:
I think I fixed notifications screen now, the problem was with dependencies. In best Android fashion, XML and programmatic APIs are not fully equivalent, in XML dependency sets the preference key to be registered later, in programmatic it tries to register it directly (which might be too early, e.g. if we didn't assign the screen yet).
I worked around this by changing the order of things: first screen is attached and then we modify it. Seems to work.

connyduck:
Nice! Do you want to replace all the string literals with the constants from `SettingsConstants` as well in this PR? If not please remove the WIP and we can do it later.

charlag:
I thought about it and I think we should use some more high-level preference interface so I don't think it makes sense to replace all constants

PR #1624, opened 2020-01-06
Fix search bugs

connyduck:
This PR

- fixes toggling media hidden state (was broken because differ received the same list again)
- fixes bookmarking in search (copy/paste mistake)
- cleans up the code a bit (actually it is already quite good compared to Java code we have in the project)

There is another bug that I did not fix because it seems to only appear on emulators: 
- when using the emulator hardware keyboard, search is triggered twice, which messes with the cache and statuses randomly appear, see https://stackoverflow.com/questions/12929590/search-widget-call-onnewintent-twice

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/components/search/SearchViewModel.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/components/search/SearchViewModel.kt
@@ -31,9 +31,9 @@ class SearchViewModel @Inject constructor(
             accountManager.activeAccount = value
         }
 
-    val mediaPreviewEnabled: Boolean = activeAccount?.mediaPreviewEnabled ?: false
-    val alwaysShowSensitiveMedia: Boolean = activeAccount?.alwaysShowSensitiveMedia ?: false
-    val alwaysOpenSpoiler: Boolean = activeAccount?.alwaysOpenSpoiler ?: false
+    val mediaPreviewEnabled = activeAccount?.mediaPreviewEnabled ?: false
+    val alwaysShowSensitiveMedia = activeAccount?.alwaysShowSensitiveMedia ?: false
+    val alwaysOpenSpoiler = activeAccount?.alwaysOpenSpoiler ?: false
 
     private val statusesRepository = SearchRepository<Pair<Status, StatusViewData.Concrete>>(mastodonApi)
     private val accountsRepository = SearchRepository<Account>(mastodonApi)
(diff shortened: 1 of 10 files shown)


app/src/main/java/com/keylesspalace/tusky/components/search/adapter/SearchDataSource.kt
```diff
@@ -51,7 +51,7 @@ class SearchDataSource<T>(
     @SuppressLint("CheckResult")
     override fun loadInitial(params: LoadInitialParams, callback: LoadInitialCallback<T>) {
         if (!initialItems.isNullOrEmpty()) {
-            callback.onResult(initialItems, 0)
+            callback.onResult(initialItems.toList(), 0)
```
charlag: I still think it's a shame that there's no more readable/performant alternative

app/src/main/java/com/keylesspalace/tusky/components/search/SearchViewModel.kt
```diff
@@ -33,49 +31,46 @@ class SearchViewModel @Inject constructor(
             accountManager.activeAccount = value
         }
 
-    val mediaPreviewEnabled: Boolean
-        get() = activeAccount?.mediaPreviewEnabled ?: false
+    val mediaPreviewEnabled: Boolean = activeAccount?.mediaPreviewEnabled ?: false
```
charlag: idk if we really need a type here

charlag: But feel free to leave it ofc

connyduck:
There was another bug: Search for a status and interact with it (e.g. toggle spoiler, fav, toggle media visibility). There will be an unnecessary network request to load another page. So I added a boolean to remember when a `DataSource` is exhausted, which fixes the problem. Not sure if this is the best way to do it though.

connyduck:
But how can I keep it local to the `DataSource`? A new one gets created when the old one is invalidated and the state is lost.

charlag:
@connyduck but do you ever invalidate it? If not, there should be no other created (if I understand this machinery correctly)

connyduck:
Its always invalidated when you interact with statuses (`repoResultStatus.value?.refresh?.invoke()` in the ViewModel points to `sourceFactory.sourceLiveData.value?.invalidate()` in the Repo)

charlag:
ahhh that might be, but I cannot find where? The only place I see is in `refresh` of Listing
(I think that we still have the case which is described in the doc and should cache it somehow mb)

connyduck:
?? The refresh function is called in the ViewModel

PR #1625, opened 2020-01-06
remove redundant fromHtml from StatusBaseViewHolder

connyduck:


connyduck:
because theres no more `<b>` in the strings

PR #1628, opened 2020-01-07
Implement local bookmarks

charlag:
We already had implementation of bookmarks which relies on the server
but most servers don't support this yet so we needed a local fallback.
This commits adds such a fallback. Bookmarked statuses are saved to
the DB and never cleaned up. We fetch them likely any other statuses
from DB with additional condition.

There's only one problem one: in timeline the bookmark state is not displayed correctly when we load statuses from network. I don't think it's a big problem but I think we could do one of two things:
1. Load all/range bookmarks in repository from db and check which are bookmarked
2. Hide bookmark button (that's what vanilla web client does now btw)

### DIFF
--- a/app/src/androidTest/java/com/keylesspalace/tusky/TimelineDAOTest.kt
+++ b/app/src/androidTest/java/com/keylesspalace/tusky/TimelineDAOTest.kt
@@ -55,6 +55,34 @@ class TimelineDAOTest {
     }
 
     @Test
+    fun insertGetBookmarked() {
+        val setOne = makeStatus(statusId = 3, bookmarked = true)
+        val setTwo = makeStatus(statusId = 20, reblog = true)
+        val ignoredOne = makeStatus(statusId = 1)
+
+        for ((status, author, reblogger) in listOf(setOne, setTwo)) {
+            timelineDao.insertInTransaction(status, author, reblogger)
+        }
+
+        val resultsFromDb = timelineDao.getStatusesForAccount(
+                account = setOne.first.timelineUserId,
+                maxId = "21",
+                sinceId = ignoredOne.first.serverId,
+                limit = 10,
+                bookmarkedOnly = true
+        )
+                .blockingGet()
+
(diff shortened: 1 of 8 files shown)


connyduck:
This idea of this is actually awesome :thinking:

I found some bugs though:
- Locally bookmarked statuses are not marked as bookmarks in timeline after restarting the app or scrolling the status out of view.

- when loading the bookmarks fails for any reason, it falls back to displaying local bookmarks, which is very weird when the instance actually supports bookmarks. I think we need to check for the specific 404 error.

> I don't think it's a big problem

I think its related to the first point above, and thats super annoying. (Why would removing the bookmark button solve it? You can't bookmark anything then?)

Also, adding more and more undeleteable statuses to the db could turn out to be problematic.

Local bookmarks will be inaccesible after an instance updates to bookmark support, right? Users will hate that.

**Maybe the better solution is to do server version detection and hide the bookmarks if not supported?**

charlag:
Thanks for the review!

So, yeah, I agree I suppose, especially with the fact that users would lose their bookmarks (we could show them below server bookmarks but it can have duplicates and stuff so it"s probably a bad idea).

Let me clarify my point with button, even though it's not relevant. My proposal with button would be to put it inside the "more" menu and make it always bookmark things and only unbookmark from bookmarks screen. Which is fairly weird experience probably but I think that you don't see the same statuses all the time.

We can never be fully offline-first app as long as Mastodon API doesn't have tool for that. *sigh*

So, for bookmark feature detection, should we kindly ask @Gargron to add a flag or that's not what vanilla Mastodon does?

connyduck:
> So, for bookmark feature detection, should we kindly ask @Gargron to add a flag or that's not what vanilla Mastodon does?

It would help to know under which version it will be released and then we use the instance endpoint and our `VersionUtils`

Gargron:
The next release of Mastodon, most likely called v3.1.0, will have the bookmarks API.

PR #1629, opened 2020-01-08
Add support for sharing audio attachments

Tak:


charlag:
Wait, that's it 🤔

Tak:
That's it for sharing from Tusky to another app that's registered for the matching mime type (e.g. mail, telegram), since we already had everything in place for video.

It does remind me that I should register Tusky as a target for audio sharing in the other branch…

Issue #1337, opened 2019-06-22
Add support for recording and uploading audio notes


connyduck:

wakest:
voice-microblogging! micropodcasting!

novadeviator:
yes, please.
reminds me of 'dial-a-poem'.
too bad opus is not really supported. such a superb codec.

bonsaiheldin:
@wakest I love this idea! Imagine you speak a short message instead of writing a toot of it and someone else reacts with spoken words, too. That would open a whole new way of communicating known only rudimentarily from the podcast scene where in podcasts people react to other podcasts or comments. Speaking transfers so much more information than text: Your mood, even facial expressions like smiling.

There would be an accessibility issue (going beyond Tusky's scope): Deafness would be a problem as there is no text to read. You would have to also write the toot, turning it into a transcript which means more effort. On the other hand blind people could just listen to the audio file instead of their screen reader, assuming one speaks clear enough.

@connyduck: Looking very very forward to this feature! 😬

bonsaiheldin:
@connyduck Looking forward to this feature. Especially the handy recording function. 😃

koyuawsmbrtn:
Still an issue. Hope Tusky 10 comes soon.

charlag:
What's the point of saying "still an issue"?
No one worked on this as far as I know

novadeviator:
Is recording pretty much synonymous with uploading here/on Android? It would be great to first work just on uploading of audio files and recording is a separate enhancement?

PR #1630, opened 2020-01-08
Add support for uploading audio attachments

Tak:
Partially addresses #1337


Tak:
Updated

connyduck:
Yeah, better color. Not perfect, I may change it before the release, but for now it does the job.

PR #1631, opened 2020-01-09
Retain content warning when replying to a CWed post

Tak:
This seems to have regressed with the `ComposeActivity` refactor - if you reply to a post with a content warning, the CW is carried over to the reply, but the content warning toggle in the newly-created compose activity isn't activated.

The proposed fix sets `showContentWarning` in the view model when it's set up with a nonnull CW.
The `setupComplete` roundtrip is there because the event for showing the content warning is dispatched _after_ the activity creation finishes, so without it, the focus is incorrectly set to the CW field.


connyduck:
Thx for the fix!

This extra LiveData just for that is really weird, but I don't have a better idea either 🤔

Tak:
Yeah, I searched a little bit to see if we could do something else, like force the dispatcher to flush outstanding events or something, but I didn't come up with anything

PR #1632, opened 2020-01-10
Add space before emojo code in editor when needed, fix #1616

### LINKED ISSUE #1616: [Bug report] Emoji short code incorrect
Insert a emoji by short code ` :emoji_code: `, there should be a white space before and after the colon `:`.

The app insert without the white space by default result in:

charlag:


app/src/main/java/com/keylesspalace/tusky/components/compose/ComposeActivity.kt
```diff
@@ -407,7 +407,11 @@ class ComposeActivity : BaseActivity(),
         // If you select "backward" in an editable, you get SelectionStart > SelectionEnd
         val start = composeEditField.selectionStart.coerceAtMost(composeEditField.selectionEnd)
         val end = composeEditField.selectionStart.coerceAtLeast(composeEditField.selectionEnd)
-        composeEditField.text.replace(start, end, text)
+        val textToInsert = if (
+                composeEditField.text.isNotEmpty()
+                && !composeEditField.text[start - 1].isWhitespace()
+        ) " $text" else text
+        composeEditField.text.replace(start, end, textToInsert)
```
Tak: This is also used for the #/@ special character buttons - does this change also make sense for those?

charlag: I've seen it and I think so? They are not pasted as special sequences without leading space, at least by our editor

PR #1643, opened 2020-01-15
Enable zooming and panning in the CaptionDialog

charlag:
At first I thought that there will be problems with sizing. Then I thought it's new behaviour that the dialog consumes the full height.

I think it just... works? If it works like it is, it will be probably the highest "changes to usefulness ratio" I contributed lol.


Tak:
It seems to work for me.
The maximum zoom level seems to be 100% (actual image resolution), maybe that's ok?

charlag:
I think that's enough but maybe there's even a way to change that

PR #1646, opened 2020-01-17
Make avatars lighter, not darker in light theme, fix #1258

### LINKED ISSUE #1258: Darker avatars on post interactions look out of place on light theme
When others interact with my posts by favouring or boosting them, Tusky shows this in the 'Notifications' column. The avatars of the accounts involved get darkened currently, which looks great on dark themes, but really out of place on the light theme.

It would be great if this could be adjusted. Just making avatars lighter, instead of darker, might do the trick.

* * * *
- Tusky Version: 6.1

charlag:
We will probably throw it away but I experimented with making avatars brighter, not darker on light theme. Didn't go that well.
Multiple can only make something darker (because white is like 1 and other colors are fractions of it) so we have to add in some smart way. Also we have to take care of rounded corners, we cannot subtract anything from them.
Here is a useful link with exmaples:
https://medium.com/better-programming/practical-image-porterduff-mode-usage-in-android-3b4b5d2e8f5f
Here are docs or PorterDuff modes:
https://developer.android.com/reference/android/graphics/PorterDuff.Mode.html#DARKEN
I think Screen mode would be perfect but it screws up corners so I'm not sure what to do about it.

Maybe we could even apply filter to the bitmap itself and not to the image.

Screenshot_20200118-001915

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/adapter/NotificationsAdapter.java
+++ b/app/src/main/java/com/keylesspalace/tusky/adapter/NotificationsAdapter.java
@@ -17,7 +17,10 @@ package com.keylesspalace.tusky.adapter;
 
 import android.content.Context;
 import android.graphics.Color;
+import android.graphics.ColorFilter;
+import android.graphics.LightingColorFilter;
 import android.graphics.PorterDuff;
+import android.graphics.PorterDuffColorFilter;
 import android.graphics.Typeface;
 import android.graphics.drawable.Drawable;
 import android.text.InputFilter;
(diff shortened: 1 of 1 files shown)


connyduck:
Ah yes, I thought this would be easier. Lighting filter does not look that good, PorterDuff.LIGHTEN or SCREEN look good but srew up the corners :/
Lets postbone it to the next release

charlag:
Yeah ideally we would just use alpha but with this small avatar in the corner it doesn't really work.
As I said, we could apply filter to the bitmap before displaying but I agree with postponing it, it's absolutely non-critical.

PR #1647, opened 2020-01-18
fix FragmentAdapter returning null fragments when activity was recreated

connyduck:
Steps to reproduce: Enable "Don't keep activities" in the developer settings and bring the main activity in the background and foreground again (or recreate it in an other way). Jumping to the top by clicking the selected tab will no longer work.

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/util/CustomFragmentStateAdapter.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/util/CustomFragmentStateAdapter.kt
@@ -19,7 +19,9 @@ import androidx.fragment.app.Fragment
 import androidx.fragment.app.FragmentActivity
 import androidx.viewpager2.adapter.FragmentStateAdapter
 
-abstract class CustomFragmentStateAdapter(val activity: FragmentActivity): FragmentStateAdapter(activity) {
+abstract class CustomFragmentStateAdapter(
+        private val activity: FragmentActivity
+): FragmentStateAdapter(activity) {
 
     fun getFragment(position: Int): Fragment?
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/keylesspalace/tusky/util/CustomFragmentStateAdapter.kt
```diff
@@ -0,0 +1,26 @@
...
+ * You should have received a copy of the GNU General Public License along with Tusky; if not,
+ * see <http://www.gnu.org/licenses>. */
+
+package com.keylesspalace.tusky.util
+
+import androidx.fragment.app.Fragment
+import androidx.fragment.app.FragmentActivity
+import androidx.viewpager2.adapter.FragmentStateAdapter
+
+abstract class CustomFragmentStateAdapter(val activity: FragmentActivity): FragmentStateAdapter(activity) {
```
connyduck: No idea what a good name for this class would be :D

charlag: `FragmentSearchFragmentStateAdapter`? `FragmentAccessorFragmentStateAdapter`?

Also I would make `activity` private.

charlag: Maybe I would even replace `FragmentActivity` with something more generic in case of nested fragments but seems like there's no interface for "FragmentContainer" :roll_eyes:

connyduck: It has to be `FragmentActivity` because the constructor of  `FragmentStateAdapter` requires it

PR #1649, opened 2020-01-18
Make image in BackgroundMessageView adapt to the height, fix #1618

### LINKED ISSUE #1618: Network error message on profile doesn’t work well in landscape
Because on the profile screen there is the tabs bar, in landscape the button is flattened and almost unreadable, at least with my screen size:

Screenshot_20200102-182128_Tusky

To reproduce, just put your telephone in landscape, disable network/enable airplane mode, and try to open your profile or somebody else’s, and scroll to the bottom.

I think the error message should make a better use of the available horizontal space (mockup):

Screenshot_20200102-182128_Tusky_mockup

* * * *
- Tusky Version: 9.1/9.1-352ff3a2
- Android Version: 9
- Android Device: One Plus 5

charlag:
I tried to fix it but not like proposed in #1618 because it would require additional layouts so I tried to just fit the Image.
The only problem is notifications screen where we have this damn bar. We need to offset top correctly and we need to fix the bar height for that but idk how it would work with text scaling.
Screenshot_20200118-220346

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/fragment/NotificationsFragment.java
+++ b/app/src/main/java/com/keylesspalace/tusky/fragment/NotificationsFragment.java
@@ -167,6 +167,7 @@ public class NotificationsFragment extends SFragment implements
     private boolean alwaysShowSensitiveMedia;
     private boolean alwaysOpenSpoiler;
     private boolean showNotificationsFilter;
+    private boolean showingError;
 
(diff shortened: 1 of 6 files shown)


connyduck:
Maybe just hide the bar in case of error? But I also have nothing against two layouts.

charlag:
Hiding a bar is a good idea actually :thinking:

charlag:
I hope it did it

PR #1650, opened 2020-01-18
Add options to confirm reblog/unreblog actions, close #460

### LINKED ISSUE #460: Feature: Reduce accidental boosts/favorites, either by hiding or requiring confirmation
In Tusky, it's easy to accidentally favorite/boost toots that you did not intend to during the normal act of scrolling.

To editorialize somewhat: this is one of the things that always bothered me about the official Twitter client: it feels more like a feature that's meant to "drive engagement" rather than one that's meant to help the user. A user that wants to boost/favorite likely knows that this action can be performed, so it can be tucked away somewhere where it's not done accidentally.

I've got a few thoughts on how this could be improved:

# Add an option to require confirmation before boost/favorite

This wouldn't change the existing design, and could be done as a preference for users that wish to control the behavior. Adding options (one for boost, one for favorite) that would add a confirmation step. 

Example workflow:

* User navigates to preferences, enables "Require confirmation before boost" preference
* User navigates to timeline, taps "boost" on a toot
* _App prompts user: "Are you sure you would like to boost this toot?"_
    * User selects "yes": toot is boosted
    * User selects "no": toot is not boosted

What's possibly neat about this is the UX for actually gaining the confirmation could be handled in clever ways, like the first tap on the "boost" button changes the button to a confirm icon, and the second tap actually boosts it. Could be handled in a plain-old dialog or some sort of pop-over, however.

# Hide the boost/favorite buttons behind some user interaction

This could possibly clean up the reading experience some, but is less discoverable: Hiding the favorite/reply/boost buttons behind some interaction, such as a long-press on the toot, or requiring a tap through to the toot to the detail view, where the buttons could be shown.

Again, the UX could be more pleasant, say: the buttons are not shown, but a long-press on the toot "expands" the buttons on the bottom of it.

---

Those are just some suggestions, but in general I think it'd be ideal to add confirmation to anything which could either alter your own timeline or notify another user in some way, and accidental favorite/boost is one that's really easy to do from the timeline.

charlag:
Add an option to confirm reblogs.
Initially I wanted to put it much further the pipeline but to get animations work right I had to work around `SparkButton` a little bit. The advantage is that it's implemented in just one place now.

It tries to mimick Masto UI but not very closely.

image

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/adapter/StatusBaseViewHolder.java
+++ b/app/src/main/java/com/keylesspalace/tusky/adapter/StatusBaseViewHolder.java
@@ -19,6 +19,7 @@ import android.widget.Toast;
 import androidx.annotation.DrawableRes;
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
+import androidx.appcompat.app.AlertDialog;
 import androidx.recyclerview.widget.DefaultItemAnimator;
 import androidx.recyclerview.widget.LinearLayoutManager;
 import androidx.recyclerview.widget.RecyclerView;
@@ -209,7 +210,7 @@ public abstract class StatusBaseViewHolder extends RecyclerView.ViewHolder {
             contentWarningDescription.setVisibility(View.VISIBLE);
             contentWarningButton.setVisibility(View.VISIBLE);
             setContentWarningButtonText(expanded);
-            contentWarningButton.setOnClickListener( view -> {
+            contentWarningButton.setOnClickListener(view -> {
                 contentWarningDescription.invalidate();
                 if (getAdapterPosition() != RecyclerView.NO_POSITION) {
                     listener.onExpandedChange(!expanded, getAdapterPosition());
@@ -227,7 +228,7 @@ public abstract class StatusBaseViewHolder extends RecyclerView.ViewHolder {
     }
 
     private void setContentWarningButtonText(boolean expanded) {
-        if(expanded) {
+        if (expanded) {
             contentWarningButton.setText(R.string.status_content_warning_show_less);
         } else {
             contentWarningButton.setText(R.string.status_content_warning_show_more);
@@ -610,7 +611,9 @@ public abstract class StatusBaseViewHolder extends RecyclerView.ViewHolder {
         sensitiveMediaShow.setVisibility(View.GONE);
     }
 
-    protected void setupButtons(final StatusActionListener listener, final String accountId) {
+    protected void setupButtons(final StatusActionListener listener, final String accountId,
+                                final String statusContent,
+                                StatusDisplayOptions statusDisplayOptions) {
 
         avatar.setOnClickListener(v -> listener.onViewAccount(accountId));
         replyButton.setOnClickListener(v -> {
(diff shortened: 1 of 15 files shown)


app/src/main/java/com/keylesspalace/tusky/adapter/StatusBaseViewHolder.java
```diff
@@ -601,12 +605,25 @@ protected void setupButtons(final StatusActionListener listener, final String ac
...
-                if (position != RecyclerView.NO_POSITION) {
-                    listener.onReblog(buttonState, position);
-                }
-            });
+            if (statusDisplayOptions.confirmReblogs()) {
+                reblogButton.onClickListener = (buttonStatus) -> {
+                    int okButtonTextId =
+                            buttonStatus ? R.string.action_reblog : R.string.action_unreblog;
+                    new AlertDialog.Builder(reblogButton.getContext())
+                            .setMessage(statusContent)
```
Tak: Shouldn't this be clamped to some maximum length? Or does the dialog do this automatically?

app/src/main/java/com/keylesspalace/tusky/fragment/TimelineFragment.java
```diff
@@ -576,6 +577,16 @@ public void onReply(int position) {
     @Override
     public void onReblog(final boolean reblog, final int position) {
         final Status status = statuses.get(position).asRight();
+        boolean confirmReblogs = PreferenceManager.getDefaultSharedPreferences(getContext())
+                .getBoolean("confirmReblogs", false);
+        if (confirmReblogs) {
+
```
connyduck: looks like something is mising here

app/src/main/java/com/keylesspalace/tusky/fragment/TimelineFragment.java
```diff
@@ -576,6 +577,16 @@ public void onReply(int position) {
...
+        } else {
```
connyduck: looks like something is mising here

charlag: I find it funny that you missed that 's' first :laughing:

connyduck: :D

connyduck:
Probably best to add a return value to the `onEvent` of the `SparkEventListener` so you can return false when you don't want the animation to play

connyduck:
Here, I made the change to `SparkButton`: https://github.com/connyduck/SparkButton/pull/2

You can try it out with ` implementation "com.github.connyduck:sparkbutton:1e0d112310"`

app/src/main/java/com/keylesspalace/tusky/adapter/StatusBaseViewHolder.java
charlag: I checked and dialog just becomes scrollable but it's a good point

charlag:
Thanks, it works now (even though it's a breaking change to the library).
I didn't find a way to play animation at first but then I added it. Works fine but you can still break it but triggering it back and forth. Go figure.

I would squash it all into one commit before merging, I just split it so that you see the diff

app/src/main/java/com/keylesspalace/tusky/view/ConditionalSparkButton.java
```diff
@@ -0,0 +1,41 @@
+package com.keylesspalace.tusky.view;
+
+import android.content.Context;
+import android.util.AttributeSet;
+import android.view.View;
+
+import at.connyduck.sparkbutton.SparkButton;
+
+public class ConditionalSparkButton extends SparkButton {
```
connyduck: this class is no longer required

charlag:
>ERROR: Failed to resolve: com.github.connyduck:sparkbutton:4.0.0

connyduck:
try again, build wasn't finished

charlag:
Updated, squashed, rebased (that card thing lead to some headache with commas)

PR #1654, opened 2020-01-24
Refresh DM column when removing toot there instead of removing conversation

ariasuni:
It was crashing Tusky. Since the conversation API is annoying, the only way to properly update the DM column is to refresh it.

Note that composing a new DM or responding to one doesn’t update the DM column, it probably should?


connyduck:
thx

ariasuni:
:slightly_smiling_face:

PR #1661, opened 2020-01-30
fix overlapping icons in ComposeActivity

### LINKED ISSUE #1639: "Hide media" button overlaps with CW button
Easy to reproduce: start composing a toot, upload an image, delete it and then compose it again. After that, "hide media" button will overlap the CW button as show below:
изображение
Not critical, since you can just press that "mixed" button and it will fix itself, but still a bug.

* * * *
- Tusky Version: 9.1-21a7201 (Nightly)
- Android Version: 9
- Android Device: Oneplus 5

connyduck:
closes #1639 

The problem is that `updateSensitiveMediaToggle()` is called way to often and that messes with the `TransitionManager`. Fixed by reducing the number of calls to `updateSensitiveMediaToggle()` and using other transition mechanism.

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/components/compose/ComposeActivity.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/components/compose/ComposeActivity.kt
@@ -113,6 +113,8 @@ class ComposeActivity : BaseActivity(),
     private var composeOptions: ComposeOptions? = null
     private lateinit var viewModel: ComposeViewModel
 
+    private var mediaCount = 0
+
     public override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
         val preferences = PreferenceManager.getDefaultSharedPreferences(this)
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/keylesspalace/tusky/components/compose/ComposeActivity.kt
```diff
@@ -304,9 +306,12 @@ class ComposeActivity : BaseActivity(),
                 setStatusVisibility(visibility)
             }
             viewModel.media.observe { media ->
-                composeMediaPreviewBar.visible(media.isNotEmpty())
                 mediaAdapter.submitList(media)
-                updateSensitiveMediaToggle(viewModel.markMediaAsSensitive.value != false, viewModel.showContentWarning.value != false)
+                if(media.size != mediaCount) {
```
connyduck: Any other idea how I could get the method to be only called when the count changes? Can you dedupe a `LiveData` like you can with Rx?

app/src/main/res/layout/activity_compose.xml
```diff
@@ -274,13 +275,12 @@
         <ImageButton
             android:id="@+id/composeHideMediaButton"
             style="?attr/image_button_style"
-            android:layout_width="40dp"
-            android:layout_height="40dp"
+            android:layout_width="36dp"
+            android:layout_height="36dp"
```
connyduck: all other buttons have 36dp, no idea why this one has 40

app/src/main/java/com/keylesspalace/tusky/components/compose/ComposeActivity.kt
charlag: Not in a built-in way

https://github.com/musichin/ReactiveLiveData/blob/master/reactivelivedata/src/main/java/com/github/musichin/reactivelivedata/ReactiveLiveData.kt#L317

https://github.com/adibfara/Lives/blob/master/lives/src/main/java/com/snakydesign/livedataextensions/Filtering.kt#L34

charlag: I'm a little bit worried that it may change based on something else than media length and theoretically we should combine `media`. `markMediaAsSensitive` and `showContentWarning` into one thing but I think it's fine for now

connyduck: The only thing from `media` that is checked inside `updateSensitiveMediaToggle` is the length so it should be fine

PR #1668, opened 2020-02-07
fix crash when opening Poll dialog and instanceParams are null

connyduck:


connyduck:
ah yes thats a better idea

PR #1675, opened 2020-02-10
simplify DI & test setup, convert TuskyApplication to Kotlin

connyduck:
✅ more Kotlin
✅ makes it way more understandable whats going on imho
✅ the generated bytecode is more efficient since the compiler and R8 can inline more methods


app/build.gradle
```diff
@@ -170,6 +170,7 @@ dependencies {
     testImplementation "org.robolectric:robolectric:4.3.1"
     testImplementation "org.mockito:mockito-inline:3.2.4"
     testImplementation "com.nhaarman.mockitokotlin2:mockito-kotlin:2.2.0"
+    testImplementation "org.conscrypt:conscrypt-openjdk-uber:2.0.0"
```
connyduck: this is so Conscrypt can be initialized when running tests on the JVM and not on Android

connyduck:
Of course the tests fail on CI. Would have been too easy 🙄

app/src/main/java/com/keylesspalace/tusky/di/AppModule.kt
```diff
@@ -64,20 +64,23 @@ class AppModule {
...
     @Singleton
     fun providesEventHub(): EventHub = EventHubImpl
 
     @Provides
     @Singleton
-    fun providesDatabase(app: TuskyApplication): AppDatabase {
-        return app.serviceLocator.get(AppDatabase::class.java)
+    fun providesDatabase(appContext: Context): AppDatabase {
+        return Room.databaseBuilder(appContext, AppDatabase::class.java, "tuskyDB")
+                .allowMainThreadQueries()
```
charlag: I would disallow allowMainThreadQueries back maybe but later

app/src/test/java/com/keylesspalace/tusky/di/AppInjector.kt
```diff
@@ -1,2 +1,11 @@
 package com.keylesspalace.tusky.di
 
+import com.keylesspalace.tusky.TuskyApplication
+
+object AppInjector {
```
charlag: Why do we even need this?

app/build.gradle
charlag: I'm not sure I like it more than stub...

app/src/test/java/com/keylesspalace/tusky/di/AppInjector.kt
connyduck: we don't

app/src/main/java/com/keylesspalace/tusky/di/AppModule.kt
connyduck: YES, but we do main thread queries and that can't be fixed easily (its actually a huge problem, we have quite some app-not-responding errors because of them)

app/build.gradle
connyduck: I replaced it with a better solution

app/src/main/java/com/keylesspalace/tusky/di/AppModule.kt
charlag: Oh, I forgot about that. That's no-no. Maybe you can point to this place later.

charlag:
If you want to move to feature-based structure, I would propose to also create separate subcomponents for each feature. It's a little bit more boilerplate but it makes much more sense imo

connyduck:
Yes thats something we could consider

Issue #1679, opened 2020-02-12
Dividers between toots are missing in black theme


kyori19:
While using black theme, I can't see dividers between toots.
It seems like it's color is set to black.
Sorry it's hard to see because of mozaic.

![Screenshot_20200212_183034](https://user-images.githubusercontent.com/27999567/74321988-8943b280-4dc6-11ea-9654-4914e3b29d95.png)

* * * *
- Tusky Version: 10.0 beta 1 @6f8e8a6
- Android Version: 8.0.0
- Android Device: G8441
- Mastodon instance (if applicable): Any

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

PR #1685, opened 2020-02-13
[Tusky10] adjust poll result background, status divider & button outline color

connyduck:
This is based on feedback from beta users. I split up the poll background and status divider colors so they can be set individually. 

closes https://github.com/tuskyapp/Tusky/issues/1679

### DIFF
--- a/app/src/main/res/values-v27/styles.xml
+++ b/app/src/main/res/values-v27/styles.xml
@@ -7,20 +7,20 @@
         <item name="android:windowNoTitle">true</item>
         <item name="android:windowLightNavigationBar">false</item>
         <item name="android:navigationBarColor">@color/tusky_grey_20</item>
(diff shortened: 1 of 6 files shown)


app/src/main/res/values-v27/styles.xml
```diff
@@ -20,7 +20,7 @@
     <style name="TuskyBlackTheme" parent="TuskyBlackThemeBase">
         <item name="android:windowLightNavigationBar">false</item>
         <item name="android:navigationBarColor">@color/black</item>
-        <item name="android:navigationBarDividerColor">@color/tusky_grey_05</item>
+        <item name="android:navigationBarDividerColor">@color/tusky_grey_10</item>
```
charlag: I'm confused why it's different than splash above and why it's not attr

connyduck: Because I was not careful enough. Thx for checking!

PR #1690, opened 2020-02-14
fix list editing by moving account_ids from query to form param

connyduck:
The param goes in the body as per api definition https://docs.joinmastodon.org/methods/timelines/lists/

The question is: Why did this work before?


connyduck:
I'm not sure about this. Sometimes it works, sometimes it doesn't. What is going on.

connyduck:
Tested it again, pretty sure it works fine. I must have confused branches or something.

PR #1691, opened 2020-02-14
Fix emoji download

charlag:
thanks @connyduck for pointing out OkHttp behavior.

There are individual commits for the fix itself and for reformatting those files


connyduck:
I still would like to know why it worked in Tusky 9.1.

What if we hardcode the size of the files? Then we don't have to rely on the header

charlag:
We can investigate
I don't like hardcoding at all. Header is how it works always with progress anyway. Here we just grab it more directly, without abstraction but there's nothing hacky about it, that' how OkHttp does it

connyduck:
👍

PR #1698, opened 2020-02-18
Warn when scheduling a post within 5 minutes

Tak:
- Pops up an explanatory long toast when scheduling a post within 5:30
- Sets the initial value of the time picker to current time + 15:00

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/components/compose/ComposeActivity.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/components/compose/ComposeActivity.kt
@@ -65,6 +65,8 @@ import com.keylesspalace.tusky.adapter.OnEmojiSelectedListener
 import com.keylesspalace.tusky.components.compose.dialog.makeCaptionDialog
 import com.keylesspalace.tusky.components.compose.dialog.showAddPollDialog
 import com.keylesspalace.tusky.components.compose.view.ComposeOptionsListener
+import com.keylesspalace.tusky.components.compose.view.ComposeScheduleView
+import com.keylesspalace.tusky.components.compose.view.ComposeScheduleView.MINIMUM_SCHEDULED_SECONDS
 import com.keylesspalace.tusky.db.AccountEntity
 import com.keylesspalace.tusky.di.Injectable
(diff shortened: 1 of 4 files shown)


connyduck:
Thx!
This is probably still not optimal, but since I have no better idea right now and its better than the current state, I would merge it for Tusky 10. What do you think, @charlag?

charlag:
I'm all for releasing Tusky 10 already, just think that this should be an alert dialog or something like that

connyduck:
Oh right, it should actually warn after setting the date, not when sending.
What about a snackbar? We use snackbars for most warnings in ComposeActivity.

charlag:
Snackbar is just a notification, this should be a confirmation IMO
Warning when sending is not that bad because the time might change before sending but might be surprising.

Tak:
I intentionally made it on send, because you can schedule a post for +15 minutes, spend 10 minutes writing the post, then send

It could be an alert dialog with the option to cancel / reschedule

connyduck:
How about putting the error message right into the bottom sheet where you edit the time?
So check on sending, when time is too short don't send, open the bottom sheet and show the warning. I think thats the most useful since you will need less clicks to fix it.

Tak:
Oh, I missed the bottom sheet suggestion - won't that make the sheet huge? Or is that ok?

connyduck:
Well it should get into the way of the users so I think its ok even if it gets big on smaller screens or some languages.

Tak:
Ok, moved to bottom sheet.
Additionally, it's getting verified when setting the time _and_ when trying to post.

connyduck:
Nice! There is a small bug though: When I schedule for >5 min then wait until its <5min to send, it is not sending, but the error message is not showing up. When scheduling for <5 it works as expected.

Tak:
Hm, I was sure I tested that case - I'll look again.

connyduck:
When can you do so? I would like to release tomorrow with this PR included.

Tak:
Sorry, I'll try to find time

Tak:
:heavy_check_mark:

connyduck:
Awesome, thx!

PR #1702, opened 2020-02-19
Add more URL cases for Pleroma, Misskey and Friendica

a1batross:
* Pleroma uses flake ids on /notice/ URL. Flake ids never consist of anything than a-zA-Z0-9
* Misskey uses ids that seems to be always in lowercase and numbers
* Friendica profiles are on /profile/%s
* Friendica posts are using typical UUID

Also, did replaced [^\] to \w for Pleroma, because \w already includes lowercase, uppercase, numbers and '_' symbols which are acceptable in handles. Don't know about Mastodon, so I didn't touched it.


connyduck:
Thank you!

PR #1707, opened 2020-02-22
Compose activity fixes

connyduck:
This PR does three things (individual commits)

- cleanup the image keyboard support related code. I tried to stick with this tutorial as closely as possible: https://developer.android.com/guide/topics/text/image-keyboard
- fix a crash when rotating the phone while taking an image with the camera
- make the poll button visible again in dark theme when it is disabled

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/components/compose/ComposeActivity.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/components/compose/ComposeActivity.kt
@@ -145,6 +145,8 @@ class ComposeActivity : BaseActivity(),
         subscribeToUpdates(mediaAdapter)
         setupButtons()
 
+        photoUploadUri = savedInstanceState?.getParcelable(PHOTO_UPLOAD_URI_KEY)
+
         /* If the composer is started up as a reply to another post, override the "starting" state
          * based on what the intent from the reply request passes. */
         if (intent != null) {
@@ -470,7 +472,7 @@ class ComposeActivity : BaseActivity(),
     }
 
     override fun onSaveInstanceState(outState: Bundle) {
-        outState.putParcelable("photoUploadUri", photoUploadUri)
+        outState.putParcelable(PHOTO_UPLOAD_URI_KEY, photoUploadUri)
         super.onSaveInstanceState(outState)
     }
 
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/keylesspalace/tusky/components/compose/ComposeActivity.kt
```diff
@@ -707,38 +698,25 @@ class ComposeActivity : BaseActivity(),
...
-
-    private fun onCommitContentInternal(inputContentInfo: InputContentInfoCompat, flags: Int): Boolean {
-        if (flags and InputConnectionCompat.INPUT_CONTENT_GRANT_READ_URI_PERMISSION != 0) {
-            try {
-                inputContentInfo.requestPermission()
-            } catch (e: Exception) {
-                Log.e(TAG, "InputContentInfoCompat#requestPermission() failed." + e.message)
-                return false
+        if(supported) {
+            val lacksPermission = (flags and InputConnectionCompat.INPUT_CONTENT_GRANT_READ_URI_PERMISSION) != 0
```
charlag: this isn't new but I'm really confused about this. Isn't flag check
`flags & myFlag == myFlag`?
https://gist.github.com/LouisCAD/04be811e22d45cbc7285ab247c6c12e7#file-bitflags-kt-L7

and here we then want to check that it's not there then it should be `0` or?..

app/src/main/java/com/keylesspalace/tusky/components/compose/ComposeActivity.kt
```diff
@@ -848,9 +826,12 @@ class ComposeActivity : BaseActivity(),
         }
     }
 
-    private fun pickMedia(uri: Uri) {
+    private fun pickMedia(uri: Uri, contentInfoCompat: InputContentInfoCompat? = null) {
         withLifecycleContext {
             viewModel.pickMedia(uri).observe { exceptionOrItem ->
+
+                contentInfoCompat?.releasePermission()
```
charlag: I'm not good at this stuff so I'm not sure why this is moved here.

connyduck:
You have to rotate while taking a picture so the result comes back to a recreated activity that does not know the `photoUploadUri`.

charlag:
Ah, yes, of course, could reproduce (sorry, I have troubles concentrating today).
Not sure we should force unwrap there in any case but maybe it's fine.

app/src/main/java/com/keylesspalace/tusky/components/compose/ComposeActivity.kt
connyduck: this is how its done in the documentation as well & it works

app/src/main/java/com/keylesspalace/tusky/components/compose/ComposeActivity.kt
connyduck: Because the permission should be released after we no longer need it, so I moved it after the upload is complete (or failed).

PR #1708, opened 2020-02-24
Upgrade deps

connyduck:
- Its now recommeded to use `FragmentContainerView` to host fragments 
- `ViewModelProviders.of` is deprecated
- Dagger migrated to AndroidX so the flags in `build.gradle` are no longer needed

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/components/conversation/ConversationsFragment.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/components/conversation/ConversationsFragment.kt
@@ -20,8 +20,8 @@ import android.os.Bundle
 import android.view.LayoutInflater
 import android.view.View
 import android.view.ViewGroup
+import androidx.fragment.app.viewModels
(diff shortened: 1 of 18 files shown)


app/src/main/java/com/keylesspalace/tusky/components/conversation/ConversationsFragment.kt
```diff
@@ -87,10 +85,10 @@ class ConversationsFragment : SFragment(), StatusActionListener, Injectable, Res
 
         initSwipeToRefresh()
 
-        viewModel.conversations.observe(this, Observer<PagedList<ConversationEntity>> {
+        viewModel.conversations.observe(viewLifecycleOwner, Observer<PagedList<ConversationEntity>> {
```
charlag: Why did this change?

connyduck: its a new inspection that suggest to do so

PR #1709, opened 2020-02-25
upgrade Gradle and Android plugin

connyduck:
Should we add the wrapper verification or the dependency verification mentioned in the releasenotes? I think both would be really cool.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -15,5 +15,5 @@ org.gradle.jvmargs=-Xmx4096m
 # use parallel execution
 org.gradle.parallel=true
 
-android.enableUnitTestBinaryResources=true
 android.enableR8.fullMode=true
+android.useAndroidX=true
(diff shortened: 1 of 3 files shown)


gradle.properties
```diff
@@ -15,5 +15,5 @@ org.gradle.jvmargs=-Xmx4096m
 # use parallel execution
 org.gradle.parallel=true
 
-android.enableUnitTestBinaryResources=true
 android.enableR8.fullMode=true
+android.useAndroidX=true
```
connyduck: for some reason this flag is required again

charlag: I wanted to say that I use no such flags in my project, I checked it and I actually have this flag. Whatever, lol

charlag:
Oops, sorry, Github doesn't feel good and I think it showed that I didn't approve it yet so I did this again I think?..

PR #1712, opened 2020-02-26
When a plaintext share contains EXTRA_TEXT and a distinct EXTRA_SUBJECT, use EXTRA_SUBJECT as the content warning for the shared status

Tak:


hinaloe:
@Tak @connyduck 
This feature feels very annoying. I don't want to hide the link with a context warning, and I can't even show the preview card. I don't understand why this is enforced when sharing a link from a browser. Should it at least exist as an optional function?

frumble:
I second @hinaloe, this feature is more a bug than an improvement. I totally don’t want titles of articles as CW of my posts. That’s not my toothing style. And it’s even illogical: CWs are typically used to warn you about a content, but with this implementation, you are pasting the headline right into the CW field.

PR #1722, opened 2020-03-03
Work around the bug in ComposeScheduleView, fix #1720

### LINKED ISSUE #1720: Posts are scheduled 24 hours before the datetime chosen in the date-picker
When I schedule a post, after going through the date-picker, if I tap the clock button again, the app shows that the post is scheduled for a time 24 hours before what I specified. (I am in Eastern Standard Time or New York time, in case this is relevant.) If I choose to schedule a post one hour from now, for example, the post will actually be scheduled 23 hours in the past. If I tap "Toot!", this results in my post being published immediately.

* * * *
- Tusky Version: 10.0
- Android Version: LineageOS 16.0 (equivalent to Android 9.0)
- Android Device: OnePlus 5T (dumpling)
- Mastodon instance (if applicable): todon.nl

charlag:
DatePicker seems to think that it's in UTC. So setting selected time
might not work as aspect and receiving value from it might be in UTC
as well. This commit fixes the second issue by interpreting the date
as UTC date. Tested with America/New_York (GMT-5 at the moment) and
Russia/Kamchatka (GMT+12).

cc @Tak, @kyori19

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/components/compose/view/ComposeScheduleView.java
+++ b/app/src/main/java/com/keylesspalace/tusky/components/compose/view/ComposeScheduleView.java
@@ -192,6 +192,9 @@ public class ComposeScheduleView extends ConstraintLayout {
     private void onDateSet(long selection) {
         initializeSuggestedTime();
         Calendar newDate = getCalendar();
+        // working around bug in DatePicker where date is UTC #1720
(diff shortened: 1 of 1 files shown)


connyduck:
> I guess someone needs to monitor the upstream issue so the workaround can be removed/guarded?

yes, I will remove the workaround again when the fix is released

PR #1729, opened 2020-03-17
Add notifications for follow requests

Tak:
First pass, hit me with the feedback


app/src/main/res/layout/item_follow_request.xml
```diff
@@ -1,77 +1,102 @@
...
-            android:id="@+id/displayNameTextView"
-            android:layout_width="wrap_content"
-            android:layout_height="wrap_content"
-            android:ellipsize="end"
-            android:maxLines="1"
-            android:textColor="?android:textColorPrimary"
-            android:textSize="?attr/status_text_large"
-            android:textStyle="normal|bold"
-            tools:text="Display name" />
+    <RelativeLayout
```
connyduck: Please don't nest a `RelativeLayout` and a `LinearLayout` inside a `ConstraintLayout`. You can place all views directly inside the `ConstraintLayout`.

Tak:
>     * The `item_follow_request.xml` is also used in `AccountsInListFragment` where it now looks weird. Can you create a separate `item_account_list.xml` please?

Oops, I thought I found all the usages - I'll make another pass

>     * I would really appreciate it if we where not using the raw okhttp calls anymore

Sorry, I just followed the pattern that `AccountListFragment` or whatever was already using, I'll fix it up

>     * Should we follow the web and make notifications/filter for this type default off?

Hm, I made the notification pref off by default, but I see I'm missing some connecting logic

>     * What do you think about using other icons for the buttons, e.g. these http://materialdesignicons.com/icon/account-check http://materialdesignicons.com/icon/account-cancel

Sure, works for me

Tak:
Hm, the new icons look a little busy in context

Tak:
So it's easy to filter follow request notifications by default for _new_ accounts, but is there a sane upgrade path for existing users? It's stored in the db as a serialized json array, is there a straightforward/sane way to do it as part of the db update?

app/src/main/res/layout/item_follow_request_notification.xml
```diff
@@ -0,0 +1,98 @@
...
+<androidx.constraintlayout.widget.ConstraintLayout xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:app="http://schemas.android.com/apk/res-auto"
+    xmlns:tools="http://schemas.android.com/tools"
+    android:layout_width="match_parent"
+    android:layout_height="wrap_content"
+    android:gravity="center_vertical"
+    android:paddingLeft="16dp"
+    android:paddingRight="16dp"
+    android:paddingBottom="10dp"
+    >
```
connyduck: please run "Reformat Code" on this file

connyduck:
> Hm, the new icons look a little busy in context

Yeah, maybe its better to use the check/x again

> is there a straightforward/sane way to do it as part of the db update?

Don't think so, but its also ok if its different for new & existing accounts

Tak:
Ok, I think it's ready to go

Issue #1731, opened 2020-03-17
Support “Mute conversation” like in Mastodon Web


ariasuni:
Tusky doesn’t provide a “Mute conversation” action. Contrarily to what the name implies, it is used to mute a thread, and not a conversation in the direct message sense.

![bildo](https://user-images.githubusercontent.com/2446451/76897570-462ca180-6894-11ea-9151-70a5bbae8973.png)

On Mastodon Web, the action is available on our own toots, and for any toots that’s displayed in the notifications (mentions, fav, boosts, poll ended, etc.).

The API are described on [joinmastodon.org](https://docs.joinmastodon.org/methods/statuses/) (search “Mute conversation” and “Unmute conversation”).

- `api/v1/statuses/:id/mute`
- `api/v1/statuses/:id/unmute`

* * * *
- Tusky Version: 10.0
- Android Version: 9
- Android Device: Oneplus 3

charlag:
Closed by #1732 I suppose

PR #1732, opened 2020-03-20
Add support for muting conversations

Tak:
Implements #1731

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/fragment/SFragment.java
+++ b/app/src/main/java/com/keylesspalace/tusky/fragment/SFragment.java
@@ -334,13 +334,11 @@ public abstract class SFragment extends BaseFragment implements Injectable {
         }
 
         for (Status.Mention mention : mentions) {
-            try {
-                if (account.getUsername().equals(mention.getUsername()) &&
-                        account.getDomain().equals(Uri.parse(mention.getUrl()).getHost())) {
+            if (account.getUsername().equals(mention.getUsername())) {
+                Uri uri = Uri.parse(mention.getUrl());
+                if (uri != null && account.getDomain().equals(uri.getHost())) {
                     return true;
                 }
-            } catch (Exception exception){
-                // This block intentionally left blank
             }
         }
         return false;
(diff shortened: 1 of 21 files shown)


Tak:
Related, `SFragment` and `SearchStatusesFragment` sure share a lot of code… :thinking:

Tak:
Oops, missed some tests

app/src/main/java/com/keylesspalace/tusky/fragment/SFragment.java
```diff
@@ -305,12 +315,37 @@ protected void more(@NonNull final Status status, View view, final int position)
...
+        }
+
+        for (Status.Mention mention : mentions) {
+            try {
+                if (account.getUsername().equals(mention.getUsername()) &&
+                        account.getDomain().equals(Uri.parse(mention.getUrl()).getHost())) {
+                    return true;
+                }
+            } catch (Exception exception){
+                // This block intentionally left blank
```
connyduck: What kind of exceptions can be thrown? It would be better to catch more specific ones and log an error message.

app/src/main/java/com/keylesspalace/tusky/fragment/TimelineFragment.java
```diff
@@ -1313,6 +1317,11 @@ private void handleBookmarkEvent(@NonNull BookmarkEvent bookmarkEvent) {
         setBookmarkForStatus(pos, status, bookmarkEvent.getBookmark());
     }
 
+    private void handleMuteConversationEvent(@NonNull MuteConversationEvent event) {
+        // TODO: Is there anything to do here but a full refresh?
```
connyduck: Is there a way to find all posts that are part of a thread and remove them without refreshing the whole timeline or doing a context request? I don't think so, you could remove the comment.

Tak: I couldn't come up with anything either

app/src/main/java/com/keylesspalace/tusky/fragment/SFragment.java
Tak: Hm, I had thought it threw an exception on parse failure, but it apparently just returns null (api confusion :man_shrugging: )

ariasuni:
Thanks!!

issueseverywhere:
Hi,

Please support muting conversations for Pleroma too. I can do it in the web interface, but not on Tusky.

Thanks

PR #1733, opened 2020-03-20
use number of voters instead of votes to calculate poll results

connyduck:
This makes a lot more sense for multiple choice polls and is how Mastodon Web handles it. Unfortunately Pleroma does not send the `voters_count` so I had to add a fallback.


connyduck:
Not now, can do when I have more time.

Issue #1737, opened 2020-03-24
No confirmation dialogue when blocking a user


ffffluffy:
Hello, I tested the block function, and there doesn't seem to be any confirmation dialogue.

Here is what is currently in Mastodon's web interface : 
![Capture+_2020-03-24-07-40-33](https://user-images.githubusercontent.com/32818316/77395910-e60d9200-6da2-11ea-8b9f-ebc426cc1ef7.png)

Such a little pop-up isn't in Tusky, which is a bit annoying, as unblocking a user doesn't restore the follows for instance, so it is not a very easy reversible operation. I feel like it would be nice to have a confirmation dialogue for blocks (and also for mutes I guess?)

Thank you very much for your efforts !

* * * *
- Tusky Version: 10.0
- Android Version: 8.0.0
- Mastodon instance (if applicable): eldritch.cafe

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

connyduck:
#1740

PR #1740, opened 2020-03-26
Confirm blocks and mutes from timelines

Tak:
Implements #1737


connyduck:
1) I don't think this should be an option, just always show the dialog
2) It would be awesome to have the name of the user in the dialog instead of "this user", like "Are you sure you want to block **@connyduck​@chaos.social**?"

Tak:
:heavy_check_mark:

PR #1741, opened 2020-03-27
replace HtmlUtils with HtmlCompat

connyduck:
Got rid of another unnecessary util class.
We actually called `trimTrailingWhitespace` after every `fromHtml`, but turns out that was entirely unnecessary as it only needs to run when loading Html from the network.
And `Account` does not need to be `Parcelable` since its never parceled.


connyduck:
oh no the tests

Issue #1540, opened 2019-10-18
Add draft auto-save (option)


w4tsn:
When writing a toot and for some reason the screen gets forcefully closed there is no draft and the writings are lost. This happens e.g. when the app crashes, the writing screen get's closed or the OS causes the app to close. This is about mitigating issues, that force-close the writing screen without any chance to save a draft.

I have some problems with my OS (I don't think it's a problem of tusky) that regularly cause screens to get closed, apps get switched etc. I know this is probably a rather rare problem, but on the other hand mitigating such issues makes the app more false tolerant.

My motivation comes from the fact that I sometimes take some time to write stuff or that I include several links in there, switching apps in the process. When having spent 5 minutes on a post and losing it is quite annoying.

Implementation
---

I propose to save on every newline character and spaces or alternatively every 5 seconds.

It might be a good idea to make this configurable via an opt-out, so if you want to save energy and don't need this feature you can turn it off.

digitalcircuit:
With [Digital Wellbeing's Focus Mode](https://support.google.com/android/answer/9346420), this data loss can happen even on a fully-functioning Android system - in my case, if I take a [Focus Mode break](https://www.androidpolice.com/2020/08/06/digital-wellbeing-lets-you-override-focus-mode-for-5-minutes-apk-mirror/) and start writing in Tusky, only for Focus Mode to close the app at the end of the break timer, there's [no prompt to save a draft](https://github.com/tuskyapp/Tusky/pull/1744#issuecomment-706646979).

I'm not sure how aggressive this should be, or what the performance/battery impact might be - perhaps consider the auto-save frequency of e.g. Google Keep or Drive/Docs?

gfxstrand:
I would really like this as well. I just lost like an hour of typing because I switched away to Signal to deal with a text message and Tusky exited my draft without saving when I switched back. I also run into this regularly when I'm in the middle of typing something up and I get a notification I want to deal with. Tapping the notifiaction takes me to the notifications drawer and throws away my toot without any ability to get it back.

PR #1744, opened 2020-04-01
Improve protection against accidentally losing drafts and redrafts

Tak:


connyduck:
I don't think this is the right fix for the problem. You end up with a saved toot even when you decide to send the status.
Can you try initializing the ViewModel differently so the `didChange` check always returns true when its a redraft? But not when its editing a draft.

Tak:
Hm, we delete drafts on sending in any case.
But I can rework it.

connyduck:
We only delete drafts when we have their id

Tak:
Reworked

connyduck:
But we already have `startingText` and `startingContentWarning` in the ViewModel, having it in the Activity as well is super confusing

Tak:
Hm, this iteration has broken mentions in replies and reply context when editing drafts 😐

Tak:
I finally got around to rereworking this.
Now the view model just sets a flag when we want to mark the initial state as modified.

We're doing this in the case of redraft, but also in the case of editing a saved draft now, because saved drafts are now deleted when you start editing them.

connyduck (approved):

digitalcircuit:
Thank you for your efforts here!

In the past, I've lost several lengthy toots due to temporarily pausing Digital Wellbeing's Focus Mode, redrafting a toot, only for Focus Mode to close Tusky when the break was over before I finished editing, thus discarding the previously deleted-for-redrafting toot.

Tak:
🤔 Does this change actually help in that case?

digitalcircuit:
@Tak Ack, sorry for the noise, in my enthusiasm upon checking the `develop` commit history, I misread the progression of changes here and the commit message.  This probably doesn't actually help the 'app is effectively swiped away' behavior of Focus Mode activating.

Tak:
There's a request for autosave ( #1540 ) - that would probably mitigate the issue somewhat

PR #1748, opened 2020-04-06
convert MainActivity to Kotlin and upgrade MaterialDrawer to version 8

connyduck:
MaterialDrawer has been completely rewritten in Kotlin, so I figured its a good idea to convert our MainActivity to Kotlin as well so we can take full advantage of it.

- looks exactly like before
- uses MaterialDrawer more like it was meant to be used, less hacks e.g. for closing the drawer on our side
- now uses our placeholders for image loading

I tried removing the header layout override but it just looks better with it


app/src/main/java/com/keylesspalace/tusky/MainActivity.kt
```diff
@@ -0,0 +1,606 @@
...
+         * drawer, though, because its callback touches the header in the drawer. */
+        fetchUserInfo()
+
+        setupTabs(showNotificationTab)
+
+        val pageMargin = resources.getDimensionPixelSize(R.dimen.tab_page_margin)
+        viewPager.setPageTransformer(MarginPageTransformer(pageMargin))
+        tabLayout.addOnTabSelectedListener(object : OnTabSelectedListener {
+            override fun onTabSelected(tab: TabLayout.Tab) {
+                if (tab.position == notificationTabPosition) {
```
charlag, line 172: it no longer does `viewPager.setCurrentItem(tab.getPosition());`, is this intended?

app/src/main/java/com/keylesspalace/tusky/MainActivity.kt
```diff
@@ -0,0 +1,606 @@
...
+        if (NotificationHelper.areNotificationsEnabled(this, accountManager)) {
+            NotificationHelper.enablePullNotifications()
+        } else {
+            NotificationHelper.disablePullNotifications()
+        }
+        eventHub.events
+                .observeOn(AndroidSchedulers.mainThread())
+                .autoDispose(this, Lifecycle.Event.ON_DESTROY)
+                .subscribe { event: Event? ->
+                    if (event is ProfileEditedEvent) {
```
charlag, line 197: This asks to be `when`

app/src/main/java/com/keylesspalace/tusky/MainActivity.kt
```diff
@@ -0,0 +1,606 @@
...
+                    return ctx.getDrawable(R.drawable.avatar_default)!!
+                }
+
+                return super.placeholder(ctx, tag)
+            }
+        })
+
+        mainDrawer.apply {
+            addItems(
+                    PrimaryDrawerItem().apply { identifier = DRAWER_ITEM_EDIT_PROFILE; nameRes = R.string.action_edit_profile; isSelectable = false; iconicsIcon = GoogleMaterial.Icon.gmd_person },
```
charlag, line 337: I'm not a fan of it being oneline. I get that it will get long my head hurts from looking at this

charlag: And if we defined click listener on it, we wouldn't even have to write another long when below

app/src/main/java/com/keylesspalace/tusky/MainActivity.kt
```diff
@@ -0,0 +1,606 @@
...
+        }
+    }
+
+    private fun handleProfileClick(profile: IProfile, current: Boolean): Boolean {
+        val activeAccount = accountManager.activeAccount
+
+        //open profile when active image was clicked
+        if (current && activeAccount != null) {
+            val intent = AccountActivity.getIntent(this, activeAccount.accountId)
+            startActivityWithSlideInAnimation(intent)
```
charlag, line 463: `new Handler().postDelayed(() -> drawer.closeDrawer(), 100);` is gone, is this intended?

connyduck: yes

app/src/main/java/com/keylesspalace/tusky/MainActivity.kt
connyduck: yes, its not necessary, its handled by the `TabLayoutMediator`

connyduck:
- no touch highlight when clicking drawer items

To fix the statusbar we need `android:windowTranslucentStatus true`, but that breaks all other activities >_>

charlag:
>     * no touch highlight when clicking drawer items
> 
> 
> To fix the statusbar we need `android:windowTranslucentStatus true`, but that breaks all other activities >_>

I thought you can set it per activity? You can even request it programmatically

connyduck:
Yes I can set it programmatically, but the stausbar will have a very different color which sucks

charlag:
ah, I get it
I checked some Android/Google apps and it seems like Google mostly picks the same color for action bar and status bar (white) or it does not overlap the status bar (like stock Files app from AOSP where there are two different blues) so I guess we don't have to worry about it

charlag:
Google Translate sets it to translucent and then the same color and because of that the status bar looks darker. Sounds smart but then you don't really see that drawer is below the status bar anyway and I again think that no one cares about this

connyduck:
Ok, I found a solution:
`window.statusBarColor = Color.TRANSPARENT` for the `MainActivity` only.
Why this works: When the status bar is transparent, you see the DrawerLayout, and that draws `colorPrimaryDark` as the background color, the same color the other activities show as the stausBarColor. But now you also see the status bar tint of the MaterialDrawer.

charlag:
I'm surprised that it works, I wonder if it is across all versions

connyduck:
Ok, I think I addressed all issues.
I "fixed" the header text color in the light theme by lightening the background, so its not the same as previously where both theme darkened the background.

The issue with no ripple highlight affects only Android 5, see https://github.com/mikepenz/MaterialDrawer/issues/2584

charlag:
Changes look great!
My profile still looks weird in light theme:
image
(it actually looks better on screenshot because it's bigger)

I think it can only be completely helped by either of
1. Not having background under this part (I'm pretty sure this drawer does not support that)
2. Having gradient (not sure drawer supports that)
3. Outlining text (pretty sure not possible)
4. Not using profile background there

We could also lighten it up even more or just leave it like it is, I think it's not that bad, it's just I have extreme case I think

connyduck:
I made the white filter even stronger and the color of the subtitle darker.

PR #1751, opened 2020-04-12
Add log sending functionality

charlag:
This might be useful when we want to debug some strange behavior/crash.
We obviously get only our own logs (unless you give Tusky permission to read all of them and I hope no one does that).

Current approach is prompting to send file. We could fill in some email address but I don't know if it's a good idea because it's quite spam prone, maybe we would tell it to people when they ask or propose some other channel.

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/util/LogReader.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/util/LogReader.kt
@@ -1,0 +1,46 @@
+/* Copyright 2017 Andrew Dawson
+ *
+ * This file is a part of Tusky.
+ *
+ * This program is free software; you can redistribute it and/or modify it under the terms of the
+ * GNU General Public License as published by the Free Software Foundation; either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * Tusky is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even
+ * the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU General
+ * Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License along with Tusky; if not,
+ * see <http://www.gnu.org/licenses>. */
+
+package com.keylesspalace.tusky.util
+
+import android.content.Context
+import android.net.Uri
+import java.io.File
+import java.io.IOException
+
+
+object LogReader {
+    fun getLogFile(context: Context): Uri {
+        return try {
+
+            val logFile = File(context.cacheDir, "log.txt")
+            logFile.delete()
+            logFile.createNewFile()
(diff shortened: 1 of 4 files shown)


connyduck:
Production builds of Tusky log almost nothing though it would only be helpful for crashes 🤔

charlag:
I mostly want it for cases upload fails, do we get something in that case?

connyduck:
R8 is configured to strip all log calls, so probably not
https://github.com/tuskyapp/Tusky/blob/develop/app/proguard-rules.pro#L74

I guess we could not do that for one or two versions? It will make the apk a bit larger though
Or if you have a specific user in mind, send them an unstripped apk for debugging?

app/src/main/java/com/keylesspalace/tusky/util/LogReader.kt
```diff
@@ -0,0 +1,46 @@
...
+
+package com.keylesspalace.tusky.util
+
+import android.content.Context
+import android.net.Uri
+import java.io.File
+import java.io.IOException
+
+
+object LogReader {
```
connyduck, line 24: Why the object? Just `getLogFile` a top level function

app/src/main/java/com/keylesspalace/tusky/util/LogReader.kt
```diff
@@ -0,0 +1,46 @@
...
+                process.waitFor()
+            } catch (ignored: InterruptedException) {
+            }
+            if (process.exitValue() != 0) {
+                val error: String = process.errorStream.bufferedReader().readText()
+                throw RuntimeException("Reading logs failed: " + process.exitValue() + ", " + error)
+            }
+            Uri.fromFile(logFile)
+        } catch (e: IOException) {
+            throw RuntimeException(e)
```
connyduck, line 43: o_0 why crash the app when you could also show a toast

app/src/main/java/com/keylesspalace/tusky/util/LogReader.kt
charlag: because I copies it from Java version 😅

app/src/main/java/com/keylesspalace/tusky/util/LogReader.kt
charlag: that's also something that was handled originally bur now is not, will change!

charlag:
I think we can check how big apk get or strip based on the log level if possible. or maybe just leave it there foe nightly?..

connyduck:
Not stripping nightly for some time seems like a good idea

charlag:
There's no big need currently I think so it can sit here for a while.
I didn't know we strip logs when I thought about it, I don't think it makes a huge difference in size but I don't have anything to back it up with.
Let's revisit it when we have mysterious issues again.

PR #1752, opened 2020-04-12
 Add preference to disable swipe gesture for switching tabs

### LINKED ISSUE #1718: [Accessibility Request] Option To Toggle Swiping Between Tabs
I recently updated to the latest version of Tusky and have been fighting with the swipe to change tab feature. It would be great if we had an option in the settings to disable swiping between tabs. Often when scrolling because of how my thumb moves, I will swipe by accident instead of scrolling.

charlag:
close #1718

I don't like adding more knobs but I dislike having accessibility problems even more and I think it should probably be optional independent of sensitivity.

It's based on `mainactivity_kotlin` branch so I will have to rebase it later


connyduck:
I hate this but I guess its our best option

PR #1756, opened 2020-04-16
Restore compose options after rotation

### LINKED ISSUE #1755: Content Warning disappears if phone orientation changes
Recently found out that if I add a content warning to a post, if my phone is rotated and the screen orientation changes, the CW disables.

* * * *
- Tusky Version: 10.0
- Android Version: 9
- Android Device: Fairphone 3

- [✔️] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

TheLastProject:
Fixes #1755

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/components/compose/ComposeActivity.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/components/compose/ComposeActivity.kt
@@ -690,8 +690,7 @@ class ComposeActivity : BaseActivity(),
 
     private fun onContentWarningChanged() {
         val showWarning = composeContentWarningBar.isGone
-        viewModel.showContentWarning.value = showWarning
-        viewModel.contentWarningChanged()
+        viewModel.contentWarningChanged(showWarning)
         updateVisibleCharactersLeft()
     }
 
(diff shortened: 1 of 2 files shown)


TheLastProject:
Sorry, I'm not sure how to do that. This was the easiest way I could think of and nothing else I'm trying seems to work. I'm willing to learn more but you're also free to fix it without me :)

connyduck:
1. ) Well, the root of the bug is that the showContentWarning.value is overridden on every rotation here https://github.com/tuskyapp/Tusky/blob/develop/app/src/main/java/com/keylesspalace/tusky/components/compose/ComposeViewModel.kt#L371
2. ) ViewModels survive rotations and are not recreated like activities are
3. ) Add a boolean flag `contentWarningStateChanged` to the ViewModel, default false
4. ) set the flag to true when the content warning changed (adding a `contentWarningChanged` method to the ViewModel makes this nicer because then the flag can be private)
5. ) don't override the state from 1) when it has been changed by the user

TheLastProject:
Thanks for the pointers, I believe I fixed it as you asked now. Tested it and still works.

I do wonder though... why is `composeOptions?.contentWarning` even null after rotating in here?
After all, that seems to be the only reason why `showContentWarning` gets set to false in the first place: https://github.com/tuskyapp/Tusky/blob/develop/app/src/main/java/com/keylesspalace/tusky/components/compose/ComposeViewModel.kt#L367

connyduck:
If you now move the `viewModel.showContentWarning.value = showWarning` from the Activity to the new `contentWarningChanged` method its perfect!

> why is composeOptions?.contentWarning even null after rotating in here?

Depends where you come from. When you open with the action button on the main activity, composeOptions will alway be null because its never set. When you open from somewhere else where it is set, it will not be null even after rotating.

Issue #1753, opened 2020-04-13
Bug: Show Less button shown in favourite/boost notifications even though the toot isn't expanded


dkter:
Steps to reproduce:

- Enable "Always expand toots marked with content warnings" in Account Preferences
- Receive a favourite or boost notification on a toot with a CW
- In the notification tab, the toot won't be expanded but the button will say `Show Less`, meaning you have to press the button twice to expand the toot

* * * *
- Tusky Version: 10.0
- Android Version: 10.0
- Android Device: Pixel 3a
- Mastodon instance (if applicable): cybre.space

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

PR #1757, opened 2020-04-17
fix content warning in notifications when alwaysShowSpoiler is on

connyduck:
closes https://github.com/tuskyapp/Tusky/issues/1753

I have no idea why `NotificationViewData` has an `expanded` attribute when its already in `StatusViewData`


charlag:
I can't reproduce original issue but the code change looks good, I"m also not sure  why we had it

Issue #1749, opened 2020-04-08
Extra blank lines are added to first post in Home timeline


MasterGroosha:
This is a strange bug, honestly. One may notice that the first post in Home timeline has extra lines in the bottom, for example:
![изображение](https://user-images.githubusercontent.com/2780044/78765277-75649900-7990-11ea-9528-b2d78782e73e.png)
After posting second post and restarting the app (sometimes twice):
![изображение](https://user-images.githubusercontent.com/2780044/78765308-801f2e00-7990-11ea-9929-f0c669c5f88d.png)

However, when you restart the app after posting the second post, you may notice that in the beginning the older post also has these blank lines, but then you see an animation of them disappear, leaving blank lines only on the most recent post. Cannot upload the video for now, but you can easily reproduce it:
1. Post something
2. Restart the app (maybe twice) to see blank lines after it.
3. Post something else.
4. Restart the app (maybe twice) and look carefully on the older post. You will notice an animation.

* * * *
- Tusky Version: 10.0-d44eada
- Android Version: 10
- Android Device: Oneplus 7T Pro
- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

connyduck:
regression from https://github.com/tuskyapp/Tusky/pull/1741
Turns out the `trimTrailingWhitespace` cannot be removed so easily

PR #1761, opened 2020-04-20
fix toots showing too much whitespace

connyduck:
Turns out `trimTrailingWhitespace` is necessary in more places, because `toHtml` adds in `<p>` tags again even if there is no newline anywhere 🙄 

closes https://github.com/tuskyapp/Tusky/issues/1749


app/src/main/java/com/keylesspalace/tusky/util/StringUtils.kt
```diff
@@ -71,3 +72,11 @@ fun String.isLessThan(other: String): Boolean {
         else -> this < other
     }
 }
+
+fun Spanned.trimTrailingWhitespace(): Spanned {
+    var i = length
+    do {
+        i--
+    } while (i >= 0 && Character.isWhitespace(get(i)))
```
charlag, line 80:
```suggestion
    } while (i >= 0 && get(i).isWhitespace())
```

PR #1773, opened 2020-05-04
Fix tag highlighting in editor

kyori19:
Tag highlighting in ComposeActivity is not same as Mastodon does, so I fixed it.
I was also tried to fix mentions, but I don't know how to match to `[:word:]` by code point...

## Screenshots

Before

After

Actual Behaviour


kyori19:
@Tak How should I add it?
Is just adding `##seven` to [here](https://github.com/tuskyapp/Tusky/pull/1773/files#diff-b2d9f7e3fddadf6e616637a2c648fe69R13) enough, or should I add new test method?

Tak:
I think a new method that checks the bounds of the highlighted span (verifying that it captures just `#tag` and not `##tag`) would be best

kyori19:
I intended to add test for japanese tags such as `#日本語` but `\w` interpreted differently in android devices and junit and test fails...

PR #1777, opened 2020-05-05
fix confirmReblogs default values

connyduck:
I think on per default makes more sense, because a accidental boost is worse than an annoying dialog.


connyduck:
yes it was not consistent

PR #1778, opened 2020-05-06
Fix RxJava UndeliverableException crashes

connyduck:
So I learned that these weird crashes can not only occur because of missing error handlers (I think we caught them all), but also when using e.g. `zipWith` on network errors 🙄 ...
Lets fix them with a global error handler

https://medium.com/@bherbst/the-rxjava2-default-error-handler-e50e0cab6f9f


charlag:
TBH I have doubts about this, it would be better to transform errors into Results or something before calling fancy operators like zip but if we want to replace most of it with coroutines anyway maybe it doesn't make much sense.

I kinda understand their reasoning behind this but also... it's so unintuitive and cumbersome.

PR #1783, opened 2020-05-09
use AndroidX WorkManager instead of Evernote Android Job

connyduck:
Android Job is deprecated https://github.com/evernote/android-job


connyduck:
Yes its necessary, I tried it out. Will wait with merging until I tested this for some time.

connyduck:
Seems to work very similarly to Android Job ✅

PR #1788, opened 2020-05-12
add missing attribute to TimelineDao query

connyduck:
Very cool that there is a warning for this, would have never found otherwise


connyduck:
Probably because its nullable. It will just always be null if its not in the query.

charlag:
ah, yes

PR #1789, opened 2020-05-12
remove usage of deprecated userVisibleHint

connyduck:


charlag:
I tested it a little bit and I have one "definitely bug" and one "It feels weird" thing.

1. Have two photos in a status. Open the second one. Scroll to the first one. Tap to hide everything. Scroll to the second one. Everything is hidden but description.

2. In the profile media list I get weird glitches when viewing media and scrolling, like when image flashes for a second. Also spinner doesn't seem to work correctly. I don't know if it's just because of the debug version but it's definitely different from Nightly.

connyduck:
I cannot reproduce either of those 😞 
The behavior is equally weird on develop and with the changes.
Can you maybe record a video?

charlag:
I wanted to see first if you can get them too. I can try to record a video

connyduck:
Yeah maybe we should leave it as it is. The Media viewing is very high on my "rewrite from scratch" list anyway since the code is just a complete clusterfuck.

charlag:
I wanted to teach PhotoView that it's inside scrolling container somehow. Do you have other plans?

connyduck:
Rewriting from scratch includes researching if PhotoView is still the way to go. It hasn't received an update in quite a while.

charlag:
I would love to replace it with something but I think there's nothing else, at least I didn't find anything else last time I researched.

PR #1790, opened 2020-05-14
enable multiple hashtags in one hashtag tab

connyduck:
Screenshot_1589480784


charlag:
>files changed: 37
oof, will take some time

app/src/main/java/com/keylesspalace/tusky/adapter/TabAdapter.kt
```diff
@@ -86,21 +88,43 @@ class TabAdapter(private var data: List<TabData>,
...
         if (!small) {
 
             if (data[position].id == HASHTAG) {
                 holder.itemView.chipGroup.show()
-                holder.itemView.actionChip.text = data[position].arguments[0]
 
-                holder.itemView.actionChip.setChipIconResource(R.drawable.ic_edit_chip)
+                data[position].arguments.forEachIndexed { i, arg ->
+
+                    val chip = holder.itemView.chipGroup.getChildAt(i).takeUnless { it.id == R.id.actionChip } as Chip?
```
charlag, line 104: It took me a while to understand what's going on here, why `takeUnless` and why `holder.itemView.chipGroup.size - 1`. A comment would be nice.

charlag:
Oh, I forgot one thing. The field in the dialog seems to hug to the edges a little bit way too much. I think it was like that before but I wanted to point out.

PR #1797, opened 2020-05-20
botatu -> bozkatu

ZiTAL:
botatu -> bozkatu


charlag:
Maybe it would be better to change it in Weblate so that others can review it?

ZiTAL:
I'm searching the string in the weblate but i can't find it

connyduck:
Its right here: https://weblate.tusky.app/translate/tusky/tusky/eu/?q=Botatu&checksum=&offset=

Issue #1487, opened 2019-09-18
[Feature Request] Display average toots per day at profile view


Hofer-Julian:
Hi!
I really enjoy the great application you created and continue to develop.
What I think would make the experience even more enjoyable if you would implement a feature similar to Twidere which displays the average toots per day when I look at a profile. 
![Screenshot_20190918-123557(1)](https://user-images.githubusercontent.com/30049909/65141812-a6d84000-da11-11e9-93e9-de37ed03d8e6.png)
Thank you for your time!

charlag:
Hu
Sorry, we can't calculate that, it should be done on the server.

Hofer-Julian:
Okay?
Well Twidere can do it for Mastodon as well. Unfortunately I can't paste a screenshot from my phone...

I was kind of hoping as Twidere is open source and written in Kotlin the code might be portable...

connyduck:
Actually, the Mastodon Api is returning a `created_at` timestamp for each account, so its possible to calculate this on the client side.

charlag:
Oh
I feel not smart rn
Sorry

PR #1799, opened 2020-05-20
Add a badge on account to show the date the user has joined mastodon

ademar111190:
The account creation date looks pretty useful to be shown to the user, it can bring valuable information especially to identify recently created accounts.

It is possible to get that information through the `created_at` field as mentioned [on this issue](https://github.com/tuskyapp/Tusky/issues/1487)

This MR unlocks the possibility to solve that issue, I did not solve that issue yet because that functionality looks a little bit advanced, so I think it should be enabled through settings option, so it is better to wait for the [settings refactor](https://github.com/tuskyapp/Tusky/pull/1615) to finish before implementing it.

Twitter for instance shows the joined date, probably to help to identify fake accounts, not sure:

twitter_example

Here some screenshots of how the app looks like with the changes:

|Before|After|
|---|---|
|before_1|after_1|
|before_2|after_2|
|before_3|after_3|

### DIFF
--- a/app/schemas/com.keylesspalace.tusky.db.AppDatabase/24.json
+++ b/app/schemas/com.keylesspalace.tusky.db.AppDatabase/24.json
@@ -1,0 +1,747 @@
+{
+  "formatVersion": 1,
+  "database": {
+    "version": 24,
+    "identityHash": "b9df42a18129a1ee90428f5b4f0939df",
+    "entities": [
+      {
+        "tableName": "TootEntity",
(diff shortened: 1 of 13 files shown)


connyduck:
I am not sure if we want this. On one hand it might help to identify spambots, on the other it might make it harder for newer accounts to be taken seriously.

charlag:
I would be opposed to it because it's easy to fake but all other numbers are easy to fake too...

charlag:
And it doesn't need a preference because there's no actual behavior change

ademar111190:
I understood the points, I do not disagree, the last thing I want is to make it harder for newcomers to fully enjoy the platform, moreover, I could create my own instance and put any date I want…

In another hand, I would love to have that piece of information, generally, that information is precise.

So would you consider the possibility of this functionality be turned off by default, but possible to enable through the settings?

I can make that change if you guys agree…

charlag:
I think we either should include it or not, pref does not make sense.
People on fedi pointed our that it's shown in web interface (and on static pages of masto too) so I would say "yes".
I think we should discuss design separately, I wouldn't put it in bubble e.g. but let's first hear more people.

a1batross:
That would make sense in some kind of "moderation mode" but it's useless for normal users IMHO.

connyduck:
There is no value in this information, and if you still need it you can get it with two clicks.

PR #1801, opened 2020-05-21
Reduce horizontal scroll sensibility of view pagers

ademar111190:
I have been using the app for some days, and I did notice a strange behaviour when you are scrolling, the horizontal scroll (ViewPager scroll) is far more sensible than the vertical scrolls (RecyclerView scroll).

A lot of times when I try to scroll down (finger from bottom to up) I did trigger the horizontal scroll and I get a change of tab instead of trigger the vertical scroll.

Android has notorious issues with the nested scroll and the API those views provide don't help much to solve the matter, but it is possible to manage the scroll sensibility, though it has a cost, it requires a piece of reflection because the relevant property is part of the internals of the recycler view.

So I'm opening this MR to the consideration of the maintainers, so you can see the UX difference and judge if it worth to adopt this reflection based code or not. I know Android apps usually don't want reflection for two main reasons, 1 performance and 2 unpredictable behaviour, so please consider the following.

About performance, in this specific case, the reflection is triggered only once at the activity creation stage, so it hits the performance but it is not a level of loss that is noticeable, in my personal experience I could not notice any difference.

About the predictability, the reflection is being made on 2 classes that are bundled in the App, the ViewPager2 and RecyclerView so we know the behaviour will be the same across different android devices and android version, but we are not in control of library update that changes those classes internals, anyway in case of update that changes the internals, at least this code will just not have any effect anymore it will stop to work but will not crash.

The last but not least, check these two videos, the blue app is from play store and green version is my build with the reflection changes. Notice that scrolls at 90º degree and 180º degree has the same behaviour on both, the difference is notable on scrolls between the 120º and 150º degrees because the current app triggers the tab switch and with my changes, it triggers the vertical list movement.

So please, test it on your own, and judge if it is worth it, In my humble opinion, it is a tremendous UX improvement if it were not so I would not be opening this MR using reflection :)

The Home:
https://vimeo.com/421289681

The search:
https://vimeo.com/421289730

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/MainActivity.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/MainActivity.kt
@@ -173,6 +173,7 @@ class MainActivity : BottomSheetActivity(), ActionButtonActivity, HasAndroidInje
         val uswSwipeForTabs = PreferenceManager.getDefaultSharedPreferences(this)
                 .getBoolean("enableSwipeForTabs", true)
         viewPager.isUserInputEnabled = uswSwipeForTabs
+        if (uswSwipeForTabs) viewPager.reduceHorizontalScrollSensibility()
 
         tabLayout.addOnTabSelectedListener(object : OnTabSelectedListener {
             override fun onTabSelected(tab: TabLayout.Tab) {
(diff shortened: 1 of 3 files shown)


connyduck:
But this is way too hacky. Who knows what side effects it introduces and it may break on any library upgrade.

ademar111190:
That is correct, all the 3 points. The reason I did open this PR anyway is "does that UX improvement worth the risk?"

I have been looking at other projects but I couldn't find any other solution than this hack. also, It looks usual just block the swipe and avoid the problem instead.

connyduck:
This is something that should be fixed in the library, not be hacked into every app that uses the library -> https://issuetracker.google.com/issues/157335285

In the meantime we have an option to disable swiping

PR #1808, opened 2020-05-26
add possibility to move the main navigation to the bottom

### LINKED ISSUE #796: Possiblity to move the header to the bottom of the screen
As someone with a 6" smartphone, it would be nice to have it as a setting to move the header bar to the bottom, like the official Twitter does it since recently.

I disagree with the Twitter's decision to simply move it for everyone by default, but it would still be nice to have it as an option within the settings.

The reasoning is simpler navigation using one hand.

connyduck:
I struggled a lot when implementing this, mostly because of the limited capabilities of `BottomAppBar` and `CoordinatorLayout`.
- In the notification fragment we have a nested `CoordinatorLayout`, making it impossible to do anything with hide-on-scroll in the `MainActivity`
- Having the `BottomAppBar` not overlap the `ViewPager2` required additional margin that needs resetting when the bar is not shown
- Having the `FloatingActionButton` not overlap the `BottomAppBar` required calculating margins in code

I decided against embedding the FAB in the bar to have enough space for the tabs.

Things that I am not yet sure of
- Include the drawer toggle in the bottom bar as well?

Closes #796


app/src/main/res/xml/preferences.xml
```diff
@@ -36,6 +36,14 @@
             android:summary="%s"
             android:title="@string/pref_status_text_size" />
 
+        <ListPreference
```
connyduck, line 39: I went for a ListPreference instead of a SwitchPreference because it makes it clearer that there are two exclusive options imho.

MasterGroosha:
I'm not a designer, but a FAB potentially could be replaced with a "+" icon in the middle of bottom bar (like Instagram or some other apps do).
However, this would introduce some problems, since Tusky allows you to hide some tabs (even Home).

charlag:
I think it's really nice I like the idea but I have a lot of small things about it:

I liked cutoff a lot from one of the earlier versions that you posted because it did not look like bottom navigation. Currently it looks like bottom navigation but it doesn't behave like it. By putting tabs in the bottom we are already going against material but I think it's fine as long as we communicate our intent properly. To do that, we need to show that things in the bottom are not independent app parts (what bottom nav is for) but tabs and one specific thing for tabs is horizontal colored line. I don't know if it should be in the bottom or at the top. If we leave them look like they are, then we should make them behave accordingly: like independent screens with no horizontal scrolling (like in Twitter).

Another thing is that I think we need top bar as well. I was saying it for a long time and I think it's a great time to have it again. We need it for accessibility reasons to be able to refresh the timeline without using gestures but also we can do other things: search button (with dope circular revel animation!, currently it's in the super awkward position) and current timeline name (which is nice for hashtags/lists). In notification screen we need coordinator only because of the hiding bar but it has two rarely used functions which could live happily in more menu. This would also solve the problem with drawer button:
Screenshot from 2020-05-26 20-35-14
I think if we would always have collapsible appbar (independent of where tabs are), it would solve many problems for us.

So, tl;dr: if we use tabs, make them look like tabs. If we use bottom navigation, make it behave accordingly. It would be really nice to have app bar and to get rid of many ad-hoc solutions.

I hope you won't dismiss it, I really care about this. I could implement a prototype for it.

charlag:
I went ahead and made my own hacky prototype on the branch bottom-nav-proto.

(sorry for the big screenshots)

device-2020-05-26-221505
device-2020-05-26-221514
device-2020-05-26-221452

You don't have to agree or accept, it's fine to turn it down but I wanted to give some perspective and back my arguments with something.

charlag:
After going with my version for a few days, I must say that I don't really miss swipe between tabs. I had to re-learn to tab tab in the bottom to scroll to the top but it seems as convenient. I kinda wish notifications were always up-to-date (especially after we fetch them in bg), I'm not sure why I started noticing it more

connyduck:
This is really cool, I like it a lot. 

Did you manage to get the top bar to hide on scroll when showing notifications?

Can you open a PR from your branch, lets close this one?

charlag:
I removed special bar for notifications in favor of menu in appbar, it hides like on other screens.
It's quite hacky but I'll open it, we can figure it out better together

connyduck:
I added the top toolbar as discussed in #1816 @charlag 

I really like it, but I am not sure what to do about the notification settings. I don't like the top menu having different options depending on the currently shown fragment. I think having the current bar just scroll with all the notifications could work.

charlag:
This might work but we could always have "notification filters" there or "this tab filters".

connyduck:
We could also move the notification filters to the Account preferences 🤔 
I don't really know what is best here from a UX perspective, but I would really like a solution that works without any hacks or communication Activity <-> fragment

charlag:
Can't we use onCreateOprionsMenu?

connyduck:
Could work, yes

Can we merge this PR while we figure something out? I think its in a good state and helps us to collect user feedback from nightly.

charlag:
You can merge, I didn't try this version yet, I can do it in an hour or so

charlag:
1. There's still that bug with preferences where they display no value but I believe it will sort itself with merge
2. I think when selecting a tab in the bottom it should switch fragment without animation but idk how it fuses with swipe gesture. What do you think? It just feels... weird but it's nothing criminal.

connyduck:
> There's still that bug with preferences where they display no value but I believe it will sort itself with merge

Yes it will

> I think when selecting a tab in the bottom it should switch fragment without animation but idk how it fuses with swipe gesture. What do you think? It just feels... weird but it's nothing criminal.

We have the option to enable swiping, we could also disable animation when it not active and bottom bar is used? When its enabled we need animations ofc.

Issue #1775, opened 2020-05-04
Access to follow request on non-locked accounts


lapete:
The menu entry "Follow requests" is only visible if the account is locked. But even with accounts that aren't locked, there may be follow request, if the follow requester is on an instance that's silcenced by your instance. You can only accept requests by locking your account and unlock it again after you finished.
As not all users are aware of this behaviour, some requests may be overseen for a long time or ever.

* * * *
- Tusky Version: 10.0
- Android Version: 9
- Android Device: Xiaomi Mi A1
- Mastodon instance (if applicable): rage.love in this case (mastodon.social is silenced instance wide to tidy up the federated tl)

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

PR #1809, opened 2020-05-26
always show follow requests in main menu

connyduck:
closes https://github.com/tuskyapp/Tusky/issues/1775

At first I tried to solve this inside the `FollowRequestsAdapter`, but it was a mess with a lot of duplicate code and edge cases, so I decided to use the new `MergeAdapter` api.

Note: Since this is not super urgent, I would wait until the RecyclerView lib is at least a release candidate. Also `ViewHolder.getAdapterPosition()` is now deprecated, should migrate before merging.


connyduck:
This is now ready for testing @charlag

charlag:
@connyduck will try to have a look today

charlag:
Seems to work! I think header formatting looks a bit strange but it's okay.
Next step: showing the number in the side menu?

charlag (approved):

connyduck:
Yes, that would be cool. I think we will need an additional network request for it though.

charlag:
I think another request is worth it, maybe with some time limit

PR #1810, opened 2020-05-27
NotificationsFragment: show appbar if there is no notifications

a1batross:
Thus, if user accidentally will filter everything, they will be able to return to initial state.

Although, I'm wondering, does "Clear" button must be shown or greyed out as well if it's useless?


connyduck:
Nice, thx!

I think the clear button could be greyed out, but it also doesn't hurt if we leave it like it is.

PR #1814, opened 2020-05-29
Move the "more" button towards the edge

### LINKED ISSUE #1813: The 'more' button [...] should move to the edge
Currently the more button (on every toot in the timeline) has a margin on the right side. This causes me to miss it and tap between the button and the edge, thus opening the toot. 

I want the button to be as close to the edge as possible.
PR is incoming

* * * *
- Tusky Version: 11.0
- Android Version: 8.1, 10
- Android Device: doesn't matter

- [x ] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

guntbert:
having space on the right side causes me to miss the button
and select the whole toot instead

fixes #1813 

Signed-off-by: Guntbert Reiter <guntbert@gmx.at>

### DIFF
--- a/app/src/main/res/layout/item_status.xml
+++ b/app/src/main/res/layout/item_status.xml
@@ -554,7 +554,6 @@
         style="@style/TuskyImageButton"
         android:layout_width="24dp"
         android:layout_height="30dp"
-        android:layout_marginEnd="8dp"
         android:contentDescription="@string/action_more"
         android:importantForAccessibility="no"
         android:padding="4dp"


connyduck:
I see the problem but I'm not sure if this is the right fix. It breaks the alignment of the buttons relative to the text and it is still possible to click right of it. I would try to increase the touch area of all the buttons without changng their position.
Also there is another more button in item_conversation.xml.

guntbert:
Defending my proposal: 
- alignment: this button is far away from any text but relatively near to the "age" of the next toot, which is completely right aligned too
- possibility to still click between button and edge: I didn't test it on my phone but for my thumb this shouldn't be a problem any more

I **should** have looked for another occurrence though :-/  

About your suggestion to make the touch area bigger - can you give me a pointer where to do this? (Although I never had any mistouching problem with other buttons - this is probably a habit I kept from using twidere, I heavily "share" links to *pocket/readitlater*)

connyduck:
The buttons have `layout_width="30dp"` and `layout_height="30dp"`. According to [Material Design guidelines](https://material.io/design/usability/accessibility.html#layout-and-typography), they should have at least 48dp. Probably the paddings and margins have to be adjusted as well.

PR #1816, opened 2020-05-31
Prototype of bottom navigation with collabible appbar

charlag:
Like proposed [here](https://github.com/tuskyapp/Tusky/pull/1808#issuecomment-634256507) it's a prototype of moving navigation to the bottom, having full app bar and moving some of the functions we need there.
The goal was to get it to work as quick as possible so there are quite a few shortcuts taken.
1. Refresh in menu doesn't work currently
2. Appbar could be moved inside the fragment
3. With bottom nav I used `order` to distinguish tabs which might be brittle.
4. I hardcoded FAB margin
5. Search icon is incorrect
6. Notifications fragment doesn't need its own `CoordinatorLayout` anymore. Bar with filter/clear was not removed correctly, it just accidentally disappeared.

I would appreciate help in some places (like how to position bottom nav/fab and with search icon) but the best thing you can do is to try it out and tell if you agree with how it works or not.

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/MainActivity.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/MainActivity.kt
@@ -33,6 +33,7 @@ import androidx.appcompat.app.ActionBarDrawerToggle
 import androidx.appcompat.app.AlertDialog
 import androidx.core.content.ContextCompat
 import androidx.core.content.pm.ShortcutManagerCompat
+import androidx.core.view.forEach
 import androidx.emoji.text.EmojiCompat
 import androidx.emoji.text.EmojiCompat.InitCallback
 import androidx.lifecycle.Lifecycle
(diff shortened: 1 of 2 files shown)


connyduck:
So would you say we should use only one layout instead of making it configurable? 🤔 
I think giving users the option to choose for themselves is very important here

charlag:
I don't know. If there's a reason to use top one and people say "yes, I need this". I'm not a good judge of it because I myself implemented this.
I can only refer to Twitter app which has very similar navigation.
Counterexample is Facebook who, I'm pretty sure, also did research and decided and put it on top (I don't like it conceptually, it's a mishmash of everything).
We need to collect feedback somehow. We can start with two options or run bottom-only in nightly/beta and ask for opinions. I don't think just asking it without letting to try is a very useful.

It wouldn't be nice to implement tab logic twice (for tabs and for bottom nav) but it's doable.

charlag:
Given more thought, it's okay to have both, esp because you also think so.
I have a question then: should bottom nav behave like... bottom nav or like tabs? Tabs is easier to implement but wouldn't work very material imo.
I think we should have separate tabs and appbar in any case and appbar should collapse, I think we agree on that?

connyduck:
> I have a question then: should bottom nav behave like... bottom nav or like tabs? Tabs is easier to implement but wouldn't work very material imo.

Well, whats the difference in behavior? Disabling animation and user input on the ViewPager?

> I think we should have separate tabs and appbar in any case and appbar should collapse, I think we agree on that?

So, to summarize: The current layout gets an additional bar on top that collapses on scroll (just like the one you did on this branch). The notification settings get moved into the overflow menu of the appbar. Users can select to show the tabs at the bottom or at the top.

I will implement this next week.

charlag:
Tabs look different (they have stripe? maybe something else) and pager, yeah. Maybe we can use tab elements in both cases, I really don't know

connyduck:
Tabs can look almost like BottomNav (see my branch) and are easier to fill programmatically imho.

charlag:
Yeah, then it's just easier to go with tabs I guess.

PR #1819, opened 2020-06-03
cleanup proguard rules

connyduck:


app/proguard-rules.pro
```diff
@@ -17,56 +17,32 @@
...
--keepattributes Exceptions
--keepattributes *Annotation*
-
--keepclasseswithmembers class * {
-    @retrofit2.http.* <methods>;
-}
-
--keep class com.keylesspalace.tusky.entity.** { *; }
+# keep members of our model classes, they are used in json de/serialization
+-keepclassmembers class com.keylesspalace.tusky.entity.* { *; }
```
connyduck, line 39: turns out its not required to keep the whole class for Gson to work, keeping class members is sufficient

app/proguard-rules.pro
```diff
@@ -17,56 +17,32 @@
...
-  public static final android.os.Parcelable$Creator CREATOR;
-}
--keepclassmembers class **.R$* {
-    public static <fields>;
+    public static final ** CREATOR;
 }
 
 # TUSKY SPECIFIC OPTIONS
 
-## for okhttp
```
connyduck, line 41: All those library-specific rules can be deleted because they are now included directly in the libraries

app/proguard-rules.pro
```diff
@@ -17,56 +17,32 @@
...
 }
 # For enumeration classes, see http://proguard.sourceforge.net/manual/examples.html#enumerations
 -keepclassmembers enum * {
     public static **[] values();
     public static ** valueOf(java.lang.String);
 }
 -keepclassmembers class * implements android.os.Parcelable {
-  public static final android.os.Parcelable$Creator CREATOR;
-}
--keepclassmembers class **.R$* {
```
connyduck, line 35: This one is the only rule that was risky to remove, but I tested it and it seems save. Gets ~100kb off the apk size.

charlag: I'm not sure for which fields it is

app/proguard-rules.pro
charlag: If we don't want it to use setters (which it does by default I think) but we probably don't need it.

app/proguard-rules.pro
connyduck: Its for the fields of the generated R file. Seems like some tools look up fields via reflection so it was in the default proguard config

PR #1820, opened 2020-06-03
cleanup bad string translations

connyduck:
they slip in via Weblate from time to time, but they are easy to find with lint

### DIFF
--- a/app/src/main/res/values-zh-rCN/strings.xml
+++ b/app/src/main/res/values-zh-rCN/strings.xml
@@ -413,11 +413,9 @@
     <string name="action_view_bookmarks">书签</string>
     <string name="action_view_domain_mutes">隐藏域名</string>
     <string name="action_add_poll">新增意见调查</string>
-    <string name="action_mute_domain">隐藏</string>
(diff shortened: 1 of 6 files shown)


app/src/main/res/values-zh-rCN/strings.xml
```diff
@@ -413,11 +413,9 @@
     <string name="action_view_bookmarks">书签</string>
     <string name="action_view_domain_mutes">隐藏域名</string>
     <string name="action_add_poll">新增意见调查</string>
-    <string name="action_mute_domain">隐藏</string>
     <string name="action_access_scheduled_toot">定时嘟文</string>
     <string name="action_schedule_toot">定时嘟文</string>
     <string name="confirmation_domain_unmuted">%s 已解除静音</string>
-    <string name="mute_domain_warning">您确定要封锁 ％s 域名吗？您将不会在任何联邦时间轴或通知中看到该域名中的内容，并且来自该域名的关注者将被删除。</string>
```
charlag, line 420: what's the problem with this file?

connyduck: its the wrong ％, the right one is %

PR #1835, opened 2020-06-18
speedup blurhash decoding by using a intermediate int array

connyduck:
`Bitmap.setPixel` is rather expensive.

Thx @ignaciotcrespo for the idea.

I think we don't need to incorporate the caching from https://github.com/woltapp/blurhash/pull/66 as well.


app/src/main/java/com/keylesspalace/tusky/util/BlurHashDecoder.kt
```diff
@@ -102,10 +102,10 @@ object BlurHashDecoder {
                         b += color[2] * basis
                     }
                 }
-                bitmap.setPixel(x, y, Color.rgb(linearToSrgb(r), linearToSrgb(g), linearToSrgb(b)))
```
charlag, line 105: huh, not only does `setPixel()` multiple checks but it's also implemented in native which means JNI. No wonder that array is faster

PR #1818, opened 2020-06-03
upgrade gradle, AGP and dependencies

connyduck:

PR #1837, opened 2020-06-18
upgrade gradle wrapper, materialdrawer and AndroidX fragment

connyduck:
Looks like I upgraded the gradle version but not the wrapper in #1818


connyduck:
no, I did that in #1818 already (but not the other gradle files)

PR #1843, opened 2020-06-21
Improve image viewer

charlag:
This commit does 3 things:
1. Replaces PhotoView (which is abandonware) with modern TouchImageView
2. Fixes an issue with panning images. Gesture was not intercepted
properly and pager was taking control instead of image being moved.
3. Adds feedback to dismissing of images with vertical gesture.

My other plans in this are include improve description view and making video view less wonky (eg it should not reset itself on screen rotation)

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -159,7 +159,7 @@ dependencies {
 
     implementation "com.github.connyduck:sparkbutton:4.0.0"
 
-    implementation "com.github.chrisbanes:PhotoView:2.3.0"
+    implementation 'com.github.MikeOrtiz:TouchImageView:3.0.1'
 
     implementation "com.mikepenz:materialdrawer:$materialdrawerVersion"
     implementation "com.mikepenz:materialdrawer-iconics:$materialdrawerVersion"
--- a/app/src/main/java/com/keylesspalace/tusky/components/compose/dialog/CaptionDialog.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/components/compose/dialog/CaptionDialog.kt
@@ -33,9 +33,9 @@ import at.connyduck.sparkbutton.helpers.Utils
 import com.bumptech.glide.Glide
 import com.bumptech.glide.request.target.CustomTarget
 import com.bumptech.glide.request.transition.Transition
-import com.github.chrisbanes.photoview.PhotoView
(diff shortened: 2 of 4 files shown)


connyduck:
I wish we would not need to write so much of this hard to understand view code, I have no better idea though

connyduck:
I will tests this on some devices and then merge

charlag:
@connyduck I took one part from the sample of TouchImageView and added bunch of my code for transform. I don't like it either and I just wanted to use `GestureDetector` at first but it was not smooth at all.

a1batross:
Why not to use a library that optimized for bigger images? Not everyone rescaling images on server and client side, so that would give a chance to users see super resolution images, like comics, memes, artwork.

Like this, found in PhotoView readme: https://github.com/davemorrissey/subsampling-scale-image-view

Also, PhotoView does not supported animation. Does TouchImageView solve this issue?

charlag:
I started with SubsamplingImageView and another library which uses it but it only works with on-disk files. The library which uses it first writes do disk and then loads from it. I believe it's not what we need, it's quite a slowdown.
And what is the use case? Multi-Gb images which don't fit into memory? Why would you casually share it like this? Even if Pleroma doesn't use previews, it still has limit on upload size, or?

a1batross:
@charlag not multi-gb but PhotoView do not allow to see non-downscaled image even if you zoom pinch the picture. The problem is not file size but resolution.

>I believe it's not what we need, it's quite a slowdown.

Point taken. I like how Glide manages the image cache and working around the cache just because library doesn't allow it doesn't seem right to me.

connyduck:
I have tested it for a while and it seems to work fine. It feels weird sometimes but thats probably because I am not used to it.

charlag:
Thanks! Do you know what feels weird? I think double tap is kinda slow to zoom and pull down could be improved

connyduck:
Yeah mostly the pull down. On my slower device sometimes transition animations are missing, but I could not yet reproduce it reliably.

charlag:
Found a post which crashes it :c
https://social.wxcafe.net/@wxcafe/104389268812639645

I removed `centerInsde()`, maybe that's the problem.
maybe @a1batross was right

connyduck:
doesn't crash on my device 🤔

PR #1846, opened 2020-06-21
Description improvements

charlag:
The branch is based on #1843

Increase description char count and use bottom sheet for viewing image descriptions.

I could use some help regarding the sheet. I don't know how to make a better border. There was also a suggestion to fade text when some part is not visible but I'm not sure how to detect that and how to display that, maybe we can omit it for now. It's just some indication of whether you need to pull the sheet or not would be nice.

### DIFF
--- a/app/src/main/res/drawable/description_bg_expanded.xml
+++ b/app/src/main/res/drawable/description_bg_expanded.xml
@@ -1,0 +1,10 @@
+<?xml version="1.0" encoding="utf-8"?>
+<shape xmlns:android="http://schemas.android.com/apk/res/android">
+    <corners
+        android:topLeftRadius="6dp"
+        android:topRightRadius="6dp" />
+    <stroke
+        android:color="?attr/colorBackgroundAccent"
+        android:width="1dp" />
+    <solid android:color="#B3000000" />
+</shape>
--- a/app/src/main/res/drawable/ic_drag_indicator_horiz_24dp.xml
+++ b/app/src/main/res/drawable/ic_drag_indicator_horiz_24dp.xml
@@ -1,0 +1,9 @@
+<?xml version="1.0" encoding="utf-8"?>
+<shape android:shape="rectangle"
+    xmlns:android="http://schemas.android.com/apk/res/android">
+    <size
+        android:width="8dp"
+        android:height="1dp" />
+    <solid android:color="#fff" />
+    <corners android:radius="1dp" />
+</shape>
--- a/app/src/main/res/drawable/ic_drag_indicator_horiz_24dp.xml
+++ b/app/src/main/res/drawable/ic_drag_indicator_horiz_24dp.xml
@@ -4,6 +4,6 @@
     <size
         android:width="8dp"
         android:height="1dp" />
-    <solid android:color="#fff" />
+    <solid android:color="?attr/colorBackgroundAccent" />
     <corners android:radius="1dp" />
 </shape>
--- a/app/src/main/res/layout/fragment_view_image.xml
+++ b/app/src/main/res/layout/fragment_view_image.xml
@@ -24,18 +24,54 @@
         app:layout_constraintRight_toRightOf="parent"
         app:layout_constraintTop_toTopOf="parent" />
 
-    <TextView
-        android:id="@+id/mediaDescription"
-        android:layout_width="match_parent"
-        android:layout_height="wrap_content"
-        android:background="#60000000"
-        android:hyphenationFrequency="full"
-        android:lineSpacingMultiplier="1.1"
-        android:padding="8dp"
-        android:textAlignment="center"
-        android:textColor="#eee"
-        android:textSize="?attr/status_text_medium"
-        app:layout_constraintBottom_toBottomOf="parent"
-        tools:text="Some media description" />
+    <!--    This should be inside CoordinatorLayout for two reasons:
 
+            1. TouchImageView really wants some constraints ans has no size otherwise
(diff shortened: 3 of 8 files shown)


connyduck:
Please rebase so its easier to review

charlag:
@connyduck done

app/src/main/res/layout/fragment_view_image.xml
```diff
@@ -24,18 +24,54 @@
...
+        <LinearLayout
+            android:id="@+id/captionSheet"
+            android:layout_width="match_parent"
+            android:layout_height="wrap_content"
+            android:background="@drawable/description_bg_expanded"
+            android:orientation="vertical"
+            app:behavior_peekHeight="90dp"
+            app:layout_behavior="com.google.android.material.bottomsheet.BottomSheetBehavior">
+
+            <ImageView
```
connyduck, line 50: Better make that just a `View` and set the handle drawable as background

app/src/main/res/layout/fragment_view_image.xml
```diff
@@ -24,18 +24,54 @@
...
         app:layout_constraintTop_toTopOf="parent" />
 
-    <TextView
-        android:id="@+id/mediaDescription"
+    <!--    This should be inside CoordinatorLayout for two reasons:
+
+            1. TouchImageView really wants some constraints ans has no size otherwise
+            2. We don't want sheet to overlap with appbar but the only way to do it with autosizing
+            is to gibe parent some margin. -->
+    <androidx.coordinatorlayout.widget.CoordinatorLayout
```
connyduck, line 32: Did you try to have the `CoordinatorLayout` as the top level layout? With `TouchImageView`, `Progressbar` and bottom sheet as childs. Should be possible imho.

app/src/main/res/drawable/description_bg_expanded.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="utf-8"?>
+<shape xmlns:android="http://schemas.android.com/apk/res/android">
+    <corners
+        android:topLeftRadius="6dp"
+        android:topRightRadius="6dp" />
+    <stroke
+        android:color="?attr/colorBackgroundAccent"
+        android:width="1dp" />
+    <solid android:color="#B3000000" />
+</shape>
```
connyduck, line 10: try this for no border at the bottom

```
<layer-list xmlns:android="http://schemas.android.com/apk/res/android">
    <item>
        <shape>
            <solid android:color="<border color>" />
            <corners
                android:topLeftRadius="6dp"
                android:topRightRadius="6dp" />
        </shape>
    </item>
    <item
        android:left="1dp"
        android:right="1dp"
        android:top="1dp">
        <shape>
            <solid android:color="<body color>" />
            <corners
                android:topLeftRadius="5dp"
                android:topRightRadius="5dp" />
        </shape>
    </item>
</layer-list>
```

app/src/main/res/drawable/ic_drag_indicator_horiz_24dp.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="utf-8"?>
+<shape android:shape="rectangle"
+    xmlns:android="http://schemas.android.com/apk/res/android">
+    <size
+        android:width="8dp"
+        android:height="1dp" />
+    <solid android:color="#fff" />
```
connyduck, line 7: maybe also use `?attr/colorBackgroundAccent` here?

app/src/main/res/layout/fragment_view_image.xml
charlag: I tried that first, I left a comment about it in the previous lines. Didn't find a way to prevent overlapping the app bar otherwise.

app/src/main/res/drawable/description_bg_expanded.xml
charlag: I will try that but would still be nice to have some indication of the bottom

connyduck: nah it does not work when the body color is transparent

app/src/main/res/layout/fragment_view_image.xml
connyduck: ah I see 🤦 
Why do Android Layouts suck so hard sometimes

charlag: Yeah it's a question to BottomSheet, you can either set upper limit manually or living auto-sizing decide. It wouldn't add much complexity to expose upper bound IMO

app/src/main/res/drawable/description_bg_expanded.xml
charlag: Any ideas what to do with it? I'm kinda okay with how it is but maybe it should transform somehow when expanded to full height, idk

connyduck: No idea, I think its fine to leave it like it is.

app/src/main/res/layout/fragment_view_image.xml
charlag: This produces a colored square for some reason, not sure why it decides to stretch

connyduck: Because of `android:layout_height="24dp"`probably. Try something lower, maybe adjust margin accordingly

charlag:
@connyduck should be good now, please review

PR #1852, opened 2020-06-30
Improve image viewer v2

a1batross:
After having crashes with too wide images with TouchImageView that I didn't figured out how to debug, I've decided to migrate to another library once again. I've found BigImageViewer that fits my needs pretty nice:
* It can open high resolution images because it uses SSIV internally
* It supports GIF and WEBP

It has Glide support but it's hidden inside of library so we can't execute Glide commands and customize them, except adding a new decoder. That something I'm not really sure about, does it fits Tusky but let's see how it behaves in the wild. I didn't noticed where it may crash or corrupt images or just don't load them. :shrug:


app/src/main/java/com/keylesspalace/tusky/components/compose/dialog/CaptionDialog.kt
```diff
@@ -50,9 +52,9 @@ fun <T> T.makeCaptionDialog(existingDescription: String?,
     dialogLayout.setPadding(padding, padding, padding, padding)
 
     dialogLayout.orientation = LinearLayout.VERTICAL
-    val imageView = TouchImageView(this).apply {
-        maxZoom = 6f
-    }
+    val imageView = BigImageView(this)
+    // imageView.ssiv.maxScale = 6f
```
a1batross, line 56: Woops. TODO for myself. ssiv is not initialized before we have an image, so it must be put in "onSuccess" method of ImageLoaderCallback.

charlag:
I've tried it out before the current one, it still saves image to the disk first. We need to solve the crash but I wouldn't like to go through the disk always.

a1batross:
@charlag yes, it is. I didn't figured out does it caching internally, it may be not that bad.

Sadly Glide don't store files in cache in a way that we can just take it from there and display.

connyduck:
> I've tried it out before the current one, it still saves image to the disk first.

Whats the problem with this? Glide caches on disk as well? Does it not cleanup saved files?

a1batross:
It seems it doesn't actually copy file but uses what Glide provides as File. So yes, it's still in the cache managed by Glide itself.

https://github.com/Piasy/BigImageViewer/blob/master/GlideImageLoader/src/main/java/com/github/piasy/biv/loader/glide/GlideImageLoader.java#L72

https://github.com/Piasy/BigImageViewer/blob/92ffdc73be9243a514d985bdd928e64605aab139/BigImageViewer/src/main/java/com/github/piasy/biv/view/BigImageView.java#L436

a1batross:
Updated PR.

connyduck:
I tested this and it feels slower to open images than what is currently on develop. All three variants (current release, develop, this) have in common that they never feel 100% right and I don't know how to objectively compare them. If @charlag is fine with it I can merge this.

charlag:
My main concerns about this library were that it will be slower (because we would wait for the disk) and the disk churn (which might result in encryption and more CPU work). I don't fully understand how this transition and file code works but if it doesn't save additional files I think it's not a problem.

app/src/main/java/com/keylesspalace/tusky/fragment/ViewImageFragment.kt
```diff
@@ -181,99 +201,18 @@ class ViewImageFragment : ViewMediaFragment() {
...
-    private inner class ImageRequestListener(
-            private val isCacheRequest: Boolean,
-            private val isThumnailRequest: Boolean) : RequestListener<Drawable> {
-
-        override fun onLoadFailed(e: GlideException?, model: Any, target: Target<Drawable>,
-                                  isFirstResource: Boolean): Boolean {
-            // If cache for full image failed complete transition
-            if (isCacheRequest && !isThumnailRequest && shouldStartTransition
-                    && !startedTransition) {
-                photoActionsListener.onBringUp()
```
charlag, line 243: Without this code it doesn't start transition immediately on cache hit/ miss, it waits for the full download before unblocking the UI. I just tried it out and it's indeed the case. That's no-no.
I guess we could manually do this part with Glide (so do cache only request) and then do full download.

a1batross:
About transitions: I didn't understood what's going on with this code, but indeed that must be fixed otherwise UI may be unresponsive until image loads from network/disk cache. BIV provides us a onCacheMiss/Hit callbacks so it must be possible to handle it from there. But I'm not sure if they actually work as their name implies.

Also, I found a bug with the photos are not being re-oriented. Probably SSIV doesn't handle it automatically, so I would fix that as well.

charlag:
@a1batross are you working on the transition? Otherwise we could revert back to PhotoView because current library clearly doesn't cut it.

a1batross:
@charlag sorry, I don't really have time to work on it this month, too much deadlines.

charlag:
@a1batross it's fine, just want to prevent parallel work

a1batross:
@charlag well... Didn't knew I would be free enough on my weekend to finalize it. :)

I've did the transition/cache lookup thing. Works pretty fine for me, GUI doesn't stuck, although my network speed is high enough to barely notice difference between preview->full image transition. 

So I request for another review, except merging this is still a bit too early, unless https://github.com/Piasy/BigImageViewer/pull/199 is merged and I can proceed to make image swiping work for GIFs and animated WebP.

a1batross:
Updated the PR, as patch to BigImageViewer was merged.

a1batross:
Closed due to reworking on the top of databinding refactor.

PR #1855, opened 2020-07-06
Make account views clickable instead of avatar only

ariasuni:
Affects:
- New follower request and New follower notifications
- Accounts in Blocked users and Muted users pages

Action buttons, when present, still work.


connyduck:
Nice, thanks

PR #1856, opened 2020-07-07
Implement optional notifications muting for account muting

### LINKED ISSUE #484: Add support for improved muting
New Mastodon 2.1 feature: https://github.com/tootsuite/mastodon/releases/tag/v2.1.0rc1

ariasuni:
fix #484

Screenshot_20200725-222304_Tusky
Screenshot_20200720-194632_Tusky

Basically, what’s left to do is to find a way to retrieve the `mutingNotifications` state inside the Muted users page (`app/src/main/java/com/keylesspalace/tusky/adapter/MutesAdapter.java`). Since I don’t have a lot of experience with the Tusky codebase, I’m stuck at the moment.

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/AccountActivity.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/AccountActivity.kt
@@ -58,6 +58,7 @@ import com.keylesspalace.tusky.interfaces.LinkListener
 import com.keylesspalace.tusky.interfaces.ReselectableFragment
 import com.keylesspalace.tusky.pager.AccountPagerAdapter
 import com.keylesspalace.tusky.util.*
+import com.keylesspalace.tusky.view.showMuteAccountDialog
 import com.keylesspalace.tusky.viewmodel.AccountViewModel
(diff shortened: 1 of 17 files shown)


connyduck:
The information if notifications are muted is not part of the `Account` object so you need to get the relationships for all users returned by `/api/v1/mutes`. The [/api/v1/accounts/relationships](https://docs.joinmastodon.org/methods/accounts/#general-account-actions) endpoint takes an array of account ids and returns the relationships.

ariasuni:
I think it’s good now. The only thing I’m not entirely satisfied with is that clicking on the notifications icon reload the view in a noticeable way, i.e. the whole account view updates with a fading effect.

Also, I added tooltips to icons, since there are two of them and the meaning of it may not be immediately obvious.

iknownothingaboutcomputers:
Hope this much-needed feature comes with the next update.

ariasuni:
> Works perfect, awesome :heart_eyes: 

Thanks!

> Also please reformat all .xml files so the attributes are in a nice order.

Like this?

PR #1858, opened 2020-07-15
remove scrolling from notification options bar

connyduck:
After careful consideration I think this is the best option to solve the problem with the double hiding bars.


charlag:
Didn't get around to try it today, will try tomorrow

PR #1863, opened 2020-07-23
Roll back image viewer but retain swiping fix

charlag:
This fixes crashes with images which were too wide (aspect ratio wise).

I've found out that there _is_ an option to prevent parent from scrolling and we have to write less brittle touch code. I left swipe gesture because I think it's nice.
Overall it's not as smooth as TouchImageView but the huge plus is that it should stop crashing (and I would be glad if anyone verified this)

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -159,7 +159,7 @@ dependencies {
 
     implementation "com.github.connyduck:sparkbutton:4.0.0"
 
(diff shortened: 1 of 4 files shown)


connyduck:
Build is failing because you still have the `TouchImageView` in the `CaptionDialog`

charlag:
Oops, sorry, forgot to check it in

PR #1871, opened 2020-08-01
Fix crash in the image viewer

charlag:
This is imperfect as it probably reduces quality of the image but it
fixes the crash for now.


connyduck:
Do you have any idea how the crash was introduced? It did not crash in Tusky 11 and it is the same ImageView library

charlag:
When I changed library I also removed `centerInside()`, I thought it's unnecessary but now I realize that it resizes the image somehow and prevents huge canvases. Without it long, narrow images create canvas which is too long. Ideally we would do this dynamically, based on zoom level with subsampling library but it's okay for now I think.

connyduck:
thx 👍

Issue #1893, opened 2020-08-16
[Feature Request] Setting for hiding the titles in the header bar


naums:
I'd like to be able to hide the titles in the header bar, making the header a bit less high. The search-button could move to left drawer menu. Please add a settings-entry for toggling a smaller header bar.

* * * *
- Tusky Version: 12.0
- Android Version: 10
- Android Device: Nokia 7.1

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

PR #1895, opened 2020-08-16
add setting to hide top toolbar

connyduck:
closes https://github.com/tuskyapp/Tusky/issues/1893


connyduck:
Same, but there where like 10 requests for it 🤷

Le1b1:
With that setting activated I'm unable to open the menu on my device with gesture navigation.
Because the swipe from the left side of the screen is the back gesture and doesn't open the menu. :-(

connyduck:
You have to long-press on the side of the screen.

PR #1904, opened 2020-08-20
Fix opening the public filter list from account preferences

Tak:
Previously, clicking "public timelines" would open the "conversations" list.

It looks like this happened during the settings refactor?


connyduck:
Thx! Some users already complained about this

PR #1905, opened 2020-08-20
fix notifications sometimes not rendering

connyduck:
Sometimes, when switching to the `NotificationsFragment`, it does not render the content of the notifications, only the dividers. Looks like this: 

I don't know why, but this fixes the problem

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/fragment/NotificationsFragment.java
+++ b/app/src/main/java/com/keylesspalace/tusky/fragment/NotificationsFragment.java
@@ -79,7 +79,6 @@ import com.keylesspalace.tusky.util.ListUtils;
 import com.keylesspalace.tusky.util.NotificationTypeConverterKt;
 import com.keylesspalace.tusky.util.PairedList;
 import com.keylesspalace.tusky.util.StatusDisplayOptions;
-import com.keylesspalace.tusky.util.ThemeUtils;
 import com.keylesspalace.tusky.util.ViewDataUtils;
 import com.keylesspalace.tusky.view.BackgroundMessageView;
 import com.keylesspalace.tusky.view.EndlessOnScrollListener;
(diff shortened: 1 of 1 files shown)


Tak:
I've seen the same thing on the federated timeline - do we need a similar fix somewhere else?

connyduck:
Yes, I have also seen it in other places, but could only reproduce it on the notifications fragment. Will keep trying.

charlag (approved):
This is super weird but whatever works

connyduck:
Wait. I can also reproduce it within `TimelineFragment`. The fix is also to not scroll up on the first load. But getting the adapter count is not working there 🤔

charlag:
Hm, probably data is loaded earlier for it and from disk? I don't know how to detect that it's "too early"
I think it would be useful to know how did you reproduce it.

connyduck:
To reproduce, set the tab you want to bug out to position 2 of 4, kill the app, reopen, wait for everything to load, go to tab 4 by clicking on the tabbar, wait, go to tab 2 by clicking on the tabbar

connyduck:
Ok, now its fixed for `TimelineFragment` as well!

Explanation: `adapter.getItemCount()` gives the number of loaded+new items, while `count` is the number of new items only. When `adapter.getItemCount() != count` we know its not the first load and can safely scroll up. Only checking for `adapter.getItemCount() > 0` was not enough for stopping the bug in the timeline.

Apparently calling `scrollBy` while the recyclerview is in the background causes it to not render items correctly 🤷‍♂️

charlag:
Cool, cannot believe you got to the bottom of this!
I think this should be a comment in code, no one would remember why this condition is like that in two weeks

Tak:
That explains why I never saw it on notifications, that's always my first tab

PR #1910, opened 2020-08-23
explicitly set text colors in LicenseCard

connyduck:
looks better this way imho


charlag (approved):

charlag:
Really looks much better

Issue #1301, opened 2019-06-01
Feature request: avatar as menu icon


turtlebird:
It seems relatively common for Mastodon users to have a number of alt accounts. It would be easier to identify at a glance which account you are browsing and posting from if the three horizontal bar menu icon in the top left hand corner were instead the avatar of the currently active account. Same functionality otherwise, just a visual change.

* * * *
- [X] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

PR #1919, opened 2020-08-28
use current user's avatar as drawer toggle

connyduck:
Screenshot_1598632002

closes https://github.com/tuskyapp/Tusky/issues/1301

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/MainActivity.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/MainActivity.kt
@@ -171,7 +171,8 @@ class MainActivity : BottomSheetActivity(), ActionButtonActivity, HasAndroidInje
         val hideTopToolbar = preferences.getBoolean(PrefKeys.HIDE_TOP_TOOLBAR, false)
         mainToolbar.visible(!hideTopToolbar)
 
-        mainToolbar.navigationIcon = FixedSizeDrawable(getDrawable(R.drawable.avatar_default), Utils.dpToPx(this, 36), Utils.dpToPx(this, 36))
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/keylesspalace/tusky/MainActivity.kt
```diff
@@ -169,6 +171,8 @@ class MainActivity : BottomSheetActivity(), ActionButtonActivity, HasAndroidInje
         val hideTopToolbar = preferences.getBoolean(PrefKeys.HIDE_TOP_TOOLBAR, false)
         mainToolbar.visible(!hideTopToolbar)
 
+        mainToolbar.navigationIcon = FixedSizeDrawable(getDrawable(R.drawable.avatar_default), Utils.dpToPx(this, 36), Utils.dpToPx(this, 36))
```
connyduck, line 174: The navigation icon is not automatically sized correctly, and it ignores bounds set on the drawable, but it works when using this `FixedSizeDrawable` from Glide.

a1batross: What does 36 mean here?

Looks like magic number to me.

a1batross:
Looks great! :)

app/src/main/java/com/keylesspalace/tusky/MainActivity.kt
charlag: I'm okay with it being a magic number but we use it 3 times so I would put it into constant.

charlag (approved):
Works well for me, I like it.
It kinda subverts my expectations of what happens when I click on it but I think it's fine.

PR #1922, opened 2020-08-31
convert EmojiPreference and EmojiCompatFont to Kotlin

connyduck:
- RxJava instead of AsyncTask
- inject the OkHttpClient instead of creating one for each request
- handles the edge-case where the download size can't be determinded by showing an indetermined progress bar
- code still does the same but hopefully is way more readable now


app/proguard-rules.pro
```diff
@@ -55,15 +55,13 @@
     public static *** v(...);
     public static *** i(...);
 }
+-assumenosideeffects class java.lang.String {
+    public static java.lang.String format(...);
```
connyduck, line 59: This is to get rid of lonely `String.format`s in the bytecode after the log statements have been stripped out.

charlag: You would expect Proguard could figure it out

app/src/main/java/com/keylesspalace/tusky/components/preference/PreferencesFragment.kt
```diff
@@ -13,22 +13,27 @@
...
 import com.mikepenz.iconics.utils.colorInt
 import com.mikepenz.iconics.utils.sizePx
+import okhttp3.OkHttpClient
+import javax.inject.Inject
 
-class PreferencesFragment : PreferenceFragmentCompat() {
+class PreferencesFragment : PreferenceFragmentCompat(), Injectable {
+
+    @Inject
+    lateinit var okhttpclient: OkHttpClient
```
a1batross, line 36: I'm maybe wrong here but isn't OkHttpClient we have there has interceptor to add user token? 
This could lead to token leakage.

connyduck: no it can't we only send the token with requests made by Retrofit

app/src/main/java/com/keylesspalace/tusky/components/preference/EmojiPreference.kt
```diff
@@ -0,0 +1,250 @@
...
+                        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.N) {
+                            progressBar.setProgress((max * progress).toInt(), true)
+                        } else {
+                            progressBar.progress = (max * progress).toInt()
+                        }
+                    } else {
+                        progressBar.isIndeterminate = true
+                    }
+                },
+                Consumer {
```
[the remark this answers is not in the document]
charlag: If we would need to cancel, we could dispose it on our own. it's more bookkeeping outside of font but I think it's cleaner

app/src/main/java/com/keylesspalace/tusky/settings/SettingsConstants.kt
```diff
@@ -17,7 +17,7 @@ object PrefKeys {
     // each preference a key for it to work.
 
     const val APP_THEME = "appTheme"
-    const val EMOJI = "emojiCompat"
+    const val EMOJI = "selected_emoji_font"
```
charlag, line 20: "selected" is redundant IMO but it's not that important

app/src/main/java/com/keylesspalace/tusky/util/EmojiCompatFont.kt
```diff
@@ -0,0 +1,363 @@
...
+     * on whether they are older than the currently available version or not.
+     *
+     * @param context The Context
+     */
+    private fun getExistingFontFiles(context: Context): List<Pair<File, List<Int>>> {
+        // Only load it once
+        existingFontFileCache?.let {
+            return it
+        }
+            // If we call this on the system default font, just return nothing...
```
charlag, line 128: pls reformat

app/src/main/java/com/keylesspalace/tusky/components/preference/PreferencesFragment.kt
a1batross: Yeah, didn't noticed the check for domain. Nice!

app/src/main/java/com/keylesspalace/tusky/util/EmojiCompatFont.kt
```diff
@@ -0,0 +1,363 @@
...
+    }
+
+    fun downloadFontFile(context: Context,
+                         okHttpClient: OkHttpClient,
+                         onNext: Consumer<Float>,
+                         onError: Consumer<Throwable>,
+                         onComplete: Action) {
+        Observable.create { emitter: ObservableEmitter<Float> ->
+            // It is possible (and very likely) that the file does not exist yet
+            val downloadFile = getFontFile(context)!!
```
charlag, line 186: tbh I'm surprised that Okio doesn't support any nio, at least I couldn't find anything.

charlag (commented):
Hey, one `AsyncTask` down, nice!

charlag (approved):

Issue #1911, opened 2020-08-24
The tab with hashtags doesn't update


Hund:
I added a tab with 3 hashtags and everything worked as expected. A few days later I added a bunch of more hashtags, the issue now is that it doesn't update that tab with new posts from the hashtags I added.

* * * *
- Tusky Version: 12.0 (F-Droid)
- Android Version: 10
- Android Device: Pixel 3a

Hund:
I added a new tab with all the hashtags at the same time and it looks like it's only showing the first 3 hashtags.

connyduck:
What Mastodon instance?

Hund:
It's Fosstodon.org.

Hund:
I noticed today that I can only remove a current hashtag if I add a new one at the same time.

It also looks like it's showing the first 4 hashtags for me now.

Hund:
This is still an issue for me.

connyduck:
which version are you on now?

Hund:
12.1 from F-Droid.

connyduck:
Can you check the 13 beta as well please?

Hund:
Sure. How do I do that without installing my current installation?

connyduck:
it should be available on Fdroid as well

Hund:
I did not know that. I upgraded, but it didn't help.

connyduck:
Oh wow I can finally reproduce it. It is an mastodon server bug that occurs on the webpage as well.
https://github.com/tootsuite/mastodon/issues/15194

Hund:
Considering that Mastodon [won't fix this issue](https://github.com/mastodon/mastodon/issues/15194#issuecomment-742015124) would you consider working around this in Tusky? I find it stupid that you're limited to **four** hashtags by default in Mastodon and that he's refusing to fix it.

charlag:
IIs not something we can fix, we do not assemble timelines.

Hund:
> IIs not something we can fix, we do not assemble timelines.

You don't assemble timelines? I'm pretty sure I have a timeline of post from people that I follow, a public timeline of the posts from the instance and a public timeline of all federated posts in Tusky. I'm also able to search for posts, users and hashtags. 

Perhaps I'm mistaken, perhaps I'm using some other client with the same name? ;)

charlag:
Perhaps you do or perhaps you misunderstand what we do.
[...]

PR #1925, opened 2020-09-01
fix various bugs when creating tab with multiple hashtags

connyduck:
There where 3 problems which led to weird glitches:

- Tabs were not saved after removing hashtags
- comparing tabs with equals is a bad idea, because they contain function types and there could be two distinct tabs with exactly the same configuration -> use the index instead
- when removing hashtag the wrong views got removed


connyduck:
I don't think this fixes https://github.com/tuskyapp/Tusky/issues/1911 (which I could not yet reproduce, but I found these other bugs while trying to do so 😅 )

charlag (approved):

PR #1931, opened 2020-09-05
Fix some of the StrictMode violations

charlag:
Mostly it's disk reads/writes. I fixed those for AccountManager which
are reported at startup.

Conscrypt reads own version on startup which reads from disk multiple
times. There's no solution for it right now.

SharedPreferences which are used in BaseActivity also read from disk
and pretty early but it shouldn't be a problem.

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/MainActivity.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/MainActivity.kt
@@ -81,6 +81,7 @@ import com.uber.autodispose.android.lifecycle.autoDispose
 import dagger.android.DispatchingAndroidInjector
 import dagger.android.HasAndroidInjector
 import io.reactivex.android.schedulers.AndroidSchedulers
+import io.reactivex.schedulers.Schedulers
 import kotlinx.android.synthetic.main.activity_main.*
 import javax.inject.Inject
 
@@ -227,8 +228,10 @@ class MainActivity : BottomSheetActivity(), ActionButtonActivity, HasAndroidInje
                     }
                 }
 
-        // Flush old media that was cached for sharing
-        deleteStaleCachedMedia(applicationContext.getExternalFilesDir("Tusky"))
+        Schedulers.io().scheduleDirect {
+            // Flush old media that was cached for sharing
+            deleteStaleCachedMedia(applicationContext.getExternalFilesDir("Tusky"))
+        }
     }
 
     override fun onResume() {
--- a/app/src/main/java/com/keylesspalace/tusky/TuskyApplication.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/TuskyApplication.kt
@@ -25,11 +25,15 @@ import androidx.work.WorkManager
 import com.keylesspalace.tusky.components.notifications.NotificationWorkerFactory
 import com.keylesspalace.tusky.di.AppInjector
 import com.keylesspalace.tusky.settings.PrefKeys
-import com.keylesspalace.tusky.util.*
+import com.keylesspalace.tusky.util.EmojiCompatFont
+import com.keylesspalace.tusky.util.LocaleManager
+import com.keylesspalace.tusky.util.ThemeUtils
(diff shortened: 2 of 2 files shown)


app/src/main/java/com/keylesspalace/tusky/MainActivity.kt
```diff
@@ -210,8 +211,10 @@ class MainActivity : BottomSheetActivity(), ActionButtonActivity, HasAndroidInje
                     }
                 }
 
-        // Flush old media that was cached for sharing
-        deleteStaleCachedMedia(applicationContext.getExternalFilesDir("Tusky"))
+        Schedulers.io().scheduleDirect {
```
connyduck, line 214: oh, didn't know about `scheduleDirect`, nice

app/src/main/java/com/keylesspalace/tusky/db/AccountManager.kt
```diff
@@ -40,11 +41,16 @@ class AccountManager @Inject constructor(db: AppDatabase) {
     private val accountDao: AccountDao = db.accountDao()
 
     init {
-        accounts = accountDao.loadAll().toMutableList()
-
-        activeAccount = accounts.find { acc ->
-            acc.isActive
-        }
+        val disposable = accountDao.loadAll()
```
connyduck, line 44: I'm afraid this could cause race conditions? So that `SplashActivity` is checking for existing accounts, but they haven't loaded yet and user would falsely be presented with the login screen?

app/src/main/java/com/keylesspalace/tusky/TuskyApplication.kt
```diff
@@ -38,11 +42,21 @@ class TuskyApplication : Application(), HasAndroidInjector {
...
     @Inject
     lateinit var androidInjector: DispatchingAndroidInjector<Any>
+
     @Inject
-    lateinit var notificationWorkerFactory: NotificationWorkerFactory
+    lateinit var notificationWorkerFactory: Lazy<NotificationWorkerFactory>
 
     override fun onCreate() {
-
+        // Uncomment me to get StrictMode violation logs
```
connyduck, line 50: How about enabling it for debug builds?

connyduck (commented):
Not sure about the account manager thing, otherwise really nice 👍

app/src/main/java/com/keylesspalace/tusky/db/AccountManager.kt
charlag: Hm yeah we should make it wait for this probably.

app/src/main/java/com/keylesspalace/tusky/TuskyApplication.kt
charlag: we get spam because of the SharedPreferences/Conscrypt so I don't know

connyduck: I wouldn't mind in debug builds, but you can also leave it commented

app/src/main/java/com/keylesspalace/tusky/db/AccountManager.kt
connyduck: It can happen everywhere, when Android recreates an Activity without going over the SplashScreen, e.g. when opening the app from recents. Maybe we should not make the initial load in account manager async.

charlag: I think the app start is the most painful place regarding blocking right now but I agree that we need to change other places then to expect it async. Should I do it or should we split it into a separate PR?

connyduck: I would leave the initial loading of the account on the main thread
- yes it blocks the thread, but it cannot do anything useful in the meantime anyway - splash screen will be shown even when another thread does the loading
- moving the loading off the main screen means making getAccount somehow async and that will be a lot of changes, it won't be worth it

charlag:
I removed all accountManager changes

Issue #1935, opened 2020-09-15
Can’t properly translate the remaining time for a poll


ariasuni:
String `poll_info_time_relative` is  `%s left`. However, this doesn’t take into account that, depending on what word is used (day, hour, minute) and if the word is plural or not, the string isn’t translated the same. In French, `%s left` is translated in four different ways depending on context, for example:

– if `%s` is 1 hour: `%s restante`
– if `%s` is 2 hours: `%s restants`
– if `%s` is 1 day: `%s restant`
– if `%s` is 2 days: `%s restants`

I’m not sure how Android handle this, but there should be at least one translation per word used and then the plural correctly handled.

* * * *
- Tusky Version: 12.1
- Android Version: 9
- Android Device: Oneplus 5 (cheeseburger)

connyduck:
That is indeed an interesting edge case. Android can handle plurals, but in this case its a plural string nested in another string.

Now, the 4 `poll_timespan_X` strings are only used inside the `poll_info_time_relative`, so how about you translate `poll_info_time_relative` with `%s` only and put the equivalence of `left` inside the `poll_timespan_X`s as well? That is kinda hacky but should work.

ariasuni:
Well, I’m proposing that all translations use the same scheme, because it’s certainly not what I would case a edge case, because it affects a lot of languages, and it would be very hard to guess that the right thing is to do this weird trick.

Starting with the English (source) translation,  `poll_info_time_relative` could be changed to `%s` and `poll_timespan_X` changed to include “left” at the end of the string.  `poll_info_time_relative` can be kept for some time to let people update translations. Both `poll_info_time_relative` and `poll_timespan_X` should be updated at the same time, but it’s much better than breaking everything.

PR #1939, opened 2020-09-21
improve relative poll timestamp string format

connyduck:
closes https://github.com/tuskyapp/Tusky/issues/1935

unfortunately I was not able to correct the strings for rtl languages

cc @BoFFire @ariasuni


charlag (commented):
It looks fine but I :shrug:

PR #1946, opened 2020-10-02
Set onClick listener for poll result view

### LINKED ISSUE #1861: Tapping a poll in a notification should expand the status
Tapping a poll in a notification doesn't expand the status for that poll

* * * *
- Tusky Version: 11.0
- Android Version: 10
- Android Device: Pixel 2 XL
- Mastodon instance (if applicable): glitch.taks.garden

- ☑️ I searched or browsed the repo’s other issues to ensure this is not a duplicate.

Tak:
Fixes #1861

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/adapter/StatusBaseViewHolder.java
+++ b/app/src/main/java/com/keylesspalace/tusky/adapter/StatusBaseViewHolder.java
@@ -915,12 +915,18 @@ public abstract class StatusBaseViewHolder extends RecyclerView.ViewHolder {
 
         if (expired || poll.getVoted()) {
             // no voting possible
-            pollAdapter.setup(poll.getOptions(), poll.getVotesCount(), poll.getVotersCount(), emojis, PollAdapter.RESULT);
+            View.OnClickListener viewThreadListener = v -> {
+                int position = getAdapterPosition();
+                if (position != RecyclerView.NO_POSITION) {
+                    listener.onViewThread(position);
+                }
+            };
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/keylesspalace/tusky/adapter/StatusBaseViewHolder.java
```diff
@@ -915,12 +915,18 @@ private void setupPoll(PollViewData poll, List<Emoji> emojis,
 
         if (expired || poll.getVoted()) {
             // no voting possible
-            pollAdapter.setup(poll.getOptions(), poll.getVotesCount(), poll.getVotersCount(), emojis, PollAdapter.RESULT);
+            View.OnClickListener viewThreadListener = v -> {
```
connyduck, line 918: There is already the same listener created in line 669 - it would be cool if we could reuse that somehow - how about just setting it on the whole RecyclerView?

Tak: You mean on the view holder? Or I'm totally misunderstanding what you're getting at?

connyduck: I meant just doing `pollOptions.setOnClickListener(viewThreadListener);` in line 677, but it does not work I tried it out. Your solution is fine 👍

Tak: Oh yeah, I looked at that too, but it apparently needs to be specifically on the result subview :shrug:

PR #1947, opened 2020-10-02
Delete scheduled post after editing

### LINKED ISSUE #1924: Editing a scheduled toot doesn't remove the previous one
When editing a scheduled toot, either by changing the content and/or the publishing time, the previous toot isn't removed so we end up with an additional one.

It gets confusing when only the publishing time is edited since we end with duplicates and can't know which toot is the right one.

* * * *
- Tusky Version: 12.1
- Android Version: 9
- Android Device: SM-G900F

garritfra:
This ensures that editing a scheduled toot will not duplicate it.

If the ID of a scheduled post is being passed to the compose activity, it will delete this scheduled post and create a new post with the same parameters.

Fixes #1924 

## Room for improvement

Deleting a toot and reposting it was the simplest way to accomplish this. There is a very slight chance of something going wrong in one of these two operations though. Alternatively, there is `PUT /scheduled_statuses/:id`, but it is not yet implemented in the app. Please let me know what you think about this, and if you need help with possibly implementing it.

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/components/compose/ComposeViewModel.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/components/compose/ComposeViewModel.kt
@@ -36,6 +36,7 @@ import com.keylesspalace.tusky.service.ServiceClient
 import com.keylesspalace.tusky.service.TootToSend
 import com.keylesspalace.tusky.util.*
 import io.reactivex.Observable.empty
+import io.reactivex.Observable.just
 import io.reactivex.disposables.Disposable
 import io.reactivex.rxkotlin.Singles
 import java.util.*
@@ -249,7 +250,7 @@ class ComposeViewModel
         val deletionObservable = if (isEditingScheduledToot) {
             api.deleteScheduledStatus(scheduledTootUid.toString()).toObservable().map { Unit }
         } else {
-            empty()
+            just(Unit)
         }.toLiveData()
 
         val sendObservable = media
--- a/app/src/main/java/com/keylesspalace/tusky/components/compose/ComposeViewModel.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/components/compose/ComposeViewModel.kt
@@ -19,6 +19,7 @@ import android.net.Uri
 import android.util.Log
 import androidx.core.net.toUri
 import androidx.lifecycle.LiveData
+import androidx.lifecycle.MediatorLiveData
 import androidx.lifecycle.MutableLiveData
 import androidx.lifecycle.Observer
 import com.keylesspalace.tusky.adapter.ComposeAutoCompleteAdapter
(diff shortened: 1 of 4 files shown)


connyduck (commented):
Deleting and recreating should be fine - there is also the possibility that the user wants to send the status immediately and then it needs to be deleted anyway.

But where do you actually pass the scheduled toot id to the `ComposeActivity`? Did you forget to commit a file? And why is `scheduledTootUid` Int? `ScheduledStatus.id` is String.

garritfra:
> But where do you actually pass the scheduled toot id to the ComposeActivity?

Whoops, this somehow slipped through. Bizarre enough though, it works without passing the id. I'll investigate this and push a commit in a minute.

garritfra:
@connyduck I went down a deep rabbithole of obervables and livedata. Been a while since I worked with it. My approach works as expected, but has a nasty implementation. There are probably better solutions, and I'd be happy to take feedback.

app/src/main/java/com/keylesspalace/tusky/components/compose/ComposeViewModel.kt
```diff
@@ -445,6 +466,10 @@ class ComposeViewModel
         scheduledAt.value = newScheduledAt
     }
 
+    private fun isEditingScheduledToot(): Boolean {
+        return !scheduledTootUid.isNullOrEmpty()
+    }
+
```
connyduck, line 472: As per Kotlin styleguide https://kotlinlang.org/docs/reference/coding-conventions.html#functions-vs-properties one could make this a property

```suggestion
    private val isEditingScheduledToot get() = !scheduledTootUid.isNullOrEmpty()
```

connyduck (commented):
This is actually quite nice. But the resulting MediatorLiveData will signal when the first of `deletionObservable` and `sendObservable` is done, maybe better signal when both are done? There is a helper function `combineLiveData` (in `LiveDataUtils`) that can help with that.

app/src/main/java/com/keylesspalace/tusky/components/compose/ComposeViewModel.kt
garritfra: Fair point. Fixed with a4b20b5

garritfra:
@connyduck could you review the changes again?

connyduck (commented):
almost - the `empty()` Observable will never emit so sending a toot without scheduling is now broken. You probably want something like `just(Unit)`.

connyduck:
And for some reason it makes the `Finishing Media Upload` dialog appear for a split second, even though there is nothing to upload 🤔

garritfra:
@connyduck I couldn't really figure out why the "Finish media upload" dialog appears only if you edit a scheduled toot🧐. I wrapped the responsible code in an (redundant?) check, which fixes it.

connyduck (approved):

Issue #1943, opened 2020-09-30
Remark negative count on toot


Porrumentzio:
When the character count gets negative (<0), it's not very evident as there appears no change on color or size.
It would be great to remark a <0 count with some color and/or bold text.

- [ ] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

connyduck:
fixed in https://github.com/tuskyapp/Tusky/pull/1952

PR #1952, opened 2020-10-04
Change characters left color to red when negative

clasick:
Implements request by @Porrumentzio in #1943

The remaining character length indicator in the compose screen turns red when the message exceeds the limit.

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/components/compose/ComposeActivity.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/components/compose/ComposeActivity.kt
@@ -682,17 +682,15 @@ class ComposeActivity : BaseActivity(),
     }
 
     private fun updateVisibleCharactersLeft() {
-        var remainingLength = maximumTootCharacters - calculateTextLength();
+        val remainingLength = maximumTootCharacters - calculateTextLength();
         composeCharactersLeftView.text = String.format(Locale.getDefault(), "%d", remainingLength)
 
-        if(remainingLength < 0) {
-            val textColor = ThemeUtils.getColor(this, android.R.attr.colorError)
-            composeCharactersLeftView.setTextColor(textColor)
+        val textColor = if (remainingLength < 0) {
+            ContextCompat.getColor(this, R.color.tusky_red)
+        } else {
+            ThemeUtils.getColor(this, android.R.attr.textColorTertiary)
         }
-        else {
-            val textColor = ThemeUtils.getColor(this, android.R.attr.textColorTertiary)
-            composeCharactersLeftView.setTextColor(textColor)
-        }
+        composeCharactersLeftView.setTextColor(textColor)
     }
 
     private fun onContentWarningChanged() {
--- a/app/src/main/java/com/keylesspalace/tusky/components/compose/ComposeActivity.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/components/compose/ComposeActivity.kt
@@ -682,7 +682,17 @@ class ComposeActivity : BaseActivity(),
     }
 
     private fun updateVisibleCharactersLeft() {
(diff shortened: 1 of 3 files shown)


Porrumentzio:
Great work! What about making it also bold?

app/src/main/java/com/keylesspalace/tusky/components/compose/ComposeActivity.kt
```diff
@@ -682,7 +682,17 @@ class ComposeActivity : BaseActivity(),
     }
 
     private fun updateVisibleCharactersLeft() {
-        composeCharactersLeftView.text = String.format(Locale.getDefault(), "%d", maximumTootCharacters - calculateTextLength())
+        var remainingLength = maximumTootCharacters - calculateTextLength();
```
connyduck, line 685: can be `val`

app/src/main/java/com/keylesspalace/tusky/components/compose/ComposeActivity.kt
```diff
@@ -682,7 +682,17 @@ class ComposeActivity : BaseActivity(),
...
+        composeCharactersLeftView.text = String.format(Locale.getDefault(), "%d", remainingLength)
+
+        if(remainingLength < 0) {
+            val textColor = ThemeUtils.getColor(this, android.R.attr.colorError)
```
connyduck, line 689: Field requires API level 26 (current min is 21): `android.R.attr#colorError` - better define your own color, ideally one for the dark themes and one for the light theme

app/src/main/java/com/keylesspalace/tusky/components/compose/ComposeActivity.kt
```diff
@@ -682,7 +682,17 @@ class ComposeActivity : BaseActivity(),
...
+            composeCharactersLeftView.setTextColor(textColor)
+        }
+        else {
+            val textColor = ThemeUtils.getColor(this, android.R.attr.textColorTertiary)
+            composeCharactersLeftView.setTextColor(textColor)
+        }
```
connyduck, line 695: You can write this a bit shorter:
```suggestion
        val textColor = if(remainingLength < 0) {
            ThemeUtils.getColor(this, android.R.attr.colorError)
        } else {
            ThemeUtils.getColor(this, android.R.attr.textColorTertiary)
        }
        composeCharactersLeftView.setTextColor(textColor)
```

connyduck (changes requested):

clasick:
Added the changes suggested by @connyduck. The new `tusky_red` color has been defined. I thought it looked fine the way it is on both Night & Day themes as is. Your thoughts?
### Day
image
### Night
image

Porrumentzio:
It is good this way. Maybe it's too aggressive in color, but it's a tiny detail.

If it's easy to make, I would like to see it in bold

clasick:
> It is good this way. Maybe it's too aggressive in color, but it's a tiny detail.
> 
> If it's easy to make, I would like to see it in bold!
> 
> And thanks for the work :)

Yeah, I thought the same myself. The red might be a bit too bright in the Light theme. Could someone suggest a color hex code for the Light one and the Dark one too perhaps? 

I think we can decide first on the exact color to be used and then try making it bold, cause that might alter the perception of the color a bit.

Porrumentzio:
This colour is which Twitter uses: #DF1553 (more or less)
I would also try this one: #CC1900

clasick:
Here's the different combinations with those two color codes with their normal/bold variance.
# Normal
## DF1553
### Dark 
image
### Light
image
## CC1900
### Dark
image
### Light
image

# Bold
## DF1553
### Dark 
image
### Light
image
## CC1900
### Dark
image
### Light
image

Personally, I think bold #DF1553 variants look pretty good.

Porrumentzio:
I also think same way. It looks clearer and less aggressive; and although it really is more pink/purple, it looks red :)

PR #1960, opened 2020-10-16
Use blurhash for preview cards

Tak:
When blurhash is enabled in the user preferences:
- Adds a blurhash placeholder for card images
- Uses blurhash for card images when the user has disabled media preview download
- Uses blurhash for card images when the post is marked sensitive (mostly useful for posts originating from other fediverse software?)


connyduck:
Very nice!

PR #1961, opened 2020-10-17
use MaterialContainerTransform for ActionButton animation

connyduck:
The thing where the CoordinatorLayout does it automatically is deprecated


charlag:
Code-wise it seems fine but I ran into a weeeeird bug when tapping away: back doesn't work (neither hardware nor app).

I just opened screen, clicked on button, clicked away and on button again and then it got buggy. Here's recording:

[device-2020-10-18-211903.zip](https://github.com/tuskyapp/Tusky/files/5398047/device-2020-10-18-211903.zip)

connyduck:
Oops, there was indeed a bug. Thx!

PR #1963, opened 2020-10-19
update Kotlin to 1.4.10 and fix some warnigs

connyduck:
mostly the new "redundant SAM constructor" one


charlag (approved):
SAM thing is such a relief
We can also probably convert some of the interfaces we wrote in Java just for SAM

PR #1976, opened 2020-11-04
Perform bidirectionality isolation manually instead of relying on `BidiFormatter`

### LINKED ISSUE #1921: RLO character in user name breaks notification text
Hi,

I recently got a notification from a user who has a RIGHT-TO-LEFT-OVERRIDE character (U+202E / \&#8238;) in their name, like this: "Name [RLO] some emojis". When receiving a "favourited" notification from this user, the expected "[Name] [emojis] has favourited your toot" (I have it in French so that may not the be the exact wording), it becomes "[Name] toot ruoy detiruovaf sah [emojis]". The web version handles this fine (that is, the emojis are displayed in reverse order, but the rest of the text is LTR as expected). 
Here's what it looks like. Tusky :
image
Web :
image

(I'd imagine a LRO character would similarly break the notification text if the system language is RTL, but I haven't tested it)

* * * *
- Tusky Version: 12.1
- Android Version: 10
- Android Device: Pixel 3

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate. (I found some issues about text direction but I didn't find any about notifications, user names or RLO, sorry if I missed one)

Tak:
Fixes #1921 

I visually verified normal notifications and follow request notifications with the case in the issue report - I'm not sure if we have other test cases we need to verify.

Before
Screenshot_20201104-134745

After
Screenshot_20201104-134835

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/adapter/FollowRequestViewHolder.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/adapter/FollowRequestViewHolder.kt
@@ -1,13 +1,15 @@
 package com.keylesspalace.tusky.adapter
 
 import android.view.View
-import androidx.core.text.BidiFormatter
 import androidx.preference.PreferenceManager
 import androidx.recyclerview.widget.RecyclerView
 import com.keylesspalace.tusky.R
 import com.keylesspalace.tusky.entity.Account
 import com.keylesspalace.tusky.interfaces.AccountActionListener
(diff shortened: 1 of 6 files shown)


connyduck:
Very nice! Thanks for figuring this out.

Tak:
Thanks to @clarfonthey for prompting me to try brute-forcing the isolator characters

clarfonthey:
Glad to help! ♥️

Issue #1956, opened 2020-10-12
Add normal emojis alongside custom emojis


Porrumentzio:
Currently to add normal/general emojis when composing, it depends entirely on the keyboard used.

With some keyboard configurations it is not easy to add emojis with it, so I thought it would be great if general emojis would be added to the Custom Emojis view, adding a button to change between Custom and General.

Maybe there could be a problem with the (in)compatibility between emoji themes.

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

hex-m:
It took me forever to realize the cause of this problem. It just didn't make sense, that some emojis are shown while others are not.

I used "Simple Keyboard" from F-Droid which [does not plan to implement emoticon support](https://github.com/rkkr/simple-keyboard/issues/13). Switching to [OpenBoard](https://github.com/openboard-team/openboard) solved this for me.

Very confusing and hard for users to figure out. So my vote goes to adding a fully-fledged emoji/emoticon picker (including search) into Tusky.

mejo-:
I have the same problem with the default keyboard on GrapheneOS.

PR #1977, opened 2020-11-04
[needs help] Support announcements

### LINKED ISSUE #1693: Add support for Announcements
Since Mastodon 3.1.0 it is possible to send Announcements to all users of an instance.

At this moment I'm not able to see these announcements in the app on my Android Phone. I'm only able to see them in the web UI which makes it pretty easy to miss them.

* * * *
- Tusky Version: 9.1
- Android Version: 9
- Android Device: Fairphone 3
- Mastodon instance (if applicable): 3.1.1

kyori19:
Closes #1693 

Drawer(with one unread announcement)
Screenshot_1604504702
AnnouncementsActivity(with emoji picker opened)
Screenshot_1604504421

## Problems

- [x] No method to add normal emojis (like 🤔 and 👍). We have to add it in emoji picker but it should be really hard work... (Related to #1956). Or add text field for now?
- [x] Hard to distinguish self reaction. I tried to fix this but couldn't do... (In screenshot above, 🤔 is not selected while 🦀 is selected)
- [ ] Weird chip size in custom emojis. Maybe bug of `emojify` method...? (This custom emoji is `:javascript:`. I think the width is calculated with shortcode.)
image
- [x] Calling `load()` method after updated reaction(AnnouncementsViewModel L89, L99).  I should call `mutableAnnouncements.postValue` with updated reaction, but it requires deep copy of announcements so I gave up. Should I fix this?
- [ ] Unread announcements badge in the drawer is not appealing. Maybe it should be displayed with colored background, or a badge on drawer toggle button...?
- [x] When to mark as read announcements. When the user opened AnnouncementsActivity...? Do you have any ideas?

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/components/announcements/AnnouncementAdapter.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/components/announcements/AnnouncementAdapter.kt
@@ -37,7 +37,7 @@ interface AnnouncementActionListener {
 }
 
 class AnnouncementAdapter(
-        private var items: List<Announcement>,
+        private var items: List<Announcement> = emptyList(),
         private val listener: AnnouncementActionListener
 ) : RecyclerView.Adapter<AnnouncementAdapter.AnnouncementViewHolder>() {
 
--- a/app/src/main/java/com/keylesspalace/tusky/components/announcements/AnnouncementAdapter.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/components/announcements/AnnouncementAdapter.kt
@@ -1,0 +1,115 @@
+/* Copyright 2020 Tusky Contributors
+ *
+ * This file is a part of Tusky.
+ *
+ * This program is free software; you can redistribute it and/or modify it under the terms of the
+ * GNU General Public License as published by the Free Software Foundation; either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * Tusky is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even
+ * the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU General
+ * Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License along with Tusky; if not,
+ * see <http://www.gnu.org/licenses>. */
+
+package com.keylesspalace.tusky.components.announcements
+
+import android.view.ContextThemeWrapper
+import android.view.LayoutInflater
+import android.view.View
+import android.view.ViewGroup
+import android.widget.TextView
+import androidx.core.view.size
+import androidx.recyclerview.widget.RecyclerView
+import com.google.android.material.chip.Chip
+import com.google.android.material.chip.ChipGroup
+import com.keylesspalace.tusky.R
+import com.keylesspalace.tusky.entity.Announcement
+import com.keylesspalace.tusky.entity.Emoji
+import com.keylesspalace.tusky.util.emojify
+import kotlinx.android.synthetic.main.item_announcement.view.*
+
+interface AnnouncementActionListener {
+    fun openReactionPicker(announcementId: String, target: View)
+    fun addReaction(announcementId: String, name: String)
+    fun removeReaction(announcementId: String, name: String)
+}
+
+class AnnouncementAdapter(
+        private var items: List<Announcement>,
+        private val listener: AnnouncementActionListener
+) : RecyclerView.Adapter<AnnouncementAdapter.AnnouncementViewHolder>() {
+
+    override fun onCreateViewHolder(parent: ViewGroup, viewType: Int): AnnouncementViewHolder {
+        val view = LayoutInflater.from(parent.context)
+                .inflate(R.layout.item_announcement, parent, false)
+        return AnnouncementViewHolder(view)
+    }
+
+    override fun onBindViewHolder(viewHolder: AnnouncementViewHolder, position: Int) {
+        viewHolder.bind(items[position])
+    }
+
+    override fun getItemCount() = items.size
+
+    fun updateList(items: List<Announcement>) {
+        this.items = items
+        notifyDataSetChanged()
+    }
+
+    inner class AnnouncementViewHolder(private val view: View) : RecyclerView.ViewHolder(view) {
+
+        private val text: TextView = view.text
+        private val chips: ChipGroup = view.chipGroup
+        private val addReactionChip: Chip = view.addReactionChip
+
+        fun bind(item: Announcement) {
+            text.text = item.content
+
+            item.reactions.forEachIndexed { i, reaction ->
+                (chips.getChildAt(i)?.takeUnless { it.id == R.id.addReactionChip } as Chip?
+                        ?: Chip(ContextThemeWrapper(view.context, R.style.Widget_MaterialComponents_Chip_Choice)).apply {
+                            isCheckable = true
+                            checkedIcon = null
+                            chips.addView(this, i)
+                        })
+                        .apply {
+                            val emojiText = if (reaction.url == null) {
+                                reaction.name
+                            } else {
+                                view.context.getString(R.string.emoji_shortcode_format, reaction.name)
+                            }
+                            text = ("$emojiText ${reaction.count}")
+                                    .emojify(
+                                            listOf(Emoji(
+                                                    reaction.name,
+                                                    reaction.url ?: "",
+                                                    reaction.staticUrl ?: "",
+                                                    null
+                                            )),
+                                            this
+                                    )
+
+                            isChecked = reaction.me
+
+                            setOnClickListener {
+                                if (reaction.me) {
+                                    listener.removeReaction(item.id, reaction.name)
+                                } else {
+                                    listener.addReaction(item.id, reaction.name)
+                                }
+                            }
+                        }
+            }
+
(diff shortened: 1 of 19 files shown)


connyduck:
> No method to add normal emojis (like 🤔 and 👍). We have to add it in emoji picker but it should be really hard work... (Related to #1956). Or add text field for now?

I think is fine if we add this functionality later

> Hard to distinguish self reaction. I tried to fix this but couldn't do... (In screenshot above, 🤔 is not selected while 🦀 is selected)

Yes. Needs a custom chip style. https://material.io/develop/android/components/chips#theming-chips

> Weird chip size in custom emojis. Maybe bug of emojify method...? (This custom emoji is :javascript:. I think the width is calculated with shortcode.)

Oh, this is a hard one. Seems like the width is indeed calculated with the shortcode. No idea how to fix it, but I can investigate further when I have more time. A workaround could be using a different shortcode.

> Calling load() method after updated reaction(AnnouncementsViewModel L89, L99). I should call mutableAnnouncements.postValue with updated reaction, but it requires deep copy of announcements so I gave up. Should I fix this?

Yes, you should. Try something like that

```
 announcementsMutable.value?.data?.map { announcement ->
            if (announcement.id == announcementId) {
                announcement.copy(
                    reactions = announcement.reactions.map { reaction ->
                        if( reaction.name == name) {
                            reaction.copy(count = reaction.count++)
                        } else { 
                            reaction
                        }
                    }
                )
            } else {
                announcement
            }
```

> Unread announcements badge in the drawer is not appealing. Maybe it should be displayed with colored background, or a badge on drawer toggle button...?

Yes. I think white on Tusky blue or background color on primary text color could look cool. You can add a `BadgeStyle` to the drawer item to achieve that. See https://github.com/mikepenz/MaterialDrawer/

> When to mark as read announcements. When the user opened AnnouncementsActivity...? Do you have any ideas?

After opening the activity should be fine. Or add a button in the toolbar "mark all as read". But thats already more effort than necessary.


Another thing I would change is the position in the drawer. I think it is pretty important, it could easily be higher up, maybe even after "edit profile".

Thoughts? @mal0ki @Tak @charlag

kyori19:
> > Calling load() method after updated reaction(AnnouncementsViewModel L89, L99). I should call mutableAnnouncements.postValue with updated reaction, but it requires deep copy of announcements so I gave up. Should I fix this?
> 
> Yes, you should. Try something like that

Updated my code, but I have to get emoji's url from emojis so it became bit longer.

kyori19:
> > Hard to distinguish self reaction. I tried to fix this but couldn't do... (In screenshot above, 🤔 is not selected while 🦀 is selected)
> 
> Yes. Needs a custom chip style. https://material.io/develop/android/components/chips#theming-chips

I read [this](https://stackoverflow.com/questions/51089150/set-com-google-android-material-chip-chip-selected-color) and I tried to add custom style like this but there's no effect.
Am I something wrong?

```
--- a/app/src/main/java/com/keylesspalace/tusky/components/announcements/AnnouncementAdapter.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/components/announcements/AnnouncementAdapter.kt
@@ -69,7 +69,7 @@ class AnnouncementAdapter(
 
             item.reactions.forEachIndexed { i, reaction ->
                 (chips.getChildAt(i)?.takeUnless { it.id == R.id.addReactionChip } as Chip?
-                        ?: Chip(ContextThemeWrapper(view.context, R.style.Widget_MaterialComponents_Chip_Choice)).apply {
+                        ?: Chip(ContextThemeWrapper(view.context, R.style.ColoredChip)).apply {
                             isCheckable = true
                             checkedIcon = null
                             chips.addView(this, i)

diff --git a/app/src/main/res/values/styles.xml b/app/src/main/res/values/styles.xml
--- a/app/src/main/res/values/styles.xml
+++ b/app/src/main/res/values/styles.xml
@@ -160,4 +160,8 @@
         <item name="materialDrawerDrawCircularShadow">false</item>
     </style>
 
+    <style name="ColoredChip" parent="Widget.MaterialComponents.Chip.Choice">
+        <item name="chipBackgroundColor">@color/compound_button_color</item>
+    </style>
+
 </resources>
```

Tak:
Can the chips not just be a constant size?

I don't have a strong feeling about the position in the menu.

Should there also be a notification for new announcements?

app/src/main/java/com/keylesspalace/tusky/components/announcements/AnnouncementAdapter.kt
```diff
@@ -0,0 +1,115 @@
...
+import kotlinx.android.synthetic.main.item_announcement.view.*
+
+interface AnnouncementActionListener {
+    fun openReactionPicker(announcementId: String, target: View)
+    fun addReaction(announcementId: String, name: String)
+    fun removeReaction(announcementId: String, name: String)
+}
+
+class AnnouncementAdapter(
+        private var items: List<Announcement>,
```
connyduck, line 40: I would rather put `var items: List<Announcement> = emptyList()` into the class body

app/src/main/java/com/keylesspalace/tusky/components/announcements/AnnouncementsActivity.kt
```diff
@@ -0,0 +1,156 @@
...
+
+    override fun removeReaction(announcementId: String, name: String) {
+        viewModel.removeReaction(announcementId, name)
+    }
+
+    companion object {
+        @JvmStatic
+        fun newIntent(context: Context): Intent {
+            return Intent(context, AnnouncementsActivity::class.java)
+        }
```
connyduck, line 154:
```suggestion
        fun newIntent(context: Context) = Intent(context, AnnouncementsActivity::class.java)
```

app/src/main/java/com/keylesspalace/tusky/entity/Announcement.kt
```diff
@@ -0,0 +1,57 @@
...
+        @SerializedName("ends_at") val endsAt: Date?,
+        @SerializedName("all_day") val allDay: Boolean,
+        @SerializedName("published_at") val publishedAt: Date,
+        @SerializedName("updated_at") val updatedAt: Date,
+        val read: Boolean,
+        val mentions: Array<Status.Mention>,
+        val statuses: Array<Status>,
+        val tags: Array<HashTag>,
+        val emojis: List<Emoji>,
+        val reactions: List<Reaction>
```
connyduck, line 35: please always use `List` (not `Array`) in model classes

app/src/main/res/layout/activity_announcements.xml
```diff
@@ -0,0 +1,51 @@
...
+    xmlns:tools="http://schemas.android.com/tools"
+    android:layout_width="match_parent"
+    android:layout_height="match_parent">
+
+    <include layout="@layout/toolbar_basic" />
+
+    <androidx.constraintlayout.widget.ConstraintLayout
+        android:layout_width="match_parent"
+        android:layout_height="match_parent"
+        app:layout_behavior="@string/appbar_scrolling_view_behavior">
```
connyduck, line 13: It should be possible to do this layout without a `ConstraintLayout`. Try putting the `app:layout_behavior="@string/appbar_scrolling_view_behavior"` on the `SwipeRefreshLayout` and center the `Progressbar` and the `BackgroundMessageView` inside the `CoordinatorLayout`.

app/src/main/res/layout/item_announcement.xml
```diff
@@ -0,0 +1,40 @@
...
+    android:layout_height="wrap_content">
+
+    <androidx.emoji.widget.EmojiTextView
+        android:id="@+id/text"
+        android:layout_width="0dp"
+        android:layout_height="wrap_content"
+        android:padding="8dp"
+        android:textSize="?attr/status_text_medium"
+        app:layout_constraintEnd_toEndOf="parent"
+        app:layout_constraintStart_toStartOf="parent"
```
connyduck, line 14: Please add `android:lineSpacingMultiplier="1.1"`. Makes multiline text easier to read.

app/src/main/res/values/strings.xml
```diff
@@ -570,6 +571,7 @@
 
     <string name="no_saved_status">You don\'t have any drafts.</string>
     <string name="no_scheduled_status">You don\'t have any scheduled statuses.</string>
+    <string name="no_announcements">There\'s no announcements.</string>
```
connyduck, line 574: `There are no announcements.`

connyduck (changes requested):
>  Am I something wrong?

Yes, you need to set `<item name="chipStyle">` on the theme overlay style and link another style there, and in that other style you can use `chipBackground` and it will work.

kyori19:
> Yes, you need to set `<item name="chipStyle">` on the theme overlay style and link another style there, and in that other style you can use `chipBackground` and it will work.

Thanks. I understood how to apply styles.
For now, I changed to use default style, and I think this is enough to distinguish.
How do you think?
image

connyduck:
> How do you think?

I think its fine. If you want more contrast you could make the background of the unselected one a bit darker though.

connyduck:
@kyori19 If you resolve the merge conflict I'd merge it so nightly testers can try it out.

kyori19:
@connyduck I resolved the conflict, but now build is failing...
I didn't touch anything in dependencies... Why is this happens...?

connyduck (approved):

connyduck:
I have no idea why it happened, but for some reason it complained about not being able to resolving a dependency. Upgrading it fixed the problem.

PR #1978, opened 2020-11-04
Personal account notes

connyduck:
- notes are saved automatically once you stop typing
- note field is only shown when server supports it
- I also cleaned up some of the old `Call` things and replaced them with RxJava `Single`

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/viewmodel/AccountViewModel.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/viewmodel/AccountViewModel.kt
@@ -11,10 +11,12 @@ import com.keylesspalace.tusky.entity.IdentityProof
 import com.keylesspalace.tusky.entity.Relationship
 import com.keylesspalace.tusky.network.MastodonApi
 import com.keylesspalace.tusky.util.*
+import io.reactivex.Single
 import io.reactivex.disposables.Disposable
 import retrofit2.Call
 import retrofit2.Callback
 import retrofit2.Response
+import java.util.concurrent.TimeUnit
 import javax.inject.Inject
 
 class AccountViewModel @Inject constructor(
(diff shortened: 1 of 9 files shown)


app/src/main/java/com/keylesspalace/tusky/viewmodel/AccountViewModel.kt
```diff
@@ -243,37 +219,35 @@ class AccountViewModel @Inject constructor(
...
-            RelationShipAction.UNMUTE -> mastodonApi.unmuteAccount(accountId)
-        }
-
-        call.enqueue(callback)
-        callList.add(call)
-
+    fun noteChanged(newNote: String) {
+        noteSaved.postValue(false)
+        noteDisposable?.dispose()
+        noteDisposable = Single.timer(1500, TimeUnit.MILLISECONDS)
```
charlag, line 233: This very much looks like some `throttle`, I think RxAndroid has a way to get a stream from a field?

charlag (commented):
Code looks good, would be nice to finally have it!
I didn't try it yet, i will post once I build and try it

app/src/main/java/com/keylesspalace/tusky/viewmodel/AccountViewModel.kt
connyduck: You mean something like [TextViewTextChangesObservable](https://github.com/JakeWharton/RxBinding/blob/master/rxbinding/src/main/java/com/jakewharton/rxbinding4/widget/TextViewTextChangesObservable.kt)? It is part of RxBinding which we currently don't have in the project, but I can add it if you think its better.

charlag: Oh, I thought we either had the library or had a home-grown alternative but maybe I'm mistaking it with something

charlag:
I tried it out and it works well! I just thought maybe we should disable the field until we load things to not overwrite by accident?

app/src/main/java/com/keylesspalace/tusky/viewmodel/AccountViewModel.kt
charlag: Maybe we should at least modify our `addTextChangedListener` from `ViewExtensions` to allow unsubscribing and use that instead?

PR #1982, opened 2020-11-12
Adds link to account page when clicking on displayName

conor-f:
After finding out [here](https://social.randombits.host/web/statuses/105187994632656075) that you can get to a user profile by clicking their avatar and not their display name I felt dumb so I made this to have what I feel is the expected behavior of also going to a profile when you click the display name.


connyduck (changes requested):
Please reuse the click listener from the avatar (one object creation less)

conor-f:
Updated to create and reuse the `onClickListener`

connyduck (approved):

Issue #1985, opened 2020-11-15
App crashes when changing accounts


isteinbach:
How to repeat:
- (at least) two accounts configured in Tusky
- switch from one account to the other
- the app crashes in many cases (not always)
- after re-opening it, the other account can be operated on normally
- [logcat_11-15-2020_12-31-40.txt](https://github.com/tuskyapp/Tusky/files/5542396/logcat_11-15-2020_12-31-40.txt) might give some more insight

* * * *
- Tusky Version: 12.1 (from F-Droid repository)
- Android Version: 7.1.2 (FairPhone build 19.11.2 with GMS)
- Android Device: FairPhone2
- Mastodon instance (if applicable): various; in my case mastodon.online and social.tchncs.de

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

connyduck:
Here is a deobfuscated stack trace

```
java.lang.IllegalArgumentException: 
  at com.bumptech.glide.manager.RequestManagerRetriever.assertNotDestroyed (RequestManagerRetriever.java:317)
  at com.bumptech.glide.manager.RequestManagerRetriever.get (RequestManagerRetriever.java:128)
  at com.bumptech.glide.manager.RequestManagerRetriever.get (RequestManagerRetriever.java:108)
  at com.bumptech.glide.Glide.with (Glide.java:776)
  at com.keylesspalace.tusky.MainActivity$setupDrawer$2.cancel (MainActivity.java:329)
  at com.mikepenz.materialdrawer.util.DrawerImageLoader.cancelImage (DrawerImageLoader.java:56)
  at com.mikepenz.materialdrawer.model.ProfileDrawerItem.unbindView (ProfileDrawerItem.java:137)
  at com.mikepenz.materialdrawer.model.ProfileDrawerItem.unbindView (ProfileDrawerItem.java:21)
  at com.mikepenz.fastadapter.listeners.OnBindViewHolderListenerImpl.unBindViewHolder (OnBindViewHolderListenerImpl.java:36)
  at com.mikepenz.fastadapter.FastAdapter.onViewRecycled (FastAdapter.java:410)
  at androidx.recyclerview.widget.RecyclerView$Recycler.dispatchViewRecycled (RecyclerView.java:6769)
  at androidx.recyclerview.widget.RecyclerView$Recycler.addViewHolderToRecycledViewPool (RecyclerView.java:6538)
[...]

PR #1989, opened 2020-11-20
fix crash when clearing glide requests

connyduck:
closes https://github.com/tuskyapp/Tusky/issues/1985

We cannot control when the `cancel` callback is called, and not cancelling would also be a bad idea, so I followed [this advice](https://github.com/bumptech/glide/issues/2690#issuecomment-349837873) in order to fix it.


charlag (approved):
There's standard lifecycle now yet we still have to manage this things :roll_eyes:

PR #1992, opened 2020-11-21
Wellbeing mode

### LINKED ISSUE #1980: [Feature Request] Digital wellbeing mode
In todays world, we're surrounded by social media noise that hurts our mental wellbeing. I'd like to have a mode that cuts down on this to focus purely on the aspect of interacting with others. This might include:

* Hide stars/boosts count on toots
* Disable "Someone stared/boosted your toot" notifications
* Disable follower/following count on profiles
* [Your Suggestion Here]

garritfra:
Closes #1980

Implements a settings switch to disable some features regarding the digital wellbeing of the user.

These features are disabled:
- Favorite/Boost/Follow notifications
- Favorite/Boost count on toots
- Follower/Post stats on profiles
- "Follows you" badge on profiles

Once enabled, the user will be prompted to review his push-notifications in the android settings.


garritfra:
This last commit (bb6f93d) should work as expected, but I'm not really happy with it, since it has a lot of code duplication. Let me know if there is a prettier way that I might have missed.

connyduck:
I'd like to avoid feature creep in Tusky as much as possible, and I am not sure how many people would really use this.
If there is demand for this, wouldn't it be better to make individual toggles than bundle it into one?
https://chaos.social/@ConnyDuck/105255335776206826

garritfra:
I had the same idea, which lead me to this PR :)

https://fosstodon.org/@garritfra/105196617347971371

garritfra:
I do agree with you. A "wellbeing" section where these things can be toggled individually would be a better approach.

connyduck:
Ok so, it seems there are no strong opinions about this feature but some users would like it so I think we should merge this.

- I'd prefer to not have translations in PRs because they make Weblate act up
- Can we maybe give the mode another name? For me it feels kinda patronizing to tell users "this is for your wellbeing".
Suggestion: Call it "limited mode". Text can be something like "Some non-essential information will be hidden from the interface. . This includes:"
- Maybe include the recently merged announcement badge in the mode as well?

Tak:
>     * Can we maybe give the mode another name? For me it feels kinda patronizing to tell users "this is for your wellbeing".

One argument for "wellbeing" is that it's consistent with android's builtin "Digital Wellbeing" thing :shrug:

garritfra:
> I'd prefer to not have translations in PRs because they make Weblate act up

What's the usual way to handle new translations from PRs?

> For me it feels kinda patronizing to tell users "this is for your wellbeing".

> One argument for "wellbeing" is that it's consistent with android's builtin "Digital Wellbeing" thing 🤷

I was also thinking about the android wellbeing thingy when developing this. I don't know if "limited mode" would quite hit the nail. For now I guess this doesn't matter though. Should we ask the fediverse about this?

> Maybe include the recently merged announcement badge in the mode as well?

I wasn't aware of this, but sounds good👍

I won't have time this weekend, but I'll push commits early next week.

garritfra:
Okay, I think everything is in place now. 

image

One thing that should be taken care of are the translations. @connyduck could you add them to Weblate, so that I can remove them from the strings.xml?

> Maybe include the recently merged announcement badge in the mode as well?

I added this to the "Hide quantitative stats on posts" toggle, since announcements are posts by trait. I don't think it's necessary to add another toggle just for this. Let me know if you disagree.

app/src/main/java/com/keylesspalace/tusky/components/preference/PreferencesFragment.kt
```diff
@@ -196,6 +203,45 @@ class PreferencesFragment : PreferenceFragmentCompat(), Injectable {
...
+                                notificationFilter.add(Notification.Type.FOLLOW)
+                                notificationFilter.add(Notification.Type.REBLOG)
+                            } else {
+                                notificationFilter.remove(Notification.Type.FAVOURITE)
+                                notificationFilter.remove(Notification.Type.FOLLOW)
+                                notificationFilter.remove(Notification.Type.REBLOG)
+                            }
+
+                            account.notificationsFilter = serialize(notificationFilter)
+                            accountManager.saveAccount(account)
```
connyduck, line 226: This is weird: general settings should affect ALL accounts. This only changes the current account. I'd argue this is not the right place to do it, rather leave the accounts alone and change the filters when they are retrieved based on the wellbeing setting.

app/src/main/java/com/keylesspalace/tusky/components/announcements/AnnouncementAdapter.kt
```diff
@@ -67,6 +71,17 @@ class AnnouncementAdapter(
         fun bind(item: Announcement) {
             text.text = item.content
 
+            val preferences: SharedPreferences = PreferenceManager.getDefaultSharedPreferences(itemView.context)
+            val wellbeingEnabled = preferences.getBoolean(PrefKeys.WELLBEING_HIDE_STATS_POSTS, false)
```
connyduck, line 75: I'm afraid, the `bind` method is not the right place to load the settings. It might be called quite often, loading preferences can be expensive (disk read!) and the value of the setting will not change. Better load it outside of the adapter and pass it through the constructor or something like that.

app/src/main/java/com/keylesspalace/tusky/adapter/StatusDetailedViewHolder.java
```diff
@@ -52,13 +55,18 @@ protected void setCreatedAt(Date createdAt, StatusDisplayOptions statusDisplayOp
 
     private void setReblogAndFavCount(int reblogCount, int favCount, StatusActionListener listener) {
 
-        if (reblogCount > 0) {
+        SharedPreferences preferences = PreferenceManager.getDefaultSharedPreferences(itemView.getContext());
+
+        // If wellbeing mode is enabled, favs and boosts should not be visible.
+        boolean wellbeingEnabled = preferences.getBoolean(PrefKeys.WELLBEING_HIDE_STATS_POSTS, false);
```
connyduck, line 61: same here, better load the setting somewhere else and pass the value to the Adapter (maybe put it in the `StatusDisplayOptions`)

connyduck (changes requested):
> One thing that should be taken care of are the translations. @connyduck could you add them to Weblate, so that I can remove them from the strings.xml?

Sure. delete them and I'll add them once this is merged

> I added this to the "Hide quantitative stats on posts" toggle, since announcements are posts by trait. I don't think it's necessary to add another toggle just for this. Let me know if you disagree.

👍

app/src/main/java/com/keylesspalace/tusky/components/preference/PreferencesFragment.kt
garritfra: I didn't wanted to straight up ignoring the user-set filter, but I definitely see your concern. What if we update the settings for all accounts in the app, as opposed to just the active one?

app/src/main/java/com/keylesspalace/tusky/components/announcements/AnnouncementAdapter.kt
garritfra: Right. I totally overlooked this. Thanks for the hint.

connyduck:
Is this ready for review again?

garritfra:
No, not yet. I was waiting for your opinion on [this comment](https://github.com/tuskyapp/Tusky/pull/1992#discussion_r540350218), and it faded out of my mind. Once that is resolved, I will remove the string resources, and it will be ready for review.

app/src/main/java/com/keylesspalace/tusky/components/preference/PreferencesFragment.kt
connyduck: that is also a possibility 👍

connyduck:
Oh, sorry, I must have missed that one

garritfra:
Should be ready for review. Build is failing due to missing translations.

connyduck:
You deleted the wrong translation, English is the default and should stay

garritfra:
Oh, does Weblate not add the translations by itself?

garritfra:
Okay, german translations are removed. The english translations can stay?

connyduck (approved):

resynth1943:
> Hide quantitative stats on posts

I honestly don't like how this is worded. It seems needlessly complex, and I'm guessing a lot of users won't actually know what it means.

resynth1943:
One more thing: is there a PR for disabling infinite scroll? This could really improve wellbeing, abnd has already been implemented by [Pinafore](https://pinafore.social).

garritfra:
I was planning to add this in the near future. I don't like adding more
features mid-PR. If you want to take this over though, feel free to do so.

Resynth <notifications@github.com> schrieb am Mi., 23. Dez. 2020, 10:19 PM:

garritfra:
I agree with you, I don't really like the wording either. Do you have other
suggestions? I can't really think of a simpler term right now.

Resynth <notifications@github.com> schrieb am Mi., 23. Dez. 2020, 10:05 PM:

PR #1993, opened 2020-11-24
load drawer avatar earlier to avoid flicker on app start

connyduck:


charlag (approved):
So we load it twice but it will actually only be once, if it didn't change, right?

connyduck:
Exactly. The second time does not cause any networking. 

I added the placeholder to the first load to prevent the title from jumping around.

PR #2007, opened 2020-12-03
WIP: Support completion for standard emoji

Tak:
I took a stab at adding standard unicode emoji to the emoji completion results.
I couldn't find emoji/shortcodes exposed anywhere in the public android api, so I grabbed a json file from the [gemoji](https://github.com/github/gemoji/blob/master/db/emoji.json) project (generated from the [official unicode document](http://unicode.org/Public/emoji/13.0/emoji-test.txt))

There are some challenges here:
- We need to remember to update these as new versions get released
- All the shortcodes are in english, so users will get a different experience than with gboard (e.g. :pain: != :bread:)
-- Maybe set up a translatable resource for the shortcodes?

There's also room for improvement:
- The emoji db could be a more "native" format than a big json blob
- We could reformat the db for our data access needs
- We could strip superfluous information (e.g. ios version) from the db
-- These all mean that we'd need our own postprocessor that we would ideally document and version somewhere…

image

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/adapter/ComposeAutoCompleteAdapter.java
+++ b/app/src/main/java/com/keylesspalace/tusky/adapter/ComposeAutoCompleteAdapter.java
@@ -40,6 +40,7 @@ import java.util.List;
 
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
+import androidx.emoji.widget.EmojiTextView;
 
 /**
  * Created by charlag on 12/11/17.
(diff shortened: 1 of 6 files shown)


a1batross:
>The emoji db could be a more "native" format than a big json blob

When I did emoji reactions, I just generated *.java file with all supported by backend emojis. I think, this JSON blob also could be converted to just Java class with unneeded data stripped.

Tak:
Applies to #1956 , I guess

connyduck:
My thoughts on this:
- We need to automate the generaton of the emoji data somehow, manual updating would be a lot of work
- A Java/Kotlin file is probably best for storing the emoji information
- How to handle the emoji support in different Android versions? Not all versions support all emojis

> Maybe set up a translatable resource for the shortcodes?

I think leaving them in English is acceptable for a first version, but they should be translated eventually. Translating like other strings could work, yes.


While the feature would be nice to have, I don't think it is worth the effort.

Fyi, I occasionally get asked by users how to input standard emoji in Tusky, and all had no problem to use their keyboard after I told them how to.

Tak:
Ok, that's fair

PR #2012, opened 2020-12-05
Notification bell

a1batross:
This is the adaption of code I wrote to support similar feature in Pleroma that was introduced long before Mastodon, so that's the reason why it's called subscriptions internally. 

This PR adds notification bell for both Mastodon 3.3.0rc1 and Pleroma. However, filtering in notification timeline does not work for Pleroma due to obvious absence of "status" notification type.

How it looks in Mastodon:

I'm not sure though if images should not be shown for subscription posts. To me it eliminates some use cases. However, Mastodon frontend hides them. :(


app/src/main/java/com/keylesspalace/tusky/entity/Notification.kt
```diff
@@ -72,4 +73,20 @@ data class Notification(
         }
 
     }
+    
+    companion object {
+
+        // for Pleroma compatibility that uses Mention type
+        @JvmStatic
+        fun rewriteToStatusTypeIfNeeded(body: Notification, accountId: String) : Notification {
```
connyduck, line 81: This could be just a regular method of Notification, or an extension function

app/src/main/res/values/strings.xml
```diff
@@ -287,7 +289,8 @@
     <string name="notification_favourite_description">Notifications when your toots get marked as favorite</string>
     <string name="notification_poll_name">Polls</string>
     <string name="notification_poll_description">Notifications about polls that have ended</string>
-
+    <string name="notification_subscription_name">Subscriptions</string>
```
connyduck, line 292: Subscriptions is the better wording, but maybe we should match Mastodon web for consistency? ("New toots")

connyduck (changes requested):
Super awesome, thank you. I love the Pleroma support.

- You also need to extend the `getNotificationText` method in `NotificationsFragment.java:762`

app/src/main/res/values/strings.xml
a1batross: I replaced it with "New toots" and replaced posts by toots in description.

connyduck (approved):

PR #2014, opened 2020-12-07
Workaround for old Androids not connecting to new Let's Encrypt hosts

connyduck:
Let's Encrypt is changing their root certificate, which is not trusted by devices prior to Android 7.1.1. Here is more info: https://letsencrypt.org/2020/11/06/own-two-feet.html
Most of the fediverse is using Let's encrypt and we have still about 10% of users on these Android versions so I think this change is worth it. Using a custom trust manager is a performance hit because it possibly loads a lot of certificates into memory, but better a slow app than one that does not work at all.
Let's encrypt will start signing certificates with the new root on 11.1.2021. 

Code is from https://stackoverflow.com/questions/64844311/certpathvalidatorexception-connecting-to-a-lets-encrypt-host-on-android-m-or-ea

As this may have a security impact, it would be cool if more people could review it @charlag @Tak @a1batross


app/src/main/java/com/keylesspalace/tusky/util/OkHttpUtils.kt
```diff
@@ -0,0 +1,114 @@
...
+        val address = InetSocketAddress.createUnresolved(httpServer, httpPort)
+        builder.proxy(Proxy(Proxy.Type.HTTP, address))
+    }
+
+    // trust the new Let's Encrypt root certificate that is not available on Android < 7.1.1
+    // see https://letsencrypt.org/2020/11/06/own-two-feet.html
+    // see https://stackoverflow.com/questions/64844311/certpathvalidatorexception-connecting-to-a-lets-encrypt-host-on-android-m-or-ea
+    if (Build.VERSION.SDK_INT < Build.VERSION_CODES.O) {
+        val isgCert = """
+        -----BEGIN CERTIFICATE-----
```
a1batross, line 70: Certificate seems the same as https://letsencrypt.org/certs/isrgrootx1.pem.

Maybe also add this link for those who want to check them manually?

charlag (approved):
I've heard about the change.
I don't mind having it for old versions, we do different TLS hacks for old OS versions anyway.
Just want to mention that it's still possible to issue certificate with old trust root for one year after they make a change but you have to opt in and I imagine many instances doing this but who knows.

PR #2015, opened 2020-12-07
remove ThemeUtils.getTintedDrawable

connyduck:
Comment says `drawableTint in xml once minSdkVersion >= 23`, but today I learned there is actually a compat version of it.


charlag (approved):
Oh, great, it was a major pain

PR #2021, opened 2020-12-11
cleanup accountManager.activeAccount usages in MainActivity

connyduck:
This gets rid of the "name shadowed" Kotlin warning.
Also something I messed up while adding the avatar to the toolbar 🙄


charlag (approved):
I think it's because this dialog is pseudo-async? It's fine in any case

PR #2022, opened 2020-12-11
fix emoji preference thumbnails

connyduck:
Bug introduced with #2015 
I really need to be more careful not to introduce regressions 😖


charlag (approved):
Don't worry, this is not a big issue

PR #2030, opened 2020-12-30
Move compose content's EditText cursor to first when shared content parsed in ComposeActivity

lenchan139:
If external apps parsed share content to Tusky, it will parse the URL to composing a new toot. Currently, the cursor will move to last by default but I saw most users and I preferred to write the comments in first and leave share content in last. 

So, I suggest after share content parsed, move the cursor to first and tapped new line for user to comments the contents. 

Also, If the user didn't type any. toot submit the new line will be trimmed by default.

After parsing:
Screenshot_20201231_004824


connyduck:
Sorry for not looking at this earlier. I'll merge it if you resolve the conflicts.

PR #2032, opened 2021-01-05
Drafts v2

connyduck:
- complete clean rewrite of the draft feature
- the old draft feature is kept intact, so users can still use their old drafts (will remove it in a later release)
- an info dialog will tell users where to find the old drafts (only if they have some)
- gets rid of ancient Java code with AsyncTasks
- all saved information (content warning, attachments, polls) is shown on the draft preview
- only one network request per media file when restoring a draft (instead of one for the file and one for the description)
- Fixes a bug where media descriptions can get lost when restoring a draft
- Fixes a bug where restored media would get uploaded multiple times when the screen is rotated on ComposeActivity
- since I did not want to write code with the deprecated kotlin syntethic views, this introduces Viewbinding. I intend to convert everything to Viewbinding
- once the old draft feature is removed, the code should be a lot cleaner and faster


app/src/main/java/com/keylesspalace/tusky/components/drafts/DraftsActivity.kt
```diff
@@ -0,0 +1,195 @@
...
+                    }, { throwable ->
+
+                        bottomSheet.state = BottomSheetBehavior.STATE_HIDDEN
+
+                        Log.w(TAG, "failed loading reply information", throwable)
+
+                        if (throwable is HttpException && throwable.code() == 404) {
+                            // the original status to which a reply was drafted has been deleted
+                            // let's open the ComposeActivity without reply information
+                            openDraftWithoutReply(draft)
```
charlag, line 156: It can probably take people by surprise, maybe we should let them know somehow?

app/src/main/java/com/keylesspalace/tusky/components/drafts/DraftMediaAdapter.kt
```diff
@@ -0,0 +1,81 @@
...
+import com.bumptech.glide.load.engine.DiskCacheStrategy
+import com.keylesspalace.tusky.R
+import com.keylesspalace.tusky.db.DraftAttachment
+
+class DraftMediaAdapter(
+        private val attachmentClick: () -> Unit
+) : ListAdapter<DraftAttachment, DraftMediaAdapter.DraftMediaViewHolder>(
+        object: DiffUtil.ItemCallback<DraftAttachment>() {
+            override fun areItemsTheSame(oldItem: DraftAttachment, newItem: DraftAttachment): Boolean {
+                return oldItem == newItem
```
charlag, line 35: Should this compare ids instead?

charlag (approved):
I am yet to test it somehow but looking at code it's really nice!
It was kinda hard to hunt down all the relevant commits so I think it would be great to squash it into 2-3 commits and maybe also rebase it?
I'm fine with view binding, I think it's the most painless options for us right now.

charlag:
Here are some other things:
 - in the list of drafts I think there's not enough space below images
 - Not all parts in the draft list can be pressed to open the draft, empty space next to the image doesn't do anything. Also I don't remember it having ripple.
 - When I add two images, don't finish upload them, save draft and reopen it we upload two images in parallel. I don't remember if we do this normally or we wait

app/src/main/java/com/keylesspalace/tusky/components/drafts/DraftMediaAdapter.kt
connyduck: they don't have ids, I think it is fine like this

app/src/main/java/com/keylesspalace/tusky/components/drafts/DraftsActivity.kt
connyduck: Very good point

app/src/main/java/com/keylesspalace/tusky/components/drafts/DraftMediaAdapter.kt
charlag: Oops, I confused it with something else then

charlag:
What about other comments I left?

connyduck:
Oh. Thx 😅 

> When I add two images, don't finish upload them, save draft and reopen it we upload two images in parallel. I don't remember if we do this normally or we wait

I changed nothing there, it was always this way. I think it is fine.

charlag (approved):
Looks good!

PR #2036, opened 2021-01-10
Fix TalkBack not initializing old views

charlag:
This is an attempt to fix the problem we have reported.
Basically it goes like this:
1. TalkBack is disabled
2. You open the app
3. You enable TalkBack
4. You try to use accessibility actions (swipe up-and-right)
5. It does not work on rows which were already displayed, only on those which are added later with scrolling

It seems like it should be fixed in libraries but I would try to bandaid it on our own. I think if we make it re-initialize all views it should work but I don't know how do it. `invalidate()` doesn't cut it.

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/fragment/TimelineFragment.java
+++ b/app/src/main/java/com/keylesspalace/tusky/fragment/TimelineFragment.java
@@ -24,11 +24,13 @@ import android.util.Log;
 import android.view.LayoutInflater;
 import android.view.View;
 import android.view.ViewGroup;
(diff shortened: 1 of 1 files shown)


connyduck:
seems fine?

charlag:
It loses all the state like scroll position, I think we can do better but I can also settle on this

connyduck:
Did you try forcing the Recyclerview to rebind all views? With `notifyDataSetChanged`

charlag:
@connyduck thanks, I've tried your suggestion and it works like a charm. Don't know why I didn't think of it myself

connyduck (approved):

PR #2041, opened 2021-01-14
Show display names for reposts

garritfra:
On the timeline, the app currently displays the username (E.g. `foo@mastodon.social`) of someone who reposted a toot.

This PR aims to show the display name (E.g. `Foo Bar 🕹`) of that user.

[This matrix conversation](https://matrix.to/#/!PWkUGGpeMZWDPmdfRE:matrix.org/$16103634431sblev:matrix.slashdev.space?via=matrix.org&via=kde.org&via=converser.eu) gives some more context.


app/src/main/java/com/keylesspalace/tusky/adapter/StatusViewHolder.java
```diff
@@ -73,10 +75,11 @@ protected void setupWithStatus(StatusViewData.Concrete status,
 
     }
 
-    private void setRebloggedByDisplayName(final String name) {
+    private void setRebloggedByDisplayName(final CharSequence name, final StatusViewData.Concrete status) {
         Context context = statusInfo.getContext();
-        String boostedText = context.getString(R.string.status_boosted_format, name);
-        statusInfo.setText(boostedText);
+        CharSequence boostedText = context.getString(R.string.status_boosted_format, name);
```
Tak, line 80: minor: I would maybe push the unicode wrap down to here

Tak (commented):
`StatusBaseViewHolder` also needs to be updated, doesn't it?

app/src/main/java/com/keylesspalace/tusky/adapter/StatusViewHolder.java
garritfra: Good point. Fixed in 221537b.

garritfra:
> StatusBaseViewHolder also needs to be updated, doesn't it?

@Tak Do you mean here?  https://github.com/tuskyapp/Tusky/blob/9cdf486ceb36852e0c07af4678eeab934fc39a98/app/src/main/java/com/keylesspalace/tusky/adapter/StatusBaseViewHolder.java#L804

As far as I can tell, the only class needing this logic is the `StatusViewHolder`. All other subclasses don't rely on this information.

image

connyduck:
We only show the reblogging information in the timeline, so this is fine. Showing it on the details screen as well is something for a future improvement.
thx @garritfra

PR #2055, opened 2021-01-25
get rid of BaseFragment by using RxJava instead of Retrofit Calls

connyduck:


charlag (approved):
Glorious day has come

PR #2059, opened 2021-01-28
move onOptionsItemSelected to BaseActivity

connyduck:


app/src/main/java/com/keylesspalace/tusky/components/compose/ComposeActivity.kt
```diff
@@ -887,15 +886,6 @@ class ComposeActivity : BaseActivity(),
 
     }
 
-    override fun onOptionsItemSelected(item: MenuItem): Boolean {
-        if (item.itemId == android.R.id.home) {
```
charlag, line 891: This is gone now?

connyduck: ohhh good catch, thx

charlag (approved):

PR #2062, opened 2021-01-31
Don't stop autocompletion after second @

a1batross:
The algorithm in tokenizer isn't changed that much, we just running it again if @ was encountered and if there is second @, the whole sequence is taken.


app/src/main/java/com/keylesspalace/tusky/util/ComposeTokenizer.kt
```diff
@@ -21,20 +21,54 @@ import android.text.TextUtils
...
+            }
+
+            // found mention symbol, override cursor
+            if(character2 == '@') {
+                i = j
+                character = character2
+            }
+        }
+
+        // Log.d("Tokenizer", "Stopped search at ${character} ${text.substring(i)}")
```
connyduck, line 70: please delete

connyduck (approved):

Issue #1959, opened 2020-10-16
Too much custom emojis may cause ANR


a1batross:
When post has too much custom emojis, it may cause application to freeze in `setText(CharSequence)` method of any TextView. The limit highly depends on Android version and hardware, on my device with Android 10 it's ~200 emojis, some users reported that even 100 emojis may cause a visible lag.

The problem happens for me both in latest Tusky release and my fork.

I found the root of this problem. It's the `LineBreaker` class. Although it was introduced pretty lately, the line breaking was always there and in API 22 it's possible to tune breaking 
By playing with TextView, I found out that:
1. Setting hyphenation frequency from High Quality to Normal or disabling it doesn't give any effect.
2. Setting break strategy from High Quality to Simple fixes the ANR.

The problem is that simple break strategy makes less enjoyable to read and disables auto-hyphentation:
<img src="https://user-images.githubusercontent.com/3213697/96272841-51d17980-0fd7-11eb-8fd3-5fc39d1ad1a8.png" width="40%"/> <img src="https://user-images.githubusercontent.com/3213697/96272854-54cc6a00-0fd7-11eb-93c6-6db5daa9d302.png" width="40%"/>

My solution would be is to subclass EmojiTextView and override setText. If there is too much spans, higher than 100, then break strategy is set to simple or high-quality otherwise.

connyduck:
Interesting find 😮 

Can you send a PR?

Issue #472, opened 2017-11-30
Support custom emoji gifs


connyduck:
For example https://cybre.space/@maunzikation/99094971933547861

charlag:
It's really not easy to do and it will be quite laggy but it's desirable of course
https://stackoverflow.com/questions/8632801/how-to-add-animated-emoticon-in-textview-or-edittext-in-android

M4lik:
Yeah, I know that this is low prio and maybe a little bit difficult, but just wanted to ask if there is any progress here? :sweat_smile:

charlag:
Nope, not yet, sorry.

aka-rin:
I don't know if I get it wrong, but custom emoji only supports png files that less than 50KB. So they are not gifs, but APNG. It seems that Line, the popular chatting application in Japan and Korea, released an APNG library for Android: https://github.com/line/apng-drawable .

Well, some users are asking if any client of Mastodon support animated emojis so I want to create a new issue but it seems duplicate ... If Tusky implement this feature it will be the first one though. 😂

hynet-mel:
It seems Fedilab already implemented custom emojis, maybe I will look if I can make a patch for both Tusky and Husky (and my own fork of Husky)
https://github.com/stom79/Fedilab/commit/0dff65192a3ace68301c02cde331dc2215db89f6

PR #2064, opened 2021-01-31
Animated emoji support

a1batross:
Only animates GIF, as it's the only animated format supported by Android's internal drawable decoder.

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/util/CustomEmojiHelper.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/util/CustomEmojiHelper.kt
@@ -16,12 +16,11 @@
 @file:JvmName("CustomEmojiHelper")
 package com.keylesspalace.tusky.util
 
-import android.graphics.Bitmap
 import android.graphics.Canvas
 import android.graphics.Paint
-import android.graphics.drawable.BitmapDrawable
-import android.graphics.drawable.Drawable
-import android.text.SpannableStringBuilder
+import android.graphics.drawable.*
+import android.text.SpannableString
+import android.text.Spanned
 import android.text.style.ReplacementSpan
 import android.view.View
 
@@ -33,6 +32,8 @@ import com.keylesspalace.tusky.entity.Emoji
 
 import java.lang.ref.WeakReference
 import java.util.regex.Pattern
+import androidx.preference.PreferenceManager
+import com.keylesspalace.tusky.settings.PrefKeys
 
 /**
  * replaces emoji shortcodes in a text with EmojiSpans
@@ -45,7 +46,9 @@ fun CharSequence.emojify(emojis: List<Emoji>?, view: View) : CharSequence {
     if(emojis.isNullOrEmpty())
         return this
 
-    val builder = SpannableStringBuilder.valueOf(this)
+    val builder = SpannableString.valueOf(this)
+    val pm = PreferenceManager.getDefaultSharedPreferences(view.context)
+    val animate = pm.getBoolean(PrefKeys.ANIMATE_CUSTOM_EMOJIS, false)
 
     emojis.forEach { (shortcode, url) ->
         val matcher = Pattern.compile(":$shortcode:", Pattern.LITERAL)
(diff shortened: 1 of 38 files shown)


a1batross:
Fixes #472

Running this always enabled since October, it doesn't seem to significantly decrease performance.

app/src/main/java/com/keylesspalace/tusky/util/CustomEmojiHelper.kt
```diff
@@ -45,7 +46,9 @@ fun CharSequence.emojify(emojis: List<Emoji>?, view: View) : CharSequence {
     if(emojis.isNullOrEmpty())
         return this
 
-    val builder = SpannableStringBuilder.valueOf(this)
+    val builder = SpannableString.valueOf(this)
```
connyduck, line 49: Why this change?

app/src/main/java/com/keylesspalace/tusky/util/CustomEmojiHelper.kt
```diff
@@ -45,7 +46,9 @@ fun CharSequence.emojify(emojis: List<Emoji>?, view: View) : CharSequence {
...
+    val pm = PreferenceManager.getDefaultSharedPreferences(view.context)
+    val animate = pm.getBoolean(PrefKeys.ANIMATE_CUSTOM_EMOJIS, false)
```
connyduck, line 51: The setting should be queried once per Activity and passed into the function as a Boolean param, as we try to do with other settings. `StatusDisplayOptions` is used in most places for this.

app/src/main/java/com/keylesspalace/tusky/util/CustomEmojiHelper.kt
```diff
@@ -16,12 +16,11 @@
...
 
-import android.graphics.Bitmap
 import android.graphics.Canvas
 import android.graphics.Paint
-import android.graphics.drawable.BitmapDrawable
-import android.graphics.drawable.Drawable
-import android.text.SpannableStringBuilder
+import android.graphics.drawable.*
+import android.text.SpannableString
+import android.text.Spanned
```
connyduck, line 23: unused import

connyduck (changes requested):
😍

app/src/main/java/com/keylesspalace/tusky/util/CustomEmojiHelper.kt
a1batross: Oh, that seems to be a part of older research when I tried to make this custom emojis to not ANR. 

There is no need in SSB, when we're not modifying the text, so we can use just SpannableString.

Despite this change doesn't seem to fix #1959 and the core issue was text line breaking, I still think using SS rather than SSB is pretty harmless. The only place where SSB must be used is LinkHelper [here](https://github.com/tuskyapp/Tusky/blob/develop/app/src/main/java/com/keylesspalace/tusky/util/LinkHelper.java#L127).

app/src/main/java/com/keylesspalace/tusky/util/CustomEmojiHelper.kt
a1batross: Correct. Gonna change it then.

a1batross:
Now it's queried only once (in most places), somewhere it's refactored to also do the same with animateAvatars option.

app/src/main/java/com/keylesspalace/tusky/util/CustomEmojiHelper.kt
connyduck: The reason why SSB was used here is because when the emojify method is used together with the one in LinkHelper, `valueOf` saves one instance creation because an existing one is reused. Not sure if it has a real impact though.

a1batross: That makes sense. Gonna revert it then.

Still I think using SSB is quite overhead and even LinkHelper should not use it but that's for another discussion.

connyduck (approved):

PR #2065, opened 2021-01-31
Notifications streaming

a1batross:
Since pull notifications seems to get unreliable more and more, I decided to implement live notifications.

Currently, it's naive implementation that just starts service which just connects to instance's WebSocket. It works. After long absense of network connection, it may die but will reconnect as soon as user will open Tusky again.

Battery drain happens but it's only 1-2% over night on my phone. 🤷🏼


connyduck:
Very cool 🧐 
I will put this on my phone for a while to test it.
Can you make a dedicated database migration in the meantime, please?

a1batross:
> Can you make a dedicated database migration in the meantime, please?

Done!

a1batross:
@Tak weird. Is sticky notification is shown up? 

I have a testing accounts on mstdn.social and mastodon.ml, both on vanilla 3.3.0. It works there.

Tak:
Yeah, the sticky notification is there.

app/schemas/com.keylesspalace.tusky.db.AppDatabase/25.json
```diff
@@ -2,7 +2,7 @@
   "formatVersion": 1,
   "database": {
     "version": 25,
-    "identityHash": "e2cb844862443c2c5cc884c11f120d43",
+    "identityHash": "81c19fca95f32273ff674ec67cb17e0e",
```
connyduck, line 5: please revert the changes to this file and check in the 26.json instead

app/src/main/java/com/keylesspalace/tusky/service/StreamingService.kt
```diff
@@ -0,0 +1,216 @@
...
+    companion object {
+        val CHANNEL_ID = "streaming"
+        val KEY_STOP_STREAMING = "stop_streaming"
+        val TAG = "StreamingService"
+
+        @JvmStatic
+        var serviceRunning = false
+
+        @JvmStatic
+        private fun startForegroundService(ctx: Context, intent: Intent) {
```
connyduck, line 152: use `ContextCompat.startForegroundService` instead

app/src/main/java/com/keylesspalace/tusky/service/StreamingService.kt
```diff
@@ -0,0 +1,216 @@
...
+            serviceRunning = true
+        }
+
+        return START_NOT_STICKY
+    }
+
+    companion object {
+        val CHANNEL_ID = "streaming"
+        val KEY_STOP_STREAMING = "stop_streaming"
+        val TAG = "StreamingService"
```
connyduck, line 146: these can be `private const val`

app/src/main/java/com/keylesspalace/tusky/service/StreamingService.kt
```diff
@@ -0,0 +1,216 @@
...
+    override fun onCreate() {
+        AndroidInjection.inject(this)
+        super.onCreate()
+    }
+
+    private fun stopStreamingForId(id: Long) {
+        if(id in sockets) {
+            sockets[id]!!.close(1000, null)
+            sockets.remove(id)
+        }
```
connyduck, line 61:
```suggestion
        sockets[id]?.close(1000, null)
        sockets.remove(id)
```

app/src/main/java/com/keylesspalace/tusky/service/StreamingService.kt
```diff
@@ -0,0 +1,216 @@
...
+import dagger.android.AndroidInjection
+import okhttp3.*
+import javax.inject.Inject
+
+class StreamingService: Service(), Injectable {
+    @Inject
+    lateinit var api: MastodonApi
+
+    @Inject
+    lateinit var eventHub: EventHub
```
connyduck, line 36: Not used?

app/src/main/java/com/keylesspalace/tusky/service/StreamingService.kt
```diff
@@ -0,0 +1,216 @@
...
+        for(sock in sockets) {
+            sock.value.close(1000, null)
+        }
+        sockets.clear()
+
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
+            ServiceCompat.stopForeground(this, ServiceCompat.STOP_FOREGROUND_DETACH)
+        }
+
+        notificationManager.cancel(1337)
```
connyduck, line 74: make that 1337 a constant please

app/src/main/java/com/keylesspalace/tusky/service/StreamingService.kt
```diff
@@ -0,0 +1,216 @@
...
+        @JvmStatic
+        private fun startForegroundService(ctx: Context, intent: Intent) {
+            if(Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
+                ctx.startForegroundService(intent)
+            } else {
+                ctx.startService(intent)
+            }
+        }
+
+        @JvmStatic
```
connyduck, line 160: `@JvmStatic` is not needed if the method is only called from Kotlin

connyduck (changes requested):
It works nicely for me, most of the time. Amazing to see real time notifications without Google stuff in Tusky. But sometimes the connection gets lost while the notification stays. I think it should either try to reconnect as long as the service runs, or kill the service and the notification as well when there is no connection.

Also, it would be nice if clicking the notifications does something, maybe just open the app.

a1batross:
Not needed anymore with the introduction of UnifiedPush support.

PR #2067, opened 2021-01-31
cleanup drafts when user logs out

connyduck:


app/src/main/java/com/keylesspalace/tusky/components/drafts/DraftHelper.kt
```diff
@@ -126,6 +127,15 @@ class DraftHelper @Inject constructor(
                 .andThen(draftDao.delete(draft.id))
     }
 
+    fun deleteAllDraftsAndAttachmentsForAccount(accountId: Long) {
+        draftDao.loadDraftsSingle(accountId)
```
charlag, line 131: I would probably rather write SQL for this but I don't think it matters much

charlag (approved):

app/src/main/java/com/keylesspalace/tusky/components/drafts/DraftHelper.kt
connyduck: I'm not sure I understand what you mean? The SQL statement is defined in DraftDao?

charlag: I mean for the whole deletion thing, sorry for being unclear

connyduck: Ah! But I need to query the drafts so I can delete the attachment files before deleting the drafts in the db. I can't query & delete in 1 statement, can I?

charlag: I think you can get all those attachments but I think it's a rather complicated query

connyduck: I rather leave it like this then

PR #2070, opened 2021-02-07
change draft attachment directory

connyduck:
This is so that when removing the old drafts feature we can savely delete the attachment files as well. When the old & new feature use the same directory, one file could belong to two drafts.

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/components/drafts/DraftHelper.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/components/drafts/DraftHelper.kt
@@ -59,11 +59,17 @@ class DraftHelper @Inject constructor(
     ): Completable {
         return Single.fromCallable {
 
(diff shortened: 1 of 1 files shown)


charlag:
I'm a bit confused. So now "Tusky" is external dir and inside of it we have drafts (for new drafts I assume). How does it help to delete old drafts? wouldn't you also delete "drafts" directory?

connyduck:
When you save a draft, the new implementation checks if the media is already in the dir, and only copies it if it is not yet there. This is to avoid copying when you edit a draft. 
To delete old draft media, I would go through the db and delete the referenced files. But if they are in the same folder as the the new drafts, it is possible that they are actually used by the new drafts (by editing and saving an old one). By using a different directory, old draft media is copied when you save it as new draft and old media can be safely deleted.

Yes it is a edge case, but still.

PR #2077, opened 2021-02-12
upgrade Material Design lib and use their new time picker

connyduck:
before & after


charlag (commented):
Looks good
Before it always used UTC and we had a hack, did we remove it already?

connyduck:
No, it is still there, and now officially recommended https://github.com/material-components/material-components-android/blob/master/docs/components/Picker.md#timezones

PR #2093, opened 2021-02-28
migrating to ViewBinding part 2: Activities

connyduck:


app/src/main/res/layout/activity_account.xml
```diff
@@ -248,20 +248,63 @@
                         app:layout_constraintTop_toBottomOf="@id/accountFieldList"
                         tools:visibility="visible" />
 
-                    <ViewStub
```
connyduck, line 251: Could not find a good way to use a ViewStub in a ViewBinding, so I inlined the referenced layout. May make AccountActivity a bit slower, but cleaner code is also worth something.

charlag: I'm not ssure I get it, ViewStub would lazily inflate the ID, how does it work now?

charlag: ah, you inlined it, gotcha

charlag (approved):

PR #2099, opened 2021-03-03
fix MainActivity switching to notification tab when opened from recents

connyduck:
This fixes the following bug:

When opening the app via notification, switching tabs and then reopening the app from recents, the app switches to notifications again instead of the last used tab.

Please confirm this also fixes the issue you are having @charlag


Tak (approved):
Makes sense to me

PR #2101, opened 2021-03-04
 Avoid using displayName for displaying purpose

kyori19:
Sometimes `displayName` was used for displaying username purposes.

(before)
image
(after)
image

I fixed them in first commit.

Second commit is my proposal. Make `displayName` private to prevent such mistakes.
If you don't like this, I will remove this commit.


connyduck:
That is a very good catch, thank you.
I prefer to have just the first commit.

kyori19:
OK, updated.

connyduck (approved):

PR #2108, opened 2021-03-07
migrating to ViewBinding part 4: Fragments

connyduck:


app/src/main/java/com/keylesspalace/tusky/fragment/ViewVideoFragment.kt
```diff
@@ -52,18 +55,18 @@ class ViewVideoFragment : ViewMediaFragment() {
     override fun setUserVisibleHint(isVisibleToUser: Boolean) {
         // Start/pause/resume video playback as fragment is shown/hidden
         super.setUserVisibleHint(isVisibleToUser)
-        if (videoView == null) {
+        if (binding.videoView == null) {
```
connyduck, line 58: I need to get rid of the 'userVisibleHint' business or this will crash the hell out of us.

PR #2113, opened 2021-03-09
upgrade viewpager2 and swiperefreshlayout to the most recent alpha

### LINKED ISSUE #2112: Pull-To-Reload often results in scrolling left/right
When attempting to use pull-To-Reload, tusky often changes to the right page instead. It seems that the angle is very narrow to make pull-To-Reload work,resulting in unwanted page changes, especially when using a phone one-handed. 

I double checked against other apps, and most seem to use another angle. Tusky seems to have its margin at 45° (sometimes even less) where other apps have an angle up to 70° to make pull To refresh work. 

* * * *
- Tusky Version: 14.0
- Android Version: 11
- Android Device: Samsung M31

connyduck:
I really don't like alpha versions, but if we can fix #2112 this way its ok I guess

SwipeRefreshLayout alpha was published on July 22, 2020 and has not been promoted since, oh Google 🙄


charlag (approved):

charlag:
@connyduck do you still want to wait with this?

connyduck:
It did not help anything, so I'm closing this

PR #2122, opened 2021-03-18
Adding disclaimer for restricted instances

waltercool:
Hi everyone,

Just saw the Tusky application deserves itself the right to restrict certain instances, and I think it's quite dishonest to not explain or make this public on README.md

Added a small change mentioning the only affected instance. To explain the situation, used the same comment used by Marie Axelsson at commit 5d04a7ccda016e1426fdc6352293c379d28312e9.

Thank you very much.


connyduck:
No thanks.

defango:
conny, it's in your best interest. Or we can let Google know about that other thing

PR #2127, opened 2021-04-03
Add missing breaks in switch statements

l2dy:
Follow https://google.github.io/styleguide/javaguide.html#s4.8.4-switch, and fix a bug (probably).


app/src/main/java/com/keylesspalace/tusky/fragment/NotificationsFragment.java
```diff
@@ -863,6 +863,7 @@ private void onPreferenceChanged(String key) {
                     adapter.setMediaPreviewEnabled(enabled);
                     fullyRefresh();
                 }
+                break;
             }
             case "showNotificationsFilter": {
```
a1batross, line 868: While you're at it, it seems there is missing break statement too.

l2dy: Fixed.

a1batross (approved):

PR #2131, opened 2021-04-07
Convert TimelineFragment to Kotlin & ViewBinding

connyduck:
This should not alter any functionality or interaction with other classes.

The goal is to make future refactorings, like getting rid of SFragment, easier.

Needs to be tested extensively before it is ready to merge.


charlag:
Yay, refactoring of that monster! :fireworks: 
I had a quick glance and it seemed fine to me but I will try to have in-depth look over the weekend

app/src/main/java/com/keylesspalace/tusky/fragment/TimelineFragment.kt
```diff
@@ -0,0 +1,1273 @@
...
+        }
+        if (kind == Kind.TAG) {
+            tags = arguments.getStringArrayList(HASHTAGS_ARG)!!
+        }
+
+        isSwipeToRefreshEnabled = arguments.getBoolean(ARG_ENABLE_SWIPE_TO_REFRESH, true)
+
+        val preferences = PreferenceManager.getDefaultSharedPreferences(activity)
+        val statusDisplayOptions = StatusDisplayOptions(
+                preferences.getBoolean("animateGifAvatars", false),
```
charlag, line 167: should we use `PrefKeys` as long as we are here?

app/src/main/java/com/keylesspalace/tusky/fragment/TimelineFragment.kt
```diff
@@ -0,0 +1,1273 @@
...
+
+    private fun setupTimelinePreferences() {
+        alwaysShowSensitiveMedia = accountManager.activeAccount!!.alwaysShowSensitiveMedia
+        alwaysOpenSpoiler = accountManager.activeAccount!!.alwaysOpenSpoiler
+        if (kind == Kind.HOME) {
+            val preferences = PreferenceManager.getDefaultSharedPreferences(context)
+            filterRemoveReplies = !preferences.getBoolean("tabFilterHomeReplies", true)
+            filterRemoveReblogs = !preferences.getBoolean("tabFilterHomeBoosts", true)
+        }
+        reloadFilters(false)
```
charlag, line 288: this complains, should change in SFragment to

```java
    @VisibleForTesting(otherwise = VisibleForTesting.PROTECTED)
    public void reloadFilters(boolean forceRefresh) {
      //...
    }
```

app/src/main/java/com/keylesspalace/tusky/fragment/TimelineFragment.kt
```diff
@@ -0,0 +1,1273 @@
...
+
+    private fun setupSwipeRefreshLayout() {
+        binding.swipeRefreshLayout.isEnabled = isSwipeToRefreshEnabled
+        binding.swipeRefreshLayout.setOnRefreshListener(this)
+        binding.swipeRefreshLayout.setColorSchemeResources(R.color.tusky_blue)
+    }
+
+    private fun setupRecyclerView() {
+        binding.recyclerView.setAccessibilityDelegateCompat(
+                ListStatusAccessibilityDelegate(binding.recyclerView, this, object : StatusProvider {
```
charlag, line 307: we can make `StatusProvider` `fun interface` now

app/src/main/java/com/keylesspalace/tusky/fragment/TimelineFragment.kt
```diff
@@ -0,0 +1,1273 @@
...
+        }.also {
+            binding.recyclerView.addOnScrollListener(it)
+        }
+
+        if (!eventRegistered) {
+            eventHub.events
+                    .observeOn(AndroidSchedulers.mainThread())
+                    .autoDispose(from(this))
+                    .subscribe { event: Event? ->
+                        if (event is FavoriteEvent) {
```
charlag, line 389: this asks to be `when`

app/src/main/java/com/keylesspalace/tusky/fragment/TimelineFragment.kt
```diff
@@ -0,0 +1,1273 @@
...
+             * should be ignored. */
+            return
+        }
+        super.viewAccount(id)
+    }
+
+    private fun onPreferenceChanged(key: String) {
+        val sharedPreferences = PreferenceManager.getDefaultSharedPreferences(context)
+        when (key) {
+            "fabHide" -> {
```
charlag, line 670: Also have `PrefKeys` for these things

app/src/main/java/com/keylesspalace/tusky/fragment/TimelineFragment.kt
```diff
@@ -0,0 +1,1273 @@
...
+        // using iterator to safely remove items while iterating
+        val iterator = statuses.iterator()
+        while (iterator.hasNext()) {
+            val status = iterator.next().asRightOrNull()
+            if (status != null && LinkHelper.getDomain(status.account.url) == instance) {
+                iterator.remove()
+            }
+        }
+        updateAdapter()
+    }
```
charlag, line 737: I wanted to rewrite these with removeIf like so:

```kotlin
    private fun removeAllByAccountId(accountId: String) {
        // using iterator to safely remove items while iterating
        statuses.removeIf { either ->
            val status = either.asRightOrNull()
            if (status != null) {
                status.account.id == accountId || status.actionableStatus.account.id == accountId
            } else {
                false
            }
        }
        updateAdapter()
    }

    private fun removeAllByInstance(instance: String) {
        // using iterator to safely remove items while iterating
        statuses.removeIf { either ->
            val status = either.asRightOrNull()
            status != null && LinkHelper.getDomain(status.account.url) == instance
        }
        updateAdapter()
    }
```

but it needs API 24. I don't know if it's handled by compile transformations these days or not, I remember they ported some APIs

app/src/main/java/com/keylesspalace/tusky/fragment/TimelineFragment.kt
```diff
@@ -0,0 +1,1273 @@
...
+        } else {
+            val iterator: ListIterator<Either<Placeholder, Status>?> = statuses.listIterator(statuses.size)
+            while (iterator.hasPrevious()) {
+                val previous = iterator.previous()
+                if (previous!!.isRight()) {
+                    bottomId = previous.asRight().id
+                    break
+                }
+            }
+        }
```
charlag, line 772: Can we do this?
```kotlin
        val bottomId: String? = if (kind == Kind.FAVOURITES || kind == Kind.BOOKMARKS) {
            nextId
        } else {
            statuses.lastOrNull { it.isRight() }?.asRight()?.id
        }
```

app/src/main/java/com/keylesspalace/tusky/fragment/TimelineFragment.kt
```diff
@@ -0,0 +1,1273 @@
...
+        if (fullFetch) {
+            newStatuses.add(placeholder)
+        }
+        statuses.addAll(pos, newStatuses)
+        removeConsecutivePlaceholders()
+        updateAdapter()
+    }
+
+    private fun findStatusOrReblogPositionById(statusId: String): Int {
```
charlag, line 1045:
```kotlin
    private fun findStatusOrReblogPositionById(statusId: String): Int {
        return statuses.indexOfFirst { either ->
            val status = either.asRightOrNull()
            status != null &&
                    (statusId == status.id ||
                            (status.reblog != null && statusId == status.reblog.id))
        }
    }
```

app/src/main/java/com/keylesspalace/tusky/fragment/TimelineFragment.kt
```diff
@@ -0,0 +1,1273 @@
...
+    private fun findStatusAndPosition(position: Int, status: Status): Pair<StatusViewData.Concrete, Int>? {
+        val statusToUpdate: StatusViewData.Concrete
+        val positionToUpdate: Int
+        val someOldViewData = statuses.getPairedItem(position)
+
+        // Unlikely, but data could change between the request and response
+        if (someOldViewData is StatusViewData.Placeholder ||
+                (someOldViewData as StatusViewData.Concrete).id != status.id) {
+            // try to find the status we need to update
+            val foundPos = statuses.indexOf(Right<Any, Status>(status))
```
charlag, line 1068: `Right<Any, Status>(status))` -> `Right(status))`

charlag (approved):
Great work!
I left some suggestions, feel free to ignore them, it's nothing important

app/src/main/java/com/keylesspalace/tusky/fragment/TimelineFragment.kt
connyduck: We need to enable desugaring to use it https://developer.android.com/studio/write/java8-support#library-desugaring

The only drawback is a little bit more app size and build time, but I guess its worth it?

app/src/main/java/com/keylesspalace/tusky/fragment/TimelineFragment.kt
connyduck: yes, good point

app/src/main/java/com/keylesspalace/tusky/fragment/TimelineFragment.kt
charlag: Probably doesn't make too much sense as we are using Kotlin for most things. Here it doesn't work because `PairedList` is our own list in Java I assume.

charlag: kotlin has `removeAll`:
https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/remove-all.html

app/src/main/java/com/keylesspalace/tusky/fragment/TimelineFragment.kt
connyduck: yes 🙏

PR #2146, opened 2021-05-04
migrate to RxJava3

connyduck:


app/src/main/java/com/keylesspalace/tusky/viewdata/NotificationViewData.java
```diff
@@ -15,13 +15,13 @@
 
 package com.keylesspalace.tusky.viewdata;
 
+import androidx.annotation.Nullable;
```
charlag, line 18: why not Java's one?

charlag (approved):
Looks good!

app/src/main/java/com/keylesspalace/tusky/viewdata/NotificationViewData.java
connyduck: I think it is the standard one we use everywhere else

PR #2163, opened 2021-05-17
get rid of jcenter

connyduck:
JCenter is getting turned off sometime next year https://jfrog.com/blog/into-the-sunset-bintray-jcenter-gocenter-and-chartcenter/

Only two libraries we use were not available on Maven central: FileEmojiCompat, which @C1710 published there (thx again!) and the Image Cropper, which I replaced with a better fork because the original library is no longer maintained.

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/EditProfileActivity.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/EditProfileActivity.kt
@@ -47,7 +47,7 @@ import com.mikepenz.iconics.IconicsDrawable
 import com.mikepenz.iconics.typeface.library.googlematerial.GoogleMaterial
 import com.mikepenz.iconics.utils.colorInt
 import com.mikepenz.iconics.utils.sizeDp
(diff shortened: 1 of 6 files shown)


connyduck:
For some reason I had to update the BottomSheetActivity Test? Probably because the new imagecropper dependency pulls in newer AndroidX stuff

app/src/main/java/com/keylesspalace/tusky/EditProfileActivity.kt
```diff
@@ -47,7 +47,7 @@ import com.mikepenz.iconics.IconicsDrawable
 import com.mikepenz.iconics.typeface.library.googlematerial.GoogleMaterial
 import com.mikepenz.iconics.utils.colorInt
 import com.mikepenz.iconics.utils.sizeDp
-import com.theartofdev.edmodo.cropper.CropImage
```
charlag, line 50: idk about changes between libraries so don't trust me on checking this :sweat_smile:

charlag (approved):

PR #2165, opened 2021-05-18
don't upscale images in caption dialog

connyduck:
In #2149 I fixed crashes with too large images, but now I'm seeing even more crashes and I think it is because smaller images get upscaled.


charlag (commented):
Why is it so brittle omg.
Can you still zoom in if you downsample?

connyduck:
Yes

PR #2167, opened 2021-05-21
Update dagger to 2.35.1

connyduck:


charlag (approved):
Now activity injection will be done during onCreate() and not before, otherwise nothing else should change for us.

connyduck:
Oh, seems like I have missed this in the changelog, thx for the info

PR #2173, opened 2021-05-22
update to SDK 30 and fix deprecations

connyduck:


connyduck:
`window.setSoftInputMode(WindowManager.LayoutParams.SOFT_INPUT_ADJUST_RESIZE)` is also deprecated, but I could not get the suggested workaround with the `OnApplyWindowInsetsListener` done. 

1) the alert dialog doesn't have a view were one could apply the inset to.
2) the insets also affect the activity behind the dialog and things get messed up.

I will leave it for now, solutions could be to switch to a DialogFrament or an different solution entirely.

app/src/main/java/com/keylesspalace/tusky/ViewMediaActivity.kt
```diff
@@ -137,7 +143,19 @@ class ViewMediaActivity : BaseActivity(), ViewImageFragment.PhotoActionsListener
...
+        ViewCompat.setOnApplyWindowInsetsListener(binding.root) { _, insets ->
+            val top = insets.getInsets(systemBars()).top
+            val toolbarParams = binding.toolbar.layoutParams as ViewGroup.MarginLayoutParams
+            toolbarParams.topMargin = top
+            WindowInsetsCompat.CONSUMED
+        }
+        // hide the system bars
+        WindowInsetsControllerCompat(window, window.decorView).run {
+            hide(systemBars())
+        }
```
connyduck, line 158: this isn't as nice as the `SYSTEM_UI_FLAG_LOW_PROFILE` unfortunately but looks there is no 100% replacement for it :(

app/src/main/java/com/keylesspalace/tusky/AccountActivity.kt
```diff
@@ -237,13 +238,11 @@ class AccountActivity : BottomSheetActivity(), ActionButtonActivity, HasAndroidI
 
     private fun setupToolbar() {
         // set toolbar top margin according to system window insets
-        binding.accountCoordinatorLayout.setOnApplyWindowInsetsListener { _, insets ->
-            val top = insets.systemWindowInsetTop
-
-            val toolbarParams = binding.accountToolbar.layoutParams as CollapsingToolbarLayout.LayoutParams
+        ViewCompat.setOnApplyWindowInsetsListener(binding.accountCoordinatorLayout) { _, insets ->
```
charlag, line 241: I thought they made extension functions instead of static methods for compat stuff now?

app/src/main/java/com/keylesspalace/tusky/ViewMediaActivity.kt
```diff
@@ -137,7 +143,18 @@ class ViewMediaActivity : BaseActivity(), ViewImageFragment.PhotoActionsListener
             true
         }
 
-        window.decorView.systemUiVisibility = View.SYSTEM_UI_FLAG_LOW_PROFILE
```
charlag, line 140: ViewMediaActivity does not work for me:

 - There's a huge lag when opening a picture
 - System bars are hidden by default (which is out of sync with viewer UI)
 - If I tap anywhere then system UI shows up. Bottom nav bar from system will be on top of the description
 - If I tap again then viewer UI disappears but system UI still shows

app/src/main/java/com/keylesspalace/tusky/AccountActivity.kt
connyduck: Apparently only sometimes? I can't find them for these.

app/src/main/java/com/keylesspalace/tusky/ViewMediaActivity.kt
connyduck: I see, that sucks. I'll try to find a better solution or revert to the deprecated flag.

connyduck: I think reverting is the best solution for now.

charlag:
@connyduck unfortunately image viewer is still wonky on me :c
try on image with caption. When I open it the system bars are hidden by default even though our UI is not. If I tap anywhere then system bars appear on top of our UI (over caption). Only the next tap actually makes them both hide.

I wonder if it changed with just SDK level?

connyduck:
Are you sure you have the latests version from this branch? I'm pretty sure it works now, you are describing how it looked before the latest change. If yes, what device are you using?

charlag:
Last time I made sure I even clean built because it sounded so suspicious and this time I also did pull --force, clean rebuild, uninstall, install.

Unfortunately it is still a thing. The only difference is that now description doesn't overlap with nav bar because nav bar does not disappear.

I have FP3 with Android 10.

https://user-images.githubusercontent.com/3099142/120225375-6b14f680-c245-11eb-9ec9-898ccb2a35cb.mp4

charlag:
Oh, it seems like it's the same in prod... sorry.

PR #2175, opened 2021-05-22
Timeline refactor

charlag:
Preview of gradual breakup of TimelineFragment into something more sensible. I don't aim for perfection here but merely for "make it work and make it testable".

First of all I moved some things into their own package.

Second I copied most of the state and loading functions to the ViewModel. Idea for now is that fragment sends request and refreshes when there's some result. As we diff whole list anyway we can do it pretty often.

What is not working yet:
 - Handling of updates
 - Handling of preference changes
 - Handling of loading state
 - Handling of error state

What is working already:
 - Initial loading
 - Refresh of cached
 - Loading above
 - Loading gaps
 - Loading below


connyduck:
I like where this is going!

Do you think we could get rid of the `PairedList` in this as well? It should be possible to just have the `StatusViewData` representation and still do everything.

charlag:
@connyduck thanks for having a look!
I could but it would be a bit bigger refactoring. Do you think we don't need the original status? Probably not. I could also think how we don't need viewData but it has some additional fields now.

charlag:
Status has completely different equality semantics :cry:

charlag:
@connyduck okay I am *hopefully* done with all the functionality, just want to add some tests and fix one or two things here and there and maybe squash some commits. I should stop soon because it is already a monster of a PR.

charlag:
Instead of a continuation-passing mess we had before when loading callbacks would look at the arguments and set the state it is now done in a more sane way: load thing, set state.

charlag:
I think I'm ready now!
Rewrote some commits.
There might be one TODO where you might know why we need a certain thing but mostly I'm pretty confident that it works.

app/src/main/java/com/keylesspalace/tusky/components/timeline/TiemlineViewModel.kt
```diff
@@ -0,0 +1,910 @@
...
+            val response = fetchStatusesForKind(maxId, sinceId, LOAD_AT_ONCE).await()
+            if (response.isSuccessful) {
+                val newNextId = extractNextId(response)
+                if (newNextId != null) {
+                    // when we reach the bottom of the list, we won't have a new link. If
+                    // we blindly write `null` here we will start loading from the top
+                    // again.
+                    nextId = newNextId
+                }
+                response.body()!!.map { Either.Right(it) }
```
connyduck, line 440: maybe better play this safe and return empty list when body is null?

app/src/main/java/com/keylesspalace/tusky/components/timeline/TiemlineViewModel.kt
```diff
@@ -0,0 +1,910 @@
...
+        this.statuses.clear()
+        return loadInitial()
+    }
+
+    private fun fetchStatusesForKind(
+        fromId: String?,
+        uptoId: String?,
+        limit: Int
+    ): Single<Response<List<Status>>> {
+        val api = mastodonApi
```
connyduck, line 603: why not change the name to `api` in the instructor instead

app/src/main/java/com/keylesspalace/tusky/components/timeline/TiemlineViewModel.kt
```diff
@@ -0,0 +1,910 @@
...
+
+    fun voteInPoll(position: Int, choices: List<Int>): Job = viewModelScope.launch {
+        val status = statuses[position].asStatusOrNull() ?: return@launch
+
+        val poll = status.status.poll ?: run {
+            Log.w(TAG, "No poll on status ${status.id}")
+            return@launch
+        }
+
+        val votedPoll = status.actionable.poll!!.votedCopy(choices)
```
connyduck, line 367: shouldn't this be `val votedPoll = poll.votedCopy(choices)`?

app/src/main/java/com/keylesspalace/tusky/fragment/SFragment.java
```diff
@@ -75,9 +71,6 @@
 
 import io.reactivex.rxjava3.android.schedulers.AndroidSchedulers;
 import kotlin.Unit;
-import retrofit2.Call;
-import retrofit2.Callback;
-import retrofit2.Response;
```
connyduck, line 80: There are even more unused imports, and `TAG` is also no longer used.

But moving out things from SFragment, yeah! This thing has to die.

app/src/main/java/com/keylesspalace/tusky/util/ViewDataUtils.kt
```diff
@@ -0,0 +1,60 @@
...
+import com.keylesspalace.tusky.viewdata.StatusViewData
+import com.keylesspalace.tusky.viewdata.toViewData
+import java.util.*
+
+/**
+ * Created by charlag on 12/07/2017.
+ */
+object ViewDataUtils {
+    @JvmStatic
+    fun statusToViewData(
```
connyduck, line 29: `fun Status.toViewData`?

Also it could be a top level funtion, no need for the object if it holds no state

connyduck (commented):
I really like this. I have tested it a bit and so far it looks fine! Will test for longer and report if I find any issues.

app/src/main/java/com/keylesspalace/tusky/components/timeline/TimelineFragment.kt
```diff
@@ -0,0 +1,567 @@
...
+        return inflater.inflate(R.layout.fragment_timeline, container, false)
+    }
+
+    override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
+        setupSwipeRefreshLayout()
+        setupRecyclerView()
+        updateViews()
+        if (viewModel.statuses.isEmpty()) {
+            viewModel.loadInitial()
+        }
```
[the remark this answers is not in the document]
charlag: It is! I wanted to change it and forgot.

app/src/main/java/com/keylesspalace/tusky/components/timeline/TimelineFragment.kt
```diff
@@ -0,0 +1,567 @@
...
+        binding.recyclerView.adapter = adapter
+    }
+
+    private fun showEmptyView() {
+        binding.statusView.show()
+        binding.statusView.setup(R.drawable.elephant_friend_empty, R.string.message_empty, null)
+    }
+
+    override fun onActivityCreated(savedInstanceState: Bundle?) {
+        super.onActivityCreated(savedInstanceState)
```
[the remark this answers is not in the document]
charlag: I didn't introduce it so I would really prefer it in another PR

app/src/main/java/com/keylesspalace/tusky/components/timeline/TiemlineViewModel.kt
charlag: Yes, this makes sense!

app/src/main/java/com/keylesspalace/tusky/components/timeline/TimelineFragment.kt
```diff
@@ -0,0 +1,562 @@
...
+    }
+
+    private fun updateViews() {
+        differ.submitList(viewModel.statuses.toList())
+
+        if (isAdded) {
+            binding.swipeRefreshLayout.isRefreshing = viewModel.isRefreshing
+            binding.progressBar.visible(viewModel.isLoadingInitially)
+            // TODO: ??
+            binding.swipeRefreshLayout.isEnabled = true
```
charlag, line 384: I still don't know what to do with this one

charlag:
Rebased on develop so should have no conflicts now

app/src/main/java/com/keylesspalace/tusky/components/timeline/TimelineFragment.kt
connyduck: what is the problem?

charlag: I have no idea when we should enable/disable it

connyduck: I think it should be enabled as long as no other way of refreshing is available. So it should only be disabled while the fullscreen statusview with a retry button is shown. Having both options is weird imho.

charlag:
@connyduck I still didn't check that place with refresher disabling/enabling and I can't make myself for a long time already. I can try to check it today or if you think it's okay you can just merge.

connyduck:
@charlag Oh, please check it, this should not be rushed.

connyduck (approved):
Thank you so much! I tested it again, looks good!

charlag:
Thanks for testing!

PR #2181, opened 2021-05-28
fix LiveData nullability issues

connyduck:
There is a new lint inspection that tells you when you set `null` into non-null LiveData. This PR fixes the issues it reports.


charlag (approved):
For Google I/O they migrated from LiveData to Flow btw but good that they catch nullability now

connyduck:
Yes, I will try to write new things with flow

PR #2187, opened 2021-05-31
Convert some adapters to Kotlin

charlag:
During timeilne refactoring I realized how annoying existing Java code is and that we should get rid of it.
I tried to not make conflicts with timeline branch so there's a bunch of other adapters.
Protip: if you look at diff from second commit only it will actually show you the proper diff.

This is ofc in no rush and can wait.


app/src/main/java/com/keylesspalace/tusky/adapter/ThreadAdapter.kt
```diff
@@ -0,0 +1,135 @@
...
+            VIEW_TYPE_STATUS_DETAILED -> {
+                val view = LayoutInflater.from(parent.context)
+                    .inflate(R.layout.item_status_detailed, parent, false)
+                StatusDetailedViewHolder(view)
+            }
+            else -> {
+                val view = LayoutInflater.from(parent.context)
+                    .inflate(R.layout.item_status, parent, false)
+                StatusViewHolder(view)
+            }
```
connyduck, line 49: I'd either put this in a if/else to not have 3 branches or throw an exception in the third branch (also in the other adapters)

app/src/main/java/com/keylesspalace/tusky/adapter/ThreadAdapter.kt
```diff
@@ -0,0 +1,135 @@
...
+        return if (position == detailedStatusPosition) {
+            VIEW_TYPE_STATUS_DETAILED
+        } else {
+            VIEW_TYPE_STATUS
+        }
+    }
+
+    override fun getItemCount(): Int {
+        return statuses.size
+    }
```
connyduck, line 68:
```suggestion
    override fun getItemCount() = statuses.size
```

app/src/main/java/com/keylesspalace/tusky/adapter/AccountAdapter.kt
```diff
@@ -0,0 +1,92 @@
+/* Copyright 2017 Andrew Dawson
```
connyduck, line 1: Maybe update these to "2021 Tusky Contributors"?

connyduck (commented):
Looks pretty good!

app/src/main/java/com/keylesspalace/tusky/adapter/AccountAdapter.kt
```diff
@@ -0,0 +1,92 @@
...
+    override fun getItemViewType(position: Int): Int {
+        return if (position == accountList.size && bottomLoading) {
+            VIEW_TYPE_FOOTER
+        } else {
+            VIEW_TYPE_ACCOUNT
+        }
+    }
+
+    fun update(newAccounts: List<Account>) {
+        accountList = removeDuplicates(newAccounts)
```
connyduck, line 44: Kotlin has a built in function for this https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/distinct.html

charlag: It produces immutable list and we need a mutable one and if I re-implement it or add toMutableList() it won't be simpler

connyduck (approved):

PR #2190, opened 2021-06-05
improve kotlin related proguard rules

connyduck:
This gives us about 80kb smaller apk size


charlag (approved):
This will make finding errors a bit harder but I'm okay with that

PR #2206, opened 2021-06-19
migrate drafts to paging 3

connyduck:


charlag (approved):
Wow that looks straightforward

connyduck:
> Wow that looks straightforward

Yeah but I forgot a part 🤦

app/src/main/java/com/keylesspalace/tusky/service/SendTootService.kt
```diff
@@ -244,8 +251,8 @@ class SendTootService : Service(), Injectable {
     }
 
     private fun saveTootToDrafts(toot: TootToSend) {
-
-        draftHelper.saveDraft(
+        serviceScope.launch {
```
charlag, line 254: Maybe easier to just block? I think that's how services are supposed to work anyways.

charlag (approved):

app/src/main/java/com/keylesspalace/tusky/service/SendTootService.kt
connyduck: No they run on the main thread 😬

charlag: Ah, true dat, we don't have separate process

PR #2217, opened 2021-06-30
Timeline fixes

charlag:


connyduck:
Yes, cached toots are now visible, but here is still something very wrong with this. Checkout this video: https://connyduck.at/device-2021-07-01-170651.webm

1) the error view is still flashing at the beginning
2) looks like the first status is diffed the wrong way because it disappears and appears again? Or is there a view update that does not include it inbetween

charlag:
I fixed flashing in the beginning. I don't like how brittle it is though.

Tak:
For what it's worth, I still experience infinite loading placeholder with https://github.com/tuskyapp/Tusky/pull/2217/commits/357250f6d5f303b6190f087d68b68a9974a2903f :neutral_face:

charlag:
@Tak maybe you could try again now? There were multiple issues

charlag:
nvm, I hit it again

connyduck:
superseded by https://github.com/tuskyapp/Tusky/pull/2238

PR #2238, opened 2021-08-29
Timeline paging

connyduck:


app/src/main/java/com/keylesspalace/tusky/components/timeline/TimelineTypeMappers.kt
```diff
@@ -0,0 +1,256 @@
...
+            pinned = status.pinned,
+            muted = status.muted,
+            poll = poll,
+            card = null
+        )
+    }
+    return StatusViewData.Concrete(
+        status = status,
+        isExpanded = this.status.expanded,
+        isShowingContent = this.status.contentHidden,
```
Tak, line 252:
```suggestion
        isShowingContent = !this.status.contentHidden,
```

app/src/main/java/com/keylesspalace/tusky/components/timeline/viewmodel/CachedTimelineViewModel.kt
```diff
@@ -0,0 +1,187 @@
...
+
+    override fun changeExpanded(expanded: Boolean, status: StatusViewData.Concrete) {
+        viewModelScope.launch {
+            db.timelineDao().setExpanded(accountManager.activeAccount!!.id, status.id, expanded)
+        }
+    }
+
+    override fun changeContentHidden(isShowing: Boolean, status: StatusViewData.Concrete) {
+        viewModelScope.launch {
+            db.timelineDao().setContentHidden(accountManager.activeAccount!!.id, status.id, isShowing)
```
Tak, line 91:
```suggestion
            db.timelineDao().setContentHidden(accountManager.activeAccount!!.id, status.id, !isShowing)
```

Tak:
- Long/collapsible statuses are expanded by default (not sure if this should happen in the timeline type mapper, or in the N places where the collapse button gets set up)
- Can't vote in polls from the timeline
- [Timeline loops](https://user-images.githubusercontent.com/142250/146362380-de304898-a02a-48c0-9a74-7545492bb507.png)
- Some(!) media on content warning posts is displayed by default in the timeline (e.g. https://weirder.earth/@distel/107456188493276288 )

Tak:
Crashed on pressing "Load More" in the middle of the home timeline, on restart there's a big chronological gap there with no placeholder

app/src/main/java/com/keylesspalace/tusky/components/timeline/viewmodel/CachedTimelineRemoteMediator.kt
```diff
@@ -0,0 +1,138 @@
...
+        loadType: LoadType,
+        state: PagingState<Int, TimelineStatusWithAccount>
+    ): MediatorResult {
+
+        try {
+            var dbEmpty = false
+            if (!initialRefresh && loadType == LoadType.REFRESH) {
+                val topId = timelineDao.getTopId(activeAccount.id)
+                topId?.let { cachedTopId ->
+                    val statusResponse = api.homeTimeline(
```
charlag, line 58: we do api.homeTimeline(...).await() + unwrap response multiple times, maybe we should wrap it in another method somewhere here

app/src/main/java/com/keylesspalace/tusky/components/timeline/viewmodel/CachedTimelineViewModel.kt
```diff
@@ -0,0 +1,187 @@
...
+        }
+    }
+
+    override fun removeStatusWithId(id: String) {
+        // handled by CacheUpdater
+    }
+
+    override fun loadMore(placeholderId: String) {
+        viewModelScope.launch {
+            val response = api.homeTimeline(maxId = placeholderId.inc(), limit = 20).await()
```
charlag, line 119: we check response in other places but not here

app/src/main/java/com/keylesspalace/tusky/components/timeline/viewmodel/CachedTimelineViewModel.kt
```diff
@@ -0,0 +1,187 @@
...
+        viewModelScope.launch {
+            db.timelineDao().deleteAllFromInstance(accountManager.activeAccount!!.id, instance)
+        }
+    }
+
+    override fun removeStatusWithId(id: String) {
+        // handled by CacheUpdater
+    }
+
+    override fun loadMore(placeholderId: String) {
```
charlag, line 117: I'm trying to wrap my head around how all of this works and this piece is confusing. I thought this is a job of pagingSource/mediator, it seems like for us its the other way around and they call viewModel?

app/src/main/java/com/keylesspalace/tusky/components/timeline/viewmodel/NetworkTimelineViewModel.kt
```diff
@@ -0,0 +1,280 @@
...
+                    alwaysShowSensitiveMedia = !activeAccount.alwaysShowSensitiveMedia && status.actionableStatus.sensitive,
+                    alwaysOpenSpoiler = activeAccount.alwaysOpenSpoiler
+                )
+            }
+
+            val index = statusData.indexOfFirst { it is StatusViewData.Placeholder && it.id == placeholderId }
+            statusData.removeAt(index)
+            statusData.addAll(index, data)
+
+            currentSource?.invalidate()
```
charlag, line 162: it's not obvious to me why it's needed

app/src/main/java/com/keylesspalace/tusky/components/timeline/TimelineFragment.kt
```diff
@@ -179,7 +231,7 @@ class TimelineFragment :
 
     private fun setupRecyclerView() {
         binding.recyclerView.setAccessibilityDelegateCompat(
-            ListStatusAccessibilityDelegate(binding.recyclerView, this) { pos -> viewModel.statuses.getOrNull(pos) }
+            ListStatusAccessibilityDelegate(binding.recyclerView, this) { null }
```
charlag, line 234: This isn't ready, or?

charlag:
Didn't finish looking yet

app/src/main/java/com/keylesspalace/tusky/components/timeline/TimelineFragment.kt
connyduck: indeed not, thx 🧐

app/src/main/java/com/keylesspalace/tusky/components/timeline/viewmodel/CachedTimelineViewModel.kt
connyduck: Mediator doesn't support loading in between, only refresh/append/prepend, thats why it is here

app/src/main/java/com/keylesspalace/tusky/components/timeline/viewmodel/CachedTimelineViewModel.kt
connyduck: thx

app/src/main/java/com/keylesspalace/tusky/components/timeline/viewmodel/NetworkTimelineViewModel.kt
connyduck: Everytime the data changes, the source needs to be invalidated, which prompts the paging system to create a new source from the source factory. Thats how this library works.

app/src/main/java/com/keylesspalace/tusky/components/timeline/viewmodel/CachedTimelineViewModel.kt
connyduck: No, that should be fine the way it is

connyduck: oh, no, because of the inverted logic 🤯 
Let me fix this properly by using the same name everywhere, this is too confusing

connyduck:
There is another bug that is quite annoying:
Whenever statuses get refreshed, their old local state (expanded, media shown, etc) is overwritten by the default values.
To fix this I need to check if the reloaded status already exists, and if yes keep the local state instead of blindly overriding it.

app/src/main/java/com/keylesspalace/tusky/components/timeline/viewmodel/NetworkTimelineViewModel.kt
charlag: We should leave a comment then

charlag:
> There is another bug that is quite annoying: Whenever statuses get refreshed, their old local state (expanded, media shown, etc) is overwritten by the default values. To fix this I need to check if the reloaded status already exists, and if yes keep the local state instead of blindly overriding it.

should we maybe separate the state into a different table and do a join request of sorts? It's complicated but it's probably more clean. Otherwise we have to insert only certain fields

connyduck:
> should we maybe separate the state into a different table and do a join request of sorts? It's complicated but it's probably more clean. Otherwise we have to insert only certain fields

I solved it now by matching statuses in-memory and copying the attributes. Might not be super elegant but definitely works.

app/src/main/java/com/keylesspalace/tusky/db/TimelineStatusEntity.kt
```diff
@@ -50,15 +65,19 @@ data class TimelineStatusEntity(
...
+    val visibility: Status.Visibility,
     val attachments: String?,
     val mentions: String?,
     val application: String?,
     val reblogServerId: String?, // if it has a reblogged status, it's id is stored here
     val reblogAccountId: String?,
     val poll: String?,
-    val muted: Boolean?
+    val muted: Boolean?,
+    val expanded: Boolean, // used as the "loading" attribute when this TimelineStatusEntity is a placeholder
```
connyduck, line 77: Could we do this nicer? Is it worth having an additional column "loading" in the db?

app/src/main/java/com/keylesspalace/tusky/components/timeline/viewmodel/CachedTimelineRemoteMediator.kt
connyduck: I think I prefer to do this later when replacing Rx with coroutines

connyduck:
Ok, I think I addressed all problems, please re-review @charlag @Tak 

Important: When reporting a bug, please tell me which timeline was affected

Tak:
I still can't vote in polls from the home timeline. I think everything else I mentioned has been resolved.

app/src/main/java/com/keylesspalace/tusky/components/timeline/TimelineTypeMappers.kt
```diff
@@ -0,0 +1,256 @@
...
+        favourited = actionableStatus.favourited,
+        bookmarked = actionableStatus.bookmarked,
+        sensitive = actionableStatus.sensitive,
+        spoilerText = actionableStatus.spoilerText,
+        visibility = actionableStatus.visibility,
+        attachments = actionableStatus.attachments.let(gson::toJson),
+        mentions = actionableStatus.mentions.let(gson::toJson),
+        application = actionableStatus.application.let(gson::toJson),
+        reblogServerId = reblog?.id,
+        reblogAccountId = reblog?.let { this.account.id },
```
charlag, line 143: should this be `it.account.id` instead? I don't remember

charlag: No, I think it's right...

app/src/main/java/com/keylesspalace/tusky/components/timeline/viewmodel/CachedTimelineRemoteMediator.kt
```diff
@@ -0,0 +1,154 @@
...
+            timelineDao.insertAccount(status.account.toEntity(activeAccount.id, gson))
+            status.reblog?.account?.toEntity(activeAccount.id, gson)?.let { rebloggedAccount ->
+                timelineDao.insertAccount(rebloggedAccount)
+            }
+
+            // check if we already have one of the newly loaded statuses cached locally
+            // in case we do, copy the local state (expanded, contentShowing, contentCollapsed) over so it doesn't get lost
+            var oldStatus: TimelineStatusEntity? = null
+            for (page in state.pages) {
+                oldStatus = page.data.find { s ->
```
charlag, line 132: This is a triple-nested loop which makes me a bit uneasy but I guess pages don't overlap so it is like 2 loops. Still not ideal but will do.

app/src/main/java/com/keylesspalace/tusky/components/timeline/viewmodel/CachedTimelineViewModel.kt
charlag: Well we have a reference to our mediator, we could put it there just so that it isn't scattered

app/src/main/java/com/keylesspalace/tusky/components/timeline/viewmodel/NetworkTimelineRemoteMediator.kt
```diff
@@ -0,0 +1,109 @@
...
+                    isExpanded = expanded,
+                    isCollapsed = contentCollapsed
+                )
+            }
+
+            if (loadType == LoadType.REFRESH && viewModel.statusData.isNotEmpty()) {
+
+                val insertPlaceholder = if (statuses.isNotEmpty()) {
+                    !viewModel.statusData.removeAll { statusViewData ->
+                        statuses.find { status -> status.id == statusViewData.asStatusOrNull()?.id } != null
```
charlag, line 82:
```suggestion
                        statuses.any { status -> status.id == statusViewData.asStatusOrNull()?.id }
```

app/src/main/java/com/keylesspalace/tusky/db/TimelineDao.kt
```diff
@@ -1,84 +1,66 @@
...
+    abstract suspend fun insertStatus(timelineStatusEntity: TimelineStatusEntity): Long
 
     @Query(
         """
 SELECT s.serverId, s.url, s.timelineUserId,
 s.authorServerId, s.inReplyToId, s.inReplyToAccountId, s.createdAt,
 s.emojis, s.reblogsCount, s.favouritesCount, s.reblogged, s.favourited, s.bookmarked, s.sensitive,
 s.spoilerText, s.visibility, s.mentions, s.application, s.reblogServerId,s.reblogAccountId,
-s.content, s.attachments, s.poll, s.muted,
+s.content, s.attachments, s.poll, s.muted, s.expanded, s.contentShowing, s.contentCollapsed, s.pinned,
```
charlag, line 39: I thought you preserve these fields in memory anyway?

app/src/main/java/com/keylesspalace/tusky/db/TimelineStatusEntity.kt
charlag: We could have it as number to map it to enum but it seems like none of these needs to be in DB anyway

charlag (approved):
I'm done.
I think we should merge it and let people test it on nightly but I will write my thoughts anyway.

My intuition tells me that there should be one viewModel and factory should decide which loader/whatever is injected into it (cached or not) but I don't know if it would be even practically possible.

app/src/main/java/com/keylesspalace/tusky/components/timeline/viewmodel/CachedTimelineRemoteMediator.kt
connyduck: yeah it is not pretty :/

app/src/main/java/com/keylesspalace/tusky/db/TimelineDao.kt
connyduck: well yes, but the UI does still only get updated through the db

PR #2240, opened 2021-08-31
Add markdown support

Aldarone:
Allow users to toot in markdown when they are on a compatible instance.

- [x] Backend detection
- [ ] Display a format selector
- [ ] Use `text/markdown` content type when selected

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/util/VersionUtils.java
+++ b/app/src/main/java/com/keylesspalace/tusky/util/VersionUtils.java
@@ -21,24 +21,54 @@ import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
 public class VersionUtils {
+    private int vanillaMajor;
+    private int vanillaMinor;
+    private int vanillaPatch;
 
-    private int major;
-    private int minor;
-    private int patch;
+    private String backend = "Mastodon";
+    private int backendMajor;
+    private int backendMinor;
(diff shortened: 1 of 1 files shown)


charlag:
Isn't it returned with the instance metadata? It's not great to just pattern match server software name tbh

Aldarone:
> Isn't it returned with the instance metadata? It's not great to just pattern match server software name tbh

I agree. I did it like that because Pleroma has everything we need in the `api/v1/instance` endpoint as well as in nodeinfo. But Mastodon isn't there yet (I'm told it's « planned »)

charlag:
Well maybe they should finish it then rather than us building hacks.

a1batross:
There is also Mastodon Glitch that supports Markdown.

I did it by comparing the version and software name and then depending on that I just returned either the field in nodeinfo, containing all supported markup languages (on Pleroma you can disable Markdown, also there is BBcode and HTML) or return just "text/markdown" for Glitch.

Tak:
Tusky works with glitch *today*, because in glitch you can specify a per-account default markup style for posts

Tak:
+1 for matching capabilities instead of backend names

connyduck:
Will you continue working on this @Aldarone ?

Aldarone:
Sorry I'd like to but life got in the way :frowning_face:

PR #2242, opened 2021-09-06
Highlight your own votes when displaying poll results

Tak:
poll-own-dark poll-result-own-light


cadadr:
Could you add something like a :heavy_check_mark: near the option(s) the user voted? It's hard to tell which option they picked, from the screenshot.

connyduck (approved):
nice 👍

PR #2245, opened 2021-09-20
Add confirmation for favourite and bookmark actions

ColinKinloch:
Adds the same confirmation dialogues for favourite and bookmark actions as boost has.

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/components/preference/PreferencesFragment.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/components/preference/PreferencesFragment.kt
@@ -174,13 +174,6 @@ class PreferencesFragment : PreferenceFragmentCompat(), Injectable {
                 }
 
                 switchPreference {
-                    setDefaultValue(false)
-                    key = PrefKeys.CONFIRM_BOOKMARKS
-                    setTitle(R.string.pref_title_confirm_bookmarks)
-                    isSingleLineTitle = false
-                }
-
-                switchPreference {
                     setDefaultValue(true)
                     key = PrefKeys.ENABLE_SWIPE_FOR_TABS
                     setTitle(R.string.pref_title_enable_swipe_for_tabs)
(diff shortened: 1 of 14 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -591,6 +592,8 @@
     <string name="warning_scheduling_interval">Mastodon has a minimum scheduling interval of 5 minutes.</string>
     <string name="pref_title_show_cards_in_timelines">Show link previews in timelines</string>
     <string name="pref_title_confirm_reblogs">Show confirmation dialog before boosting</string>
+    <string name="pref_title_confirm_favourites">Show confirmation dialog before favouriting</string>
```
connyduck, line 595: American English please (British has its own strings.xml)

app/src/main/java/com/keylesspalace/tusky/components/preference/PreferencesFragment.kt
```diff
@@ -166,6 +166,20 @@ class PreferencesFragment : PreferenceFragmentCompat(), Injectable {
                     isSingleLineTitle = false
                 }
 
+                switchPreference {
+                    setDefaultValue(true)
```
connyduck, line 170: I'd make it default off

connyduck (changes requested):
Works great!
I have to question the bookmark one though - bookmarking a post does not cause a notification anywhere so one can just un-bookmark if accidentally done so.

app/src/main/java/com/keylesspalace/tusky/fragment/NotificationsFragment.java
```diff
@@ -255,6 +255,8 @@ public View onCreateView(@NonNull LayoutInflater inflater, @Nullable ViewGroup c
                 preferences.getBoolean("useBlurhash", true),
                 CardViewMode.NONE,
                 preferences.getBoolean("confirmReblogs", true),
+                preferences.getBoolean("confirmFavourites", true),
+                preferences.getBoolean("confirmBookmarks", true),
```
connyduck, line 259:
```suggestion
```

connyduck (commented):
It is not building right now
`NotificationsFragment.java:250: error: constructor StatusDisplayOptions in class StatusDisplayOptions cannot be applied to given types`

connyduck (approved):
Thank you!

Issue #2257, opened 2021-10-17
Compose button on profile page is half hidden


nailyk-fr:
While opening a profile, the button to compose a toot while mentioning the account is half hidden (but still usable) : 

<a href="https://user-images.githubusercontent.com/16372405/137637623-0f553920-7a4e-45dd-9fb0-1fac48093997.png"><img src="https://user-images.githubusercontent.com/16372405/137637623-0f553920-7a4e-45dd-9fb0-1fac48093997.png" width="250"></a>

* * * *
- Tusky Version: 15.1-0a992480
- Android Version: 10 (Lineage)
- Android Device: H4413
* * * *
- Tusky Version: 15.1-821411b0
- Android Version: 10 (Lineage)
- Android Device: G8142
* * * *
- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.
* * * *

nailyk-fr:
Thank you @connyduck ! 🙂

PR #2267, opened 2021-11-16
fix compose button on profile page being half hidden

connyduck:
closes https://github.com/tuskyapp/Tusky/issues/2257


app/src/main/java/com/keylesspalace/tusky/AccountActivity.kt
```diff
@@ -141,6 +142,7 @@ class AccountActivity : BottomSheetActivity(), ActionButtonActivity, HasAndroidI
         animateEmojis = sharedPrefs.getBoolean(PrefKeys.ANIMATE_CUSTOM_EMOJIS, false)
         hideFab = sharedPrefs.getBoolean("fabHide", false)
 
+        handleWindowInsets()
```
charlag, line 145: is it enough to call it in `onCreate()` only? if yes, I have no other notes. Otherwise I would put it to something like `onResume()`

PR #2282, opened 2021-12-27
fix confirmFavourites default

connyduck:
It was `false` in `PreferencesFragment`, but `true` everywhere else.

Apparently was overlooked when the default was changed https://github.com/tuskyapp/Tusky/pull/2245


charlag (approved):
Oof we really need to refactor all the prefs, don't we?

connyduck:
Yes we should. It is on my list but let's get a release out first.

PR #2285, opened 2021-12-31
filter messages by also looking at media descriptions

### LINKED ISSUE #2284: Filters should be applied to media descriptions
Currently, Tusky doesn’t filter a toot if the word of the filter isn’t in the toot itself or in the CW, but in the media description.

It wasn’t working in Mastodon web when doing client-side filtering until very recently, but that was an oversight and it’s now fixed, see: https://github.com/mastodon/mastodon/pull/17206

* * * *
- Tusky Version: 15.1
- Android Version: 10
- Android Device: Oneplus 5 (cheeseburger)

ariasuni:
Fix #2284

I have not tried it yet, I’ll see if I can test that quickly.

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/network/FilterModel.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/network/FilterModel.kt
@@ -29,9 +29,16 @@ class FilterModel @Inject constructor() {
         }
 
         val spoilerText = status.actionableStatus.spoilerText
+        val attachmentsDescriptions = status.attachments
+            .mapNotNull { it.description }
+
         return (
             matcher.reset(status.actionableStatus.content).find() ||
-                spoilerText.isNotEmpty() && matcher.reset(spoilerText).find()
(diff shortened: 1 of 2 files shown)


Tak:
Makes sense to me

app/src/main/java/com/keylesspalace/tusky/network/FilterModel.kt
```diff
@@ -29,9 +29,17 @@ class FilterModel @Inject constructor() {
         }
 
         val spoilerText = status.actionableStatus.spoilerText
+        val attachmentsDescriptions = status.attachments
+            .filter { it.description != null }
+            .map { it.description as String }
```
connyduck, line 34:
```suggestion
            .mapNotNull { it.description }
```

connyduck (changes requested):
Can you please extend the `FilterTest` with a testcase for this?

ariasuni:
I added the test and improved the code as suggested :slightly_smiling_face:

connyduck:
Thank you!

PR #2291, opened 2022-01-13
Update gradle, kotlin and other dependencies

### LINKED ISSUE #2233: Crash when press the "Mute" button on profile
crash log: https://paste.crdroid.net/Mq6xOb

* * * *
- Tusky Version: **15.1-821411b0 (nightly)**
  - **15.1 (release) is not affected.**
- Android Version: 11
- Android Device: POCO F2 Pro

connyduck:
fixes #2233

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -130,6 +130,8 @@ dependencies {
 
     implementation "com.google.android.material:material:1.4.0"
 
+    implementation "com.google.code.gson:gson:2.8.9"
+
     implementation "com.squareup.retrofit2:retrofit:$retrofitVersion"
     implementation "com.squareup.retrofit2:converter-gson:$retrofitVersion"
     implementation "com.squareup.retrofit2:adapter-rxjava3:$retrofitVersion"
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -83,22 +83,26 @@ android {
             enableSplit = false
         }
     }
+    kotlinOptions {
+        freeCompilerArgs += "-opt-in=kotlin.RequiresOptIn"
+    }
 }
 
+ext.coroutinesVersion = "1.6.0"
 ext.lifecycleVersion = "2.3.1"
 ext.roomVersion = '2.3.0'
 ext.retrofitVersion = '2.9.0'
-ext.okhttpVersion = '4.9.1'
+ext.okhttpVersion = '4.9.3'
 ext.glideVersion = '4.12.0'
-ext.daggerVersion = '2.37'
-ext.materialdrawerVersion = '8.4.1'
+ext.daggerVersion = '2.40.5'
+ext.materialdrawerVersion = '8.4.5'
 
 // if libraries are changed here, they should also be changed in LicenseActivity
 dependencies {
     implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk8:$kotlin_version"
 
-    implementation 'org.jetbrains.kotlinx:kotlinx-coroutines-android:1.5.0'
-    implementation 'org.jetbrains.kotlinx:kotlinx-coroutines-rx3:1.5.0'
+    implementation "org.jetbrains.kotlinx:kotlinx-coroutines-android:$coroutinesVersion"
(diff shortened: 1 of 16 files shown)


app/build.gradle
```diff
@@ -83,30 +83,34 @@ android {
             enableSplit = false
         }
     }
+    kotlinOptions {
+        freeCompilerArgs += "-opt-in=kotlin.RequiresOptIn"
```
connyduck, line 87: Kotlin 1.6 is way stricter with how experimental annotations are handled, so this is now necessary.

app/build.gradle
```diff
@@ -83,30 +83,34 @@ android {
...
 dependencies {
     implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk8:$kotlin_version"
 
-    implementation 'org.jetbrains.kotlinx:kotlinx-coroutines-android:1.5.0'
-    implementation 'org.jetbrains.kotlinx:kotlinx-coroutines-rx3:1.5.0'
+    implementation "org.jetbrains.kotlinx:kotlinx-coroutines-android:$coroutinesVersion"
+    implementation "org.jetbrains.kotlinx:kotlinx-coroutines-rx3:$coroutinesVersion"
 
     implementation "androidx.core:core-ktx:1.5.0"
     implementation "androidx.appcompat:appcompat:1.3.0"
```
connyduck, line 108: Can't update most AndroidX dependencies yet, because they require SDK 31

app/src/main/java/com/keylesspalace/tusky/components/account/AccountActivity.kt
```diff
@@ -68,7 +68,9 @@ import com.keylesspalace.tusky.interfaces.LinkListener
 import com.keylesspalace.tusky.interfaces.ReselectableFragment
 import com.keylesspalace.tusky.settings.PrefKeys
 import com.keylesspalace.tusky.util.DefaultTextWatcher
+import com.keylesspalace.tusky.util.Error
```
connyduck, line 71: We had the wrong import for this all the time 😱 
This new inspection highlighted it as a warning, I would have never seen that otherwise

charlag (approved):
I didn't check changelogs, just fyi

connyduck:
With coroutines 1.6 the testing library got refactored, so TimelineViewModelTest has some deprecated stuff now, I could not figure out how to migrate it yet.

connyduck:
I tested this now for a while and I am pretty sure it is fine

app/proguard-rules.pro
```diff
@@ -47,6 +52,18 @@
     public *;
 }
 
+# https://github.com/google/gson/blob/master/examples/android-proguard-example/proguard.cfg
+# Retain generic signatures of TypeToken and its subclasses with R8 version 3.0 and higher.
+-keep,allowobfuscation,allowshrinking class com.google.gson.reflect.TypeToken
+-keep,allowobfuscation,allowshrinking class * extends com.google.gson.reflect.TypeToken
+
+# Retain generic signaturs of classes used in MastodonApi so Retrofit works
```
charlag, line 59: >signaturs

app/proguard-rules.pro
```diff
@@ -5,6 +5,7 @@
 -optimizationpasses 6
 -allowaccessmodification
 -dontpreverify
+-dontobfuscate
```
charlag: Does it remove obfuscate or adds it?

connyduck: This flag makes R8 skip obfuscation. But it will still do the other shrink optimisations. I use it to debug crashes caused by too much shrinking, because "class x.z.y not found" is not helpful and with this flag you get the full classname again. But it shouldn't be used in production imho.

app/proguard-rules.pro
connyduck: thx

app/proguard-rules.pro
charlag: That's what I thought. I don't know how much obfuscation matters for the size. I'm also a bit confused, you want to roll it back before the release or have another proguard file or?..

PR #2292, opened 2022-01-13
fix muting accounts

connyduck:
For some reason the dialog layout is removed during the build which causes the app to crash when muting 

closes https://github.com/tuskyapp/Tusky/issues/2233


charlag (approved):
I don't know how databinding works, I would feel better if we knew why it disappears

connyduck:
Indeed. I will investigate further before merging this 🧐

connyduck:
Ok so the thing is that R8 (the code shrinker in the Android toolchain) got way more agressive in recent versions and requires more keep rules. Since we do not use the official `proguard-android-optimize.txt` but our own optimization rules we have to add then manually. 

I will do this in #2291 because there the Android Gradle Plugin is updated (and R8 with it) which requires even more rules.

Closing this.

PR #2300, opened 2022-01-21
Cleanup dagger setup

connyduck:


app/src/main/java/com/keylesspalace/tusky/components/compose/MediaUploader.kt
```diff
@@ -61,21 +62,16 @@ fun createNewImageFile(context: Context): File {
 
 data class PreparedMedia(val type: QueuedMedia.Type, val uri: Uri, val size: Long)
 
-interface MediaUploader {
```
connyduck, line 64: Having these interfaces has become an antipattern imho. They used to be helpful years ago, but now with Testframeworks that can mock final classes and powerful refactoring tools available they are really just boilerplate.

charlag (approved):

PR #2303, opened 2022-01-24
Push notifications support via UnifiedPush

### LINKED ISSUE #793: Push notifications?
Currently, only fetch notifications are implemented, and the minimum frequency is 15 minutes. Would it be possible to implement push notifications?

* * * *
- Tusky Version: 2.3
- Android Version: 7.1
- Android Device: Xperia Z5

- [✔️] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

PeterCxy:
Fixes #793.

This is an implementation for push notifications based on UnifiedPush for Tusky. No push gateway (other than UP itself) is needed, since UnifiedPush is simple enough such that it can act as a catch-all endpoint for WebPush messages. When a UnifiedPush distributor is present on-device, we will by default register Tusky as a receiver; if no UnifiedPush distributor is available, then pull notifications are used as a fallback mechanism. 

Because WebPush messages are encrypted, and Mastodon does not send the keys and IV needed for decryption in the request body, for now the push handler simply acts as a trigger for the pre-existing NotificationWorker which is also used for pull notifications. Nevertheless, I have implemented proper key generation and storage, just in case we would like to implement full decryption support in the future when Mastodon upgrades to the latest WebPush encryption scheme that includes all information in the request body.

For users with existing accounts, push notifications will not be enabled until all of the accounts have been re-logged in to grant the new `push` OAuth scope. A small prompt will be shown (until dismissed) as a `Snackbar` to explain to the user about this, and an option is added in Account Preferences to facilitate re-login without deleting local drafts and cache.

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/components/login/LoginActivity.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/components/login/LoginActivity.kt
@@ -92,10 +92,15 @@ class LoginActivity : BaseActivity(), Injectable {
 
         if (savedInstanceState == null &&
             BuildConfig.CUSTOM_INSTANCE.isNotBlank() &&
-            !isAdditionalLogin()
+            !isAdditionalLogin() && !isAccountMigration()
         ) {
             binding.domainEditText.setText(BuildConfig.CUSTOM_INSTANCE)
             binding.domainEditText.setSelection(BuildConfig.CUSTOM_INSTANCE.length)
+        }
+
+        if (isAccountMigration()) {
+            binding.domainEditText.setText(accountManager.activeAccount!!.domain)
+            binding.domainEditText.isEnabled = false
         }
 
         if (BuildConfig.CUSTOM_LOGO_URL.isNotBlank()) {
@@ -120,7 +125,7 @@ class LoginActivity : BaseActivity(), Injectable {
             textView?.movementMethod = LinkMovementMethod.getInstance()
         }
 
-        if (isAdditionalLogin()) {
+        if (isAdditionalLogin() || isAccountMigration()) {
             setSupportActionBar(binding.toolbar)
             supportActionBar?.setDisplayHomeAsUpEnabled(true)
             supportActionBar?.setDisplayShowTitleEnabled(false)
@@ -135,7 +140,7 @@ class LoginActivity : BaseActivity(), Injectable {
 
     override fun finish() {
         super.finish()
-        if (isAdditionalLogin()) {
+        if (isAdditionalLogin() || isAccountMigration()) {
             overridePendingTransition(R.anim.slide_from_left, R.anim.slide_to_right)
         }
     }
@@ -230,7 +235,7 @@ class LoginActivity : BaseActivity(), Injectable {
             domain, clientId, clientSecret, oauthRedirectUri, code, "authorization_code"
         ).fold(
             { accessToken ->
-                accountManager.addAccount(accessToken.accessToken, domain)
+                accountManager.addAccount(accessToken.accessToken, domain, OAUTH_SCOPES)
 
                 val intent = Intent(this, MainActivity::class.java)
                 intent.flags = Intent.FLAG_ACTIVITY_NEW_TASK or Intent.FLAG_ACTIVITY_CLEAR_TASK
@@ -262,19 +267,28 @@ class LoginActivity : BaseActivity(), Injectable {
     }
 
     private fun isAdditionalLogin(): Boolean {
-        return intent.getBooleanExtra(LOGIN_MODE, false)
+        return intent.getIntExtra(LOGIN_MODE, MODE_DEFAULT) == MODE_ADDITIONAL_LOGIN
+    }
+
+    private fun isAccountMigration(): Boolean {
+        return intent.getIntExtra(LOGIN_MODE, MODE_DEFAULT) == MODE_MIGRATION
     }
 
     companion object {
         private const val TAG = "LoginActivity" // logging tag
-        private const val OAUTH_SCOPES = "read write follow"
+        private const val OAUTH_SCOPES = "read write follow push"
         private const val LOGIN_MODE = "LOGIN_MODE"
         private const val DOMAIN = "domain"
         private const val CLIENT_ID = "clientId"
         private const val CLIENT_SECRET = "clientSecret"
 
+        const val MODE_DEFAULT = 0
+        const val MODE_ADDITIONAL_LOGIN = 1
+        // "Migration" is used to update the OAuth scope granted to the client
+        const val MODE_MIGRATION = 2
+
         @JvmStatic
-        fun getIntent(context: Context, mode: Boolean): Intent {
+        fun getIntent(context: Context, mode: Int): Intent {
             val loginIntent = Intent(context, LoginActivity::class.java)
             loginIntent.putExtra(LOGIN_MODE, mode)
             return loginIntent
--- a/app/src/main/java/com/keylesspalace/tusky/components/notifications/PushNotificationHelper.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/components/notifications/PushNotificationHelper.kt
@@ -1,0 +1,220 @@
+/* Copyright 2022 Tusky contributors
+ *
+ * This file is a part of Tusky.
+ *
+ * This program is free software; you can redistribute it and/or modify it under the terms of the
+ * GNU General Public License as published by the Free Software Foundation; either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * Tusky is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even
+ * the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU General
+ * Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License along with Tusky; if not,
+ * see <http://www.gnu.org/licenses>. */
+
+@file:JvmName("PushNotificationHelper")
+package com.keylesspalace.tusky.components.notifications
+
+import android.app.NotificationManager
+import android.content.Context
+import android.os.Build
+import android.util.Log
+import android.view.View
+import androidx.appcompat.app.AlertDialog
+import androidx.preference.PreferenceManager
+import com.google.android.material.snackbar.Snackbar
+import com.keylesspalace.tusky.R
+import com.keylesspalace.tusky.components.login.LoginActivity
+import com.keylesspalace.tusky.db.AccountEntity
+import com.keylesspalace.tusky.db.AccountManager
+import com.keylesspalace.tusky.entity.Notification
+import com.keylesspalace.tusky.network.MastodonApi
+import com.keylesspalace.tusky.util.CryptoUtil
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.withContext
+import org.unifiedpush.android.connector.UnifiedPush
+import retrofit2.HttpException
+
+private const val TAG = "PushNotificationHelper"
+
+private const val KEY_MIGRATION_NOTICE_DISMISSED = "migration_notice_dismissed"
+
+private fun anyAccountNeedsMigration(accountManager: AccountManager): Boolean =
+    accountManager.accounts.any(::accountNeedsMigration)
+
+private fun accountNeedsMigration(account: AccountEntity): Boolean =
+    !account.oauthScopes.contains("push")
+
+fun currentAccountNeedsMigration(accountManager: AccountManager): Boolean =
+    accountManager.activeAccount?.let(::accountNeedsMigration) ?: false
+
+fun showMigrationNoticeIfNecessary(context: Context, parent: View, accountManager: AccountManager) {
+    // No point showing anything if we cannot enable it
+    if (!isUnifiedPushAvailable(context)) return
+    if (!anyAccountNeedsMigration(accountManager)) return
+
+    val pm = PreferenceManager.getDefaultSharedPreferences(context)
+    if (pm.getBoolean(KEY_MIGRATION_NOTICE_DISMISSED, false)) return
+
+    Snackbar.make(parent, R.string.tips_push_notification_migration, Snackbar.LENGTH_INDEFINITE).apply {
+        setAction(R.string.action_details) { showMigrationExplanationDialog(context, accountManager) }
+        show()
+    }
+}
+
+private fun showMigrationExplanationDialog(context: Context, accountManager: AccountManager) {
+    AlertDialog.Builder(context).apply {
+        if (currentAccountNeedsMigration(accountManager)) {
+            setMessage(R.string.dialog_push_notification_migration)
+            setPositiveButton(R.string.title_migration_relogin) { _, _ ->
+                context.startActivity(LoginActivity.getIntent(context, LoginActivity.MODE_MIGRATION))
+            }
+        } else {
+            setMessage(R.string.dialog_push_notification_migration_other_accounts)
+        }
+        setNegativeButton(R.string.action_dismiss) { dialog, _ ->
+            val pm = PreferenceManager.getDefaultSharedPreferences(context)
+            pm.edit().putBoolean(KEY_MIGRATION_NOTICE_DISMISSED, true).apply()
+            dialog.dismiss()
+        }
+        show()
+    }
+}
+
+private suspend fun enableUnifiedPushNotificationsForAccount(context: Context, api: MastodonApi, accountManager: AccountManager, account: AccountEntity) {
+    if (isUnifiedPushNotificationEnabledForAccount(account)) {
+        // Already registered, update the subscription to match notification settings
+        updateUnifiedPushSubscription(context, api, accountManager, account)
+    } else {
+        UnifiedPush.registerAppWithDialog(context, account.id.toString())
+    }
+}
+
+fun disableUnifiedPushNotificationsForAccount(context: Context, account: AccountEntity) {
+    if (!isUnifiedPushNotificationEnabledForAccount(account)) {
+        // Not registered
+        return
+    }
+
+    UnifiedPush.unregisterApp(context, account.id.toString())
+}
+
+fun isUnifiedPushNotificationEnabledForAccount(account: AccountEntity): Boolean =
+    account.unifiedPushUrl.isNotEmpty()
+
+private fun isUnifiedPushAvailable(context: Context): Boolean =
+    UnifiedPush.getDistributors(context).isNotEmpty()
+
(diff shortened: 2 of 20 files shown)


connyduck:
This is awesome 😍 

Regarding the migration of old accounts - prompting the user to relogin is probably the best idea. Be careful not to delete the users local settings & drafts.

Please add no new code with Java & Rx, we are trying to replace it all with Kotlin & Coroutines right now 😅 
Instead of adding to NotificationHelper.java I would add a new file PushNotificationHelper.kt.

Mek101:
Following this

karmanyaahm:
> Because WebPush messages are encrypted, and the UnifiedPush specification does not handle binary messages yet (but will eventually), for now the push handler simply acts as a trigger for the pre-existing NotificationWorker which is also used for pull notifications.

Just an update @PeterCxy: v2.0 of UP android-connector now handles binary messages properly

PeterCxy:
@karmanyaahm Thank you, I've been waiting for their binary message support for a while. I'll see if I can make use of that to do direct notification delivery later (after I finish my work for this semester).

connyduck:
I would really really like to have this feature in Tusky - will you continue working on it @PeterCxy? Is there anything I can help you with?

PeterCxy:
@connyduck I will be working on it -- hopefully I'll be back in the next few weeks or so, if everything goes well :)

PeterCxy:
I've rebased the changes to the latest develop branch + latest UnifiedPush connector 2.0. Currently looking into potentially implementing WebPush decryption (and re-registering the push notification types after changing it in settings)

karmanyaahm:
I don't know if you already have this, but this was posted to the UP chat a little while back by @emersion. It might be a good reference https://git.sr.ht/~emersion/PushGardenTestApp/tree/master/item/app/src/main/java/com/example/pushgardentestapp/WebPush.kt

PeterCxy:
@karmanyaahm Yes, I do have this :)

PeterCxy:
@connyduck Turns out we cannot implement webpush decryption support right now as Mastodon is using an old version of the webpush library that does not include a needed nonce in the notification payload. So we'll have to settle on using UnifiedPush purely as a trigger to pull notifications right now.

PeterCxy:
Two issues remaining for this PR:

1. The migration flow for existing users (to change the OAuth scope)
2. Updating subscription settings on the Mastodon server side according to local notification settings (i.e. whether each channel is enabled / disabled)

PeterCxy:
@connyduck For migration, I just added a prompt for now to let user know that they need to re-login to use push notifications. I'll definitely make dismissing that prompt permanent. The next question is how we would like to actually perform the migration -- do we just let the user re-login on their own, or do we add a new option (in Settings, for example, and direct users there in the prompt) specifically for this purpose?

connyduck:
> @connyduck For migration, I just added a prompt for now to let user know that they need to re-login to use push notifications. I'll definitely make dismissing that prompt permanent. The next question is how we would like to actually perform the migration -- do we just let the user re-login on their own, or do we add a new option (in Settings, for example, and direct users there in the prompt) specifically for this purpose?

A one time prompt that explains users they should relogin without logging out (or their draft & cache will be gone) should be enough. If you want to be fancy, add an special option on top of account preferences that opens the `LoginActivity` with the instance prefilled. Hide the option from migrated users.

PeterCxy:
@connyduck This PR is now ready for review / merge

app/src/main/java/com/keylesspalace/tusky/components/login/LoginActivity.kt
```diff
@@ -228,7 +233,15 @@ class LoginActivity : BaseActivity(), Injectable {
             domain, clientId, clientSecret, oauthRedirectUri, code, "authorization_code"
         ).fold(
             { accessToken ->
-                accountManager.addAccount(accessToken.accessToken, domain)
+                if (intent.getIntExtra(LOGIN_MODE, MODE_DEFAULT) == MODE_MIGRATION) {
+                    accountManager.activeAccount?.let {
+                        it.accessToken = accessToken.accessToken
+                        it.oauthScopes = OAUTH_SCOPES
+                        accountManager.saveAccount(it)
```
connyduck, line 240: We cannot be sure the user logged in with the same account, they could have mulltiple accounts on the same instance so this is dangerous. Just using `accountManager.addAccount` should work though? `updateActiveAccount` will fix it if the account already exists

app/src/main/java/com/keylesspalace/tusky/receiver/NotificationBlockStateBroadcastReceiver.kt
```diff
@@ -0,0 +1,61 @@
...
+import javax.inject.Inject
+
+class NotificationBlockStateBroadcastReceiver : BroadcastReceiver() {
+    @Inject
+    lateinit var mastodonApi: MastodonApi
+
+    @Inject
+    lateinit var accountManager: AccountManager
+
+    override fun onReceive(context: Context?, intent: Intent?) {
```
connyduck, line 36: the params can both be non-null and you save the !!s below

app/src/main/java/com/keylesspalace/tusky/components/notifications/PushNotificationHelper.kt
```diff
@@ -0,0 +1,211 @@
...
+import com.keylesspalace.tusky.db.AccountEntity
+import com.keylesspalace.tusky.db.AccountManager
+import com.keylesspalace.tusky.entity.Notification
+import com.keylesspalace.tusky.network.MastodonApi
+import com.keylesspalace.tusky.util.CryptoUtil
+import io.reactivex.rxjava3.schedulers.Schedulers
+import org.unifiedpush.android.connector.UnifiedPush
+import retrofit2.HttpException
+
+object PushNotificationHelper {
```
connyduck, line 37: If it does not hold any state, there is no reason to create an object in Kotlin. Just make everything top level and annotate the file with `@file:JvmName("PushNotificationHelper")` to make it nice to call from Java.

app/src/main/java/com/keylesspalace/tusky/components/notifications/PushNotificationHelper.kt
```diff
@@ -0,0 +1,211 @@
...
+            show()
+        }
+    }
+
+    private fun showMigrationExplanationDialog(context: Context, accountManager: AccountManager) {
+        AlertDialog.Builder(context).apply {
+            if (currentAccountNeedsMigration(accountManager)) {
+                setMessage(R.string.dialog_push_notification_migration)
+                setPositiveButton(R.string.action_view_account_preferences) { _, _ ->
+                    PreferencesActivity.newIntent(context, PreferencesActivity.ACCOUNT_PREFERENCES).let {
```
connyduck, line 70: Why not send the user straight to login here? 🤔

app/src/main/java/com/keylesspalace/tusky/receiver/UnifiedPushBroadcastReceiver.kt
```diff
@@ -0,0 +1,72 @@
...
+    override fun onReceive(context: Context, intent: Intent) {
+        super.onReceive(context, intent)
+        AndroidInjection.inject(this, context)
+    }
+
+    override fun onMessage(context: Context, message: ByteArray, instance: String) {
+        AndroidInjection.inject(this, context)
+        Log.d(TAG, "New message for account $instance: $message")
+        val workManager = WorkManager.getInstance(context)
+        val request = OneTimeWorkRequest.from(com.keylesspalace.tusky.components.notifications.NotificationWorker::class.java)
```
connyduck, line 50: import NotificationWorker?

app/src/main/java/com/keylesspalace/tusky/components/login/LoginActivity.kt
```diff
@@ -260,19 +273,28 @@ class LoginActivity : BaseActivity(), Injectable {
...
         private const val CLIENT_SECRET = "clientSecret"
 
+        const val MODE_DEFAULT = 0
+        const val MODE_ADDITIONAL_LOGIN = 1
+        // "Migration" is used to update the OAuth scope granted to the client
+        const val MODE_MIGRATION = 2
+
         @JvmStatic
-        fun getIntent(context: Context, mode: Boolean): Intent {
+        fun getIntent(context: Context, mode: Int): Intent {
```
connyduck, line 297: default mode to MODE_DEFAULT so not all callsites have to be changed?

app/src/main/java/com/keylesspalace/tusky/entity/NotificationSubscribeResult.kt
```diff
@@ -0,0 +1,24 @@
+/* Copyright 2022 Conny Duck
```
connyduck, line 1:
```suggestion
/* Copyright 2022 Tusky Contributors
```

(or you name)

app/src/main/res/drawable/ic_logout_24dp.xml
```diff
@@ -0,0 +1,10 @@
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
+    android:width="24dp"
+    android:height="24dp"
+    android:viewportWidth="24"
+    android:viewportHeight="24"
+    android:tint="?attr/colorControlNormal">
+  <path
+      android:fillColor="@android:color/white"
+      android:pathData="M10.09,15.59L11.5,17l5,-5 -5,-5 -1.41,1.41L12.67,11H3v2h9.67l-2.58,2.59zM19,3H5c-1.11,0 -2,0.9 -2,2v4h2V5h14v14H5v-4H3v4c0,1.1 0.89,2 2,2h14c1.1,0 2,-0.9 2,-2V5c0,-1.1 -0.9,-2 -2,-2z"/>
+</vector>
```
connyduck, line 10: We already have the same drawable, can't we reuse it?

connyduck (changes requested):
Very nice job!

Which provider did you use to test this? I tried it with the [UP-FCM Distributor](https://unifiedpush.org/users/distributors/fcm/) and it was not as fast or reliable as I hoped.

As mentioned above, please don't add fresh RxJava code, we are trying to get rid of it entirely (if possible, I'd still merge this with RxJava)

PeterCxy:
@connyduck These days most people use the Ntfy distributor for UP. The FCM distributor could work but I have not tested it and nor do I plan to integrate it into this PR.

PeterCxy:
@connyduck I believe I have addressed most of the concerns in the previous review. Before this PR is merged, I will rebase and squash all the commits into one.

connyduck:
Looks pretty ok - I'd be very intereseted in @charlag's opinion as well

app/src/main/java/com/keylesspalace/tusky/components/notifications/PushNotificationHelper.kt
```diff
@@ -0,0 +1,216 @@
...
+import kotlinx.coroutines.withContext
+import org.unifiedpush.android.connector.UnifiedPush
+import retrofit2.HttpException
+
+private const val TAG = "PushNotificationHelper"
+
+private const val KEY_MIGRATION_NOTICE_DISMISSED = "migration_notice_dismissed"
+
+private fun anyAccountNeedsMigration(accountManager: AccountManager): Boolean =
+    accountManager.accounts.map(::accountNeedsMigration).any { it }
```
charlag, line 44: why `.map(...).any()` instead of just `.any()`

app/src/main/java/com/keylesspalace/tusky/components/notifications/PushNotificationHelper.kt
```diff
@@ -0,0 +1,216 @@
...
+    }
+}
+
+fun disableAllNotifications(context: Context, accountManager: AccountManager) {
+    disablePushNotifications(context, accountManager)
+    NotificationHelper.disablePullNotifications(context)
+}
+
+private fun buildSubscriptionData(context: Context, account: AccountEntity): Map<String, Boolean> =
+    HashMap<String, Boolean>().apply {
```
charlag, line 146:
```suggestion
    buildMap {
```

app/src/main/java/com/keylesspalace/tusky/components/notifications/PushNotificationHelper.kt
```diff
@@ -0,0 +1,216 @@
...
+        api.unsubscribePushNotifications("Bearer ${account.accessToken}", account.domain)
+            .onFailure {
+                Log.d(TAG, "Error unregistering push endpoint for account " + account.id)
+                Log.d(TAG, Log.getStackTraceString(it))
+                Log.d(TAG, (it as HttpException).response().toString())
+            }
+            .onSuccess {
+                Log.d(TAG, "UnifiedPush unregistration succeeded for account " + account.id)
+                // Clear the URL in database
+                account.unifiedPushUrl = ""
```
charlag, line 212: do we need to clear other parts?

app/src/main/java/com/keylesspalace/tusky/receiver/NotificationBlockStateBroadcastReceiver.kt
```diff
@@ -0,0 +1,67 @@
...
+            NotificationManager.ACTION_NOTIFICATION_CHANNEL_GROUP_BLOCK_STATE_CHANGED -> {
+                intent.getStringExtra(NotificationManager.EXTRA_NOTIFICATION_CHANNEL_GROUP_ID)
+            }
+            else -> null
+        } ?: return
+
+        accountManager.getAccountByIdentifier(gid)?.let { account ->
+            if (isUnifiedPushNotificationEnabledForAccount(account)) {
+                // Update UnifiedPush notification subscription
+                GlobalScope.launch { updateUnifiedPushSubscription(context, mastodonApi, accountManager, account) }
```
charlag, line 63: will the process live long enough for this? Do we need to hold something?

app/src/main/java/com/keylesspalace/tusky/receiver/UnifiedPushBroadcastReceiver.kt
```diff
@@ -0,0 +1,80 @@
...
+        val request = OneTimeWorkRequest.from(NotificationWorker::class.java)
+        workManager.enqueue(request)
+    }
+
+    override fun onNewEndpoint(context: Context, endpoint: String, instance: String) {
+        AndroidInjection.inject(this, context)
+        Log.d(TAG, "Endpoint available for account $instance: $endpoint")
+        accountManager.getAccountById(instance.toLong())?.let {
+            // Launch the coroutine in global scope -- it is short and we don't want to lose the registration event
+            // and there is no saner way to use structured concurrency in a receiver
```
charlag, line 65: we could probably just block here?

charlag (approved):
Looks interesting!
I'm not super happy that we have to include BC tho. Do we really need it? Is it for older versions? I would imagine it should be possible to generate EC key with plain Java (I don't remember which EC keys are compatible with what)

app/src/main/java/com/keylesspalace/tusky/components/notifications/PushNotificationHelper.kt
PeterCxy: We don't really need to -- but just to be safe I changed it to clear everything

PeterCxy:
@charlag Regarding the GlobalScope usage in BroadcastReceiver: the process should be alive when it receives the event because push registration happens when the app UI is launched, and it happens quick enough so that I don't think the process being killed is a problem. However, blocking would result in a noticeable lag in the UI because the receiver runs on the UI thread. But even if the process is killed (because the user somehow closed the UI quick enough), it still does not really matter as the registration will be triggered again at the next launch anyway (we only write to the database when we are sure the notifications are set up properly).

Regarding BouncyCastle, unfortunately I'm not very familiar with the Java crypto API and have no idea how to implement this with plain JDK crypto.

@connyduck I've rebased my changes again to the tip of develop, but commit b4eda5ea65e71419233b6a350ade04baf4c95e1f introduced another db version 35. Since it is not released yet, I just merged both together. However that commit did not generate a JSON schema file -- I included that in mine.

charlag:
@PeterCxy I didn't realize that the event is received together with the app launch, I have no concerns then.

About key generation: I quickly searched and found this example (and others):

https://github.com/JZ-Darkal/AndroidHttpCapture/blob/9afc8805e8210df5d72e9718a745d802c59126f0/app/src/main/java/net/lightbody/bmp/mitm/keys/ECKeyGenerator.java#L37-L49

I think we just need to check that it works on our minimum version. I don't want to be the pain in the butt here but I think avoiding including something like BC is worth it.

charlag:
Related example from Android docs:
https://developer.android.com/reference/android/security/keystore/KeyGenParameterSpec#example:-nist-p-256-ec-key-pair-for-signingverification-using-ecdsa

PeterCxy:
@charlag I'd prefer not to use the Android keystore one as it can make backup & restore very painful. However I'll check if the plain JDK implementation exists on the current minSDK.

charlag:
@PeterCxy you don't need (and shouldn't really) use keychain for this, the relevant part of the interface is the same, all I wanted to show.

PeterCxy:
@charlag I've tried using the plain Java JCE APIs without BC but unfortunately there are a few obstacles. The first one is JCE is missing the same encoded form for the plain ECPoints which is needed by WebPush. This can be fixed by simply re-implementing it, which is fine. However, the Java-generated prime256v1 (aka secp256r1 in JCE) keys seem to sometimes be longer than 32 bytes, and that is incompatible with WebPush. I'm not sure why this happens yet...

PeterCxy:
@charlag Turns out WebPush requires the public key to be encoded in the X9.62 uncompressed point form, which is unimplemented in plain Java JCE (nor is it implemented in Android keystore). What's worse is that we cannot seem to even control whether the ECPoint objects are represented in their compressed forms or not. To avoid BouncyCastle, we would have to re-implement both X9.62 and the conversion between compressed / uncompressed representation by hand, which I think is a worse idea than including BouncyCastle.

charlag:
@PeterCxy thanks for checking it. I have a feeling I run into it sometime before. I think it's fine to use BC then, at least there's a reason to do that

PeterCxy:
@charlag Thank you for your review.

@connyduck I believe all of the concerns in the reviews have been addressed. Do you think this is ready for merge now? If so I will just go ahead and squash everything for you to merge.

connyduck (approved):

PeterCxy:
@connyduck Rebased and squashed.

PeterCxy:
@connyduck Wait actually, we need a database version 36 for this as I just realized you made a new release before this. I'll do this real quick (BTW: the database schema version 35 still seems to be missing from the develop branch?)

PeterCxy:
Fixed. Now using database version 36.

connyduck:
@PeterCxy thank you! I added 35.json to develop.
Yes, using db version 36 is correct, even if there is no release in between you can never add to prevoius migrations because each commit to develop is published as nightly release.

PR #2304, opened 2022-01-25
Fix some weird behavior when clicking links in statuses

connyduck:
There is some weird behavior in some edgecases:

1) Apparently there are statuses that contain links to themselves, so one can open the detail view in an infinite recursion. Example: https://anonsys.net/display/bf69967c-1961-efc6-22a9-867784751114
2) Pixelfed links are opened in the browser. Example: https://mastodon.social/@ConnyDuck/107684681310479962
3) Clicking on a card will always open the browser, even when the link could be opened in Tusky

This PR fixes those problems and cleans `BottomSheetActivityTests` up a bit

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/BottomSheetActivity.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/BottomSheetActivity.kt
@@ -180,6 +180,8 @@ abstract class BottomSheetActivity : BaseActivity() {
 // https://friendica.foo.bar/profile/user
 // https://friendica.foo.bar/display/d4643c42-3ae0-4b73-b8b0-c725f5819207
 // https://misskey.foo.bar/notes/83w6r388br (always lowercase)
+// https://pixelfed.social/p/connyduck/391263492998670833
+// https://pixelfed.social/connyduck
 fun looksLikeMastodonUrl(urlString: String): Boolean {
     val uri: URI
     try {
@@ -203,7 +205,9 @@ fun looksLikeMastodonUrl(urlString: String): Boolean {
         path.matches("^/objects/[-a-f0-9]+$".toRegex()) ||
         path.matches("^/notes/[a-z0-9]+$".toRegex()) ||
         path.matches("^/display/[-a-f0-9]+$".toRegex()) ||
-        path.matches("^/profile/\\w+$".toRegex())
+        path.matches("^/profile/\\w+$".toRegex()) ||
+        path.matches("^/p/\\w+/\\d+$".toRegex()) ||
+        path.matches("^/\\w+$".toRegex())
 }
 
 enum class PostLookupFallbackBehavior {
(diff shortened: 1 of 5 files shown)


connyduck:
Wait there is a better way to solve the infinite recursion. Apparently this happens when Friendica federates just a preview of a post. Instead of doing nothing when the link is clicked we should open the browser so the user can read the whole post.

app/src/main/java/com/keylesspalace/tusky/BottomSheetActivity.kt
```diff
@@ -203,7 +205,9 @@ fun looksLikeMastodonUrl(urlString: String): Boolean {
         path.matches("^/objects/[-a-f0-9]+$".toRegex()) ||
         path.matches("^/notes/[a-z0-9]+$".toRegex()) ||
         path.matches("^/display/[-a-f0-9]+$".toRegex()) ||
-        path.matches("^/profile/\\w+$".toRegex())
+        path.matches("^/profile/\\w+$".toRegex()) ||
+        path.matches("^/p/\\w+/\\d+$".toRegex()) ||
+        path.matches("^/\\w+$".toRegex())
```
Tak, line 210: This is getting pretty close to "just try opening all links in Tusky first" :grinning:

app/src/main/java/com/keylesspalace/tusky/adapter/StatusBaseViewHolder.java
```diff
@@ -1034,7 +1034,12 @@ private CharSequence getPollInfoText(long timestamp, PollViewData poll,
         return pollDescription.getContext().getString(R.string.poll_info_format, votesText, pollDurationInfo);
     }
 
-    protected void setupCard(StatusViewData.Concrete status, CardViewMode cardViewMode, StatusDisplayOptions statusDisplayOptions) {
+    protected void setupCard(
+            StatusViewData.Concrete status,
+            CardViewMode cardViewMode,
+            StatusDisplayOptions statusDisplayOptions,
+            final StatusActionListener listener
+    ) {
```
Tak, line 1042: Oh, nice find

Tak (approved):

app/src/main/java/com/keylesspalace/tusky/BottomSheetActivity.kt
connyduck: It will indeed led to more false positives, but there are still a lot of links that are not matched, like ones with queries, multiple path segments, special characters, etc

Tak: I was only half-joking - maybe it would be cleaner to actually do this (guarded by a preference?)

charlag (approved):

app/src/main/java/com/keylesspalace/tusky/BottomSheetActivity.kt
connyduck: This is one of the things that should just work, without users thinking about it (and I don't know how to name that preference)

PR #2305, opened 2022-01-27
fix placeholder not disappearing when no statuses are loaded

connyduck:
This fixes the problem at one end, I am still thinking if I could somehow just detect that a placeholder will not load anything and not even insert it.


connyduck:
We could replace one loaded status with a placeholder instead of additionally adding a placeholder, this way it would be almost guaranteed that every placeholder loads at least one status 🤔

charlag (approved):

PR #2309, opened 2022-02-02
Fix disappearing placeholders

connyduck:
A disappearing placeholder could be reproduced like this on the home timeline:

Load a lot of new statuses so you get a full page (30 statuses) and a placeholder displayed.
One or more of the just loaded statuses gets deleted.
Refresh the timeline -> the deleted status disappears, but so does the placeholder

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/components/timeline/viewmodel/CachedTimelineRemoteMediator.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/components/timeline/viewmodel/CachedTimelineRemoteMediator.kt
@@ -53,11 +53,19 @@ class CachedTimelineRemoteMediator(
 
         try {
             var dbEmpty = false
+
+            val topPlaceholderId = if (loadType == LoadType.REFRESH) {
+                timelineDao.getTopPlaceholderId(activeAccount.id)
+            } else {
+                null  // don't execute the query if it is not needed
+            }
+
             if (!initialRefresh && loadType == LoadType.REFRESH) {
                 val topId = timelineDao.getTopId(activeAccount.id)
                 topId?.let { cachedTopId ->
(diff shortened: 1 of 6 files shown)


app/src/main/java/com/keylesspalace/tusky/components/timeline/viewmodel/CachedTimelineRemoteMediator.kt
```diff
@@ -53,11 +53,19 @@ class CachedTimelineRemoteMediator(
...
+            } else {
+                null // don't execute the query if it is not needed
+            }
+
             if (!initialRefresh && loadType == LoadType.REFRESH) {
                 val topId = timelineDao.getTopId(activeAccount.id)
                 topId?.let { cachedTopId ->
                     val statusResponse = api.homeTimeline(
                         maxId = cachedTopId,
+                        sinceId = topPlaceholderId, // so already existing placeholders don't get accidentally overwritten
```
charlag, line 68: So it will load up from which point? Sorry I'm a bit lost. As far as I understand since topmost placeholder. It might be confusing if it's near the top I guess because we might refresh only the small part but at your discretion.

connyduck: Yes, it will only refresh above the topmost placeholder. This will almost always be a full page, because for less statuses to be loaded one must have been deleted. Yes it may be weird in case a lot of statuses have been deleted but I think that is such a rare case that it is acceptable.

charlag: You are right, this should be very rare.

PR #2311, opened 2022-02-03
Fix announcement reactions

### LINKED ISSUE #2308: Announcement reactions for custom emoji are oversized
Announcement reaction buttons for custom emoji are wider than regular emoji. Looks to me as if they're being sized according to the shortcode.

Pictured below: :chitter:, unicode bee, and :BigBlobhajHug: 

the first one is a little extra wide and the third one is very extra wide

* * * *
- Tusky Version: 15.1
- Android Version: 11
- Android Device: Poco X3

connyduck:
- closes #2308 
- fixes an issue that causes announcement reactions to not properly update when clicked

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/components/announcements/AnnouncementAdapter.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/components/announcements/AnnouncementAdapter.kt
@@ -31,6 +31,7 @@ import com.keylesspalace.tusky.interfaces.LinkListener
 import com.keylesspalace.tusky.util.BindingHolder
 import com.keylesspalace.tusky.util.EmojiSpan
 import com.keylesspalace.tusky.util.LinkHelper
+import com.keylesspalace.tusky.util.emojify
 import java.lang.ref.WeakReference
 
 interface AnnouncementActionListener : LinkListener {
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/keylesspalace/tusky/components/announcements/AnnouncementAdapter.kt
```diff
@@ -69,42 +72,38 @@ class AnnouncementAdapter(
...
+                        isCheckable = true
+                        checkedIcon = null
+                        chips.addView(this, i)
+                    }
+                )
+                .apply {
+                    if (reaction.url == null) {
+                        this.text = "${reaction.name} ${reaction.count}"
+                    } else {
+                        val spanBuilder = SpannableStringBuilder("E ${reaction.count}")
```
charlag, line 87: I got confused by this `E` for quite a bit but then I realized it's for emoji. Why can't we put the emoji itself there? Do we know the name or shortcode? This might be better for the screen reader

connyduck: Yes but putting the name there is what causes the sizing bug :/
Spans have a content description, maybe that's where the name should go? No idea what a screenreader does with it though.

charlag: I think putting it on a span wouldn't hurt. Putting something that is usually not pronounced (like fixed-width space) instead of E could be good as well. Also a comment why we do any of this, otherwise next clueless person who touches this will break it.

connyduck:
Meh, span content descriptions are API 30+ only. Still, better than before.

Gipsysoftware:
I hope you will do the update is that the problem of playing the video is solved

charlag (approved):

PR #2312, opened 2022-02-03
fix apng emojis not rendered when animation is turned off

connyduck:
After #2244, animated emojis were not rendered at all when the animation setting was turned off. This fixes that problem.


Tak (approved):

charlag:
I don't get why they would not be rendered at all before?

connyduck:
I think it is because it is still decoded as an animated drawable but we don't call `start` on it. Another solution would be decoding it as Bitmap like we do in other places.

charlag:
Ah I get it now, we do it elsewhere, I think it's fine

charlag (approved):

Issue #2306, opened 2022-01-28
Inconsistent star-state in notification


garritfra:
### Reproduction

1. Open a reply and star it
2. View reply in notifications
3. Star is not visible

### Context

App-Version: Nightly

![Screenshot_20220128-155404.jpg](https://user-images.githubusercontent.com/32395585/151570000-5db0defd-d8d9-48ba-b78e-d65cf2993073.jpg)
![Screenshot_20220128-155410.jpg](https://user-images.githubusercontent.com/32395585/151570012-bc0df33e-e36b-4472-97b6-8fd58fd37a41.jpg)

PR #2318, opened 2022-02-07
Fix comparing statuses and diffing in NotificationsFragment

connyduck:
I don't remember why equals/hashcode was overwritten in `Status` 🤔 

Anyway, _something_ in #2238 broke the diffing in NotificationsFragment, but I don't yet know what. Actually it is weird that it works in other places.

closes https://github.com/tuskyapp/Tusky/issues/2306


charlag (approved):
I think it was overriden long ago? Probably before converting anything to Kotlin even. I don't know if it's acceptable performance-wise to do full comparison but should be okay, it's in background anyway

Issue #995, opened 2019-01-17
Media Upload Issues


gudenau:
When I attempt to upload pictures from my cell phone to https://mastodon.technology they almost always fail. No error other than "media upload failed".

* * * *
- Tusky Version: 4.1
- Android Version: 9.0
- Android Device: Pixel XL
- Mastodon instance (if applicable): https://mastodon.technology

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

connyduck:
Yes, there seem to be problems with our upload code. I will add better error messages so we can finally nail them down.

nevillepark:
I'm also experiencing this. 

_______________

- Tusky version: 4.1
- OS: LineageOS 15.1 (Android 8.1)
- Device: Moto G5 Plus
- Mastodon instance: social.coop

feld:
> almost always fail

 Are the files that work PNG, GIF, or videos -- something that "send reduced" wouldn't touch?

mihudevcom:
@gudenau Can you try the latest version? Seems to be working fine to me.

I've had a similar issue on other instances but it was because of the instance itself, not the app.

mihudevcom:
@nevillepark Can you try the latest version?
@gudenau @nevillepark For both of you: when you experience trouble uploading images try the instance in your browser too. If the issue happens on both then it's a problem with the instance. If it happens only on the app then it might be a problem with the app.

nevillepark:
@mihudevcom It only happens with the app.

wjt:
On a Pixel 3 running Tusky 5.1, with an account on mastodon.social, some (but not all) image uploads fail with “The upload failed” after a second. It seems that it's only screenshots (PNG) which I have cropped with the Pixel's built-in screenshot editor: full-screen screenshots (also PNG), new photographs, and existing photographs (JPEG) succeed.

freyja-wildes:
Hello there,
I encounter issues too. On a private instance (only a few accounts) image upload on toot creation is inconsistent.
Just now, I tried to upload twice it failed.
Looked around in options, wrote a toot. Then I tried again, with the same image, and it worked.

Tusky version: 12
Android version: 9
Android device: Asus Zenfone 5

l1ving:
Can reproduce, Pixel 2, Android 11, Tusky 15.1

Video uploads fail 3/3 tries, photo uploads work but still randomly fail

tokenwizard:
[...]

PR #2327, opened 2022-02-10
use POST api/v2/media to upload media

connyduck:
Why didn't we do this sooner 😲 
v2 is async and should be much faster.
Hopefully fixes stuff like #995 
I think v2 is only available since Mastodon 3.1.3


Tak:
> I think v2 is only available since Mastodon 3.1.3

I guess that's why :grinning: 
Should there be a fallback for people using older versions? Do forks/misskey/goto/pleroma support this api yet?

connyduck:
It is supported by Pleroma and Mastodon 3.1.3 was released almost 2 years ago, I think it is fine.

PR #2332, opened 2022-02-11
merge ModalTimelineActivity & ViewTagActivity into StatusListActivity

connyduck:
Also fixes how StatusListActivity handles Fragment creation (before it was unconditionally creating the TimelineFragment, which caused two Fragments being created on screen rotation)


charlag (approved):
Didn't test but looks good, good to get rid of redundant buggy activities

PR #2334, opened 2022-02-11
Don't use notifyDataSetChanged in TimelineFragment

connyduck:
It is super inefficient.


charlag (approved):
It is, it probably should be redone with differ or something along those lines

connyduck:
I could also submit the latest `PagingData` to the differ again, but it is a little bit more involved since `Flow.last()` is suspending 🤔

charlag:
@connyduck probably not worth it yet

PR #2336, opened 2022-02-12
use ViewData only instead of Pair<Status, ViewData> in Search

connyduck:
Status is included in StatusViewData.Concrete. It makes no sense to have an additional reference to Status.


charlag (approved):
Just need to make sure it's an up-to-date but I think we are good

PR #2339, opened 2022-02-14
fix some mentions not being opened in Tusky

connyduck:
I don't really know what is going on here, maybe Mastodon changed something in how they render Mentions or it never worked.

Anyway, the thing is that fully qualified mentions (e.g. `@connyduck@chaos.social`) would not open directly, but were looked up via the search first (that is why we only noticed when somebody made such a mention that failed the `looksLikeMastodonUrl` in `BottonSheetActivity`).

Thx for reporting, @mal0ki

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/util/LinkHelper.java
+++ b/app/src/main/java/com/keylesspalace/tusky/util/LinkHelper.java
@@ -87,17 +87,11 @@ public class LinkHelper {
                     public void onClick(@NonNull View widget) { listener.onViewTag(tag); }
                 };
             } else if (text.charAt(0) == '@' && mentions != null && mentions.size() > 0) {
-                String accountUsername = text.subSequence(1, text.length()).toString();
-                /* There may be multiple matches for users on different instances with the same
-                 * username. If a match has the same domain we know it's for sure the same, but if
-                 * that can't be found then just go with whichever one matched last. */
                 String id = null;
                 for (Status.Mention mention : mentions) {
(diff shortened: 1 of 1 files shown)


charlag (approved):
I am a bit confused but seems like how it is now (before your change) was just a hack for a hack. I've found commit b6e72a94be90ae7bffddf4ce06ea135130118871 which introduced LinkHelper and it was doing simply

```java
for (Status.Mention mention : mentions) {
    if (mention.localUsername.equals(accountUsername)) {
        id = mention.id;
    }
}
```

which was fixed later by you in 41233a837b1b970370c5629e4b7c3f2732794ffd

Oddly enough `url` field on `Mention` was always there but wasn't used for this purpose. I wonder if can different from the HTML url in some cases.

I think I am for trying out your fix and seeing if it breaks anything. In any case, please include as much info as possible in the commit (like the text in PR is a good candidate) or at least a short summary.

connyduck:
Thx for investigating. I think when the code was originally written there was not much thought put into it, and later when I fixed the other problem I assumed the general idea is fine and added to it.

I just checked how Mastodon does it in the frontend and it they also compare by url https://github.com/mastodon/mastodon/blob/52e5c07948c4c91b73062846e1f19ea278ec0e24/app/javascript/mastodon/components/status_content.js#L49

>  In any case, please include as much info as possible in the commit (like the text in PR is a good candidate) or at least a short summary.

I'll link this PR

charlag:
@connyduck thanks for checking masto code as well, it should be very safe then

PR #2344, opened 2022-02-18
Use tags from status when adding handlers to hashtag spans in status content

### LINKED ISSUE #2283: Hyperlinks beginning with # are treated as hashtags
Consider this post: <https://likeable.space/notice/AEumkGbxcWAwPmyg7c>. It has a hyperlink to a matrix room and one to an IRC channel in it. Both begin with a `#`. If I click on them in Tusky, it thinks they are hashtags. 

In Pleroma they behave like normal hyperlinks and in Mastodon the Matrix-link works like a hashtag when I left-click but opens the Matrix-URL when I open it in a new tab.

* * * *
- Tusky Version: 15.1
- Android Version: 11 (LineageOS 18.1)
- Android Device: Xiaomi Mi 8
- Mastodon instance (if applicable): likeable.space

Tak:
Fixes #2283


app/src/main/java/com/keylesspalace/tusky/util/LinkHelper.kt
```diff
@@ -0,0 +1,251 @@
...
+ * @param mentions any '@' mentions which are known to be in the content
+ * @param listener to notify about particular spans that are clicked
+ */
+fun setClickableText(view: TextView, content: CharSequence, mentions: List<Mention>?, tags: List<HashTag>?, listener: LinkListener) {
+    val builder = SpannableStringBuilder.valueOf(content)
+    val urlSpans = builder.getSpans(0, content.length, URLSpan::class.java)
+
+    for (span in urlSpans) {
+        setClickableText(span, builder, mentions, tags, listener)
+    }
```
connyduck, line 72:
```suggestion
    builder.getSpans(0, content.length, URLSpan::class.java)
        .forEach { span ->
            setClickableText(span, builder, mentions, tags, listener)
        }
```

app/src/main/java/com/keylesspalace/tusky/util/LinkHelper.kt
```diff
@@ -0,0 +1,251 @@
...
+
+private fun getCustomSpanForTag(tags: List<HashTag>?, span: URLSpan, listener: LinkListener): ClickableSpan? {
+    return when (val tag = tags?.firstOrNull { it.url == span.url }) {
+        null -> null
+        else -> object : NoUnderlineURLSpan(span.url) {
+            override fun onClick(view: View) {
+                listener.onViewTag(tag.name)
+            }
+        }
+    }
```
connyduck, line 119:
```suggestion
    return tags?.firstOrNull { it.url == span.url }?.let { tag ->
        object : NoUnderlineURLSpan(span.url) {
                        override fun onClick(view: View) = listener.onViewTag(tag.name)
        }
    }
```

(similar in other places)

app/src/main/java/com/keylesspalace/tusky/util/LinkHelper.kt
```diff
@@ -0,0 +1,251 @@
...
+ * @param mentions any '@' mentions which are known to be in the content
+ * @param listener to notify about particular spans that are clicked
+ */
+fun setClickableMentions(view: TextView, mentions: List<Mention>?, listener: LinkListener) {
+    if (mentions?.isEmpty() != false) {
+        view.text = null
+        return
+    }
+
+    val builder = SpannableStringBuilder()
```
connyduck, line 152: There is a `buildSpannedString` top level function that would make this function a little bit easier to write https://developer.android.com/reference/kotlin/androidx/core/text/package-summary#buildSpannedString(kotlin.Function1)

It kinda sucks that it converts the result to `SpannedString` which is an additional object allocation we don't need though

app/src/main/java/com/keylesspalace/tusky/util/LinkHelper.kt
```diff
@@ -0,0 +1,251 @@
...
+}
+
+@VisibleForTesting
+fun setClickableText(
+    span: URLSpan,
+    builder: SpannableStringBuilder,
+    mentions: List<Mention>?,
+    tags: List<HashTag>?,
+    listener: LinkListener
+) {
```
connyduck, line 84: `) = builder.apply {` maybe? Would get rid of the repeated `builder.`

app/src/main/java/com/keylesspalace/tusky/util/LinkHelper.kt
```diff
@@ -0,0 +1,251 @@
...
+    val start = builder.getSpanStart(span)
+    val end = builder.getSpanEnd(span)
+    val flags = builder.getSpanFlags(span)
+    val text = builder.subSequence(start, end)
+
+    val customSpan = when {
+        text[0] == '#' -> getCustomSpanForTag(tags, span, listener)
+        text[0] == '@' -> getCustomSpanForMention(mentions, span, listener)
```
connyduck, line 92:
```suggestion
    val customSpan = when(text[0]) {
        '#' -> getCustomSpanForTag(tags, span, listener)
        '@' -> getCustomSpanForMention(mentions, span, listener)
```

connyduck (changes requested):
😍

app/src/main/java/com/keylesspalace/tusky/util/LinkHelper.kt
```diff
@@ -0,0 +1,251 @@
...
+import com.keylesspalace.tusky.R
+import com.keylesspalace.tusky.entity.HashTag
+import com.keylesspalace.tusky.entity.Status.Mention
+import com.keylesspalace.tusky.interfaces.LinkListener
+import java.net.URI
+import java.net.URISyntaxException
+
+fun getDomain(urlString: String?): String {
+    val uri: URI = try {
+        URI(urlString)
```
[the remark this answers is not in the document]
charlag: I would rather use java.net.URI in generic code

app/src/main/java/com/keylesspalace/tusky/util/LinkHelper.kt
```diff
@@ -0,0 +1,231 @@
...
+    return SpannableStringBuilder(text).apply {
+        setSpan(NoUnderlineURLSpan(link), 0, text.length, Spanned.SPAN_INCLUSIVE_EXCLUSIVE)
+    }
+}
+
+/**
+ * Opens a link, depending on the settings, either in the browser or in a custom tab
+ *
+ * @param url a string containing the url to open
+ * @param context context
```
connyduck, line 174:
```suggestion
 * @receiver the Context to open the link from
```

connyduck (changes requested):
Thanks for the changes!

I think I the non-null `tags: List<HashTag>` has to be changed again though.
Hashtags in profile bios no longer open in Tusky.

The right behavior probably is
- We have a tag list -> only open tags in Tusky that match (=what you added with this PR)
- We have no tag list (it is null) -> open all tags within Tusky (what we did before this PR)

connyduck (approved):

PR #2348, opened 2022-02-19
fix cache cleanup deleting more statuses than it should

connyduck:


app/src/main/java/com/keylesspalace/tusky/components/timeline/viewmodel/CachedTimelineViewModel.kt
```diff
@@ -207,4 +220,8 @@ class CachedTimelineViewModel @Inject constructor(
             }
         }
     }
+
+    companion object {
+        private const val MAX_STATUSES_IN_CACHE = 1000
```
connyduck, line 225: Not sure what a good number is here (this is per logged in user)

app/src/main/java/com/keylesspalace/tusky/components/timeline/viewmodel/CachedTimelineViewModel.kt
```diff
@@ -81,6 +84,16 @@ class CachedTimelineViewModel @Inject constructor(
         }
         .cachedIn(viewModelScope)
 
+    init {
+        viewModelScope.launch {
+            delay(5.toDuration(DurationUnit.SECONDS)) // delay so the db is not locked during initial
```
connyduck, line 89: So the cleanup will run on every account switch in Tusky, maybe that is too often? But it won't have to clean that much when it runs often

charlag: I don't know if it's good that viewModel does such cleanup tasks but otherwise I think it's totally fine to do it every time

charlag (approved):
I think some tests for queries might be good otherwise it should be fine. It doesn't have millions of statuses to run over anyway.

In the long term I think we should consider splitting up databases per logged in account?

PR #2351, opened 2022-02-23
update Android Image Cropper and get rid of deprecated onActivityResult

connyduck:
Turns out that
- the cropping library can handle both picking & resizing for us
- no permission is necessary since the picking is done by a system app

200 lines of code less for us 😍

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/EditProfileActivity.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/EditProfileActivity.kt
@@ -15,28 +15,26 @@
 
 package com.keylesspalace.tusky
 
-import android.Manifest
-import android.app.Activity
 import android.content.Intent
-import android.content.pm.PackageManager
 import android.graphics.Bitmap
 import android.graphics.Color
 import android.net.Uri
 import android.os.Bundle
+import android.util.Log
 import android.view.Menu
 import android.view.MenuItem
 import android.view.View
 import android.widget.ImageView
 import androidx.activity.viewModels
-import androidx.core.app.ActivityCompat
-import androidx.core.content.ContextCompat
 import androidx.core.view.isVisible
 import androidx.lifecycle.LiveData
 import androidx.recyclerview.widget.LinearLayoutManager
 import com.bumptech.glide.Glide
+import com.bumptech.glide.load.engine.DiskCacheStrategy
 import com.bumptech.glide.load.resource.bitmap.FitCenter
 import com.bumptech.glide.load.resource.bitmap.RoundedCorners
-import com.canhub.cropper.CropImage
+import com.canhub.cropper.CropImageContract
+import com.canhub.cropper.options
 import com.google.android.material.snackbar.Snackbar
 import com.keylesspalace.tusky.adapter.AccountFieldEditAdapter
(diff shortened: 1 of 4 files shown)


ThatOneCalculator (approved):

app/src/main/java/com/keylesspalace/tusky/EditProfileActivity.kt
```diff
@@ -218,90 +208,58 @@ class EditProfileActivity : BaseActivity(), Injectable {
...
-                            it.consumed = true
-                        }
-                    }
-                }
+            this
+        ) { imageUri ->
+            val glide = Glide.with(imageView)
+                .load(imageUri)
+                .skipMemoryCache(true)
+                .diskCacheStrategy(DiskCacheStrategy.NONE)
```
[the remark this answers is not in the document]
charlag: This should definitely be a comment, it also applies for memory cache

app/src/main/java/com/keylesspalace/tusky/viewmodel/EditProfileViewModel.kt
```diff
@@ -92,62 +79,20 @@ class EditProfileViewModel @Inject constructor(
...
+                .addTo(disposables)
         }
     }
 
-    fun newAvatar(uri: Uri, context: Context) {
-        val cacheFile = getCacheFileForName(context, AVATAR_FILE_NAME)
-
-        resizeImage(uri, context, AVATAR_SIZE, AVATAR_SIZE, cacheFile, avatarData)
-    }
+    fun getAvatarUri(context: Context) = getCacheFileForName(context, AVATAR_FILE_NAME).toUri()
```
charlag, line 86: We did this before but it's a bit eh that we pass context here, especially because any context will do here, including global

charlag (approved):
From a glance looks good. Always nice to delete some code.

charlag:
btw, do we still need storage permissions in general?

app/src/main/java/com/keylesspalace/tusky/viewmodel/EditProfileViewModel.kt
connyduck: Yes I know, but injecting context into viewmodel gives a warning because of leaked context 🙄

connyduck:
> btw, do we still need storage permissions in general?

Probably not 🧐

app/src/main/java/com/keylesspalace/tusky/viewmodel/EditProfileViewModel.kt
charlag: AndroidViewModel but it has constructor restrictions
https://developer.android.com/reference/androidx/lifecycle/AndroidViewModel

connyduck: well, why not just inject an Application instead of a context into the regular Viewmodel

connyduck: looks good

charlag: Yup, I think that's better than risking to leak Activity

PR #2353, opened 2022-02-25
Improve timeline dao

connyduck:
This doesn't fix anything, just makes the code a little bit nicer than before

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/db/TimelineDao.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/db/TimelineDao.kt
@@ -20,6 +20,7 @@ import androidx.room.Dao
 import androidx.room.Insert
 import androidx.room.OnConflictStrategy.REPLACE
 import androidx.room.Query
+import androidx.room.Transaction
 
 @Dao
 abstract class TimelineDao {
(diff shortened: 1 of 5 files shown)


app/src/main/java/com/keylesspalace/tusky/db/TimelineDao.kt
```diff
@@ -86,18 +87,39 @@ WHERE timelineUserId = :accountId AND (serverId = :statusId OR reblogServerId =
     )
     abstract fun removeAllByUser(accountId: Long, userId: String)
 
+    /**
+     * Removes everything in the TimelineStatusEntity and TimelineAccountEntity tables for one user account
+     * @param accountId id of the account for which to clean tables
+     */
+    @Transaction
```
charlag, line 94: We've been doing this before but now I realize that this transaction is *huge* and we don't even really care about it

app/src/main/java/com/keylesspalace/tusky/db/TimelineDao.kt
```diff
@@ -86,18 +87,39 @@ WHERE timelineUserId = :accountId AND (serverId = :statusId OR reblogServerId =
...
 AND serverId = :statusId"""
     )
     abstract fun delete(accountId: Long, statusId: String)
 
+    /**
+     * Cleans the TimelineStatusEntity and TimelineAccountEntity tables from old entries.
+     * @param accountId id of the account for which to clean tables
+     * @param limit how many statuses to keep
+     */
+    @Transaction
```
charlag, line 117: Same goes for this transaction. I don't think we really need transaction guarantees here

charlag (approved):

PR #2354, opened 2022-02-26
fix database migration to v29

connyduck:
cc @Tak

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/entity/Status.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/entity/Status.kt
@@ -42,7 +42,7 @@ data class Status(
     val visibility: Visibility,
     @SerializedName("media_attachments") var attachments: ArrayList<Attachment>,
     val mentions: List<Mention>,
-    val tags: List<HashTag>,
+    val tags: List<HashTag>?,
     val application: Application?,
     val pinned: Boolean?,
     val muted: Boolean?,
--- a/app/schemas/com.keylesspalace.tusky.db.AppDatabase/29.json
+++ b/app/schemas/com.keylesspalace.tusky.db.AppDatabase/29.json
@@ -2,7 +2,7 @@
   "formatVersion": 1,
   "database": {
     "version": 29,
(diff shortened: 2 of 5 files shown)


app/src/main/java/com/keylesspalace/tusky/entity/Status.kt
```diff
@@ -42,7 +42,7 @@ data class Status(
     val visibility: Visibility,
     @SerializedName("media_attachments") var attachments: ArrayList<Attachment>,
     val mentions: List<Mention>,
-    val tags: List<HashTag>,
+    val tags: List<HashTag>?,
```
connyduck, line 45: I think it is better to have the tags nullable, because there will already be statuses in the database and if we assume they have no tags they will not open in Tusky.
(empty List vs null, see here https://github.com/tuskyapp/Tusky/blob/develop/app/src/main/java/com/keylesspalace/tusky/util/LinkHelper.kt#L103-L104 )

connyduck:
Ah, and here is the stacktrace , thx @nailyk-fr for reporting

```
java.lang.RuntimeException: Unable to create application com.keylesspalace.tusky.TuskyApplication: android.database.sqlite.SQLiteException: Cannot add a NOT NULL column with default value NULL (code 1 SQLITE_ERROR): , while compiling: ALTER TABLE `ConversationEntity` ADD COLUMN `s_tags` TEXT NOT NULL
at android.app.ActivityThread.handleBindApplication(ActivityThread.java:6724)
at android.app.ActivityThread.access$1300(ActivityThread.java:237)
at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1913)
at android.os.Handler.dispatchMessage(Handler.java:106)
at android.os.Looper.loop(Looper.java:223)
at android.app.ActivityThread.main(ActivityThread.java:7664)
at java.lang.reflect.Method.invoke(Native Method)
at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run](RuntimeInit.java:592)
at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:947)
```

Tak (approved):

PR #2355, opened 2022-02-26
fix String.inc() and String.dec() not being inverse operations

connyduck:
Problem: `String.inc()` and `String.dec()` were not inverse in all cases, so a placeholder id created by status id dec() could not be turned into the status id again with inc(), which caused the timeline to miss statuses.

Example:

`"1230".dec() = "122z"`
`"122z".inc() = "122{"`
`api/v1/timelines/home?max_id=122{` returns empty array.

This PR fixes the problem by making sure `"AnyAlphanumericString".dec().inc() == "AnyAlphanumericString"`.

(And I also improved the TimelineDaoTest while looking for the bug)

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/components/timeline/viewmodel/CachedTimelineViewModel.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/components/timeline/viewmodel/CachedTimelineViewModel.kt
@@ -143,7 +143,7 @@ class CachedTimelineViewModel @Inject constructor(
 
                 val nextPlaceholderId = timelineDao.getNextPlaceholderIdAfter(activeAccount.id, placeholderId)
 
-                val response = api.homeTimeline(maxId = placeholderId.inc(), sinceId = nextPlaceholderId, limit = 20).await()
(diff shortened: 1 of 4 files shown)


Tak (approved):

Tak (approved):

Tak:
I've been running this for a few days - I haven't *observed* any gaps in that time

app/src/main/java/com/keylesspalace/tusky/components/timeline/viewmodel/CachedTimelineViewModel.kt
```diff
@@ -143,7 +143,7 @@ class CachedTimelineViewModel @Inject constructor(
 
                 val nextPlaceholderId = timelineDao.getNextPlaceholderIdAfter(activeAccount.id, placeholderId)
 
-                val response = api.homeTimeline(maxId = placeholderId.inc(), sinceId = nextPlaceholderId, limit = 20).await()
+                val response = api.homeTimeline(maxId = placeholderId.inc(), sinceId = nextPlaceholderId, limit = LOAD_AT_ONCE).await()
```
charlag, line 146: random change in this commit but ok :smile:

charlag (approved):
Takes a second to figure what it does but looks good! Awesome catch as well

PR #2357, opened 2022-02-27
Support new signup notifications

kyori19:
Latest Mastodon has new notification type: `admin.sign_up` 

* mastodon/mastodon#16953

Reused styles of new follower notifications as WebUI does: 

image


connyduck:
Do I understand this correctly, this is a notification admins & moderators will get for each signup on their instance?

kyori19:
Yes, all admins and moderators receives these notifications.

app/src/main/res/values/strings.xml
```diff
@@ -295,6 +297,8 @@
     <string name="notification_poll_description">Notifications about polls that have ended</string>
     <string name="notification_subscription_name">New toots</string>
     <string name="notification_subscription_description">Notifications when somebody you\'re subscribed to published a new toot</string>
+    <string name="notification_sign_up_name">Sign ups</string>
```
connyduck, line 300: Do you know how Mastodon Web names this in their interface? I feel like we need to call it "Signups (for Admins)" or something to not confuse regular users. 
Or even better, is there a way to detect if the current user is admin? Then we could just hide this notification type completely if not.

kyori19: In the WebUI, it is called 'New sign-ups' and hidden for non-admin users. (https://github.com/mastodon/mastodon/pull/16953/files#diff-0b4b4ff412a8d0db0b4625619d3633b2cc295c4e7f21316e7941bb4728446d50R169)

image


WebUI takes `is_staff` property from `#initial-state` element but this is only available for WebUI.

image

connyduck:
Ok, since this new notification type is not yet released on Mastodon's side, I will wait with merging this.

Btw here is Gargrons comment on why we don't get admin/mod status from the api: https://mastodon.social/@Gargron/107874999990771749

connyduck:
We may be getting a way to detect if a user can get this notification type: https://github.com/mastodon/mastodon/pull/17692

kyori19:
Mastodon v3.5.0 was released and I resolved conflicts.

PR #2369, opened 2022-03-04
fix various timeline related crashes

connyduck:
found in the Google Play crash reports

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/components/timeline/TimelineFragment.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/components/timeline/TimelineFragment.kt
@@ -217,7 +217,7 @@ class TimelineFragment :
             }
         })
 
-        lifecycleScope.launch {
+        viewLifecycleOwner.lifecycleScope.launch {
             viewModel.statuses.collectLatest { pagingData ->
                 adapter.submitData(pagingData)
             }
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/keylesspalace/tusky/components/timeline/viewmodel/CachedTimelineViewModel.kt
```diff
@@ -70,7 +70,14 @@ class CachedTimelineViewModel @Inject constructor(
     override val statuses = Pager(
         config = PagingConfig(pageSize = LOAD_AT_ONCE),
         remoteMediator = CachedTimelineRemoteMediator(accountManager, api, db, gson),
-        pagingSourceFactory = { db.timelineDao().getStatuses(accountManager.activeAccount!!.id) }
+        pagingSourceFactory = {
+            val activeAccount = accountManager.activeAccount
+            if (activeAccount == null) {
+                EmptyTimelinePagingSource()
```
connyduck, line 76: this can happen when a user logs out and the paging tries to reload from an empty database

app/src/main/java/com/keylesspalace/tusky/components/timeline/TimelineFragment.kt
```diff
@@ -217,7 +217,7 @@ class TimelineFragment :
             }
         })
 
-        lifecycleScope.launch {
+        viewLifecycleOwner.lifecycleScope.launch {
```
connyduck, line 220: There seems to be an edge case where the lifecycleScope is still valid but the Fragment view is already destroyed, this should fix it

charlag (approved):
Comments in code, not on github :wink: 

Looks good

Issue #2134, opened 2021-04-15
implement login via app


dioraman:
I have the same problem as https://github.com/tuskyapp/Tusky/issues/1609 . As per https://github.com/tuskyapp/Tusky/issues/1609#issuecomment-571240178 , implementing login via app would be much appreciated. Hence, I open the current issue with due title.
**EDIT:** if the default browser is not supported, one shall remove the default behavior, then login, then restore the default browser. Newcomers will think the app does not work.

* * * *
- Tusky Version: 14.0
- Android Version: 10QKQ1.191215.002
- [yes] I searched or browsed the repo’s other issues to ensure this is not a duplicate.
- Suggested labels: priority-medium, bug, refactoring, accessibility, privacy

connyduck:
https://github.com/tuskyapp/Tusky/pull/2371

Issue #2040, opened 2021-01-13
During initial login, Tusky sometimes opens authentication window in external browser


sorenstoutner:
When starting Tusky for the first time and logging into an instance, if certain browsers are set as the default browser for the OS, Tusky opens them after selecting the instance instead of handling authentication within the app.  This results in the browser being logged into the instance instead of Tusky.

For example, if Privacy Browser is installed and set as the default browser this behavior is manifested.

https://f-droid.org/en/packages/com.stoutner.privacybrowser.standard/

However, if the default browser is set to be Chrome, while leaving Privacy Browser installed, this doesn't happen.

Without looking at the code, my guess is that, for some reason, after the instance is selected, an intent is called that, depending on the intent filters of the default browser, will launch it instead of handling the authentication within Tusky.  If such is the case, the solution would be to modify Tusky so that it does not use an intent that attempts to match against an external app.

* * * *
- Tusky Version:  13.1
- Android Version:   11
- Android Device:  Pixel 5
- Mastodon instance (if applicable):  fosstodon.org

- [ X ] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

connyduck:
https://github.com/tuskyapp/Tusky/pull/2371

PR #2371, opened 2022-03-06
Implement Login via WebView

charlag:
I've tested all the scenarios I can think of but I would appreciate more testing.


connyduck:
closes https://github.com/tuskyapp/Tusky/issues/2134
closes https://github.com/tuskyapp/Tusky/issues/2040

app/src/main/java/com/keylesspalace/tusky/LoginWebViewActivity.kt
```diff
@@ -0,0 +1,142 @@
...
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+
+        val data = OauthLogin.parseData(intent)
+
+        setContentView(binding.root)
+
+        setSupportActionBar(binding.loginToolbar)
+        supportActionBar?.setDisplayHomeAsUpEnabled(true)
+        supportActionBar?.setDisplayShowTitleEnabled(false)
```
connyduck, line 81: I'd give it a title that says "Login"

app/src/main/java/com/keylesspalace/tusky/LoginWebViewActivity.kt
```diff
@@ -0,0 +1,142 @@
...
+
+        val webView = binding.loginWebView
+        webView.settings.allowContentAccess = false
+        webView.settings.allowFileAccess = false
+        webView.settings.databaseEnabled = false
+        webView.settings.displayZoomControls = false
+        webView.settings.javaScriptCanOpenWindowsAutomatically = false
+        webView.settings.userAgentString += "Tusky ${BuildConfig.VERSION_NAME}"
```
connyduck, line 89: maybe add the Android Version like here https://github.com/tuskyapp/Tusky/blob/develop/app/src/main/java/com/keylesspalace/tusky/di/NetworkModule.kt#L81

app/src/main/java/com/keylesspalace/tusky/LoginActivity.kt
```diff
@@ -137,130 +159,92 @@ class LoginActivity : BaseActivity(), Injectable {
...
-
-                redirectUserToAuthorizeAndLogin(domain, clientId)
-            }
+        setLoading(true)
 
-            override fun onFailure(call: Call<AppCredentials>, t: Throwable) {
+        lifecycleScope.launch {
+            val credentials: AppCredentials = try {
+                mastodonApi.authenticateApp(
+                    domain, getString(R.string.app_name), oauthRedirectUri.toString(),
```
connyduck, line 167: `oauthRedirectUri` already is a String
(also 3 more times below)

app/src/main/java/com/keylesspalace/tusky/LoginActivity.kt
```diff
@@ -137,130 +159,92 @@ class LoginActivity : BaseActivity(), Injectable {
...
+            .scheme("https")
+            .host(domain)
+            .addPathSegments(MastodonApi.ENDPOINT_AUTHORIZE)
+            .addQueryParameter("client_id", clientId)
+            .addQueryParameter("redirect_uri", oauthRedirectUri.toString())
+            .addQueryParameter("response_type", "code")
+            .addQueryParameter("scope", OAUTH_SCOPES)
+            .build()
+        val uri = Uri.parse(url.toString())
+        doWebViewAuth.launch(LoginData(uri, Uri.parse(oauthRedirectUri.toString())))
```
connyduck, line 208: `oauthRedirectUri.toUri()` 
https://developer.android.com/reference/kotlin/androidx/core/net/package-summary#(kotlin.String).toUri()

connyduck (commented):
Works very well 😍

connyduck:
Oh, and can you please move all Login things into the `components.login` package?

app/src/main/java/com/keylesspalace/tusky/LoginWebViewActivity.kt
charlag: Good call!

app/src/main/java/com/keylesspalace/tusky/LoginActivity.kt
charlag: Oh yeah I made it a `HttpUrl` but it's not actually a "http" URL so I rolled it back.

app/src/main/java/com/keylesspalace/tusky/LoginWebViewActivity.kt
charlag: We only have `action_login` I'm a bit hesitant to add a string just for that. If you say yes, I will do it

app/src/main/java/com/keylesspalace/tusky/LoginWebViewActivity.kt
charlag: I now injected exactly the same UA but then
 - Default UA has a lot of useful info e.g. engine version
 - Default UA already has Android version in it
 - We don't need OkHttp version for this UA

But I've changed a format now to be closer to what UA should be

connyduck:
You need to fix the code formatting so the CI check passes, looks good otherwise

charlag:
@connyduck thanks, I missed that. Can't squash commits nicer now because there's a move to package inbetween but eh

PR #2372, opened 2022-03-07
Don't hide potential timeline bugs by catching all exceptions

connyduck:
There might be implementation bugs that we are missing if we just catch all exceptions.

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/components/timeline/util/TimelineUtils.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/components/timeline/util/TimelineUtils.kt
@@ -5,12 +5,12 @@ import java.io.IOException
 
 fun Throwable.isExpected() = this is IOException || this is HttpException
 
-inline fun ifExpected(
+inline fun <T> ifExpected(
     t: Throwable,
-    cb: () -> Unit
-) {
+    cb: () -> T
+): T {
     if (t.isExpected()) {
-        cb()
+        return cb()
     } else {
         throw t
     }
--- a/app/src/main/java/com/keylesspalace/tusky/components/timeline/util/TimelineUtils.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/components/timeline/util/TimelineUtils.kt
@@ -1,0 +1,17 @@
+package com.keylesspalace.tusky.components.timeline.util
+
+import retrofit2.HttpException
+import java.io.IOException
+
+fun Throwable.isExpected() = this is IOException || this is HttpException
+
+inline fun ifExpected(
+    t: Throwable,
+    cb: () -> Unit
+) {
+    if (t.isExpected()) {
+        cb()
+    } else {
+        throw t
+    }
+}
--- a/app/src/main/java/com/keylesspalace/tusky/components/timeline/viewmodel/CachedTimelineRemoteMediator.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/components/timeline/viewmodel/CachedTimelineRemoteMediator.kt
@@ -23,7 +23,7 @@ import androidx.room.withTransaction
 import com.google.gson.Gson
 import com.keylesspalace.tusky.components.timeline.Placeholder
 import com.keylesspalace.tusky.components.timeline.toEntity
-import com.keylesspalace.tusky.components.timeline.util.isExpected
+import com.keylesspalace.tusky.components.timeline.util.ifExpected
 import com.keylesspalace.tusky.db.AccountManager
 import com.keylesspalace.tusky.db.AppDatabase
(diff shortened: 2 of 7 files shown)


app/src/main/java/com/keylesspalace/tusky/components/timeline/util/TimelineUtils.kt
```diff
@@ -0,0 +1,17 @@
+package com.keylesspalace.tusky.components.timeline.util
+
+import retrofit2.HttpException
+import java.io.IOException
+
+fun Throwable.isExpected() = this is IOException || this is HttpException
+
+inline fun ifExpected(
```
charlag, line 8: Just an idea: we could re-introduce `catchExpected` but control flow analysis might not work in some cases...

charlag (approved):

app/src/main/java/com/keylesspalace/tusky/components/timeline/util/TimelineUtils.kt
connyduck: How exactly would that look?

charlag: I tried few things....

```kotlin

/// Option 1
/// Simplest to write but a bit wonky to use
// Best performance

@JvmName("noNameIJustWantThisToComplie")
inline fun <T> catchExpected(block: () -> T, catch: (t: Throwable) -> T): T {
    return try {
        block()
    } catch (e: Throwable) {
        if (e.isExpected()) {
            catch(e)
        } else {
            throw e
        }
    }
}

fun test1() {
    val result = catchExpected({
        iMightThrow()
    }) { t ->
        // error handing
        return
    }
}

/// Option 2a
/// A bit more complex, I took generic catch and specialized it for our case.
/// Unfortunately when I checked bytecode I realized that it will allocate the block before
/// calling methods on it which makes little sense. and will also not inline all the way through
/// (not talking about JVM, just Kotlin inlining).

typealias Thunk<T> = () -> T

inline fun <T> Thunk<T>.catch(
    predicate: (Throwable) -> Boolean,
    catchBlock: (Throwable) -> T
): T {
    return try {
        this()
    } catch (e: Throwable) {
        if (predicate(e)) catchBlock(e) else throw e
    }
}

inline fun <T> Thunk<T>.catchExpected(
    catchBlock: (Throwable) -> T
): T {
    return this.catch(
        { t -> t.isExpected() },
        catchBlock
    )
}

fun test2a() {
    val result = {
        iMightThrow()
    }.catchExpected { t ->
        // error handling
        return
    }
}

/// Option 2b
/// A bit better syntax but I'm not sure whether it is actually inlined
/// Sam problems with inlining as in 2a, it reports correctly that it won't inline thunk

inline fun <T, F: Thunk<T>> doTry(thunk: F): F {
    return thunk
}

fun test2b() {
    val result = doTry {
        iMightThrow()
    }.catchExpected {
        // error handling
        return
    }
}


/// ---

fun Throwable.isExpected() = false
fun iMightThrow(): String {
    throw Exception()
}
```

Unfortunately, Kotiln compiler is :hankey: 

Parts of decompiled bytecode:

```java
   public static final void test1() {
      boolean var1 = false;

      try {
         int var2 = false;
         String var7 = iMightThrow();
      } catch (Throwable var6) {
         if (isExpected(var6)) {
            int var5 = false;
            return;
         }

         throw var6;
      }

   }

   public static final void test2a() {
      Function0 $this$catchExpected$iv = (Function0)null.INSTANCE;
      int $i$f$catchExpected = false;
      Function0 $this$catch$iv$iv = $this$catchExpected$iv;
      boolean var4 = false;

      Object var5;
      try {
         var5 = $this$catch$iv$iv.invoke();
      } catch (Throwable var11) {
         int var8 = false;
         if (isExpected(var11)) {
            int var10 = false;
            return;
         }

         throw var11;
      }

      String result = (String)var5;
   }


   public static final void test2b() {
      Function0 thunk$iv = (Function0)null.INSTANCE;
      int $i$f$catchExpected = false;
      $i$f$catchExpected = false;
      Function0 $this$catch$iv$iv = thunk$iv;
      boolean var4 = false;

      Object var5;
      try {
         var5 = $this$catch$iv$iv.invoke();
      } catch (Throwable var11) {
         int var8 = false;
         if (isExpected(var11)) {
            int var10 = false;
            return;
         }

         throw var11;
      }

      String result = (String)var5;
   }
```

Unfortunately variants 2a and 2b will do a lot of unnecessary work even though there's every possibility to inline block away. I don't know if there's some magic switch to make it happen but I wouldn't recommend those variants unless we find a way around it.

connyduck:
This gave me an idea, by having `isExpected` return something the code gets even shorter

charlag:
Yeah, that's better

PR #2373, opened 2022-03-07
Fix some network timeline bugs

kyori19:
I found that `NetworkTimelineViewModel` has some incomplete methods so tried to fix them.
I referred `CachedTimelineViewModel` and carefully implemented for network timeline, but it might have some mistakes so please check my codes.

Mistakes I found:

* `loadMore` does not handles statuses overlaps
* `loadMore` does not inserts new placeholder after fetched statuses when the number of statuses in the gap is more than fetch limit.
* `fullReload` does not fully refresh timeline. Statuses they are already fetched are not included after refresh.

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/components/timeline/viewmodel/NetworkTimelineViewModel.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/components/timeline/viewmodel/NetworkTimelineViewModel.kt
@@ -234,7 +234,9 @@ class NetworkTimelineViewModel @Inject constructor(
 
     override fun fullReload() {
         statusData.clear()
+        nextKey = null
         currentSource?.invalidate()
+        // FIXME: auto reload needed
     }
 
     suspend fun fetchStatusesForKind(
--- a/app/src/main/java/com/keylesspalace/tusky/components/timeline/viewmodel/NetworkTimelineViewModel.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/components/timeline/viewmodel/NetworkTimelineViewModel.kt
@@ -37,6 +37,8 @@ import com.keylesspalace.tusky.network.TimelineCases
 import com.keylesspalace.tusky.util.dec
 import com.keylesspalace.tusky.util.getDomain
 import com.keylesspalace.tusky.util.inc
+import com.keylesspalace.tusky.util.isLessThan
+import com.keylesspalace.tusky.util.isLessThanOrEqual
 import com.keylesspalace.tusky.util.toViewData
 import com.keylesspalace.tusky.viewdata.StatusViewData
 import kotlinx.coroutines.flow.map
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/keylesspalace/tusky/components/timeline/viewmodel/NetworkTimelineViewModel.kt
```diff
@@ -211,7 +233,9 @@ class NetworkTimelineViewModel @Inject constructor(
 
     override fun fullReload() {
         statusData.clear()
+        nextKey = null
         currentSource?.invalidate()
+        // FIXME: auto reload needed
```
kyori19, line 238: I cannot find an ideal way to request refresh to mediator, so timeline won't loaded after `fullReload` and error message 'Nothing here' will be shown.

app/src/main/java/com/keylesspalace/tusky/components/timeline/viewmodel/NetworkTimelineViewModel.kt
```diff
@@ -145,28 +150,46 @@ class NetworkTimelineViewModel @Inject constructor(
...
+                                }
+                                data[dataIndex] = data[dataIndex]
+                                    .copy(
+                                        isShowingContent = oldStatus.isShowingContent,
+                                        isExpanded = oldStatus.isExpanded,
+                                        isCollapsed = oldStatus.isCollapsed,
+                                    )
+                            }
+
+                            statusData.removeAt(overlappedFrom)
```
connyduck, line 185: Why is it not enough to insert this somewhere?
```kotlin
if (oldStatus != null) {
     statusData.remove(oldStatus)
}
```

connyduck (commented):
Thanks for identifying these problems 🙏

app/src/main/java/com/keylesspalace/tusky/components/timeline/viewmodel/NetworkTimelineViewModel.kt
connyduck: Try setting the `nextKey` to the id of the first status in `statusData` before clearing it

kyori19: Thanks, the problem seems to be solved

app/src/main/java/com/keylesspalace/tusky/components/timeline/viewmodel/NetworkTimelineViewModel.kt
kyori19: If some statuses in overlapped range removed by author, they could be correctly removed by this process.
I implemented this behaviour because I thought that we should remove statuses which we could know they are removed.
If you think we don't need this behaviour, I can remove this.

connyduck: Ah I see, yes we should have this if possible.

But I think there are two problems with your code
1) bothering with list indizes makes it hard to understand
2) `hashcode()` ordering does not equal object ordering, we must compare by actual status id length & order

Here is what I would try to do

```kotlin
if (statuses.isNotEmpty()) {
    val firstId = statuses.first().id
    val lastId = statuses.last().id
    statusData.removeAll { status ->
        val s = status.asPlaceholderOrNull()
        s != null && s.id.isLessOrEqualTo(firstId) && s.id.isMoreOrEqualTo(lastId)
    }
}
```
where `isLessOrEqualTo` and `isMoreOrEqualTo` are custom extension functions similar to [this one](https://github.com/tuskyapp/Tusky/blob/b145fc9d506875b7d1ad6f2786d6c3c6708b7407/app/src/main/java/com/keylesspalace/tusky/util/StringUtils.kt#L73)

kyori19: I improved the logic bit. Also stopped using `hashCode` for ordering.

connyduck (approved):

PR #2386, opened 2022-03-14
Fix status diffing and improve timeline performance

connyduck:
- Fix how `Status` and `StatusViewData` are compared to get rid of unnecessary view updates
- introduce `TimelineAccount` which is the same as `Account` but with less attributes so it is way faster
- add tests to make sure it really works as intended
- comparing statuses is slower than before, but the performance gain from `TimelineAccount` and less view updates is definitely worth it.

Decoding a full page (30 statuses) from json

|    | before - full `Account` | after - `TimelineAccount`|
| ------------- | ------------- | ------------- |
| Nokia 1  | 156 216 015 ns  | 86 186 701 ns|
| Pixel 2  | 45 626 983 ns | 23 611 981 ns|
| Fairphone 4 | 17 599 792 ns | 9 814 219 ns

Measured with this https://github.com/tuskyapp/tusky-json-benchmark

Memory usage - before / after
Screenshot 2022-03-12 at 19 06 21
Screenshot 2022-03-12 at 17 40 44

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/components/conversation/ConversationEntity.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/components/conversation/ConversationEntity.kt
@@ -28,7 +28,6 @@ import com.keylesspalace.tusky.entity.Poll
 import com.keylesspalace.tusky.entity.Status
 import com.keylesspalace.tusky.entity.TimelineAccount
 import com.keylesspalace.tusky.util.shouldTrimStatus
-import com.mikepenz.iconics.typeface.library.googlematerial.GoogleMaterial.url
 import java.util.Date
 
 @Entity(primaryKeys = ["id", "accountId"])
@@ -53,7 +52,7 @@ data class ConversationAccountEntity(
             id = id,
             username = username,
             displayName = displayName,
-            url = url,
+            url = "",
             avatar = avatar,
             emojis = emojis,
             localUsername = "",
--- a/app/src/main/java/com/keylesspalace/tusky/components/conversation/ConversationEntity.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/components/conversation/ConversationEntity.kt
@@ -16,19 +16,19 @@
 package com.keylesspalace.tusky.components.conversation
 
 import android.text.Spanned
-import android.text.SpannedString
 import androidx.room.Embedded
 import androidx.room.Entity
 import androidx.room.TypeConverters
 import com.keylesspalace.tusky.db.Converters
-import com.keylesspalace.tusky.entity.Account
 import com.keylesspalace.tusky.entity.Attachment
 import com.keylesspalace.tusky.entity.Conversation
 import com.keylesspalace.tusky.entity.Emoji
 import com.keylesspalace.tusky.entity.HashTag
 import com.keylesspalace.tusky.entity.Poll
 import com.keylesspalace.tusky.entity.Status
+import com.keylesspalace.tusky.entity.TimelineAccount
 import com.keylesspalace.tusky.util.shouldTrimStatus
+import com.mikepenz.iconics.typeface.library.googlematerial.GoogleMaterial.url
 import java.util.Date
 
 @Entity(primaryKeys = ["id", "accountId"])
@@ -48,17 +48,15 @@ data class ConversationAccountEntity(
     val avatar: String,
     val emojis: List<Emoji>
 ) {
-    fun toAccount(): Account {
-        return Account(
+    fun toAccount(): TimelineAccount {
+        return TimelineAccount(
             id = id,
             username = username,
             displayName = displayName,
+            url = url,
             avatar = avatar,
             emojis = emojis,
-            url = "",
             localUsername = "",
-            note = SpannedString(""),
-            header = ""
         )
     }
 }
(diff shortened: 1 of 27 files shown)


ariasuni:
Is this fixing the fact that refreshing my timeline re-hides messages behind their CW?

connyduck:
> Is this fixing the fact that refreshing my timeline re-hides messages behind their CW?

Is that a thing? It happened in some 16 beta version but it should be fixed in 16.0. If it still happens I need more info please.

app/src/main/java/com/keylesspalace/tusky/components/conversation/ConversationEntity.kt
```diff
@@ -100,7 +97,7 @@ data class ConversationStatusEntity(
         if (inReplyToId != other.inReplyToId) return false
         if (inReplyToAccountId != other.inReplyToAccountId) return false
         if (account != other.account) return false
-        if (content.toString() != other.content.toString()) return false // TODO find a better method to compare two spanned strings
+        if (content.toString() != other.content.toString()) return false
```
charlag, line 100: Here and in `hashCode()`:
allocating is eh. It is actually `SpannableStringBuilder` as far as I understand and it *will* allocate new array *and* new string.
Probably the best way to do this would be to implement a custom function/class on top of `charAt()` that would compare or hashCode.
Alternatively we could delay parsing HTML until `StatusViewData` which also kinda makes sense to me.

It can be a separate task but I think this would also improve performance quite a bit.

app/src/main/java/com/keylesspalace/tusky/entity/Status.kt
```diff
@@ -149,6 +149,71 @@ data class Status(
         return builder.toString()
     }
 
+    /**
+     * overriding equals & hashcode because Spanned does not always ompare correctly otherwise
```
charlag, line 153: typo in "compare"

charlag: Same thing about `content` as before, I think it should either be a concrete impl of `Spanned` or just plain text.

app/src/test/java/com/keylesspalace/tusky/StatusComparisonTest.kt
```diff
@@ -0,0 +1,216 @@
...
+import com.keylesspalace.tusky.json.SpannedTypeAdapter
+import com.keylesspalace.tusky.viewdata.StatusViewData
+import org.junit.Assert.assertEquals
+import org.junit.Assert.assertNotEquals
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.robolectric.annotation.Config
+
+@Config(sdk = [28])
+@RunWith(AndroidJUnit4::class)
```
charlag, line 16: I assume we need it because of `Spanned` too?

app/src/test/java/com/keylesspalace/tusky/StatusComparisonTest.kt
```diff
@@ -0,0 +1,216 @@
...
+    }
+
+    private fun createStatus(
+        id: String = "123456",
+        content: String = """
+            \u003cp\u003e\u003cspan class=\"h-card\"\u003e\u003ca href=\"https://mastodon.social/@ConnyDuck\" class=\"u-url mention\" rel=\"nofollow noopener noreferrer\" target=\"_blank\"\u003e@\u003cspan\u003eConnyDuck@mastodon.social\u003c/span\u003e\u003c/a\u003e\u003c/span\u003e Hi\u003c/p\u003e
+        """.trimIndent(),
+        note: String = ""
+    ): Status {
+        val statusJson = """
```
charlag, line 110: Did you make is JSON on purpose?

charlag (approved):
Great job!
Would also be nice if API didn't send huge accounts in timeline in the first place tho.

app/src/test/java/com/keylesspalace/tusky/StatusComparisonTest.kt
connyduck: Yes, so it gets created by Gson like in the real app.

app/src/main/java/com/keylesspalace/tusky/components/conversation/ConversationEntity.kt
connyduck: I'll look into it

app/src/test/java/com/keylesspalace/tusky/StatusComparisonTest.kt
charlag: We could have a separate tests where we check that we parse correctly because in some cases statuses come not from JSON. It's not very important tho.

app/src/main/java/com/keylesspalace/tusky/components/conversation/ConversationEntity.kt
charlag: Hmmm, that's quite interesting, good call to actually test it. I wonder if it uses cached `hashCode()` on the instance for quick equality check.

charlag: It might be because of dynamic dispatch on `Spanned` when you call `get()`.

connyduck: Yeah my test is also not super accurate probably. I'll figure it out when i have more time.

charlag: It was quite a moonshot anyway, I wish they just implemented it properly, then they could do it directly their buffer without dispatch and additional checks but we live in society.

PR #2390, opened 2022-03-17
handle screen rotation in LoginWebViewActivity

connyduck:


app/src/main/java/com/keylesspalace/tusky/components/login/LoginWebViewActivity.kt
```diff
@@ -127,19 +127,29 @@ class LoginWebViewActivity : BaseActivity(), Injectable {
             }
         }
         webView.setBackgroundColor(Color.TRANSPARENT)
-        webView.loadUrl(data.url.toString())
+
+        if (savedInstanceState == null) {
+            webView.loadUrl(data.url.toString())
+        } else {
+            webView.restoreState(savedInstanceState)
```
charlag, line 134: never had to do this before but it doesn't hurt I guess

charlag (approved):

app/src/main/java/com/keylesspalace/tusky/components/login/LoginWebViewActivity.kt
connyduck: Without it state is lost on rotation and user has to enter password again

PR #2393, opened 2022-03-19
WIP: prefs

charlag:


connyduck:
Is this an opportunity to migrate from shared prefs to [AndroidX DataStore](https://developer.android.com/topic/libraries/architecture/datastore) or would that be too complicated for now?

charlag:
:thinking: I did not consider that.

As far as I understand, there are two variants:
1. Untyped one via `Preferences` a-la `SharedPreferences`. Has nice observation mechanism if we wan to replace our own event bus for preference event but is generally still untyped and has no default values so we would have to wrap that
2. Typed one via protobuf. Might be more interesting with structured data and default values, also can observe it. If protobuf doesn't handle migrations automatically we would have to do that manually.

So I think it would make sense to use the protobuf variant instead of what I started because it would be a very similar API, perhaps even better one with plain object instead of some delegated properties. It also gives me a nice idea: before I thought that we need `PrefKeys` to observe them via bus events but we could have just one event for when preferences change (or just use their `Flow` API which has the same result).

In any case I think there are few things that will make it hard to adopt. First, It's async API on top of `Flow`. I'm not sure what would be the best way to access that, especially because we don't control lifecycle of activities: the value might or might not be there yet.

In general would be nice to get rid of blocking disk reads whenever possible e.g. in `TuskyApplication` but it might be a separate task, we have other blocking violation.

So what are your thoughts on that? I think it might be a good idea, I just don't know how to deal with async stuff best.

charlag:
Update: I've found out proto 3 doesn't have default values :roll_eyes:

charlag:
I could probably specify it in default value but I'm thinking whether it is overkill to specify it as a protobuf.
There are certain advantages:
 1. Enums
 2. Nested types (like our proxy)

Otherwise I could wrap around untyped API somehow

charlag:
I tried out protobuf a bit more and here are my thoughts:
 1. It needs a gradle plugin and runtime dependency which I don't like that much. Gradle plugin assumes `protoc` binary too.
 2. If we want to specify default values we need to write code for serializer which will set them. This is a separate place than the place which defines the field (.proto file) so it can be easily forgotten which I don't like.
 3. Generated file is a bit heavier than what I would like to see

So probably "untyped" variant would work for us better.

connyduck:
Yes, Preferences DataStore is the one we should use. It should be possible to make most preference calls async, with the exception of the language, font size, emoji style, and theme because we need that before we draw anything, right?

charlag:
Preferences is a singleton and it caches stuff in memory. If we use that then we would "really" block only at app start (because emoji/themes etc) and later calls will not actually do I/O. I think this should be fine.

charlag:
Reading their code it's actually only singleton as much as you instantiate it so it doesn't matter *that* much.

i kind of want to try and implement `DataStore` for just a Serializable or Parceable, I wonder why they didn't do it in the first place

charlag:
What if we just...

```kotlin
// TODO: name and stuff
data class PrefData(
    val testProp: String
)

object PrefDataSerializer : Serializer<PrefData> {
    private val gson = Gson()

    override val defaultValue: PrefData
        get() = PrefData(
            testProp = "test",
        )

    override suspend fun readFrom(input: InputStream): PrefData {
        return gson.fromJson(input.reader(), PrefData::class.java)
    }

    override suspend fun writeTo(t: PrefData, output: OutputStream) {
        gson.toJson(t, output.writer())
    }
}

fun test(context: Context) {
    val store = DataStoreFactory.create(
        PrefDataSerializer
    ) {
        context.preferencesDataStoreFile("prefs")
    }
    store.data.map { prefs -> 
        prefs.testProp
    }
    
    runBlocking {
        store.updateData { prefs ->
            prefs.copy(testProp = "new!")
        }
    }
}
```

Best of all worlds IMO.
(I tried to use Parcelable instead but it's kinda weird)

charlag:
Phew this was more effort than I wanted but we have DataStore. I can implement writing to prefs and change events now.

I must say, I think we have *way* to many preferences now. 
 * Custom tab preference? Really? Do most people even understand what it does? I checked and it was there from the beginning, might have just happened
 * hideFab?
 * hideTopToolbar???
 * showBotOverlay???
 * showNotificationsFilter?
 * useBlurhash?

But links is the most painful one because we have million of spans and we do not properly delegate link handling to a single place. I cannot create another `DataStore` in parallel so I would need to make it a global singleton which is also ugly and error-prone. I would rather get rid of link preference but I don't know if it's something people feel strongly about

charlag:
I did not figure out how to do editing yet with this approach though. If we don't use SharedPreferences then the whole `androidx.preference` becomes useless (we already reimplement good part of it).

charlag:
We already do hacks to make account preferences work and from what I see now we actually *do write values accidentally* for account preferences even though we don't want it.
I would vote for replacing it with our custom impl, it's not too much GUI

connyduck:
> I must say, I think we have way to many preferences now.

Agreed. But I'm sure users will complain if we remove one of them.
- Custom chrome tabs -> That one needs to be renamed, it is definitely confusing. But I would keep it.
- hideFab -> It is there because users complained the fab hides stuff, and others complained because it moved. When removed, which of the two options should we take?
- hideTopToolbar -> highly requested by users that want more screen space
- showBotOverlay -> that one is the only one I would be comfortable removing (always show the overlay)
- showNotificationsFilter -> again, more screen space
- useBlurhash -> Mastodon has it as well, some people don't like it


> I did not figure out how to do editing yet with this approach though. If we don't use SharedPreferences then the whole androidx.preference becomes useless (we already reimplement good part of it).

There is `PreferenceDataStore` which allows you to provide a custom data source for the preference screen, but it does not look like it would nicely integrate with `DataStore`. Weird that Google did not think of that. (And confusing that they have similar names)

> We already do hacks to make account preferences work and from what I see now we actually do write values accidentally for account preferences even though we don't want it.
I would vote for replacing it with our custom impl, it's not too much GUI

Yes.


Btw, the "Tabs" setting in Preferences (where you can filter replies and boosts in the home timeline) should actually be an Account Preference. But we can also move it so this does not get too big.

charlag:
Yeah you theoretically can use `PreferenceDataStore` but then you are back to untyped API and we don't want that.

Poked into UI a bit, I'm not sure how to do it best, maybe you have an idea.

I didn't want to rewrite all the places where we use settings DSL twice and also it's kinda nice to only have one place to add UI for prefs but it's a bit of a PITA to make views programmatically. What do you think, is it wort it or should I just add UI in xml and then link it up?

connyduck:
> I didn't want to rewrite all the places where we use settings DSL twice and also it's kinda nice to only have one place to add UI for prefs but it's a bit of a PITA to make views programmatically. What do you think, is it wort it or should I just add UI in xml and then link it up?

Not sure. Both sounds reasonable.

charlag:
I have a bit of a problem with the language picker, I think because it is accessed so early it gets some wrong instance of preferences, need to debug...

charlag:
I am thinking about splitting it into two parts: one is adapting the app and the other is changing the actual backing store and preference screens, what do you think? I am afraid of getting stuck and falling behind again

Tak:
Yeah, I think that makes sense

charlag:
I made it compile again but I'm missing few preference screens and new preferences, I will see if I can just finish it next week and if not I'll try to split

nikclayton:
Some unsolicited feedback.

# Migrating to Datastore

This appears to be reimplementing a lot of the existing Preference machinery to migrate to Datastore.

I'm not sure that's necessary. You can change the backend that the entire preference machinery writes to with `preferenceManager.preferenceDataStore`, see https://developer.android.com/develop/ui/views/components/settings/use-saved-values#custom-data-store.

However, I think that's probably one of the last steps to do.

# Reading preferences in Activity / Fragment

In the existing code (and this is continued in this PR), activities and fragments read preferences directly.

In most, possibly all cases, I think this should be changed. The preferences they access are used to determine aspects of how the UI should be displayed (`showBotOverlay`, `useBlurHash`, `absoluteTimeView`, etc).

Therefore, these should really be exposed to activities and fragments as part of the view model.

This is where a translation can be done between e.g., the string values of some properties, and actual enums, to provide better type safety.

This would simplify the code in activities and fragments. Many view models already access preferences, so while this does this move code in to the view model, I don't think it's a significant complication to the existing code.

This could be done as a single PR, or one PR per activity or fragment.

This makes it easy to shard the work out amongst multiple people, or do it increments as and when people have time.

# Introduce a preference repository

Once the activity/fragments have been decoupled from the preferences, a preference repository can be introduced.

This can continue to use SharedPreferences, and can be injected in to the `ViewModel` implementations.

Activites and Fragments will not need to be changed because their interface to the preference values (via the view models) does not change.

This will simplify the code in the view models, and move all the code for reading preferences to a single place.

# Introduce a schema version number

[This is not strictly tied to this work, and could be done now. But this would also be a good point to do it]

Preferences need a schema version number. This doesn't need to be exposed to to the view models, only the preference repository.

It's to cover the case where we decide to change how a preference is represented, or change a default value, while keeping the old default for people who are upgrading.

Example 1: suppose there's a decision to change the default app theme from "dark" to "use the system theme".

For people installing Tusky after this change there's nothing that needs to be done.

For people upgrading, if their preference value for this is null, the upgrade needs to write "dark" in to that preference so they get their current behaviour.

Example 2: Changing from a boolean to an enum

The "Always show sensitive content" option is currently a boolean. Suppose we wanted to make that a tri-state, "Always", "Never", "Only from people I follow" (i.e., hide sensitive content on statuses they have reblogged, or from people who have posted a status that includes a hashtag I follow).

[I'm not suggesting this is an option we would provide, it's just an example of how a boolean option can become more complicated]

The underlying pref will change from a bool to a string.

In both cases, the decision to perform the upgrade is driven by the schema version changing (similar to how the database migration stuff works).

# Migrate to a different preference backend

At this point:

1. All preferences are encapsulated in a repository that provides strongly typed access to them
2. Preferences are accessed only by view models (and the activity that provides the UI to modify them)
3. All activities / fragments get preference values indirectly, as part of the view model.

All of the above works with the existing SharedPreferences machinery.

If necessary, migrating to a different preferences backend (datastore (KV, or protobuf), or database) is much simpler. The view models and the activity/fragments don't need to be changed, all the work happens in the preference repository code.

And you may discover, after doing the previous steps, that there's no need to do any migration, because now the SharedPreference stuff is nicely encapsulated where the vast majority of the code doesn't need to interact with it.

charlag:
Thanks for the feedback. I agree to some of it. I am not entirely certain that views should get no access to preferences ever. Some preferences are quite specific to displaying of things and I really don't want each viewModel to translate untyped prefs to typed ones.

I am also a bit on the fence about wrapping prefs in yet another repository but if it's a very small thing which *doesn't* repeat all preferences it might work.

I didn't push my latest state yet but I clearly can't keep up with the project pace right now nor just simply work on it. I mentioned before and I am really starting to think that the translation to a more safe API should be done first (be it PrefStore or anything else) and settings via DSL should be done later.

Lakoja:
Just as an observation: This (already merged) PR implements some settings as DataStore: https://github.com/tuskyapp/Tusky/pull/3205

Issue #2382, opened 2022-03-13
Can't refresh timeline: non-alphabet digit found: `


tastytea:
I'm getting this error on the server when refreshing the home timeline in Tusky:

```
2022-03-13 11:33:43 [pleroma] request_id=FtvqOgPmfwDg3cIAAoRB [debug] GET /api/v1/timelines/home
2022-03-13 11:33:43 [pleroma] request_id=FtvqOgPmfwDg3cIAAoRB [debug] Processing with 
2022-03-13 11:33:43 [pleroma] request_id=FtvqOgPmfwDg3cIAAoRB [debug] Processing with Pleroma.Web.MastodonAPI.TimelineController.home/2_  Parameters: %{"limit" => "30", "since_id" => "AHKkQNnhEXixxt80Q`"}_  Pipelines: [:authenticated_api]
[…]
2022-03-13 11:33:43 [pleroma] request_id=FtvqOgPmfwDg3cIAAoRB [error] Internal server error: %ArgumentError{message: "non-alphabet digit found: `"}
```

The progress-animation appears for a short time but nothing happens. Force-closing the app and deleting the cache doesn't fix it. The notification timeline still works and the Pleroma FE works normally. I'm not sure if Pleroma sent the wrong ID in the first place or if Tusky modified it.

[Issue in Pleroma bug tracker](https://git.pleroma.social/pleroma/pleroma/-/issues/2850)

<details><summary>Full log:</summary>

```
2022-03-13 11:33:43 [pleroma] [info] GET /api/v1/timelines/home
2022-03-13 11:33:43 [pleroma] request_id=FtvqOgPmfwDg3cIAAoRB [debug] GET /api/v1/timelines/home
2022-03-13 11:33:43 [pleroma] request_id=FtvqOgPmfwDg3cIAAoRB [debug] Processing with Pleroma.Web.MastodonAPI.TimelineController.home/2_  Parameters: %{"limit" => "30", "since_id" => "AHKkQNnh
EXixxt80Q`"}_  Pipelines: [:authenticated_api]
2022-03-13 11:33:43 [pleroma] request_id=FtvqOgPmfwDg3cIAAoRB [debug] QUERY OK source="oauth_tokens" db=2.3ms queue=0.1ms idle=373.6ms_SELECT o0."id", o0."token", o0."refresh_token", o0."scope
[...]

PR #2398, opened 2022-03-21
Fix Timeline not loading

connyduck:
Getting rid of all client generated ids as they only cause trouble.
From now on, we don't display the last loaded status and use its id for the placeholder. This way we can safely send placeholder ids into the api.

Requires a db cache cleanup to make sure no old placeholder ids are cached anymore. I'm using this opportunity to make tags in the db non-null, fyi @Tak 

fixes https://github.com/tuskyapp/Tusky/issues/2382

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/components/timeline/viewmodel/CachedTimelineRemoteMediator.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/components/timeline/viewmodel/CachedTimelineRemoteMediator.kt
@@ -30,7 +30,6 @@ import com.keylesspalace.tusky.db.TimelineStatusEntity
 import com.keylesspalace.tusky.db.TimelineStatusWithAccount
 import com.keylesspalace.tusky.entity.Status
 import com.keylesspalace.tusky.network.MastodonApi
-import com.keylesspalace.tusky.util.dec
 import kotlinx.coroutines.rx3.await
 import retrofit2.HttpException
 
@@ -103,8 +102,12 @@ class CachedTimelineRemoteMediator(
                 val overlappedStatuses = replaceStatusRange(statuses, state)
 
                 if (loadType == LoadType.REFRESH && overlappedStatuses == 0 && statuses.isNotEmpty() && !dbEmpty) {
+                    /* This overrides the last of the newly loaded statuses with a placeholder
+                       to guarantee the placeholder has an id that exists on the server as not all
+                       servers handle client generated ids as expected
+                     */
                     timelineDao.insertStatus(
-                        Placeholder(statuses.last().id.dec(), loading = false).toEntity(activeAccount.id)
+                        Placeholder(statuses.last().id, loading = false).toEntity(activeAccount.id)
                     )
                 }
             }
--- a/app/src/main/java/com/keylesspalace/tusky/components/timeline/viewmodel/CachedTimelineViewModel.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/components/timeline/viewmodel/CachedTimelineViewModel.kt
@@ -41,8 +41,6 @@ import com.keylesspalace.tusky.entity.Poll
 import com.keylesspalace.tusky.network.FilterModel
 import com.keylesspalace.tusky.network.MastodonApi
 import com.keylesspalace.tusky.network.TimelineCases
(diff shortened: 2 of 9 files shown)


app/src/main/java/com/keylesspalace/tusky/components/timeline/viewmodel/CachedTimelineRemoteMediator.kt
```diff
@@ -103,8 +102,12 @@ class CachedTimelineRemoteMediator(
                 val overlappedStatuses = replaceStatusRange(statuses, state)
 
                 if (loadType == LoadType.REFRESH && overlappedStatuses == 0 && statuses.isNotEmpty() && !dbEmpty) {
+                    /* This overrides the last of the newly loaded statuses with a placeholder
+                       to guarantee the placeholder has an id that exists on the server as not all
+                       servers handle client generated ids as expected
+                     */
```
Tak, line 108: What happens when there's only one status?

connyduck: When the timline is loaded, and a refresh returns only one status, and the status is not known already, it would turn into a placeholder at the top of the timeline. Clicking it would load the gap (that one status) normally. (If the one status is known, or the timeline is empty, no placeholder shows)
The Api returning only one unknown status is almost impossible though, because for that you need to already have a placeholder at the top of the timeline, which you can only get by having a placeholder at the usual 30th position, having all 29 statuses before it deleted on the server, and then refreshing (Having everything filtered doesn't count because the filtered statuses are technically still in the local database).  

Not sure if this is 100% the right behavior, maybe it should show the actual status in this case and not a placeholder. I think I might need to test this, thanks for pointing it out.

connyduck: I think the check wheter to insert a placeholder or not must be "we got a full page with no overlap" instead of "we got something with no overlap". Works perfect in my tests, but maybe I am missing another edgecase now 🤔

Tak: That makes sense to me.
Is there any reason the server returns less than a full page other than "there's nothing else to load in this range" ?

ClearlyClaire: I think that depends on the software and the particular timeline you're fetching. In mainline Mastodon, this should currently be fine, but I wouldn't expect that to never change, as there are codepaths (not for timelines as far as I can tell) where we fetch from database first, and do some filtering on top of that, so we may end up with less posts than the page's limit, while there are still underlying posts that weren't returned.

connyduck: D:

charlag: We have been detecting the end of timeline like that for ages tho

ClearlyClaire: For detecting end of timelines, a more reliable way could probably be to check `next` or `prev` headers. But in any case, the current Mastodon behavior is safe, so changing that is not urgent.

ClearlyClaire: For the record, I found this commit because I recently experienced “missing chunks” of my timeline, which [I identified](https://social.sitedethib.com/@Claire/108083719157658964) to be Mastodon returning empty pages to Tusky queries mentioning invalid IDs. So that most definitely sounded like #2382 and I expected the fix to that issue to also fix the “missing chunks” issue.

However, I have seen [a report](https://eldritch.cafe/@AgathaSorceress/108089564990952239) of a similar issue from someone using a self-built Tusky that should include that fix. I am myself still running Tusky 16.0, so I have not been able to investigate the issue, but it is possible that there are still issues with the fetching logic.

PR #2400, opened 2022-03-22
RFC: display embedded html in preview cards

Tak:
We've had a number of people ask for inline embeds for peertube, funkwhale, etc., so I took a stab at loading them with a webview. Is this an approach we'd want to use (guarded by a preference)? :shrug: 

Screenshot_20220322-152155~2
Screenshot_20220322-152247~2
Screenshot_20220322-152317~2
Screenshot_20220322-152028~2
Screenshot_20220322-152226

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/adapter/StatusBaseViewHolder.java
+++ b/app/src/main/java/com/keylesspalace/tusky/adapter/StatusBaseViewHolder.java
@@ -1,6 +1,7 @@
 package com.keylesspalace.tusky.adapter;
 
 import static com.keylesspalace.tusky.viewdata.PollViewDataKt.buildDescription;
+import android.annotation.SuppressLint;
 
 import android.content.Context;
 import android.graphics.drawable.BitmapDrawable;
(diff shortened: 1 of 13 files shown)


Tak:
Even when they're hidden with no content loaded 99% of the time?
What about putting a placeholder into the layout xml and dynamically instantiating the webview when needed?

connyduck:
Fancy 🧐, but
- I'm very concerned about performance as well
- Some of the embeds do not look mobile friendly at all (small buttons/font)

I'd rather not have this in the timeline, but if you really want it I would merge it if it is hidden behind a setting and the webviews are not initialized when it is off

Tak:
I don't feel super strongly about whether we should have it or not, but imo it only really makes sense in the timeline (if I'm going to click to the detailed view to see the embed, I might as well just click the link).

I'm running it now, I can't feel any performance difference - although I'm on a pixel 2, which is still *fairly* high end, and my timelines aren't super heavy on embeds

charlag:
I don't want any more obscure preferences. If we think that something might kind work sometimes it's probably not a good idea to include it?

Oblomov:
Personal preference would be to have an option with the choice to show embed
* never
* only in the detailed view
* always

I'd like to point out that “only in detailed view” does make sense, because the oembed itself is still usually more lightweight than loading the remote site (which will usually require launching a separate app).

Tak:
I had left this here as something to point to when people came by and requested, but now it's just cluttering up the PR list

PR #2396, opened 2022-03-19
Rename Toots to Posts

connyduck:
Following https://github.com/mastodon/mastodon/pull/16080 and a lot of user requests

In Code: Toot -> Status
User visible text and string names: Toot -> Post

Did not rename: maximumTootCharacters, Toot Button, constant used in notification channel

PR #2405, opened 2022-03-26
Replace “status” by “post” in strings

ariasuni:
«Toot» has been replaced by «post» in strings, but there was a few «status» left as well. Now everything is called a post in the UI.

For reference, the same change in Mastodon: https://github.com/mastodon/mastodon/pull/16271


connyduck:
Oops I forgot those in https://github.com/tuskyapp/Tusky/pull/2396
Can you also rename all the string names like I did there, please?

ariasuni:
I think it’s good now? Since I used search and replace, it changed a lot of files (also changed a view ID), but I double-checked everything. Also fixed a «postss» into «posts» (thanks to @gunchleoc for the comment on Weblate)

connyduck (approved):
Thank you!

PR #2410, opened 2022-03-30
Fix saving failed status to drafts

### LINKED ISSUE #2399: Miscounts URL contribution to character limit, sometimes fails to save draft on error 422
## In brief
* Tusky doesn't correctly count characters when mixed with a URL
  * Despite notification, **failed toot may be lost, not saved to drafts**
  * Affects glitchsoc-based instance with raised custom toot length (`"max_toot_chars":1312`)
  * Example given below

### Steps
1.  Open Tusky, start composing a toot on an instance
    * In this case, https://vulpine.club/ with a length limit of 1312 characters
2.  Set privacy to Direct (to avoid spamming others)
3.  Add a content warning:
```
Test post
```
4.  **Test A:** Fill the body of the toot with the following:
```
https://example.com 11111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111
```
5.  **Add or remove `1`'s as needed to have exactly zero characters available/remaining**
6.  Try to send toot in Tusky
7.  Check drafts if sending toot fails
8.  Paste this into the web UI of the instance, compare character count
     * On vulpine.club, the above results in Tusky showing 0 characters left, while the web UI shows it as -4 characters (i.e. 4 characters over)
9.  **Test B:** Repeat steps 1-8 with the following toot, using the same content warning as before, **adding or removing `1`'s as needed to have zero characters available**, like before:
```
https://example.com/?testing_post_chars_here 1111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111
```

### Expected
Tusky shows the test toot as hitting 0 characters remaining.  The toot posts successfully.

If it fails to post, Tusky successfully saves the toot as a draft.

### Actual
Tusky gives 4 more characters than the web UI allows.  The toot fails to send.

When the toot fails to send, Tusky **occasionally does not save the toot in drafts, losing work** despite the notification's claim!  It seems to be random, usually not saving to drafts but sometimes succeeding.

(Note: I have other draft messages in Tusky's list of drafts.  Tusky does properly save toots to drafts if e.g. sending a toot while offline, then canceling the operation.)

### Extra details

**NOTE:** This appears to break the glitchsoc web UI, too!  Trying to post the above statuses when the web UI claims there's 0 characters remaining results in…
`422: Validation failed: Text character limit of 1312 exceeded`

…despite the length limit appearing to be set correctly (both web UI and backend report a limit of 1312 characters).

After filling the toot with `1`'s until the web UI reports 0 characters remaining, I have to delete a single `1` from the toot for it to post on the web UI (1 character free) for both **Test A** and **Test B**.

For Tusky to post correctly, after filling the toot with `1`'s until Tusky reports 0 characters remaining, I have to delete five `1`'s from the toot content in **Test A**, and only one `1` from the toot content in **Test B**.

I think there's multiple bugs here, possibly including a bug in glitchsoc or Mastodon itself.

* * * *
- Tusky Version: `16.0`
- Android Version: `12 (SP2A.220305.012)`
- Android Device: `Pixel 4 XL`
- Mastodon instance (if applicable): https://vulpine.club/ (features custom max toot length, 1312 characters as shown in web UI and Tusky)

connyduck:
It was a race condition - we started the async database query and then cancelled the service without waiting for the database query to return. This fixes the problem so that statuses that fail to send will always be saved to drafts.

I also refactored the code to use `kotlinx.coroutines.delay` instead of `java.util.Timer` which is way nicer.

closes #2399

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/service/SendStatusService.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/service/SendStatusService.kt
@@ -155,39 +155,44 @@ class SendStatusService : Service(), Injectable {
         val callback = object : Callback<Status> {
             override fun onResponse(call: Call<Status>, response: Response<Status>) {
 
-                val scheduled = !statusToSend.scheduledAt.isNullOrEmpty()
-                statusesToSend.remove(statusId)
+                serviceScope.launch {
 
-                if (response.isSuccessful) {
-                    // If the status was loaded from a draft, delete the draft and associated media files.
-                    if (statusToSend.draftId != 0) {
-                        serviceScope.launch {
+                    val scheduled = !statusToSend.scheduledAt.isNullOrEmpty()
+                    statusesToSend.remove(statusId)
+
+                    if (response.isSuccessful) {
+                        // If the status was loaded from a draft, delete the draft and associated media files.
+                        if (statusToSend.draftId != 0) {
                             draftHelper.deleteDraftAndAttachments(statusToSend.draftId)
                         }
-                    }
 
-                    if (scheduled) {
-                        response.body()?.let(::StatusScheduledEvent)?.let(eventHub::dispatch)
+                        if (scheduled) {
+                            response.body()?.let(::StatusScheduledEvent)?.let(eventHub::dispatch)
+                        } else {
+                            response.body()?.let(::StatusComposedEvent)?.let(eventHub::dispatch)
+                        }
+
+                        notificationManager.cancel(statusId)
                     } else {
-                        response.body()?.let(::StatusComposedEvent)?.let(eventHub::dispatch)
+                        // the server refused to accept the status, save status & show error message
+                        saveStatusToDrafts(statusToSend)
+
+                        val builder = NotificationCompat.Builder(this@SendStatusService, CHANNEL_ID)
+                            .setSmallIcon(R.drawable.ic_notify)
+                            .setContentTitle(getString(R.string.send_post_notification_error_title))
+                            .setContentText(getString(R.string.send_post_notification_saved_content))
+                            .setColor(
+                                ContextCompat.getColor(
+                                    this@SendStatusService,
+                                    R.color.notification_color
+                                )
+                            )
+
+                        notificationManager.cancel(statusId)
+                        notificationManager.notify(errorNotificationId--, builder.build())
                     }
-
-                    notificationManager.cancel(statusId)
-                } else {
-                    // the server refused to accept the status, save status & show error message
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/keylesspalace/tusky/service/SendStatusService.kt
```diff
@@ -154,56 +151,56 @@ class SendStatusService : Service(), Injectable {
 
         val callback = object : Callback<Status> {
```
charlag, line 152: yikes, it's still super old network API, should we also replace it? Might make the rest much easier to read, idk

charlag (approved):
Looks good.

Once again I would recommend putting PR message into the commit.

PR #2413, opened 2022-04-02
Improve time format of posts when using absolute time

ariasuni:
- Factorize code: now everything is handled at the same place, instead of being copy-pasted.
- Change how absolute time is displayed:
  - When time of the post is on the same day as today, `HH:mm` is used instead of `HH:mm:ss` (seconds are not necessary the vast majority of the time and are available in detailed view if needed anyway)
  - When time of the post is on a different day but in the same month as today, `MM-dd HH:mm` is used instead of `MM/dd HH:mm:ss` (I changed from `/` to `-` to better reflect the fact the format put the month before the day, like ISO 8601 and and RFC 3339)
  - When time of the post is on a different month, `yyyy-MM-dd` is used instead of `MM/dd HH:mm:ss` as above (the fact that the year is not displayed at all in this case is confusing and not practical, and I didn’t want the time to be too large on screen so I restricted it to not show hours and minutes)
  - When the time is unknown, `??` is used instead of `??:??:??`, since it should already be clear with only two interrogation points, and showing more serve no purpose.

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/util/AbsoluteTimeFormatter.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/util/AbsoluteTimeFormatter.kt
@@ -15,55 +15,45 @@
 
 package com.keylesspalace.tusky.util
 
-import androidx.test.ext.junit.runners.AndroidJUnit4
 import java.text.SimpleDateFormat
-import java.time.ZoneId
 import java.util.Calendar
 import java.util.Date
 import java.util.Locale
-import org.junit.Assert.assertFalse
-import org.junit.Assert.assertTrue
-import org.junit.Before
-import org.junit.Test
-import org.junit.runner.RunWith
-import org.robolectric.annotation.Config
+import java.util.TimeZone
 
-public class AbsoluteTimeFormatter {
-    private val sameDaySdf = SimpleDateFormat("HH:mm", Locale.getDefault())
-    private val sameYearSdf = SimpleDateFormat("MM-dd HH:mm", Locale.getDefault())
-    private val otherYearSdf = SimpleDateFormat("yyyy-MM-dd", Locale.getDefault())
-    private val otherYearCompleteSdf = SimpleDateFormat("yyyy-MM-dd HH:mm", Locale.getDefault())
+class AbsoluteTimeFormatter @JvmOverloads constructor(timeZone: TimeZone = TimeZone.getDefault()) {
+    private val sameDaySdf = SimpleDateFormat("HH:mm", Locale.getDefault()).apply { this.timeZone = timeZone }
+    private val sameYearSdf = SimpleDateFormat("MM-dd HH:mm", Locale.getDefault()).apply { this.timeZone = timeZone }
+    private val otherYearSdf = SimpleDateFormat("yyyy-MM-dd", Locale.getDefault()).apply { this.timeZone = timeZone }
+    private val otherYearCompleteSdf = SimpleDateFormat("yyyy-MM-dd HH:mm", Locale.getDefault()).apply { this.timeZone = timeZone }
 
-    fun format(time: Date?, shortFormat: Boolean = true): String {
-        return if (time == null) {
-            "??"
-        } else if (android.text.format.DateUtils.isToday(time.time)) {
-            sameDaySdf.format(time)
-        } else if (AbsoluteTimeFormatter.isThisYear(time)) {
-            sameYearSdf.format(time)
-        } else if (shortFormat) {
-            otherYearSdf.format(time)
-        } else {
-            otherYearCompleteSdf.format(time)
+    @JvmOverloads
+    fun format(time: Date?, shortFormat: Boolean = true, now: Date = Date()): String {
+        return when {
+            time == null -> "??"
+            isSameDate(time, now) -> sameDaySdf.format(time)
+            isSameYear(time, now) -> sameYearSdf.format(time)
+            shortFormat -> otherYearSdf.format(time)
+            else -> otherYearCompleteSdf.format(time)
         }
     }
 
     companion object {
-        private fun isThisYear(time: Date): Boolean {
-            val dateYear = time.toInstant().atZone(ZoneId.systemDefault()).toLocalDate().getYear()
-            val currentYear = Calendar.getInstance().get(Calendar.YEAR)
-            return dateYear == currentYear
+
+        private fun isSameDate(dateOne: Date, dateTwo: Date): Boolean {
+            val calendarOne = Calendar.getInstance().apply { time = dateOne }
+            val calendarTwo = Calendar.getInstance().apply { time = dateTwo }
+
+            return calendarOne.get(Calendar.YEAR) == calendarTwo.get(Calendar.YEAR) &&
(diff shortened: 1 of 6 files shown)


app/src/main/java/com/keylesspalace/tusky/util/AbsoluteTime.kt
```diff
@@ -0,0 +1,42 @@
+/* Copyright 2017 Andrew Dawson
```
connyduck, line 1:
```suggestion
/* Copyright 2022 Tusky Contributors
```

app/src/main/java/com/keylesspalace/tusky/util/AbsoluteTime.kt
```diff
@@ -0,0 +1,42 @@
...
+package com.keylesspalace.tusky.util
+
+import java.text.SimpleDateFormat
+import java.util.Calendar
+import java.util.Date
+import java.util.Locale
+
+public class AbsoluteTime {
+    companion object {
+        private val sameDaySdf = SimpleDateFormat("HH:mm", Locale.getDefault())
```
connyduck, line 25: Quoting the Android Studio warning: 
> Assigning Locale.getDefault() to a final static field will not work correctly if the user changes locale while the app is running

I'd create one `AbsoluteTime` instance for each `StatusBaseViewHolder`/`StatusViewHelper` it is used in to fix that.

app/src/main/java/com/keylesspalace/tusky/util/AbsoluteTime.kt
```diff
@@ -0,0 +1,42 @@
...
+        private val sameYearSdf = SimpleDateFormat("MM-dd HH:mm", Locale.getDefault())
+        private val otherYearSdf = SimpleDateFormat("yyyy-MM-dd", Locale.getDefault())
+
+        @JvmStatic
+        fun format(time: Date?): String {
+            return if (time == null) {
+                "??"
+            } else if (android.text.format.DateUtils.isToday(time.time)) {
+                this.sameDaySdf.format(time)
+            } else if (time.getYear() == (Calendar.getInstance().get(Calendar.YEAR) - 1900)) {
```
connyduck, line 35: `getYear()` is deprecated, please use something else

connyduck (changes requested):
Oh I like this idea very much 😍 

But the code needs some improvements:

- Can you add unit tests for `AbsoluteTime` please?
- Maybe better call it `AbsoluteTimeFormatter` so it is clear what it is for?

ariasuni:
Ok so I fixed everything you mentioned, and also taken into account that for polls we probably always want the hours and minutes, besides year-month-day.

app/src/main/java/com/keylesspalace/tusky/util/AbsoluteTimeFormatter.kt
```diff
@@ -0,0 +1,50 @@
...
+        } else if (shortFormat) {
+            otherYearSdf.format(time)
+        } else {
+            otherYearCompleteSdf.format(time)
+        }
+    }
+
+    companion object {
+        @JvmStatic
+        private fun getYear(time: Date): Int {
```
connyduck, line 46: Could make that an extension function `private fun Date.year(): Int`.

Also, @JvmStatic is not needed when only called from Kotlin.

connyduck (changes requested):
- Unit tests? 🥺 
- KTlint complains about an unused import

ariasuni:
> * Unit tests? pleading_face

Ah yeah I’ll do that. It feels like it would easier to change `getYear()` to `isSameYear()` instead of your proposal, so that it can contain all functions related to «today» and timezones, and I can mock it (except `android.text.format.DateUtils.isToday()` which I guess I can mock as well).

>     * KTlint complains about an unused import

It doesn’t complain locally, and well I could remove the import but it’s used and would break the code.
https://github.com/tuskyapp/Tusky/pull/2413/files#diff-11b44ae8dd2d90d9e4ea45a8ca43079b3a0773c4cb42eda514d491e6c24b5884R32

ariasuni:
Advice on how to create unit test would help. I tried to copy-paste things from other files and it just doesn’t work and it’s painful. Trying to create a class with `@RunWith(AndroidJUnit4::class)` with inside a function with `@Test`, and compilation spit me a `error: incompatible types: NonExistentClass cannot be converted to Annotation @error.NonExistentClass()` for some reason.

Since there’s not test in the files of the same directory, I’m assuming something’s not properly set up to be able to have unit test in this directory.

charlag:
Unit tests (in `test` directory) don't need `AndroidJunit` stuff

ariasuni:
> Unit tests (in `test` directory) don't need `AndroidJunit` stuff

OK but I still got the error with `@Test`. Or should I not annotate the function and there is another way for Junit to now what’s a test or not?

charlag:
I'm not sure what the problem is tbh, do you get the same error?
I believe you do need `@Test`

connyduck:
Oh no the time apis you used are not available on Android until api level 26. I think we have to introduce [desugaring](https://developer.android.com/studio/write/java8-support#library-desugaring) to get this to work.

> It doesn’t complain locally, and well I could remove the import but it’s used and would break the code.

Yes, it is used, but because the classes are in the same package it can be removed without breaking anything

> Advice on how to create unit test would help

Timebased unit tests are usually done by giving the unit under test a "now" parameter. This way you don't need to mock anything.

Something like this should work:

```
package com.keylesspalace.tusky.util

import org.junit.Assert.assertEquals
import org.junit.Test

class AbsoluteTimeFormatterTest {

    private val formatter = AbsoluteTimeFormatter()

    @Test
    fun `test formatting`() {
        assertEquals("...", formatter.format(time = ..., now = ...))
    }

    ...
}

```

I'm sorry if this is a bit complicated. If you want I can take this feature over.

ariasuni:
> Oh no the time apis you used are not available on Android until api level 26. I think we have to introduce [desugaring](https://developer.android.com/studio/write/java8-support#library-desugaring) to get this to work.

Yeah it looks like something that I couldn’t do alone.

For the rest, I pushed a version with the attempted unit tests and latest changes.

connyduck:
So this is how I'd do it.
Only one thing im not sure yet, this seems to be very American (?) formatting, I as an Austrian would expect e.g. days as DD.MM.YYY and not YYYY-MM-DD 🤔

charlag:
YYYY-mm-d is ISO

ariasuni:
In French I would expect dd/MM/YYYY (if anything putting year then month then day is more Japanese…) but I expect absolute time as ISO in general I’d say? Anyway, as written in my first comment:
> I changed from / to - to better reflect the fact the format put the month before the day, like ISO 8601 and and RFC 3339

Another possibility is to use a localized short format, but I’m not sure that exists. Using the ISO/RFC format is better than the current way it’s done, anyway.

connyduck:
Ok it seems there is really no way in Java (even with the new time apis) to get only day + month formatted in a localized way?? This sucks. 

Anyway this PR makes things better than they were before, I'd merge it like this if you agree? @charlag @ariasuni

ariasuni:
Can’t approve my own PR, but I reviewed the changes and I’m satisfied with them :slightly_smiling_face:

PR #2414, opened 2022-04-02
move Html parsing to ViewData

connyduck:
advantages:
- no more custom `equals` and `hashcode` due to no more `Spanned` in models
- performance improvement since status content is not parsed before being written to the database

To make this work, an additional mapping had to be introduced for conversations and reports. But I think this is actually good, as we now have a clearer separation of concerns: `ViewData` classes are for displaying, `Entity` classes for database and classes in the `model` package for json.


app/src/main/java/com/keylesspalace/tusky/components/conversation/ConversationsViewModel.kt
```diff
@@ -44,104 +46,117 @@ class ConversationsViewModel @Inject constructor(
         pagingSourceFactory = { database.conversationDao().conversationsForAccount(accountManager.activeAccount!!.id) }
     )
         .flow
+        .map { pagingData ->
+            pagingData.map { conversation -> conversation.toViewData() }
```
charlag, line 50: Does this convert all of them each time any item changes? It's most likely okay, just for my understanding

app/src/main/java/com/keylesspalace/tusky/db/AppDatabase.java
```diff
@@ -483,4 +483,41 @@ public void migrate(@NonNull SupportSQLiteDatabase database) {
             database.execSQL("DELETE FROM `TimelineStatusEntity`");
         }
     };
+
+    public static final Migration MIGRATION_31_32 = new Migration(31, 32) {
```
charlag, line 487: Is it an error to have a column that's not in the model?

app/src/main/java/com/keylesspalace/tusky/util/StatusParsingHelper.kt
```diff
@@ -0,0 +1,62 @@
...
+package com.keylesspalace.tusky.util
+
+import android.text.SpannableStringBuilder
+import android.text.Spanned
+import androidx.core.text.parseAsHtml
+
+/**
+ * parse a String containing html from the Mastodon api to Spanned
+ */
+fun String.toSpanned(): Spanned {
```
charlag, line 25: I think the name `toSpanned()` on all Strings might be a bit confusing. I think it should donate that this is HTML and/or that we are parsing it.
`parseSpannedHtml()` or something like that? `parseMastodonHTML()`?

app/src/main/java/com/keylesspalace/tusky/components/conversation/ConversationsViewModel.kt
connyduck: I think so, but i'm not 100% sure

app/src/main/java/com/keylesspalace/tusky/util/StatusParsingHelper.kt
connyduck: true

app/src/main/java/com/keylesspalace/tusky/components/conversation/ConversationsViewModel.kt
charlag: I'm afraid that it will be parsing HTML a lot in a few places

connyduck: I'll check how it actually works

connyduck: Ok I put a counter and a log statement into the mapping function, and it looks like there is some kind of working set that needs to be remapped on every revalidation. It is not the whole timeline but around 100 statuses. This is really not ideal as it also seems to run on the main thread. I need to dig deeper into this, thanks for pointing it out.

charlag: Thanks for checking this! I think it's good that it's not the whole set and 100 statuses seems fine but main thread is not great ofc. Can we use `withContext()` or `flowOn()`?

connyduck: For some reason `flowOn` alone does not work, there needs to be an executor passed into the mapping and filter functions, which is weird. 

I will refactor it to filter before mapping in another PR.

connyduck:
Here is a google engineer explaining why it works like that https://issuetracker.google.com/u/1/issues/228337195

charlag:
@connyduck I see, that's really confusing

PR #2415, opened 2022-04-02
introduce KotlinResultCallAdapter for nice suspending network calls

connyduck:


app/src/main/java/com/keylesspalace/tusky/components/announcements/AnnouncementsViewModel.kt
```diff
@@ -56,8 +57,9 @@ class AnnouncementsViewModel @Inject constructor(
             appDatabase.instanceDao().loadMetadataForInstance(accountManager.activeAccount?.domain!!)
                 .map<Either<InstanceEntity, Instance>> { Either.Left(it) }
                 .onErrorResumeNext {
-                    mastodonApi.getInstance()
-                        .map { Either.Right(it) }
+                    rxSingle {
```
charlag, line 60: honestly should just replace `mastodonApi.getCustomEmojis(),` with coroutines too and just remove this whole mess

app/src/main/java/com/keylesspalace/tusky/components/login/LoginActivity.kt
```diff
@@ -166,32 +165,33 @@ class LoginActivity : BaseActivity(), Injectable {
...
-                .putString(DOMAIN, domain)
-                .putString(CLIENT_ID, credentials.clientId)
-                .putString(CLIENT_SECRET, credentials.clientSecret)
-                .apply()
-
-            redirectUserToAuthorizeAndLogin(domain, credentials.clientId)
+            mastodonApi.authenticateApp(
+                domain, getString(R.string.app_name), oauthRedirectUri,
+                OAUTH_SCOPES, getString(R.string.tusky_website)
+            ).fold(
```
charlag, line 171: I must say I don't like using `fold` except for one very specific case: when you assign to a variable and provide default value. I would argue that `try/catch` is there for a reason.

app/src/main/java/com/keylesspalace/tusky/components/announcements/AnnouncementsViewModel.kt
connyduck: I was planning to do this in another PR, but I can also do it with this one

app/src/main/java/com/keylesspalace/tusky/components/login/LoginActivity.kt
connyduck: But `fold` returns a value, you can do something like

```
val data  = mastodonApi.getData().fold(
    { // map data
    },
    { // return default value
    }
)
```

If you just want a default value you can do 

```
val data = mastodonApi.getData().getOrDefault(/* whatever default you need */)
```

app/src/main/java/com/keylesspalace/tusky/components/announcements/AnnouncementsViewModel.kt
charlag: I think either way is fine

app/src/main/java/com/keylesspalace/tusky/components/login/LoginActivity.kt
charlag: Ooooh I got some language soup in my head, I'm sorry :sweat_smile: 
Of course it's `fold()`, it's a `Result`! 
Don't listen to me, everything is great.

PR #2420, opened 2022-04-06
fix crash when applying new emoji font on Android 12

connyduck:


charlag (approved):
Ah it's not available yes

PR #2424, opened 2022-04-08
fix black theme on Android 12

connyduck:
This new splashscreen is causing so much trouble 😠 

before & after

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -22,6 +22,22 @@
         android:usesCleartextTraffic="false">
 
         <activity
+            android:name=".SplashActivity"
+            android:theme="@style/SplashTheme"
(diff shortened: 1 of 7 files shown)


charlag (approved):
I don't know what the bug was but okay

charlag:
I mean I *see* it but I don't know *why*

connyduck:
Wait the splash screen does still not work as intended. It shows on some devices after Login. Putting it on LoginActivity instead of MainActivity would show it when adding an account. 🙄 
SplashActivity has to come back.

charlag:
Multiple activities was a mistake

connyduck:
> Multiple activities was a mistake

Yes. Another thing we should refactor D:

charlag (approved):

PR #2434, opened 2022-04-15
upgrade Kotlin and Coroutines

connyduck:


app/build.gradle
```diff
@@ -99,8 +99,6 @@ ext.materialdrawerVersion = '8.4.5'
 
 // if libraries are changed here, they should also be changed in LicenseActivity
 dependencies {
-    implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk8:$kotlin_version"
```
connyduck, line 102: The Kotlin Plugin automatically adds stdlib dependency to the build, no need to specify it explicitly

charlag: oh, it's new, right?

charlag (approved):

app/build.gradle
connyduck: yes, since a few versions

PR #2446, opened 2022-04-19
refactor compose & announcements to coroutines

connyduck:
follow-up to https://github.com/tuskyapp/Tusky/pull/2415

- no more `AsyncTask` (we didn't even use it anymore, but is was still there)
- no more database queries with RxJava
- refactored the emoji & instance data caching out into a dedicated `InstanceInfoRepository`


app/src/main/java/com/keylesspalace/tusky/components/compose/ComposeViewModel.kt
```diff
@@ -99,74 +87,40 @@ class ComposeViewModel @Inject constructor(
...
-                { throwable ->
-                    // this can happen on network error when no cached data is available
-                    Log.w(TAG, "error loading instance data", throwable)
-                }
-            )
-            .autoDispose()
     }
 
-    fun pickMedia(uri: Uri, description: String? = null): LiveData<Either<Throwable, QueuedMedia>> {
+    fun pickMedia(mediaUri: Uri, description: String? = null): LiveData<Either<Throwable, QueuedMedia>> {
```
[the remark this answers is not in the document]
charlag: Aaaactually we should just make it a suspending function that returns `Result` (to make clear that it can fail)

app/src/main/java/com/keylesspalace/tusky/components/compose/ImageDownsizer.kt
```diff
@@ -0,0 +1,113 @@
...
+    tempFile: File
+): Boolean {
+
+    var inputStream: InputStream?
+    inputStream = try {
+        contentResolver.openInputStream(uri)
+    } catch (e: FileNotFoundException) {
+        return false
+    }
+    // Initially, just get the image dimensions.
```
charlag, line 53: some comments are twice here

app/src/main/java/com/keylesspalace/tusky/components/compose/ImageDownsizer.kt
```diff
@@ -0,0 +1,113 @@
...
+ */
+fun downsizeImage(
+    uri: Uri,
+    sizeLimit: Int,
+    contentResolver: ContentResolver,
+    tempFile: File
+): Boolean {
+
+    var inputStream: InputStream?
+    inputStream = try {
```
charlag, line 48: why do we use the same variable twice?
Can we use `with` or `use`?

app/src/main/java/com/keylesspalace/tusky/components/compose/MediaUploader.kt
```diff
@@ -72,61 +77,40 @@ class MediaUploader @Inject constructor(
...
 ) {
-    fun uploadMedia(media: QueuedMedia): Observable<UploadEvent> {
-        return Observable
-            .fromCallable {
-                if (shouldResizeMedia(media)) {
-                    downsize(media)
-                } else media
+
+    @OptIn(ExperimentalCoroutinesApi::class)
+    fun uploadMedia(media: QueuedMedia): Flow<UploadEvent> {
```
charlag, line 82: I don't think we need a `Flow` here, do we ever emit more than one event? I don't think so

app/src/main/java/com/keylesspalace/tusky/components/compose/dialog/CaptionDialog.kt
```diff
@@ -77,12 +77,11 @@ fun <T> T.makeCaptionDialog(
     input.filters = arrayOf(InputFilter.LengthFilter(MEDIA_DESCRIPTION_CHARACTER_LIMIT))
 
     val okListener = { dialog: DialogInterface, _: Int ->
-        onUpdateDescription(input.text.toString())
-        withLifecycleContext {
-            onUpdateDescription(input.text.toString())
-                .observe { success -> if (!success) showFailedCaptionMessage() }
+        lifecycleScope.launch {
```
charlag, line 80: how does `lifecycleScope` work if we dismiss it right away?

app/src/main/java/com/keylesspalace/tusky/components/instanceinfo/InstanceInfoRepository.kt
```diff
@@ -0,0 +1,93 @@
...
+class InstanceInfoRepository @Inject constructor(
+    private val api: MastodonApi,
+    db: AppDatabase,
+    accountManager: AccountManager
+) {
+
+    private val dao = db.instanceDao()
+    private val instanceName = accountManager.activeAccount!!.domain
+
+    suspend fun getEmojis() = withContext(Dispatchers.IO) {
```
charlag, line 43: I would declare return type tbh, it's non-trivial enough

app/src/main/java/com/keylesspalace/tusky/util/Either.kt
```diff
@@ -45,3 +45,6 @@ sealed class Either<out L, out R> {
         }
     }
 }
+
+fun <V> V.left(): Either<V, Nothing> = Either.Left(this)
```
charlag, line 49: maybe `asLeft()` ? otherwise it looks like getter

app/src/main/java/com/keylesspalace/tusky/components/instanceinfo/InstanceInfoRepository.kt
```diff
@@ -0,0 +1,93 @@
...
+import kotlinx.coroutines.withContext
+import javax.inject.Inject
+
+class InstanceInfoRepository @Inject constructor(
+    private val api: MastodonApi,
+    db: AppDatabase,
+    accountManager: AccountManager
+) {
+
+    private val dao = db.instanceDao()
```
charlag, line 35: We use dao only as a fallback but we could also have some cache interval in which we don't re-load it again. It almost never changes anyway. But this is for another day I think.

charlag (approved):
I love this refactoring!

app/src/main/java/com/keylesspalace/tusky/components/compose/dialog/CaptionDialog.kt
connyduck: if the scope is cancelled, all calls in it are as well

app/src/main/java/com/keylesspalace/tusky/util/Either.kt
connyduck: Ah, I did not end up using those extensions.

app/src/main/java/com/keylesspalace/tusky/components/compose/dialog/CaptionDialog.kt
charlag: That's what I mean, we fire it and then dismiss the dialog right away which will close the scope

connyduck: It is the scope of the activity, it is fine

app/src/main/java/com/keylesspalace/tusky/components/compose/ImageDownsizer.kt
connyduck: Hmm, with use/with it gets a bit complicated since there are three streams involved, I'll leave it like it is

charlag (approved):

Issue #2444, opened 2022-04-18
If launched landscape on mobile, bottom bar becomes tablet style


KSPAtlas:
I set Tusky to have the timelines at the bottom, but when I start the app in landscape mode, I get a tablet style bar that persists when you switch to portrait.

* * * *
- Tusky Version: 16.0
- Android Version: 12 (GrapheneOS)
- Android Device: Google Pixel 3
- Mastodon instance (if applicable):

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

PR #2447, opened 2022-04-19
never collapse bottom app bar tabs

connyduck:
closes https://github.com/tuskyapp/Tusky/issues/2444

### DIFF
--- a/app/src/main/res/layout/activity_main.xml
+++ b/app/src/main/res/layout/activity_main.xml
@@ -58,6 +58,7 @@
                 android:layout_width="match_parent"
                 android:layout_height="?attr/actionBarSize"
                 app:tabIndicator="@null"
+                app:tabGravity="fill"
                 app:tabMode="fixed" />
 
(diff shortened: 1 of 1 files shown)


app/src/main/res/layout/activity_main.xml
```diff
@@ -58,6 +58,7 @@
                 android:layout_width="match_parent"
                 android:layout_height="?attr/actionBarSize"
                 app:tabIndicator="@null"
+                app:tabGravity="fill"
```
charlag, line 61: god, docs say *nothing* about it. What does it do? what's `tabGravity` even?

connyduck: I don't know. We had the same bug for top tabs and it was the same fix https://github.com/tuskyapp/Tusky/pull/239 (must have been one of my first PRs here :D)

charlag: Ah, I understand the issue Noe that there's a screenshot! LGTM! (also, woah!)

Issue #2441, opened 2022-04-16
Tusky crashes when trying to view a notification


rudolfmu:
Tusky crashes when i try to view a specific notification.
Details:
https://fosstodon.org/web/@rudolf/108139387376584278
Logcat:
<<< log_count = 39 >>>
[04-16 21:25:03.615 1326:1326 D/PhoneStatusBar]
disable: < expand icons* alerts system_info* back home recent clock search quick_settings >

[04-16 21:25:18.652 755:755 V/NotificationService]
pkg=com.dp.logcatapp canInterrupt=false intercept=true

[04-16 21:25:22.125 755:2546 I/ActivityManager]
START u0 {flg=0x10804000 cmp=com.android.systemui/.recents.RecentsActivity} from uid 10033 on display 0

[04-16 21:25:22.152 17399:17399 D/LogcatReader]
[Logcat] onActivityInBackground

[04-16 21:25:22.301 755:765 I/art]
Background partial concurrent mark sweep GC freed 71079(4MB) AllocSpace objects, 15(364KB) LOS objects, 33% free, 13MB/20MB, paused 1.871ms total 126.904ms

[04-16 21:25:24.359 755:2553 E/ActivityManager]
applyOptionsLocked: Unknown animationType=0

[04-16 21:25:26.760 17483:17488 I/art]
Do full code cache collection, code=441KB, data=505KB

[04-16 21:25:26.762 17483:17488 I/art]
Starting a blocking GC JitCodeCache

[04-16 21:25:26.762 17483:17488 I/art]
After code cache collection, code=385KB, data=390KB

[04-16 21:25:27.226 17483:17483 D/AndroidRuntime]
Shutting down VM

[04-16 21:25:27.229 17483:17483 E/AndroidRuntime]
FATAL EXCEPTION: main
Process: com.keylesspalace.tusky, PID: 17483
java.lang.NullPointerException: Attempt to read from field 'boolean g7.g.d' on a null object reference
	at w5.s0.O(SourceFile:1)
	at w5.z.w(SourceFile:182)
	at w5.z.n(SourceFile:1)
	at l1.h1.a(SourceFile:9)
	at l1.a2.j(SourceFile:162)
	at l1.s0.c(SourceFile:9)
	at androidx.recyclerview.widget.LinearLayoutManager.g1(SourceFile:1)
	at androidx.recyclerview.widget.LinearLayoutManager.T0(SourceFile:12)
[...]

Issue #2252, opened 2021-10-05
Notifications tab crashes on status `null`?


janboddez:
Had this happen a few times, and I'm fairly sure the object below is what triggers the latest occurrence (the web app seems to display it fine, Tusky just crashes):
```
{
	"3": {
		"id": "1139",
		"type": "poll",
		"created_at": "2021-10-02T09:16:03.213Z",
		"account": {
			"id": "7178",
			"username": "brainblasted",
			"acct": "brainblasted@social.libre.fi",
			"display_name": "Chris",
			"locked": true,
			"bot": false,
			"discoverable": false,
			"group": false,
			"created_at": "2020-02-28T00:00:00.000Z",
			"note": "Admin of social.libre.fi. GNOME developer.<br>DMs from mutuals are okay, and boosts are okay for all public posts.<br><br>Blank profiles and FOSS-only profiles will be rejected.<br><br><a class=\"hashtag\" href=\"https://social.libre.fi/tag/actuallyautistic\" rel=\"nofollow noopener noreferrer\" target=\"_blank\">#actuallyautistic</a>",
			"url": "https://social.libre.fi/users/brainblasted",
			"avatar": "https://media.geekcompass.com/geekcompass/cache/accounts/avatars/000/007/178/original/aaf07701b4726dcb.png",
			"avatar_static": "https://media.geekcompass.com/geekcompass/cache/accounts/avatars/000/007/178/original/aaf07701b4726dcb.png",
			"header": "https://media.geekcompass.com/geekcompass/accounts/headers/000/007/178/original/712f28da941cf933.jpeg",
			"header_static": "https://media.geekcompass.com/geekcompass/accounts/headers/000/007/178/original/712f28da941cf933.jpeg",
			"followers_count": 1685,
			"following_count": 643,
			"statuses_count": 90,
			"last_status_at": "2021-10-04",
			"emojis": [],
			"fields": [
				{
					"name": "Patreon",
					"value": "<a href=\"https://www.patreon.com/chrisgnome\" rel=\"nofollow noopener noreferrer\" target=\"_blank\">https://www.patreon.com/chrisgnome</a>",
					"verified_at": null
				},
				{
					"name": "Pronouns",
					"value": "he/him/his or they/them/theirs",
					"verified_at": null
				},
				{
					"name": "GitHub Sponsors",
[...]

PR #2449, opened 2022-04-19
guard against the status of a notification being null in rare cases

connyduck:
closes https://github.com/tuskyapp/Tusky/issues/2441
closes https://github.com/tuskyapp/Tusky/issues/2252

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/adapter/NotificationsAdapter.java
+++ b/app/src/main/java/com/keylesspalace/tusky/adapter/NotificationsAdapter.java
@@ -183,7 +183,9 @@ public class NotificationsAdapter extends RecyclerView.Adapter {
                     if (status == null) {
                         holder.showStatusContent(false);
                     } else {
-                        holder.showStatusContent(true);
+                        if (payloads == null) {
+                            holder.showStatusContent(true);
+                        }
                         holder.setupWithStatus(status, statusListener, statusDisplayOptions, payloadForHolder);
                     }
                     if (concreteNotificaton.getType() == Notification.Type.POLL) {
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/keylesspalace/tusky/adapter/StatusBaseViewHolder.java
```diff
@@ -1133,6 +1133,28 @@ protected void setupCard(
...
+        avatar.setVisibility(visibility);
+        avatarInset.setVisibility(visibility);
+        displayName.setVisibility(visibility);
+        username.setVisibility(visibility);
+        timestampInfo.setVisibility(visibility);
+        contentWarningDescription.setVisibility(visibility);
+        contentWarningButton.setVisibility(visibility);
+        content.setVisibility(visibility);
+        cardView.setVisibility(visibility);
+        itemView.findViewById(R.id.status_media_preview_container).setVisibility(visibility);
```
charlag, line 1147: uhm wouldn't `findViewById()` be *super* slow, especially for each element?

charlag:
Also, do we know when it's null? If yes, we should leave a comment

app/src/main/java/com/keylesspalace/tusky/adapter/StatusBaseViewHolder.java
connyduck: Not super slow, but it is indeed not ideal. Will change.

connyduck:
> Also, do we know when it's null? If yes, we should leave a comment

I think it happens in race condition scenarios on the server, e.g. when a poll is deleted at the same moment it ends or incompatibilities between different fedi software or something. Could never reproduce, or I would have reported it. Also users from both Mastodon and Pleroma had this happen to them.

app/src/main/java/com/keylesspalace/tusky/adapter/NotificationsAdapter.java
```diff
@@ -180,8 +180,14 @@ private void bindViewHolder(@NonNull RecyclerView.ViewHolder viewHolder, int pos
                 case VIEW_TYPE_STATUS: {
                     StatusViewHolder holder = (StatusViewHolder) viewHolder;
                     StatusViewData.Concrete status = concreteNotificaton.getStatusViewData();
-                    holder.setupWithStatus(status,
-                            statusListener, statusDisplayOptions, payloadForHolder);
+                    if (status == null) {
```
charlag, line 183:
```suggestion
                    // this can happen rarely on the server, seems to be a race condition
                    if (status == null) {
```

charlag (approved):

PR #2468, opened 2022-04-27
EmojiCompat fix

C1710:
- Added back the InitCallback
- Made extra sure that EmojiCompat.process (in the MainActivity) is only called if EmojiCompat has gotten to LOAD_STATE_SUCEEDED

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/MainActivity.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/MainActivity.kt
@@ -154,6 +154,14 @@ class MainActivity : BottomSheetActivity(), ActionButtonActivity, HasAndroidInje
     // We need to know if the emoji pack has been changed
     private var selectedEmojiPack: String? = null
 
+    private val emojiInitCallback = object : EmojiCompat.InitCallback() {
+        override fun onInitialized() {
+            if (!isDestroyed) {
+                updateProfiles()
+            }
+        }
+    }
+
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
 
@@ -349,6 +357,11 @@ class MainActivity : BottomSheetActivity(), ActionButtonActivity, HasAndroidInje
         }
     }
 
+    override fun onDestroy() {
+        super.onDestroy()
+        EmojiCompat.get().unregisterInitCallback(emojiInitCallback)
+    }
+
     private fun forwardShare(intent: Intent) {
         val composeIntent = Intent(this, ComposeActivity::class.java)
         composeIntent.action = intent.action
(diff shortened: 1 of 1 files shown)


connyduck:
Wait a second... EmojiCompat.process was only there because we could not change the library to use an `EmojiTextView`, but now with emoji2 that is no longer necessary and we can just remove it and it will still show the right emojis

C1710:
> Wait a second... EmojiCompat.process was only there because we could not change the library to use an `EmojiTextView`, but now with emoji2 that is no longer necessary and we can just remove it and it will still show the right emojis

I had the same thought but wasn't really sure whether a normal `TextView` is used there 😅

C1710:
I just checked and right now it does _not_ work without `EmojiCompat.process`. I guess the library used for the Navigation Drawer does not use AppCompat 1.4.x yet. It's the same with the Google Material TextInputViews (e.g. on the profile editing screen)

C1710:
So, materialdrawer 9.0.0 seems to cause a crash...

C1710:
So, for now I would just leave it like this and once the materialdrawer library is/can be updated it can be removed

connyduck:
What did you try? This works on Android 12 emulator and on Android 12 Fairphone 4

```kotlin
        val profiles: MutableList<IProfile> = accountManager.getAllAccountsOrderedByActive().map { acc ->
            ProfileDrawerItem().apply {
                isSelected = acc.isActive
                nameText = acc.displayName.emojify(acc.emojis, header, animateEmojis)
                iconUrl = acc.profilePictureUrl
                isNameShown = true
                identifier = acc.id
                descriptionText = acc.fullName
            }
        }.toMutableList()
```

The right emojis are displayed: 



The drawer library does not have to depend on a certain AppCompat version, as long as we do it is used everywhere in the app.

C1710:
> What did you try? This works on Android 12 emulator and on Android 11 Fairphone 4
> 
> ```kotlin
>         val profiles: MutableList<IProfile> = accountManager.getAllAccountsOrderedByActive().map { acc ->
>             ProfileDrawerItem().apply {
>                 isSelected = acc.isActive
>                 nameText = acc.displayName.emojify(acc.emojis, header, animateEmojis)
>                 iconUrl = acc.profilePictureUrl
>                 isNameShown = true
>                 identifier = acc.id
>                 descriptionText = acc.fullName
>             }
>         }.toMutableList()
> ```
> 
> The right emojis are displayed: 
> 
> The drawer library does not have to depend on a certain AppCompat version, as long as we do it is used everywhere in the app.

Interesting. I tried it that way (without EmojiCompat.process) and it did not use EmojiCompat.  
And when I changed the dependency version to 9.0.0, it crashed 🤔

C1710:
Wait a minute... It must have reset the selected emoji pack for some reason 😅
Sorry!

connyduck (approved):

Issue #2439, opened 2022-04-16
Remove version detection for post scheduling


connyduck:
We currently check the version of the active Mastodon instance to decide whether to enable post scheduling or not. 
This was relevant 2 years agon when post scheduling was new, but now virtually no instance runs a version below 2.7.0 anymore.
The [code](https://github.com/tuskyapp/Tusky/blob/develop/app/src/main/java/com/keylesspalace/tusky/util/VersionUtils.java) and the corresponding tests can be removed.

Arteneko:
Trying my hand at it right now

guntbert:
@connyduck Do you want to keep the no longer needed member `supportsScheduled` in `InstanceInfo` (and just set its value to `true` when calling the constructor)?
If we remove it that might have consequences for code calling `InstanceInfo()`

guntbert:
@Arteneko are you working on this? I don´t want to step on your toes :wink:

connyduck:
> @connyduck Do you want to keep the no longer needed member `supportsScheduled` in `InstanceInfo` (and just set its value to `true` when calling the constructor)?
> If we remove it that might have consequences for code calling `InstanceInfo()`

I'd remove it (and adjust the callsites accordingly ofc)

Arteneko:
I am not, tried to work on it but I can't seem to even compile the app or run the tests due to random JVM errors

PR #2469, opened 2022-04-27
Remove code to check the server version

guntbert:
will solve #2439


app/src/main/java/com/keylesspalace/tusky/components/compose/ComposeActivity.kt
```diff
@@ -335,7 +335,7 @@ class ComposeActivity :
                 maximumTootCharacters = instanceData.maxChars
                 charactersReservedPerUrl = instanceData.charactersReservedPerUrl
                 updateVisibleCharactersLeft()
-                binding.composeScheduleButton.visible(instanceData.supportsScheduled)
+                binding.composeScheduleButton.visible(true)
```
connyduck, line 338: I think this line is entirely unnecessary as the button is visible after layout inflation (if not, change it in the layout)

connyduck (changes requested):
Thanks for the help!

app/src/main/java/com/keylesspalace/tusky/components/compose/ComposeActivity.kt
guntbert: Ah, I found the button - lets see if it stays visible when I remove the line

guntbert: All looks ok - the button is visible

connyduck (approved):

guntbert:
@connyduck thank you for the guidance :bowing_man:

Issue #2379, opened 2022-03-11
Please add an option to select text in post


joenepraat:
Sometimes I want to research a word or a line, in e.g. Wikipedia or I want to translate a word using a translation app. This is not possible, because you can't only select all text or nothing. So please make text selectable, like in a web-browser.

* * * *
- Tusky Version: 16.0

- [X] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

connyduck:
https://github.com/tuskyapp/Tusky/pull/2472

PR #2472, opened 2022-04-28
Issue 2379: Make it possible to select text in posts.

Gryzor:
### Objective
* Allow the user to use native Android gestures (long press-drag) to start selecting text.
* Remove conflicting listeners that prevent users from selecting text.

### Issue
* #2379 

### Description
* As simple as it sounds, setting `textIsSelectable` to `true` in the XML should be enough, but in some complex layouts and hierarchies, other gestures may be taking precedence and prevent this from happening.
* This is my 1st PR and I am very new to this codebase, so please _let me know if I am missing something_, as it took me a bit to figure out which view Holders were being used where/when. So I may have modified one that I didn't need to, or missed one altogether.
* The LongPress click listener in the ThreadView has been _removed_ from the content. Although I think it's a nice to have feature, if the need to select text manually is preferred (this PR) then the long press would conflict (and take precedence), making it impossible for users to manually select text. Please let me know if this is ok.

### Known issues
* I haven't been able to make it _reliably_ work in the "main/home" list, because of the conflict of gestures (I believe Android is having hard time deciding where the gesture goes). On one hand there's the Drawer which takes swipe to the right, then there's the TabLayouts/ViewPager with the different tabs, this occupies the swipe to the left gesture; then there's the actual recyclerView with a SwipeToRefresh layout, driving swipe up/down gestures, so all four directions are taken. On Android, long press and drag over text is how you initiate the selection mechanism, so naturally this conflicts with the above more often than not.
That being said, it *does work* sometimes, so any pointers or ideas are appreciated.
I believe that even then, as a workaround, users can always tap the post, and manually copy in the Thread view, where it works all the time.

### Tested
* OnePlus 6p / Android 11
* Samsung S20 / Android 12


connyduck:
Please only change the detailed status, I think it is enough if there is one place to copy the content.

Also, I would remove the long press listener and the copy_to_clipboard_success string entirely, having multiple ways to copy is not necessary.

connyduck:
Please also remove the now unused string. It should be enough to remove it from the default translation, Weblate will send a PR to remove it from the others, but you can also do all 45 translations.
But the rest looks good!

Gryzor:
@connyduck done, did all translations as it was easy to do. Had to solve a merge conflict but it was just a coincidence that the string above changed in that locale `hu`. /shrug

connyduck (approved):

joenepraat:
Great! So now we can just copy a word or a sentence, search that string in e.g. Wikipedia and a translate app. Thanks!

Issue #2478, opened 2022-04-29
Captions for multiple images


qoheniac:
When I delete and re-draft a comment with multiple images, I can only add a caption to the first one. If I try to add a caption to the second one, a pop-up opens as expected and I can write a caption, but when I hit OK instead of adding the caption to the second image, it overwrites that of the first one.

* * * *
- Tusky Version: 17.0
- Android Version: 10
- Android Device: Fairphone 2
- Mastodon instance (if applicable): layer8.space

- [X] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

PR #2479, opened 2022-04-29
fix race condition where multiple uploaded media can get same internal id

connyduck:
closes https://github.com/tuskyapp/Tusky/issues/2478

Basically it is a very bad idea to use `System.currentTimeMillis()` for creating unambiguous ids. And if the media all has the same id, all edits go to the first one. 
I also refactored the media `LiveData` to `StateFlow` to get rid of all the `!!`. Will follow up with another PR where all `LiveData`s get replaced by `StateFlow`.

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/components/compose/ComposeViewModel.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/components/compose/ComposeViewModel.kt
@@ -21,6 +21,7 @@ import androidx.core.net.toUri
 import androidx.lifecycle.LiveData
 import androidx.lifecycle.MutableLiveData
 import androidx.lifecycle.ViewModel
+import androidx.lifecycle.asLiveData
 import androidx.lifecycle.viewModelScope
 import com.keylesspalace.tusky.components.compose.ComposeActivity.QueuedMedia
 import com.keylesspalace.tusky.components.drafts.DraftHelper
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/keylesspalace/tusky/components/compose/ComposeViewModel.kt
```diff
@@ -118,29 +120,30 @@ class ComposeViewModel @Inject constructor(
...
 
-    private fun addMediaToQueue(
+    private suspend fun addMediaToQueue(
         type: QueuedMedia.Type,
         uri: Uri,
         mediaSize: Long,
         description: String? = null
-    ): QueuedMedia {
+    ): QueuedMedia = withContext(Dispatchers.Main) {
+        // running on main dispatcher to avoid race condition when adding multiple media at once, e.g. when restoring from drafts
```
charlag, line 129: how about either `syncrhonized` for id generation or better yet just make an `AtomicInteger` as a field for max id? Seems more robust

app/src/main/java/com/keylesspalace/tusky/components/compose/ComposeViewModel.kt
```diff
@@ -149,29 +152,34 @@ class ComposeViewModel @Inject constructor(
...
+        val mediaItem = QueuedMedia(
+            localId = (media.value.maxOfOrNull { it.localId } ?: 0) + 1,
+            uri = uri,
+            type = type,
+            mediaSize = 0,
+            uploadPercent = -1,
+            id = id,
+            description = description
+        )
+        media.value = media.value + mediaItem
```
charlag, line 177: why do we `synchronized(media)` in one place but not all other places where we modify it? Isn't it defeating the point a bit? There's also [Mutex](https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.sync/-mutex/) which has nifty [withLock](https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.sync/with-lock.html) (it's kind of a lame lock that doesn't contain data but still better than nothing)

charlag (approved):
I have some suggestions at your discretion.

connyduck:
`MutableStateFlow` has extensions to atomically update it 🤯

charlag (approved):
:+1:
question: can it be that the last media is removed, then we add another one, it gets the same id and then we do something to it (as it has the same id)?
I would still feel safer with separate atomic id counter that only increases but I don't think it's a deal-breaker.

connyduck:
I checked, and we do nothing once we discard an item so I think it is fine.

Issue #2477, opened 2022-04-28
Profile screen improvement: show account creation date


JPLeBreton:
When viewing another user's profile, their account's creation date should appear beneath everything else in the profile view (and above their toots etc). Mastodon's web UI has this info, and it's useful - particularly during periods like now (late April 2022) when lots of accounts are following you.

- [X] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

connyduck:
https://github.com/tuskyapp/Tusky/pull/2480

PR #2480, opened 2022-04-29
Issue 2477: Show account's creation date in Profile.

Gryzor:
### Objective
* Show the account's creation date in the profile, similar to what the Web version of Mastodon does.

### Issue
* #2477 

### Description
* According to [mastodon's API](https://docs.joinmastodon.org/methods/accounts/#retrieve-information), the date the account is created comes as part of the payload as `created_at`.
* I parsed using `SimpleDateFormat` because that's what I've seen the rest of the project is doing. I'd prefer java.time APIs, but for consistency, I used the same.
* I tried to account for potential parse failures (in which case I hide the field).
* I am very new to this project, so please make sure the field is using the correct format/font/etc.
* ~Updated the Room Database with migration to include and store this createdAt string in the DB.~ No need for this, as `AccountEntity` is only for logged in users and the date of creation is not shown anywhere.

### How does it look?
<details>
  <summary>A Random Profile</summary>

image

</details>

### Tested
* OnePlus 6p / Android 11
* Samsung S20 / Android 12


Gryzor:
I'll look into the test later today.

app/src/main/java/com/keylesspalace/tusky/entity/Account.kt
```diff
@@ -23,6 +23,7 @@ data class Account(
     @SerializedName("username") val localUsername: String,
     @SerializedName("acct") val username: String,
     @SerializedName("display_name") val displayName: String?, // should never be null per Api definition, but some servers break the contract
+    @SerializedName("created_at") val createdAt: String,
```
connyduck, line 26: just use (java.util.)Date here

connyduck (commented):
I'd show only month+year, aligned at the start, below the accountFieldList and above the remote account info text, in normal font with textColorSecondary.

app/src/main/java/com/keylesspalace/tusky/entity/Account.kt
Gryzor: Ahh ok, I wasn't sure if storing Date on the DB was ok (some people prefer to store the raw Strings/values). I'll update it alongside the MIgration. Seems like I pushed _seconds_ before you reviewed it (unfortunately).

app/src/main/java/com/keylesspalace/tusky/db/AccountEntity.kt
```diff
@@ -41,6 +41,7 @@ data class AccountEntity(
     var accountId: String = "",
     var username: String = "",
     var displayName: String = "",
+    var createdAt: String = "",
```
connyduck, line 44: But where is this ever displayed to the user

Gryzor: I am not sure, are you suggesting we don't store the date in the DB at all then? (Sorry, I'm new here so I have no idea what I'm doing)

connyduck: The `com.keylesspalace.tusky.db.AccountEntitiy` is for storing the users logged into the app. I don't see why we would need to store their account creation date. We only want to show in on the profiles, where data is alwasy fetched from the api (and not cached), so it is enough to have it in `com.keylesspalace.tusky.entity.Account`.

Gryzor: Gotcha, i'll remove all the DB modifications. It makes sense.

app/src/main/java/com/keylesspalace/tusky/entity/Account.kt
Gryzor: Done

app/src/main/java/com/keylesspalace/tusky/db/AccountEntity.kt
Gryzor: Reverted.

app/src/main/java/com/keylesspalace/tusky/components/account/AccountActivity.kt
```diff
@@ -135,6 +138,8 @@ class AccountActivity : BottomSheetActivity(), ActionButtonActivity, HasAndroidI
 
     private lateinit var adapter: AccountPagerAdapter
 
+    private val dateJoinedFormatter by lazy { SimpleDateFormat("MMM, yyyy", Locale.getDefault()) }
```
connyduck, line 141: Why lazy?

connyduck:
There is enough space to show the full month

app/src/main/java/com/keylesspalace/tusky/components/account/AccountActivity.kt
Gryzor: Habit. As far as I remember, constructing SimpleDateFormats is an expensive operation, so it made sense to delay its construction until it was needed (if needed).

connyduck: I don't think it is a problem to immediately initialize it. If you want to delay the construction you could also move it into the `updateAccountJoinedDate` method, it is not called that often

Gryzor: Done, since it's not often called, then we can just locally create it and call it a day. Updated.

connyduck (approved):
Thank you! I will merge it for Tusky 19

PR #2483, opened 2022-04-30
fix timeline refresh indicator finishing before updates are visible

connyduck:
The refresh indicator was hidden a split second before the new statuses were visible, which was annoying because one thinks that there is nothing new but then the timeline jumps. The problem was the indicator was already hidden when network loading was finished. With this change it will wait until database loading is finished as well.


Tak (approved):

charlag:
OMG finally this was so annoying!

charlag (approved):

Issue #2397, opened 2022-03-20
[wellbeing] With all settings enabled, still receiving notifications


Doomsdayrs:
I have enabled all wellbeing functionality.

The problem is that I still receiving notifications about new likes / follows. But in my notification tab those are hidden. Thus my brain constantly wants to find the event.

The solution is to fix whatever class that makes the notifications by having it respect my wellbeing settings.

* * * *
- Tusky Version: 16.0
- Android Version: 11
- Android Device: OP8T
- Mastodon instance (if applicable): fostodon

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

connyduck:
System notifications need to be disabled seperately, go to Account Preferences -> Notifications

Doomsdayrs:
> System notifications need to be disabled seperately, go to Account Preferences -> Notifications

Why do they need to be disabled separately?

connyduck:
I think the idea of the wellbeing setting was to provide additional control for things not already controlled by a setting.

And in case of notifications, where the settings are provided by the Android system, a different behavior would have been quite complicated to implement.

o-jasper:
I doubt that's complicated, but think you shouldn't do it because letting android handle it is superior? This way it's not needed to let every app re-implement it..

However, discoverability could be better? Is there a way to add a link to the Tusky-app-notification-settings dialog directly? Suppose could also mention that exists. Or at the bottom a link/button that shows some text about (other)configuration.

PR #2490, opened 2022-05-01
Add preference setting to go to native notifications.

Gryzor:
### Objective
* Add an option in Preferences to link to the Android native Notification Settings so users can further refine their notifications.

### Description
* Added a `preference {}` with a new string that links to the native preferences. 
* Code is prepared to handle channels and I took the idea from [this StackOverflow](https://stackoverflow.com/q/32366649/2684) question/answer.

### Issue
* #2397 

<details>
  <summary>The new profile option</summary>

image

</details>

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/components/preference/PreferencesFragment.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/components/preference/PreferencesFragment.kt
@@ -15,7 +15,11 @@
 
 package com.keylesspalace.tusky.components.preference
 
+import android.content.Context
+import android.content.Intent
+import android.os.Build
 import android.os.Bundle
+import android.provider.Settings
 import androidx.preference.Preference
 import androidx.preference.PreferenceFragmentCompat
 import com.keylesspalace.tusky.R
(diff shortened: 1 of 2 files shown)


Gryzor:
@connyduck What do you think of this?

connyduck:
We already have a link to notifications in the account preferences. Yes it is confusing which setting is found where but I don't think it is solved by adding even more options 😬

Gryzor:
> We already have a link to notifications in the account preferences. Yes it is confusing which setting is found where but I don't think it is solved by adding even more options 😬

It appears users could benefit from a way to go from App Settings to Android Notifications but this link exists (I just realized) from the Account's preferences like you said, so I will close this. Should we perhaps link from Wellbeing to the account's preferences? This would aid in discoverability; perhaps also renaming Preferences to Settings and having both Settings (general app stuff) and Account Preferences (differentiated). What do you think?

connyduck:
I don't really know tbh. What I know is that there are multiple problems with the settings and at this point we should probably redesign them from ground up instead of applying band-aid fixes.

Gryzor:
@connyduck I'll take a look at all the settings and maybe make a proposal.

connyduck:
let's continue the discussion here https://github.com/tuskyapp/Tusky/issues/2504

PR #2492, opened 2022-05-01
Improve UX when Login WebView fails to load the page

charlag:
Previously we simply closed the screen with the login WebView which
could cause confusion. Now we specify that page could not be loaded.

As a side effect it will also show the error message which the server
returns (if any).


connyduck (approved):
thank you!

PR #2502, opened 2022-05-04
Fix crash when saving redrafted media to drafts

### LINKED ISSUE #2499: Crash when saving draft with media created from delete & redraft
Steps to reproduce: 
Delete & Redraft a post with media. Exit the compose view, on the dialog that asks for saving the draft, click save. Crash.

```
 java.lang.IllegalStateException: unknown media type
        at com.keylesspalace.tusky.components.drafts.DraftHelper$saveDraft$2.invokeSuspend(DraftHelper.kt:88)
        at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
        at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:106)
        at kotlinx.coroutines.internal.LimitedDispatcher.run(LimitedDispatcher.kt:42)
        at kotlinx.coroutines.scheduling.TaskImpl.run(Tasks.kt:95)
        at kotlinx.coroutines.scheduling.CoroutineScheduler.runSafely(CoroutineScheduler.kt:570)
        at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.executeTask(CoroutineScheduler.kt:749)
        at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.runWorker(CoroutineScheduler.kt:677)
        at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.run(CoroutineScheduler.kt:664)
```

The cause is obviously the media uri not being a local "content://" but a "https://" one.

The way to properly fix this is to download media when it comes from redrafting, but how do we show this in the interface and handle errors? Or do we save the draft in the background and ignore errors?

connyduck:
This is not ideal, but hey it is better than crashing.

It eats error, so users media may be lost, but copying local media does as well.

closes #2499

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/components/drafts/DraftHelper.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/components/drafts/DraftHelper.kt
@@ -35,6 +35,7 @@ import okhttp3.Request
 import okio.buffer
 import okio.sink
 import java.io.File
+import java.io.IOException
 import java.text.SimpleDateFormat
 import java.util.Date
 import java.util.Locale
@@ -76,7 +77,7 @@ class DraftHelper @Inject constructor(
 
         val uris = mediaUris.map { uriString ->
             uriString.toUri()
-        }.map { uri ->
+        }.mapNotNull { uri ->
             if (uri.isInFolder(draftDirectory)) {
                 uri
             } else {
@@ -152,7 +153,7 @@ class DraftHelper @Inject constructor(
         return File(filePath).parentFile == folder
     }
 
-    private fun Uri.copyToFolder(folder: File): Uri {
+    private fun Uri.copyToFolder(folder: File): Uri? {
         val contentResolver = context.contentResolver
         val timeStamp: String = SimpleDateFormat("yyyyMMdd_HHmmss", Locale.US).format(Date())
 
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/keylesspalace/tusky/components/drafts/DraftHelper.kt
```diff
@@ -133,33 +140,55 @@ class DraftHelper @Inject constructor(
...
-        IOUtils.copyToFile(contentResolver, this, file)
+
+        if (scheme == "https") {
+            // saving redrafted media
+            try {
+                val request = Request.Builder().url(toString()).build()
+
+                val response = okHttpClient.newCall(request).execute()
+
+                val sink = file.sink().buffer()
```
charlag, line 178: could probably write directly but this will do

app/src/main/res/values/strings.xml
```diff
@@ -640,5 +640,6 @@
     <string name="action_unsubscribe_account">Unsubscribe</string>
 
     <string name="tusky_compose_post_quicksetting_label">Compose Post</string>
+    <string name="saving_draft">Saving draft...</string>
```
charlag, line 643: I like … more but as you see fit

charlag (approved):

PR #2510, opened 2022-05-09
modernize autocomplete

connyduck:
improved:
- emoji result sorting
- account results now show the bot badge, if enabled
- `ComposeAutoCompleteAdapter` -> Kotlin
- layouts (no more `RelativeLayout`, yay)
- network calls no longer wrapped into RxJava stuff
- `ComposeTokenizer` moved into the compose package

removed:
- there is no more separator between emoji results that match from the start and others
- why was there a list icon in hashtag results 😝 

Unfortunately it is not possible to write a adapter for Retrofit that can do `fun x(): Result<X>` because `Result` is an inline class that the compiler can optimze away. `suspend fun x(): Result<X>` works though because for the adapter it is `Call<Result<X>>` and `Call` is available at runtime. Calling the suspending function inside `runBlocking` is inefficient so I decided to solve the problem by adding the `fun <T> Call<T>.result(): Result<T>` extension. (Ultimately this is the fault of the `BaseAdapter`/`Filterable` api that is just bad and supports no async loading)


charlag:
Before I go into review, why is `runBlocking()` more inefficient than just a blocking call? Context switching?

app/src/main/java/com/keylesspalace/tusky/components/compose/ComposeViewModel.kt
```diff
@@ -325,48 +326,39 @@ class ComposeViewModel @Inject constructor(
...
-                        results.add(ComposeAutoCompleteAdapter.EmojiResult(emoji))
-                    } else if (shortcode.indexOf(incomplete, 1) != -1) {
-                        resultsInside.add(ComposeAutoCompleteAdapter.EmojiResult(emoji))
-                    }
-                }
-                if (results.isNotEmpty() && resultsInside.isNotEmpty()) {
-                    results.add(ComposeAutoCompleteAdapter.ResultSeparator())
+                val incomplete = token.substring(1)
+
+                return emojiList.filter { emoji ->
```
charlag, line 355: I wonder if it's the case for `sequence()`

charlag (approved):
Works for me!
I'm so glad the list icon is gone

connyduck:
> Before I go into review, why is runBlocking() more inefficient than just a blocking call? Context switching?

Yes. Check the source of runBlocking, and what Retrofit does to make a call suspending also has some overhead

app/src/main/java/com/keylesspalace/tusky/components/compose/ComposeViewModel.kt
connyduck: Should I benchmark it? My guess it that there is virtually no difference

charlag: Just avoiding to allocate 2 lists, probably no difference.

charlag (approved):

PR #2514, opened 2022-05-10
Allow build on systems without git

mcclure:
app/build.gradle invokes git on the command line to get the current changeset. If the git invocation fails, the entire build fails.
I attempted to build Tusky on a Windows system without Git (I checked out the repo using Mercurial and the hg-git bridge, running inside WSL-- so Git would have failed even if it were installed, just like it would attempting to build from a tgz downloaded from Github). Android Studio failed. Worse, it failed with a very unhelpful 200-line error message ([build-3-expanded.txt](https://github.com/tuskyapp/Tusky/files/8665469/build-3-expanded.txt)), which did not indicate the line in build.gradle causing the git requirement, rather it appeared the git-related failure occurred inside something called "net.rubygrapefruit.platform" which is apparently Android Studio-internal.

This patch changes getGitSha to set the git sha to "unknown" if the git invocation fails. This allows Android Studio to build th eapk on my system.

This change might be sub-optimal because most users **are** using git, and might prefer to get an error rather than having their git hash silently set to "unknown". Some better versions of this same patch I can imagine might be:
- When the "unknown" path is hit, flag a build warning rather than a build error (I don't know how to do this)
- Allow some build-specific config option (which?) to override the git sha value, meaning non-git users have an off switch and everyone else gets a clean failure when git fails (I'm not sure I know how to do this)
- When the "unknown" path is hit, still error, but at least do it in a way where the build error flags the relevant line in the build.gradle rather than a 200-line uncaught exception (I don't know how to do this). It might make sense to specialize the error caught to something more specific than Exception (I don't know where to look up what kinds of exceptions can be thrown at this point).
I'd be happy to change the patch in whatever way, but I might need a bit of guidance on how, groovy is mysterious to me.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -7,9 +7,13 @@ apply from: "../instance-build.gradle"
 
 def getGitSha = {
     def stdout = new ByteArrayOutputStream()
-    exec {
-        commandLine 'git', 'rev-parse', '--short', 'HEAD'
-        standardOutput = stdout
+    try {
+        exec {
+            commandLine 'git', 'rev-parse', '--short', 'HEAD'
+            standardOutput = stdout
+        }
+    } catch (Exception e) {
+        return "unknown"
     }
     return stdout.toString().trim()
 }


connyduck:
This is definitely better than before, thank you

Issue #1968, opened 2020-10-23
While redrafting a message, removing attachments removes the wrong one


TooManyBees:
After deleting and redrafting a message with multiple attached images, attempting to remove any image always removes the first one.

To reproduce:
1. Post a new message with multiple images attached.
2. Immediately use the "delete and redraft" feature
3. Remove any image except for the first one (which on my device is the left-most one)

Instead of removing the image you chose, it will remove the first (left-most) image.

In this screenshot I try to remove the 3rd image in a message I deleted and redrafted:
![Series of 3 screenshots demonstrating the problem](https://user-images.githubusercontent.com/5176286/97059998-13e8ce00-1560-11eb-84a6-d11ed2373daa.png)

* * * *
- Tusky Version: 12.1
- Android Version: 8.0.0
- Android Device: Samsung Galaxy S7
- Mastodon instance (if applicable):

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

codl:
I can reproduce this, I've also ran into an issue that I have to assume must be related, where redrafting and changing the image caption of an image other than the first changes the first image's caption instead.

charlag:
I believe this is fixed, can anyone confirm this?

codl:
I can still reproduce in both Tusky 17.0 and Tusky Test 17.0-b4eda5ea, hitting delete on the second image while redrafting deletes the first instead

PR #2517, opened 2022-05-11
fix deleting media attachments removing the wrong ones

connyduck:
this is embarassing
closes https://github.com/tuskyapp/Tusky/issues/1968


charlag (approved):
Haha classic! And I missed that too.
Maybe we need more tests :sweat_smile:

Issue #2521, opened 2022-05-12
Notification pane has usernames improperly bolded


IonCharger:
I'm unsure how to describe the error, except the bolding occurs where it's unexpected when I use Gàidhlig as the app language.

I've attached two screenshots to show the bug. Instead of the name being in bold, it's either partially bolded or unbolded.

'Chuir' means placed or put, so unsure how that grammar works in other languages when it comes to Tusky.
It would read 'Placed -name- on your post a favourite'.
![Screenshot in English UK](https://user-images.githubusercontent.com/105426732/168050656-2a0d425a-8376-4b91-881f-f505ba2a8337.png)
![Screenshot in Gàidhlig showing improper bolding](https://user-images.githubusercontent.com/105426732/168050665-04edb53f-28a7-40d8-84fc-60f785c1c17c.png)

* * * *
- Tusky Version: 18
- Android Version: 11
- Android Device: Fairphone 3
- Mastodon instance (if applicable): tested on mastodon.scot and social.linux.pizza

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

PR #2522, opened 2022-05-12
fix notification message formatting when username is not at the beginning of the message

connyduck:
closes https://github.com/tuskyapp/Tusky/issues/2521


Tak (approved):
I guess it could still go wrong if the text preceding the display name happens to also contain the display name, but that seems extremely unlikely

charlag (approved):
I'm not sure if it would be more reliable to look for the placeholder position in the template? Probably not.

connyduck:
> I'm not sure if it would be more reliable to look for the placeholder position in the template? Probably not.

Good idea. It's better because it does not rely on user controlled input.

charlag (approved):
Seems okay, I hope it works with RTL too, RTL should be about just display, not order of characters

connyduck:
yes it does, I tried it out

PR #2524, opened 2022-05-13
update android material to 1.6.0

connyduck:


charlag (approved):
We should update to M3 at some point, huh?

connyduck:
I already tried and it makes the bottom nav bar really ugly and big

PR #2531, opened 2022-05-15
Ability to crop images attached to posts

mcclure:
It should be possible to crop images. There is a crop library already in place.

This is a work in progress. I have created a pull request so I can ask questions on the Matrix channel.

Current status: The menu item for the cropper is created, the local content URL is updated, but the server is not notified.

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/components/compose/ComposeActivity.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/components/compose/ComposeActivity.kt
@@ -29,7 +29,6 @@ import android.net.Uri
 import android.os.Build
 import android.os.Bundle
 import android.os.Parcelable
-import android.provider.MediaStore
 import android.util.Log
 import android.view.KeyEvent
 import android.view.MenuItem
@@ -871,21 +870,49 @@ class ComposeActivity :
         binding.addPollTextActionTextView.compoundDrawablesRelative[0].colorFilter = PorterDuffColorFilter(textColor, PorterDuff.Mode.SRC_IN)
     }
 
+    // Use to pass state from cropImage caller to result function
+    private var cropImageItemOld: QueuedMedia? = null
+    private var cropImageUriNew: Uri? = null
+    private var cropImageFileNew: File? = null
+
     private val cropImage = registerForActivityResult(CropImageContract()) { result ->
         if (result.isSuccessful) {
-            Log.w("CROPTEST","SUCCESS")
+            cropImageItemOld?.let { itemOld ->
+                cropImageFileNew?.let { fileNew ->
+                    cropImageUriNew?.let { uriNew ->
+                        val size = fileNew.length()
+
+                        lifecycleScope.launch {
+                            viewModel.addMediaToQueue(
+                                itemOld.type,
+                                uriNew,
+                                size,
+                                itemOld.description,
+                                itemOld
+                            )
+                        }
+                    }
+                }
+            }
         } else {
-            Log.w("CROPTEST","FAIL")
+            // TODO: In else case check result.getError() to see if this was a voluntary cancel, display error if involuntary
+            Log.w("ComposeActivity", "Edit image aborted: " + result.error)
         }
+        cropImageItemOld = null
+        cropImageUriNew = null
+        cropImageFileNew = null
     }
 
     private fun editImageInQueue(item: QueuedMedia) {
+        cropImageItemOld = item
+        val tempFile = createNewImageFile(getApplicationContext(), "")
+        val uriNew = Uri.fromFile(tempFile)
+        cropImageFileNew = tempFile
+        cropImageUriNew = uriNew
         cropImage.launch(
             options(uri=item.uri) {
-                //setRequestedSize(AVATAR_SIZE, AVATAR_SIZE)
-                //setAspectRatio(AVATAR_SIZE, AVATAR_SIZE)
-                //setImageSource(includeGallery = true, includeCamera = false)
-                setOutputUri(item.uri)
+                setOutputUri(uriNew)
+                // TODO: Should compress format be set or will there be a sensible default?
                 //setOutputCompressFormat(Bitmap.CompressFormat.PNG)
             }
         )
--- a/app/src/main/java/com/keylesspalace/tusky/components/compose/ComposeActivity.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/components/compose/ComposeActivity.kt
@@ -29,6 +29,7 @@ import android.net.Uri
 import android.os.Build
 import android.os.Bundle
 import android.os.Parcelable
+import android.provider.MediaStore
 import android.util.Log
 import android.view.KeyEvent
 import android.view.MenuItem
@@ -56,6 +57,8 @@ import androidx.lifecycle.lifecycleScope
 import androidx.preference.PreferenceManager
 import androidx.recyclerview.widget.LinearLayoutManager
 import androidx.transition.TransitionManager
+import com.canhub.cropper.CropImageContract
+import com.canhub.cropper.options
 import com.google.android.material.bottomsheet.BottomSheetBehavior
 import com.google.android.material.snackbar.Snackbar
 import com.keylesspalace.tusky.BaseActivity
@@ -186,6 +189,7 @@ class ComposeActivity :
                     viewModel.updateDescription(item.localId, newDescription)
                 }
             },
+            onEditImage = this::editImageInQueue,
             onRemove = this::removeMediaFromQueue
         )
         binding.composeMediaPreviewBar.layoutManager =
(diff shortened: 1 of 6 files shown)


mcclure:
This now works, I have tested it in a variety of ways and I think it is potentially ready to commit. There are two further changes that I think should be made, but one or both could be deferred to a second PR. See below:

app/src/main/java/com/keylesspalace/tusky/components/compose/ComposeActivity.kt
```diff
@@ -867,6 +870,54 @@ class ComposeActivity :
         binding.addPollTextActionTextView.compoundDrawablesRelative[0].colorFilter = PorterDuffColorFilter(textColor, PorterDuff.Mode.SRC_IN)
     }
 
+    // Use to pass state from cropImage caller to result function
+    private var cropImageItemOld: QueuedMedia? = null
+    private var cropImageUriNew: Uri? = null
+    private var cropImageFileNew: File? = null
```
mcclure, line 876: I could not find a way to pass "user data" from an ActivityResultContract's caller to the ActivityResult callback. That was a surprising omission, so maybe there is a way in the API I do not know. Because, I had to create these temporary variables to pass the data. If there is a better way to pass the data the cropImage result callback could be simplified.

app/src/main/java/com/keylesspalace/tusky/components/compose/ComposeActivity.kt
```diff
@@ -867,6 +870,54 @@ class ComposeActivity :
...
+                                itemOld.description,
+                                itemOld
+                            )
+                        }
+                    }
+                }
+            }
+        } else {
+            // TODO: In else case check result.getError() to see if this was a voluntary cancel, display error if involuntary
+            Log.w("ComposeActivity", "Edit image aborted: " + result.error)
```
mcclure, line 899: This should be changed either in this PR or a future PR. Will ask about this in Matrix.

app/src/main/java/com/keylesspalace/tusky/components/compose/ComposeViewModel.kt
```diff
@@ -121,22 +121,36 @@ class ComposeViewModel @Inject constructor(
...
+    suspend fun addMediaToQueue(
         type: QueuedMedia.Type,
         uri: Uri,
         mediaSize: Long,
-        description: String? = null
+        description: String? = null,
+        replaceItem: QueuedMedia? = null
     ): QueuedMedia {
-        val mediaItem = media.updateAndGet { mediaValue ->
+        var stashMediaItem: QueuedMedia? = null
```
mcclure, line 131: This is needed because the QueuedMedia can now be inserted in the middle of the list, so we can no longer use last() to fetch the newly added item.

app/src/main/java/com/keylesspalace/tusky/components/compose/MediaPreviewAdapter.kt
```diff
@@ -43,12 +44,16 @@ class MediaPreviewAdapter(
         val item = differ.currentList[position]
         val popup = PopupMenu(view.context, view)
         val addCaptionId = 1
-        val removeId = 2
+        val editImageId = 2
+        val removeId = 3
         popup.menu.add(0, addCaptionId, 0, R.string.action_set_caption)
+        if (item.type == ComposeActivity.QueuedMedia.Type.IMAGE)
+            popup.menu.add(0, editImageId, 0, R.string.action_edit_image)
```
mcclure, line 51: In a future PR, I would like to also add an "edit media" item that appears on audio and video, which would call out to an external activity (for example to trim an attached video clip).

app/src/main/java/com/keylesspalace/tusky/components/compose/ComposeActivity.kt
connyduck: The member variables are lost when the activity is recreated though, e.g. on screen rotation so you should either save them in the ViewModel (preferred) or save & restore them on activity recreation.

connyduck (commented):
Works like a charm!

- Maybe update the cropping library to 4.2.1, looks like there were some critical fixes 
- Maybe it helps with this when storing additional metadata in `QueuedMedia` when it is first created. E.g. `MediaUploader.prepareMedia` checks for the file extension, which could then be used to create the temp file and set the right file format for the cropper.

app/src/main/java/com/keylesspalace/tusky/components/compose/ComposeActivity.kt
```diff
@@ -867,6 +870,54 @@ class ComposeActivity :
         binding.addPollTextActionTextView.compoundDrawablesRelative[0].colorFilter = PorterDuffColorFilter(textColor, PorterDuff.Mode.SRC_IN)
     }
 
+    // Use to pass state from cropImage caller to result function
+    private var cropImageItemOld: QueuedMedia? = null
+    private var cropImageUriNew: Uri? = null
+    private var cropImageFileNew: File? = null
+
+    private val cropImage = registerForActivityResult(CropImageContract()) { result ->
```
connyduck, line 878: small nitpick on code structure: I'd like to have all contracts in one place so maybe move this up to the others

app/src/main/res/values/strings.xml
```diff
@@ -14,6 +14,7 @@
     <string name="error_image_upload_size">The file must be less than 8MB.</string>
     <string name="error_video_upload_size">Video files must be less than 40MB.</string>
     <string name="error_audio_upload_size">Audio files must be less than 40MB.</string>
+    <string name="error_media_edit_failed">The attachment could not be edited.</string>
```
mcclure, line 17: Open to suggestions on wording here. "Editing this attachment failed."?

app/src/main/java/com/keylesspalace/tusky/components/compose/ComposeViewModel.kt
```diff
@@ -121,22 +121,36 @@ class ComposeViewModel @Inject constructor(
...
+            if (replaceItem != null) {
+                mediaToJob[replaceItem.localId]?.cancel()
+                mediaValue.map {
+                    if (it.localId == replaceItem.localId) mediaItem else it
+                }
+            } else { // Append
+                mediaValue + mediaItem
+            }
+        }
+        val mediaItem = stashMediaItem!! // stashMediaItem is always non-null and uncaptured at this point, but Kotlin doesn't know that
```
connyduck, line 152: `if` can return a value, you can do something like

```
val mediaItem = if (...) {
    ...
    returnValueA
} else {
    ...
    returnValueB
}
```

mcclure: Not through the updateAndGet . The value returned by the updateAndGet block is the new value of the List inside "media", so afaict we have to use a var to smuggle a "second" value (like the QueuedMedia) out. But we want to create the new QueuedMedia inside the updateAndGet block so the localId is created atomically with adding the item to the list.

mcclure (commented):
i didn't mean to start a review

app/src/main/java/com/keylesspalace/tusky/components/compose/ComposeActivity.kt
```diff
@@ -867,6 +873,55 @@ class ComposeActivity :
...
+    private fun editImageInQueue(item: QueuedMedia) {
+        // If input image is lossless, output image should be lossless.
+        // Currently the only supported lossless format is png.
+        val mimeType:String? = contentResolver.getType(item.uri)
+        val isPng:Boolean = mimeType != null && mimeType.endsWith("/png")
+        val context = getApplicationContext()
+        val tempFile = createNewImageFile(context, if (isPng) ".png" else ".jpg")
+
+        // "Authority" must be the same as the android:authorities string in AndroidManifest.xml
+        val uriNew = FileProvider.getUriForFile(context, BuildConfig.APPLICATION_ID + ".fileprovider", tempFile)
```
mcclure, line 913: The fancy method for getting a URI was necessary to get a content URI.
Without doing this, the mime type gets lost, and then we convert to JPEG if we ever re-crop the photo or even upload to the server.
I test this section with this code:
```
Log.w("ComposeActivity", "Input file (" + mimeType + ", " + (if (isPng) "(PNG)" else "") + "): " + item.uri + " Output file: " + uriNew)
```
I know the suggestion was to go PNG on transparent files specifically, but this will cover all transparent files and more besides (my phone makes screenshots as PNGs and I don't want those getting changed to JPEGs just because I cropped them).

Note this code will do the wrong thing on WEBP (these are sometimes lossy and sometimes lossless) but in my testing Tusky refused to attach those anyway. (GIFs get attached as videos so they don't matter either.)

mcclure:
Okay I think with that last change that covers all requested changes (except as discussed on Matrix)

connyduck (approved):

PR #2537, opened 2022-05-17
migrate Lists from RxJava to Kotlin coroutines

connyduck:
slowly but surely getting rid of Rx

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/viewmodel/ListsViewModel.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/viewmodel/ListsViewModel.kt
@@ -16,19 +16,22 @@
 
 package com.keylesspalace.tusky.viewmodel
 
+import androidx.lifecycle.ViewModel
+import androidx.lifecycle.viewModelScope
 import com.keylesspalace.tusky.entity.MastoList
 import com.keylesspalace.tusky.network.MastodonApi
-import com.keylesspalace.tusky.util.RxAwareViewModel
 import com.keylesspalace.tusky.util.replacedFirstWhich
 import com.keylesspalace.tusky.util.withoutFirstWhich
-import io.reactivex.rxjava3.core.Observable
-import io.reactivex.rxjava3.subjects.BehaviorSubject
-import io.reactivex.rxjava3.subjects.PublishSubject
+import kotlinx.coroutines.channels.BufferOverflow
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.MutableSharedFlow
+import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.launch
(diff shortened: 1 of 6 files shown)


app/src/main/java/com/keylesspalace/tusky/viewmodel/ListsViewModel.kt
```diff
@@ -39,86 +42,94 @@ internal class ListsViewModel @Inject constructor(private val api: MastodonApi)
...
     data class State(val lists: List<MastoList>, val loadingState: LoadingState)
 
-    val state: Observable<State> get() = _state
-    val events: Observable<Event> get() = _events
-    private val _state = BehaviorSubject.createDefault(State(listOf(), LoadingState.INITIAL))
-    private val _events = PublishSubject.create<Event>()
+    val state: Flow<State> get() = _state
+    val events: Flow<Event> get() = _events
+    private val _state = MutableStateFlow(State(listOf(), LoadingState.INITIAL))
+    private val _events = MutableSharedFlow<Event>(replay = 0, extraBufferCapacity = 1, onBufferOverflow = BufferOverflow.DROP_LATEST)
```
charlag, line 48: Shouldn't it be `DROP_OLDEST`?
Also
>Values in state flow are conflated using [Any.equals](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-any/equals.html) comparison in a similar way to [distinctUntilChanged](https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/distinct-until-changed.html) operator. It is used to conflate incoming updates to [value](https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/-mutable-state-flow/value.html) in [MutableStateFlow](https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/-mutable-state-flow/index.html) and to suppress emission of the values to collectors when new value is equal to the previously emitted one

which is not what we want? We want to receive each event I think?

charlag: Oh sorry it's `StateFlow` that compares

charlag: which it is :facepalm:

charlag: No it's `MutableStateFlow` :facepalm: :facepalm:

connyduck: now it should be ok

charlag (approved):
Yeah looks good!

Issue #2540, opened 2022-05-18
Tusky Nightly Crashes When Using UnifiedPush


IonCharger:
Using Gotify-UP 2.2 (f-droid, registered gotify1.unifiedpush.org) and the app crashes very shortly after opening. This happens on multiple instances of masto.

I have an adb logcat, which I hope is helpful. I'm unsure if I'm using the correct unifiedpush =)

5-18 08:46:05.035 21017 21017 I Register: REGISTER
05-18 08:46:05.055 21017 21255 I RegisterService: registering com.keylesspalace.tusky.test token: 00a95895-5efd-4f60-8fbf-157fb59c4edf
05-18 08:46:05.065 21017 21255 I RegisterService: com.keylesspalace.tusky.test already registered
05-18 08:46:05.065 21017 21255 I RegisterService: Registration is finished
05-18 08:46:05.119   491   491 E SELinux : avc:  denied  { find } for pid=4743 uid=10068 name=tethering scontext=u:r:permissioncontroller_app:s0:c68,c256,c512,c768 tcontext=u:object_r:tethering_service:s0 tclass=service_manager permissive=0
05-18 08:46:05.391 21179 21231 E AndroidRuntime: FATAL EXCEPTION: DefaultDispatcher-worker-3
05-18 08:46:05.391 21179 21231 E AndroidRuntime: Process: com.keylesspalace.tusky.test, PID: 21179
05-18 08:46:05.391 21179 21231 E AndroidRuntime: java.security.NoSuchAlgorithmException: no such algorithm: EC for provider BC
05-18 08:46:05.391 21179 21231 E AndroidRuntime: 	at sun.security.jca.GetInstance.getService(GetInstance.java:87)
05-18 08:46:05.391 21179 21231 E AndroidRuntime: 	at sun.security.jca.GetInstance.getInstance(GetInstance.java:206)
05-18 08:46:05.391 21179 21231 E AndroidRuntime: 	at java.security.KeyPairGenerator.getInstance(KeyPairGenerator.java:304)
05-18 08:46:05.391 21179 21231 E AndroidRuntime: 	at a7.b.q(SourceFile:28)
05-18 08:46:05.391 21179 21231 E AndroidRuntime: 	at s9.a.i(SourceFile:3)
05-18 08:46:05.391 21179 21231 E AndroidRuntime: 	at ha.h0.run(SourceFile:22)
[...]

PR #2542, opened 2022-05-18
add shrinker rules to keep bouncycastle EC classes

connyduck:
every. single. time.

closes https://github.com/tuskyapp/Tusky/issues/2540


charlag (approved):
I'm not sure that this is enough but let's hope it is

Issue #2476, opened 2022-04-28
Crash when opening app ("Failed to parse date")


z3ntu:
Since a few days when opening Tusky it crashes pretty much immediately after opening. I assume app data reset will "fix" it. Here's the stack trace:
```
04-28 15:15:40.426 13527 13527 E AndroidRuntime: FATAL EXCEPTION: main
04-28 15:15:40.426 13527 13527 E AndroidRuntime: Process: com.keylesspalace.tusky, PID: 13527
04-28 15:15:40.426 13527 13527 E AndroidRuntime: com.google.gson.q: Failed parsing 'Apr 26, 2022 23:40:26' as Date; at path $.expires_at
04-28 15:15:40.426 13527 13527 E AndroidRuntime:        at com.google.gson.internal.bind.DateTypeAdapter.b(SourceFile:9)
04-28 15:15:40.426 13527 13527 E AndroidRuntime:        at com.google.gson.internal.bind.ReflectiveTypeAdapterFactory$Adapter.b(SourceFile:9)
04-28 15:15:40.426 13527 13527 E AndroidRuntime:        at com.google.gson.k.c(SourceFile:9)
04-28 15:15:40.426 13527 13527 E AndroidRuntime:        at com.google.gson.k.b(SourceFile:1)
04-28 15:15:40.426 13527 13527 E AndroidRuntime:        at t6.t.o(SourceFile:29)
04-28 15:15:40.426 13527 13527 E AndroidRuntime:        at t6.t.l(SourceFile:2)
04-28 15:15:40.426 13527 13527 E AndroidRuntime:        at i1.n1.b(SourceFile:13)
04-28 15:15:40.426 13527 13527 E AndroidRuntime:        at i1.x3.b(SourceFile:14)
04-28 15:15:40.426 13527 13527 E AndroidRuntime:        at g9.d.B(SourceFile:14)
04-28 15:15:40.426 13527 13527 E AndroidRuntime:        at z9.l.o(SourceFile:1)
04-28 15:15:40.426 13527 13527 E AndroidRuntime:        at i9.a.h(SourceFile:3)
04-28 15:15:40.426 13527 13527 E AndroidRuntime:        at w9.g0.run(SourceFile:22)
04-28 15:15:40.426 13527 13527 E AndroidRuntime:        at w9.o0.k0(SourceFile:8)
04-28 15:15:40.426 13527 13527 E AndroidRuntime:        at w9.i.p(SourceFile:16)
[...]

PR #2549, opened 2022-05-22
Add UtcDateTypeAdapter for Gson

connyduck:
Basically the built-in `Date` handling sucks, see https://github.com/google/gson/issues/935

closes https://github.com/tuskyapp/Tusky/issues/2476

I was thinking we could also migrate to Moshi, but

➕ Moshi has better Kotlin support
➖ According to my measurements [here](https://github.com/tuskyapp/tusky-json-benchmark) it is slower than Gson
➖ Migration is quite an effort and we could break something

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/json/UtcDateTypeAdapter.java
+++ b/app/src/main/java/com/keylesspalace/tusky/json/UtcDateTypeAdapter.java
@@ -1,0 +1,284 @@
+// https://github.com/google/gson/blob/master/extras/src/main/java/com/google/gson/typeadapters/UtcDateTypeAdapter.java
+
+/*
+ * Copyright (C) 2011 Google Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.keylesspalace.tusky.json;
+
+import java.io.IOException;
+import java.text.ParseException;
+import java.text.ParsePosition;
+import java.util.Calendar;
+import java.util.Date;
+import java.util.GregorianCalendar;
+import java.util.Locale;
+import java.util.TimeZone;
+
+import com.google.gson.JsonParseException;
+import com.google.gson.TypeAdapter;
+import com.google.gson.stream.JsonReader;
+import com.google.gson.stream.JsonWriter;
+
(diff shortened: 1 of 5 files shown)


app/src/main/java/com/keylesspalace/tusky/json/UtcDateTypeAdapter.java
```diff
@@ -0,0 +1,284 @@
...
+  public Date read(JsonReader in) throws IOException {
+    try {
+      switch (in.peek()) {
+      case NULL:
+        in.nextNull();
+        return null;
+      default:
+        String date = in.nextString();
+        // Instead of using iso8601Format.parse(value), we use Jackson's date parsing
+        // This is because Android doesn't support XXX because it is JDK 1.6
```
charlag, line 58: doesn't support what? Android is much higher than JDK 1.6 too?

charlag (commented):
Hey, I think I understand the original issue but I'm still confused, sorry:
 - We do normally serialize dates as Long in the DB. Where do we have dates otherwise? In polls?
 - Why do we need to embed ISO date parser? I understand why default date format doesn't work but doesn't it work if we use it directly?

About encoding libraries, did you consider kotlinx.serialization? It generates more code but should be the fastest. I also like Moshi because, well, square and Jack

app/src/main/java/com/keylesspalace/tusky/json/UtcDateTypeAdapter.java
connyduck: (This is ancient code I copied from Gson)
X is the format string placeholder for timezone, see here https://docs.oracle.com/javase/10/docs/api/java/text/SimpleDateFormat.html

connyduck: Should I try improving this code? Or do the Moshi migration, it will make it obsolete

charlag: What are your thoughts on kotlinx?
I honestly don't know. Here's how Moshi does it:
https://github.com/square/moshi/blob/484d525db44f7d7b1c5a6f6f30008b12f405e1e4/moshi-adapters/src/main/java/com/squareup/moshi/adapters/Iso8601Utils.kt#L79

maybe keep it as is, just comment on the class or where it's used why we need it.

connyduck: I had a look at kotlinx.serialization and it seems awesome, but maybe more for serverside. For our usecase I see two problems
- no first-party support for Retrofit
- no support for java.util.Date, so we need to either write our own adapter again or use another date framework

charlag: I used it on the client but with another network framework and yes, no date adapter kind of kills it.
Let's just go with what we have here (or take Kotlin version from Moshi, whatever you prefer)

connyduck: I think I'll merge it like it is now so we get the crash fixed

PR #2551, opened 2022-05-23
Ability to crop/trim videos attached to posts via intent [draft]

mcclure:
I'm not sure about this, I think I might need help to get it working.

The idea was to have a code path that passes media off via implicit intent to an external program, and that would images would continue to be edited by Android-Image-Cropper but video and audio would instead invoke the intent path. If the intent path turned out to work well for video, I was going to propose a setting that would optionally allow the user to configure Tusky to handle images through intents also.

But, in my testing, the intent path works with images but *not* video. If you change `val editInternally = isImage` to `val editInternally = false`, the intent path will invoke even for images and it can pass off an image to an image editor (Photos, on my phone) and get a successful result; I know this because it prints
```
2022-05-22 23:53:04.542 32174-32174/com.keylesspalace.tusky_DEV D/ComposeActivity: User Response received: Intent { dat=content://com.keylesspalace.tusky_DEV.fileprovider/my_images/Android/data/com.keylesspalace.tusky_DEV/files/Pictures/Tusky_8XR95ZBCENS1_6584364492142417520.jpg typ=image/jpeg }
```
in logcat.

However, when I try to edit a *video* attachment with this code, startActivityForResult raises a ActivityNotFoundException saying there is no installed program that can edit video/mp4. Photos is definitely able to edit mp4 files, and I have a strange Sony program called Cinema Pro installed that I think can also edit video, though I do not know whether either of these programs correctly register themselves as intents that can edit mp4 video.

Either I haven't found the right program for editing videos, or something is wrong with the PR code.

The current code prints a nice error message in the failure case. In the success case it currently only logcats and drops the value on the floor (it would be straightforward to wire through the content URL in the result object to `addMediaToQueue`, but if the code can't edit videos there may not be a point).

There is an additional problem described below:

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/components/compose/ComposeActivity.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/components/compose/ComposeActivity.kt
@@ -16,6 +16,7 @@
 package com.keylesspalace.tusky.components.compose
 
 import android.Manifest
+import android.app.Activity
 import android.app.NotificationManager
 import android.app.ProgressDialog
 import android.content.ClipData
@@ -30,6 +31,7 @@ import android.net.Uri
 import android.os.Build
 import android.os.Bundle
 import android.os.Parcelable
+import android.provider.MediaStore
 import android.util.Log
 import android.view.KeyEvent
 import android.view.MenuItem
@@ -182,6 +184,21 @@ class ComposeActivity :
         viewModel.cropImageItemOld = null
     }
 
+    // Function to catch the response
+    // Warning: Comment out this function or it will swallow the image picker result!!
+    val MEDIA_EDIT_REQUESTCODE = 0x1001
+    override fun onActivityResult(requestCode: Int, resultCode: Int, data: Intent?) {
+        when(requestCode){
+            MEDIA_EDIT_REQUESTCODE -> {
+                if (resultCode == Activity.RESULT_OK && data != null) {
+                    Log.d(TAG, "User Response received: $data")
+                }else{
+                    Log.d(TAG,"User Response is empty")
+                }
+            }
+        }
+    }
+
     public override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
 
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/keylesspalace/tusky/components/compose/ComposeActivity.kt
```diff
@@ -899,25 +919,57 @@ class ComposeActivity :
...
+                val packageName = resolveInfo.activityInfo.packageName;
+                context.grantUriPermission(packageName, uriNew, Intent.FLAG_GRANT_WRITE_URI_PERMISSION or Intent.FLAG_GRANT_READ_URI_PERMISSION);
             }
-        )
+            try {
+                startActivityForResult(intent, MEDIA_EDIT_REQUESTCODE) // FIXME: Would using Intent.createChooser(intent) be better?
+            } catch (e: ActivityNotFoundException) {
+                // FIXME: Would be nice if error message could be specialized on (or flat out include) the mime type
+                displayTransientError(R.string.error_media_edit_intent_failed)
+            }
```
mcclure, line 971: Ideally, this code would use a contract, not `startActivityForResult`:
* startActivityForResult is actually deprecated. They want you to use contracts.
* using startActivityForResult requires overriding onActivityResult, which is a little unpleasant (and is also deprecated)
* startActivityForResult requires some very bad permissions code (see FIXME above). Basically, when using startActivity any URL the target activity is going to edit must be specifically granted for its use with `grantUriPermission`. **But**, since we are using an implicit intent, we do not know ahead of time what activity is going to handle the url! So the code here (adapted from StackOverflow) grants edit permissions to *every program on the system* that has the ability to edit URLs of the appropriate MIME type *at all*, just in case it is the one the user chooses. This is not as secure as it should be. Comments on StackOverflow suggest that using Intent.createChooser() might magically wire the permissions up correctly, but the correct way to do this is with a contract. 

Unfortunately, there is no [standard contract](https://developer.android.com/reference/androidx/activity/result/contract/ActivityResultContracts) that handles the case we are looking for (`android.intent.action.EDIT`). There is a `ActivityResultContracts.TakePicture` that will take a picture and write it to a URL, but we need something that accepts one input URL and also writes to one output URL. We would need to create a custom contract. I do not know where to start with that :(

connyduck: Check `com.keylesspalace.tusky.util.PickMediaFiles`, it is a custom contract for the picker, and make a similar one for media editing

PR #2553, opened 2022-05-23
Update FilemojiCompat to version 3.2.2

C1710:
- Fixes the bug on a non-listable emoji directory
- Reduces margins in the dialog a bit (it was there two times)


connyduck (approved):
Thank you

Issue #2554, opened 2022-05-25
Relative status timestamps not updated directly on resume


Lakoja:
When returning to Tusky from other apps on the phone the relative timestamps in a timeline are still the old ones.
For example returning to it in the morning they are still labeled "3 min" when instead "10 h" would be correct.
This changes after 1 minute. (#452)

* * * *
- Tusky Version: 18.0
- Android Version: 10
- Android Device: Motorola G7
- Mastodon instance (if applicable):

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

PR #2555, opened 2022-05-25
Fire observable to update timestamps directly on resume (#2554)

Lakoja:
An initial delay (0) can be specified.

Fixes https://github.com/tuskyapp/Tusky/issues/2554


Lakoja:
Hm, just wondering now how this might relate to the
`if (talkBackWasEnabled && !wasEnabled) adapter.notifyItemRangeChanged(`
some lines above.

That _could_ have also triggered it so there would be double update.

connyduck (approved):
Thank you! I don't think notifying the adapter multiple times is a problem, it does not necessarily mean that views will be rebound multiple times.

Issue #2552, opened 2022-05-23
 Replace federated timeline with direct messages as default


connyduck:
I'd like to change the default tabs from [Home, Notifications, Local, Federated] to [Home, Notifications, Local, Direct] because I think it is more useful to new users.

Changing the default tabs will not interfere with the setup of already logged in users or remove any feature from Tusky.

Additionally, it might make sense to put timelines into the side menu when they are not set as tab (similar to how the search goes into the side menu when the toop bar is disabled).

Thoughts?

charlag:
I think this is a good idea

Doomsdayrs:
I disagree. Direct messages are second to federation.

You're more likely to communicate on federation via post replies then DM. And many instances rely on the federation for its content stream to react to, since they are small.

Tak:
Is it feasible to make the tab bar scrollable and include all of them by default?

charlag:
Nooo, that's bad!
It mightvbe subjective but I never got any value out of federated timeline. I can't say confidently what is more useful for new people

Doomsdayrs:
> Is it feasible to make the tab bar scrollable and include all of them by default?

Navigation bars should never be scrollable, it violates material design guidelines.

> Nooo, that's bad! It mightvbe subjective but I never got any value out of federated timeline. I can't say confidently what is more useful for new people

While you might not be getting value, For the vast majority of instances federated is life and blood.

---

I personally believe this trend of trying to get rid of federated timelines is toxic to the federation.

The belief that it is "easier" for new users to not see the federation is catering to centralization, and Mastodon was built to be against centralization, as it is a *federated* platform.

--- 

A far more elegant solution is to simply append the navigation bar with a direct message icon.

connyduck:
Of course we could just add the direct messages tab (at which position?), but I feel like 5 tabs is already very overloaded on small devices.

a1batross:
Federated timeline is usually cluttered with random content. I can understand if users don't need it, direct messages are usually more useful.

But speaking about tabs, 5 tabs limit is not enough, I've increased the limit to 9 in my builds but actually never used more than 7, which is enough for most modern devices. Even with the narrowest screen aspect ratio, it isn't much of a problem.

Doomsdayrs:
> Of course we could just add the direct messages tab (at which position?), but I feel like 5 tabs is already very overloaded on small devices.

[...]

PR #2556, opened 2022-05-25
Fix conversations

connyduck:
This fixes several problems with conversations (the code is basically just taken over from the newer timeline code)

- timestamps now update on pull-to-refresh
- the view now jumps up a bit to reveal newly loaded items after pull-to-refresh
- "nothing here" shown when conversations are empty instead of nothing
- when replying from conversations, the "replying to" field will be correctly filled
- the "fabHide" setting is now respected in the ConversationsFragment as well
- and most importantly, pagination now works as expected

kind of a prerequisite to https://github.com/tuskyapp/Tusky/issues/2552

I'd like to merge this at the same time as https://github.com/tuskyapp/Tusky/pull/2549 because the migrations overlap


connyduck:
draft because some details still aren't perfect

connyduck:
While this is a hundred times better than before, there are still 2 glitches:
- When the database is empty and conversations load for the first time, the list does not start at the top and one has to scroll up. I think is because of the load state footer that is shown and the recyclerview trying to keep the position. No idea how to turn this behavior off, I tried a lot.
- When having a long list of conversations and being way down, suddenly jumping up shows the wrong views for a split second. This is because paging from database does not hold all data in memory. We have the same behavior in timelines.

charlag (commented):
I'm a not so confident with all the paging APIs but maybe it's worth leaving a comment on how/why `order` works? I don't understand it fully, does server not reorder them all the time?

connyduck:
> I'm a not so confident with all the paging APIs but maybe it's worth leaving a comment on how/why order works? I don't understand it fully, does server not reorder them all the time?

Yes it does reorder it all the time but I also reload it all the time. After a pull to refresh there is only the current page in the cache.

PR #2562, opened 2022-05-27
2554 refresh timestamps on resume

Lakoja:
An initial delay (0) can also be specified for the older Notifications list.

Fixes https://github.com/tuskyapp/Tusky/issues/2554


connyduck:
please rebase

PR #2567, opened 2022-05-28
Fix unintended [mismatched] show-replies preference

mcclure:
Also add a comment to prevent confusion.

In my testing, this works with new app installs but **not** with existing app installs, I assume because they've already saved the default as an actual key. See second PR.


connyduck:
Yes, let's merge the other one.

PR #2569, opened 2022-05-29
replace kotlin-result-calladapter with networkresult-calladapter

connyduck:
The main difference between the two is that `NetworkResult` is no inline class. This has two advantages

- No weird `ClassCastException` crashes all over the place
- It also works without `suspend`

Steps to reproduce one instance of the crash: https://chaos.social/@ConnyDuck/108353117857711488
Code of NetworkResult Calladapter: https://github.com/connyduck/networkresult-calladapter

Regarding the weird crash, maybe I will try to isolate it so I can report it to Kotlin and/or Retrofit


charlag (approved):
Seems straightforward enough!

PR #2573, opened 2022-05-30
load filters before anything else

connyduck:
Calling `invalidate` on the room paging source too early apparently stops it from refreshing stuff from network, so let's try this?

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/components/timeline/viewmodel/CachedTimelineViewModel.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/components/timeline/viewmodel/CachedTimelineViewModel.kt
@@ -45,8 +45,11 @@ import com.keylesspalace.tusky.network.MastodonApi
 import com.keylesspalace.tusky.network.TimelineCases
 import com.keylesspalace.tusky.viewdata.StatusViewData
 import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.FlowPreview
 import kotlinx.coroutines.asExecutor
 import kotlinx.coroutines.delay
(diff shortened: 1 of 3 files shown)


charlag:
I think I tried something like this before, I think it's a bad idea tbh. It means no cached content would be displayed until filters right? Maybe after we could check that whether there was already something emitted and only then invalidate?

connyduck:
That could work as well. I'll try.

digitalcircuit:
Low priority note responding to @charlag's remark:
If filters are to be applied to cached posts after filters load, I'd suggest adding caching filters to the roadmap to minimize the likelihood of showing filtered posts.

My Home timeline is reasonably slow, and I've often had filtered posts show for long enough (before filters loaded) that I could skim the post text/photo, making filters not as useful for avoiding being reminded of emotionally difficult matters.

Pardon if this is out of place - I'd be happy to file a new issue or such if desired.  This happened before the recent filter breakage, I merely kept forgetting to file an issue when I had the energy (I was usually winding down when the problem happened).

charlag:
@digitalcircuit it's alright, we said the same thing on Matrix and I also mentioned in in the [commit](https://github.com/tuskyapp/Tusky/pull/2566/commits/dc47d03b316938fb6077d18fe401701125f86d7f) for the [original PR](https://github.com/tuskyapp/Tusky/pull/2566)

connyduck:
I tried only invalidating when data is already displayed, but it does not work. AndroidX paging is really bad for such things. The only way to get the load state is with a load state listener and that requires an adapter, so no way to do that in the ViewModel. 😢 
Any other ideas? Or should I try routing the load state from the Fragment to the ViewModel? @charlag

Issue #2576, opened 2022-06-05
Image upload broken (develop)


Lakoja:
After https://github.com/tuskyapp/Tusky/pull/2569 the image upload is broken for me.
It is uploaded but on finish only a transient error is displayed.

It works when reverting to above commit.

The result object in kotlinx.coroutines.flow.FlowKt#callbackFlow is null with this (leading to a NullPointerException).

* * * *
- Tusky Version: develop (> 30.05.2021)
- Android Version: 10

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

Is it ok to post issues with the develop version here (as it is no officially released version)?

PR #2577, opened 2022-06-05
Use NetworkResult in MediaUploadApi

tsmethurst:
This PR updates the MediaUploadApi logic to use Conny's `NetworkResult` instead of the previous `Result` that was being returned.

closes https://github.com/tuskyapp/Tusky/issues/2576


Lakoja:
Nice, I guessed it was something like this (but what? :-)

PR #2579, opened 2022-06-07
improve logout

connyduck:
Fixes these problems:
- timelines reloading in the background so when user relogins with another account, posts from the old account are shown
- token not being invalidated
- requests to dummy.placeholder that sometimes occur on logout

It is a bit hacky but definitely makes things better than before

Still open:
- [x] database migration
- [ ] maybe error handling for the token invalidation, although I'd prefer to do it later in another PR


app/src/main/java/com/keylesspalace/tusky/network/InstanceSwitchAuthInterceptor.java
```diff
@@ -55,13 +54,26 @@ public Response intercept(@NonNull Chain chain) throws IOException {
...
                 builder.removeHeader(MastodonApi.DOMAIN_HEADER);
             } else if (currentAccount != null) {
-                //use domain of current account
-                builder.url(swapHost(originalRequest.url(), currentAccount.getDomain()))
-                        .header("Authorization",
-                                String.format("Bearer %s", currentAccount.getAccessToken()));
+                String accessToken = currentAccount.getAccessToken();
+                if (!accessToken.isEmpty()) {
+                    //use domain of current account
+                    builder.url(swapHost(originalRequest.url(), currentAccount.getDomain()))
```
charlag, line 60: I understand that we can only add access token when we have it but what about switching domains? Should it also depend on token or just on account?

app/src/main/java/com/keylesspalace/tusky/usecase/LogoutUsecase.kt
```diff
@@ -0,0 +1,69 @@
...
+
+            // disable pull notifications
+            if (!NotificationHelper.areNotificationsEnabled(context, accountManager)) {
+                NotificationHelper.disablePullNotifications(context)
+            }
+
+            // clear notification channels
+            NotificationHelper.deleteNotificationChannelsForAccount(activeAccount, context)
+
+            // locally delete the access token
```
charlag, line 51: this comment doesn't match anymore

charlag (approved):

app/src/main/java/com/keylesspalace/tusky/network/InstanceSwitchAuthInterceptor.java
connyduck: I think it is fine like that, this branch really is only for authenticated requests to the current instance. For everything else one would use the DOMAIN_HEADER.

PR #2581, opened 2022-06-08
UnifiedPush: FEATURE_BYTES_MESSAGE distributors

### LINKED ISSUE #2547: UnifiedPush registration missing FEATURE_BYTES_MESSAGE
https://github.com/tuskyapp/Tusky/blob/9ec5d6e3b032a582b6d95091e56122e80ae07c9c/app/src/main/java/com/keylesspalace/tusky/components/notifications/PushNotificationHelper.kt#L90

TL;DR I'm guessing that line just needs to be replaced to something like:
```kotlin
 UnifiedPush.registerAppWithDialog(context, account.id.toString(), features = [UnifiedPush.FEATURE_BYTES_MESSAGE]) 
```
, but I don't know Kotlin and don't have a build environment set up.

---

https://github.com/UnifiedPush/android-connector/blob/d9bf79a8dfe20d3ed85771f01c7fb706896a096a/connector/src/main/java/org/unifiedpush/android/connector/Utils.kt#L42-L51

The 'features' argument needs to be set to `[FEATURE_BYTES_MESSAGE]` as mentioned on the UP Android docs. This is because without this argument, UnifiedPush v1 distributors like Gotify-UP are still allowed. The problem with the existing v1 version of Gotify-UP is that when there's (certain types of, I haven't narrowed down what) binary data, it sends no notification at all, rather than sending a corrupted notification. Since Gotify-UP (probably) won't be updated, users might accidentally select it and will be confused about why their notifications don't work.

cc @PeterCxy

* * * *
- Tusky Version: Tusky Test (recent version, it doesn't matter which)
- Android Version: n/a
- Android Device: n/a
- Mastodon instance (if applicable): n/a

karmanyaahm:
Only use distributors which are compatible with FEATURE_BYTES_MESSAGE. This is because Mastodon sends notifications that are arbitrary bytes, not UTF-8. This causes issues in older versions of UnifiedPush distributors and providers that don't support FEATURE_BYTES_MESSAGE, specifically Gotify w/ Postgres, see #2547 for more info.

I have tested this on my device.
Closes #2547

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/components/notifications/PushNotificationHelper.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/components/notifications/PushNotificationHelper.kt
@@ -89,7 +89,7 @@ private suspend fun enableUnifiedPushNotificationsForAccount(context: Context, a
         // Already registered, update the subscription to match notification settings
         updateUnifiedPushSubscription(context, api, accountManager, account)
     } else {
-        UnifiedPush.registerAppWithDialog(context, account.id.toString())
+        UnifiedPush.registerAppWithDialog(context, account.id.toString(), features = arrayListOf(UnifiedPush.FEATURE_BYTES_MESSAGE))
     }
 }
 


connyduck (approved):
Thank you!

PR #2584, opened 2022-06-09
use moshis Rfc3339DateJsonAdapter for json date parsing

connyduck:
- it is in Kotlin and not in ancient Java 6
- it seems to be more permissive
- easier than migrating to moshi as a whole


charlag (approved):
Seems fine

PR #2596, opened 2022-06-22
add test for InstanceSwitchAuthInterceptor and convert it to Kotlin

connyduck:


app/src/test/java/com/keylesspalace/tusky/network/InstanceSwitchAuthInterceptorTest.kt
```diff
@@ -0,0 +1,142 @@
...
+import org.mockito.kotlin.doAnswer
+import org.mockito.kotlin.mock
+
+class InstanceSwitchAuthInterceptorTest {
+
+    private val mockWebServer = MockWebServer()
+
+    @Before
+    fun setup() {
+        mockWebServer.start(80)
```
charlag, line 23: Isn't 80 a privileged port?
They also say
>Automated tests should always use port 0 to avoid flakiness when a specific port is unavailable.

charlag (approved):

app/src/test/java/com/keylesspalace/tusky/network/InstanceSwitchAuthInterceptorTest.kt
connyduck: yes it is the standard http port, using a different one require changes to the InstanceSwitchAuthInterceptor though

connyduck: With some creativity I got it to work 👏 
Thank you!

PR #2620, opened 2022-07-16
Add UI for image-attachment "focus"

mcclure:
Choose "Set focus" in the attachment menu, tap once to select focus point, tap "OK". Has been [successfully tested](https://mastodon.social/@mcc_test_tusky/108655193725406904) with mastodon.social (image has two identical very tall image attachments, but in one the focus point is set near the top and in the other it is set near the bottom).

Currently very ugly, no visual feedback of where/whether you tapped and I think a naive user would be totally confused by what the dialog even does:

[](https://user-images.githubusercontent.com/277318/179340919-7f0f185d-9f94-470a-9631-421de9a600b5.png)
 
But, it works, so should be a good starting point for a more complete attempt.

Things I think should change before submit:
- Thumbnails in compose view should refocus on focus point
- Some visual feedback of tap, a bullseye or something
- Currently if you tap outside the photo view (like slightly to the right or left) the tap does not register. That's very bad (may require replacing chrisbanes.photoview)
- I need to do more aggressive testing (changing focus points multiple times, changing caption before and after etc)

Things I'm 50% sure should change before submit:

- It would be nice if you could drag a finger to move around the focus point instead of only taps working (this may require replacing chrisbanes.photoview)
- Some text above the image would be good. Possibly either a full explanation of how focus points work or a link to one. It could be an unfamiliar concept.
- Some sort of preview would be nice. Ideally the preview would let you choose a couple different aspect ratios like 16:9 or 1:1. Maybe that could be saved for a followup PR.

Things I'm not sure about at all:
- What should the menu item say? Currently it's "Set focus". Would something like "Image focus" or "Set focus point" be better?
- TODO item in [ComposeViewModel.kt](https://github.com/tuskyapp/Tusky/compare/develop...mcclure:image-focus?expand=1#diff-979cd489c1b985730744fe1ffea7cfc58c035627b13189e21602de4942ccd474)
- Currently the pane is inset (like "set caption"), maybe it would be good if the pane were "full screen" (like the cropper)

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/components/compose/ComposeActivity.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/components/compose/ComposeActivity.kt
@@ -26,6 +26,7 @@ import android.content.pm.PackageManager
 import android.graphics.Bitmap
 import android.graphics.PorterDuff
 import android.graphics.PorterDuffColorFilter
+import android.graphics.PointF
 import android.net.Uri
 import android.os.Build
 import android.os.Bundle
@@ -68,6 +69,7 @@ import com.keylesspalace.tusky.R
 import com.keylesspalace.tusky.adapter.EmojiAdapter
 import com.keylesspalace.tusky.adapter.OnEmojiSelectedListener
 import com.keylesspalace.tusky.components.compose.dialog.makeCaptionDialog
+import com.keylesspalace.tusky.components.compose.dialog.makeFocusDialog
 import com.keylesspalace.tusky.components.compose.dialog.showAddPollDialog
 import com.keylesspalace.tusky.components.compose.view.ComposeOptionsListener
 import com.keylesspalace.tusky.components.compose.view.ComposeScheduleView
@@ -214,6 +216,11 @@ class ComposeActivity :
             onAddCaption = { item ->
                 makeCaptionDialog(item.description, item.uri) { newDescription ->
                     viewModel.updateDescription(item.localId, newDescription)
+                }
+            },
+            onAddFocus = { item ->
+                makeFocusDialog(item.focus, item.uri) { newFocus ->
+                    viewModel.updateFocus(item.localId, newFocus)
                 }
             },
             onEditImage = this::editImageInQueue,
@@ -1065,7 +1072,8 @@ class ComposeActivity :
         val mediaSize: Long,
         val uploadPercent: Int = 0,
         val id: String? = null,
-        val description: String? = null
+        val description: String? = null,
+        val focus: PointF? = null // Range -1..1 y-up
     ) {
         enum class Type {
             IMAGE, VIDEO, AUDIO;
--- a/app/src/main/java/com/keylesspalace/tusky/components/compose/ComposeViewModel.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/components/compose/ComposeViewModel.kt
@@ -15,6 +15,7 @@
 
 package com.keylesspalace.tusky.components.compose
 
+import android.graphics.PointF
 import android.net.Uri
 import android.util.Log
 import androidx.core.net.toUri
@@ -337,11 +338,37 @@ class ComposeViewModel @Inject constructor(
 
         val updatedItem = newMediaList.find { it.localId == localId }
         if (updatedItem?.id != null) {
-            return api.updateMedia(updatedItem.id, description)
+            return api.updateMediaDescription(updatedItem.id, description)
                 .fold({
                     true
                 }, { throwable ->
                     Log.w(TAG, "failed to update media", throwable)
+                    false
+                })
+        }
+        return true
+    }
+
+    // TODO: Factor this and updateDescription into a single function?
+    suspend fun updateFocus(localId: Int, focus: PointF): Boolean {
+        val newMediaList = media.updateAndGet { mediaValue ->
+            mediaValue.map { mediaItem ->
+                if (mediaItem.localId == localId) {
+                    mediaItem.copy(focus = focus)
+                } else {
+                    mediaItem
+                }
+            }
+        }
+
+        val updatedItem = newMediaList.find { it.localId == localId }
+        if (updatedItem?.id != null) {
+            val focusString = "${focus.x},${focus.y}"
+            return api.updateMediaFocus(updatedItem.id, focusString)
+                .fold({
+                    true
+                }, { throwable ->
+                    Log.w(TAG, "failed to update media focus point", throwable)
                     false
                 })
         }
--- a/app/src/main/java/com/keylesspalace/tusky/components/compose/dialog/FocusDialog.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/components/compose/dialog/FocusDialog.kt
@@ -1,0 +1,110 @@
+/* Copyright 2019 Tusky Contributors
+ *
+ * This file is a part of Tusky.
+ *
+ * This program is free software; you can redistribute it and/or modify it under the terms of the
+ * GNU General Public License as published by the Free Software Foundation; either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * Tusky is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even
+ * the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU General
+ * Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License along with Tusky; if not,
+ * see <http://www.gnu.org/licenses>. */
+
+package com.keylesspalace.tusky.components.compose.dialog
+
+import android.app.Activity
+import android.content.DialogInterface
+import android.graphics.PointF
+import android.graphics.drawable.Drawable
+import android.net.Uri
+import android.text.InputFilter
+import android.text.InputType
+import android.util.Log // ANDI SHOULD NOT CHECK THIS LINE IN TO GIT
+import android.view.WindowManager
+import android.widget.EditText
+import android.widget.ImageView;
+import android.widget.LinearLayout
+import android.widget.Toast
+import androidx.appcompat.app.AlertDialog
+import androidx.lifecycle.LifecycleOwner
+import androidx.lifecycle.lifecycleScope
+import at.connyduck.sparkbutton.helpers.Utils
+import com.bumptech.glide.Glide
+import com.bumptech.glide.load.resource.bitmap.DownsampleStrategy
+import com.bumptech.glide.request.target.CustomTarget
+import com.bumptech.glide.request.transition.Transition
+import com.github.chrisbanes.photoview.OnPhotoTapListener
+import com.github.chrisbanes.photoview.PhotoView
+import com.keylesspalace.tusky.R
+import kotlinx.coroutines.launch
+
+fun <T> T.makeFocusDialog(
+    existingFocus: PointF?,
+    previewUri: Uri,
+    onUpdateFocus: suspend (PointF) -> Boolean
+) where T : Activity, T : LifecycleOwner {
+    var focus = existingFocus ?: PointF(0.0f, 0.0f) // Default to center
+
+    val dialogLayout = LinearLayout(this)
+    val padding = Utils.dpToPx(this, 8)
+    dialogLayout.setPadding(padding, padding, padding, padding)
+
+    dialogLayout.orientation = LinearLayout.VERTICAL
+    val imageView = PhotoView(this).apply {
+        maximumScale = 6f
+        setOnPhotoTapListener(object : OnPhotoTapListener {
+            override fun onPhotoTap(view: ImageView, x:Float, y:Float) {
+                focus = PointF(x*2-1, 1-y*2) // PhotoView range is 0..1 Y-down but Mastodon API range is -1..1 Y-up
+            }
+        })
+    }
+
+    val margin = Utils.dpToPx(this, 4)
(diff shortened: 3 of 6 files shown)


app/src/main/java/com/keylesspalace/tusky/components/compose/ComposeViewModel.kt
```diff
@@ -324,11 +325,12 @@ class ComposeViewModel @Inject constructor(
         return combineLiveData(deletionObservable, sendFlow.asLiveData()) { _, _ -> }
     }
 
-    suspend fun updateDescription(localId: Int, description: String): Boolean {
+    // Updates a QueuedMedia item arbitrarily, then sends description and focus to server
+    suspend fun updateMediaItem(localId: Int, mutator: (QueuedMedia) -> QueuedMedia): Boolean {
```
mcclure, line 329: Should this be private?

mcclure:
Commit [613358e](https://github.com/tuskyapp/Tusky/pull/2620/commits/613358ef83e5065e66a2e5c2b5e4871cb43efc4c) removes the code duplication in ComposeViewModel which I wanted, removes the function duplication in MastodonApi which Conny wanted, and has now been aggressively tested (IE change caption then focus, change focus then caption, change both more than once, everything is correct in the end)

app/src/main/java/com/keylesspalace/tusky/components/compose/ComposeViewModel.kt
connyduck: yes

connyduck:
> - Thumbnails in compose view should refocus on focus point
> - Some visual feedback of tap, a bullseye or something
> - Currently if you tap outside the photo view (like slightly to the right or left) the tap does not register. That's very bad (may require replacing chrisbanes.photoview)

Refocusing thumbnails and visual feedback definitely, the third is not that important for a first version. Maybe also show the selected focus point on the thumbnail? How about a simple circle in primary color with theme background color as outline as indicator?

> What should the menu item say? Currently it's "Set focus". Would something like "Image focus" or "Set focus point" be better?

I like "set focus point" most.

> Currently the pane is inset (like "set caption"), maybe it would be good if the pane were "full screen" (like the cropper)

I think fullscreen would be better, but it isn't important.

> Should there be a focus checkmark image like there is for captions?

If you want, definitely not needed for a first version

mcclure:
The attachment previews in the compose view now focus properly where appropriate. ProgressImageView now inherits from MediaPreviewImageView to accomplish this. I also discovered Attachment.Focus and replaced my uses of PointF with it.

mcclure:
Screenshot_20220903-182223

Screenshot_20220903-182254

Screenshot_20220903-182120

Will edit this post from PC momentarily

mcclure (commented):
asdfadsffg i didn't mean to click this button

app/src/main/java/com/keylesspalace/tusky/components/compose/dialog/FocusDialog.kt
```diff
@@ -0,0 +1,235 @@
...
+import com.github.chrisbanes.photoview.OnPhotoTapListener
+import com.github.chrisbanes.photoview.PhotoView
+import com.keylesspalace.tusky.R
+import com.keylesspalace.tusky.entity.Attachment.Focus
+import kotlinx.coroutines.launch
+import java.nio.ByteBuffer
+import java.security.MessageDigest
+import java.util.concurrent.locks.Lock
+
+// Private, but necessary to implement BitmapTransformation, function extracted from Glide
```
mcclure, line 54: This file now contains rather a lot of code adapted from Glide. There was no sample code for creating a custom BitmapTransformation in their docs so I used the RoundedCorners class and the TransformationUtils.roundedCorners method as bases. `getAlphaSafeBitmap` and `getAlphaSafeConfig` are taken direct from Glide TransformationUtils (but converted to Kotlin) and the "boilerplate" methods are almost identical to the ones from RoundedCorners. I do not think this introduces license issues because (1) we are already bound to the Glide license and (2) the code taken is "minimal" in the sense that it practically could not be any different than it is.

I am going to file a bug on Glide asking them to make getAlphaSafeBitmap and getAlphaSafeConfig public methods as I do not think it is possible to implement a BitmapTransformation without them.

app/src/main/java/com/keylesspalace/tusky/components/compose/dialog/FocusDialog.kt
```diff
@@ -0,0 +1,235 @@
...
+        val bitmapDrawableLock: Lock = TransformationUtils.getBitmapDrawableLock()
+        val safeConfig: Bitmap.Config = getAlphaSafeConfig(inBitmap)
+        val toTransform: Bitmap = getAlphaSafeBitmap(pool, inBitmap)
+        val result = pool[toTransform.width, toTransform.height, safeConfig]
+
+        val plainPaint = Paint()
+        val strokePaint = Paint(Paint.ANTI_ALIAS_FLAG)
+        strokePaint.setAntiAlias(true)
+        strokePaint.setStyle(Paint.Style.STROKE)
+        val strokeWidth = 10.0f
```
mcclure, line 94: As described above this should not be a fixed number

app/src/main/java/com/keylesspalace/tusky/components/compose/dialog/FocusDialog.kt
```diff
@@ -0,0 +1,235 @@
...
+class HighlightFocus(val focus: Focus) : BitmapTransformation() {
+    override fun transform(pool: BitmapPool, inBitmap: Bitmap, outWidth: Int, outHeight: Int): Bitmap {
+        // Draw overlaid target
+        val bitmapDrawableLock: Lock = TransformationUtils.getBitmapDrawableLock()
+        val safeConfig: Bitmap.Config = getAlphaSafeConfig(inBitmap)
+        val toTransform: Bitmap = getAlphaSafeBitmap(pool, inBitmap)
+        val result = pool[toTransform.width, toTransform.height, safeConfig]
+
+        val plainPaint = Paint()
+        val strokePaint = Paint(Paint.ANTI_ALIAS_FLAG)
```
mcclure, line 91: Very unclear if ANTI_ALIAS_FLAG is required, I tried it both ways and I THINK it looks better with but it's hard to tell

app/src/main/java/com/keylesspalace/tusky/components/compose/dialog/FocusDialog.kt
```diff
@@ -0,0 +1,235 @@
...
+            canvas.drawCircle(x, y, circleRadius, strokePaint)
+            canvas.drawCircle(x, y, strokeWidth / 2.0f, strokePaint)
+
+            canvas.setBitmap(null)
+        } finally {
+            bitmapDrawableLock.unlock()
+        }
+
+        if (!toTransform.equals(inBitmap)) {
+            pool.put(toTransform)
```
mcclure, line 129: I am very, VERY confused if this is desirable or not. I *think* it's good because this will cache the image with the circle/shadow drawn on so if the set focus dialog is opened a second time it will open faster.

mcclure:
Took in Conny's commit and I think this is mergeable? One last question:

app/src/main/java/com/keylesspalace/tusky/components/compose/view/FocusIndicatorView.kt
```diff
@@ -0,0 +1,104 @@
...
+
+    private val curtainPaint = Paint(Paint.ANTI_ALIAS_FLAG)
+    private val strokePaint = Paint(Paint.ANTI_ALIAS_FLAG)
+
+    init {
+        curtainPaint.setColor(transparentDarkGray)
+        curtainPaint.style = Paint.Style.FILL
+
+        strokePaint.setStyle(Paint.Style.STROKE)
+        strokePaint.setStrokeWidth(strokeWidth)
```
mcclure, line 77: Will this be in display-independent pixels or do I need to add `* getResources().getDisplayMetrics().density` ?

Tak:
Is it possible to apply the focus point to the preview in the compose dialog? (could be a followup after merging)

Tak:
Oh, I think I see - is it possible that the focal point is getting *reversed* in this scenario? Here, I'm setting the focal point in the upper left, and the preview is panning to the lower right. (This is against https://github.com/tuskyapp/Tusky/pull/2620/commits/f4e4f8b2b99b1f1a089ed024eaf98145f9b65144 )

Screenshot_20220907-190449
Screenshot_20220907-190515

mcclure:
@Tak Ahh, thank you, good catch-- the X axis was inverted but the testcase I used on the previous push only tested the Y axis...!

I have a fix which I will push tomorrow.

app/src/main/java/com/keylesspalace/tusky/components/compose/ComposeActivity.kt
```diff
@@ -169,6 +170,7 @@ class ComposeActivity :
                         uriNew,
                         size,
                         itemOld.description,
+                        null, // Intentionally reset focus when cropping
```
mcclure, line 173: Is this the best approach? We could do something weird where we try to calculate where/whether the focal point is within the cropped area, but this feels like second guessing the user so I just reset.

app/src/main/java/com/keylesspalace/tusky/components/drafts/DraftMediaAdapter.kt
```diff
@@ -42,24 +42,34 @@ class DraftMediaAdapter(
...
+        return DraftMediaViewHolder(MediaPreviewImageView(parent.context))
     }
 
     override fun onBindViewHolder(holder: DraftMediaViewHolder, position: Int) {
         getItem(position)?.let { attachment ->
             if (attachment.type == DraftAttachment.Type.AUDIO) {
+                holder.imageView.clearFocus()
                 holder.imageView.setImageResource(R.drawable.ic_music_box_preview_24dp)
             } else {
-                Glide.with(holder.itemView.context)
```
mcclure, line 53: Long term observation: This clause has (and had before I started changing it) significant code duplication VS MediaPreviewAdapter. Maybe it's worth a refactor here

mcclure:
New changes:
- Fix x flipping bug
- Drafts, scheduled tweets, . Focus properly focuses in the draft view preview.
- Crop window narrows itself vertically so it no longer has large amounts of empty space. Note, this ONLY affects vertical width, not horizontal width, horizontal width seems fine already.

Will post some screenshots of the new window sizing after I fix merging and do some exhaustive tests.

mcclure:
How the new dialog sizes look
Screenshot_20220918-193722
Screenshot_20220918-193755
Screenshot_20220918-193824

mcclure:
Screenshot_20220918-193755

mcclure:
I think this is a candidate for merging. **The only remaining issues/questions:**

- Window disappears when the phone autorotates the screen. (Caption window now resizes itself properly.)
- Should explanation text go above the image or is it obvious enough?

connyduck:
It works so well now, I love it. Thank you very much.

> Window disappears when the phone autorotates the screen. (Caption window now resizes itself properly.)

We had the same problem for the caption dialog, see [here ](https://github.com/tuskyapp/Tusky/pull/2693) how it was fixed. It is no big deal though.

> Should explanation text go above the image or is it obvious enough?

You mean the "Click or drag the circle on the preview to choose the focal point which will always be in view on all thumbnails." that the web interface has? Would be nice, yes but it is also no blocker for merging this.

Android studio gives me a lot of code style warnings, can you fix those please? It is mostly property access syntax vs getter/setter and Kotlin functions instead of java `Math` ones.

mcclure:
Looking through the warnings now, I think I was confused how to get Android Studio to show you warnings at first...

Is it okay to leave these ones? "Tusky" being a typo is obviously wrong, and I think the soft input thing is not trivial to fix. That code was given to me and I don't know what it does exactly.

image

mcclure:
I'm finding it a little hard to get full lists of Android Studio warnings, but I *think* those last two commits fix all the warnings in code the PR touches.

I left some warnings as-is:

- Many claims that "Tusky" is a typo 🙄 
- Ignored an if-null foldable in FocusIndicatorView because the suggested alternative would be pretty ugly
- Ignored "'SOFT_INPUT_ADJUST_RESIZE: Int' is deprecated. Deprecated in Java" in FocusDialog; I copied this code from CaptionDialog and I don't at present know either what it does or what to replace it with

I also explicitly @SuppressLint'd and commented an accessibility warning because FocusIndicatorView implements a touch event but not the "click" event from the screen reader and there's legitimately nothing that can be done with the screen reader click. FocusIndicatorView interaction is not meaningful without an x,y position. If there's something better I should do here instead (implement performClick but return false because we didn't handle it? hide focus select from screen readers completely? or is there actually a way to make this pane accessible?) I'd be happy to do that instead.

I wouldn't mind trying to replicate the CaptionDialog refactor at some later point, but I'd prefer to merge this PR first so that the PR does not develop conflicts with the develop branch.

connyduck (approved):

connyduck:
> I also explicitly @SuppressLint'd and commented an accessibility warning because FocusIndicatorView implements a touch event but not the "click" event from the screen reader and there's legitimately nothing that can be done with the screen reader click. FocusIndicatorView interaction is not meaningful without an x,y position. If there's something better I should do here instead (implement performClick but return false because we didn't handle it? hide focus select from screen readers completely? or is there actually a way to make this pane accessible?) I'd be happy to do that instead.

I'm not sure. I'd leave it for now and wait for feedback from people using Tusky with screenreaders.

Issue #2527, opened 2022-05-15
Tapping "Add Poll" with keyboard open, then cancelling poll, glitches out menu


mcclure:
Repro steps: Begin a new "toot" compose. Tap in the text field so that the keyboard comes up. Tap the attach icon and choose to attach a poll. The keyboard will slide away and the attach menu will get "stuck", hovering partway up the screen. If you cancel poll creation at this point the menu will continue hovering until you dismiss it by swiping down. The hitbox on the menu items is strange at this point, I cannot activate them by tapping on the right side of the menu but it is easy to accidentally trigger them.

Video:

https://user-images.githubusercontent.com/277318/168452207-216cce3a-e6e1-487e-92a2-158c20c932ab.mp4

Note the following points in the video:

0:42: This is the basic behavior of the bug
0:39: I attempt to reproduce the bug **without** the keyboard and the bug does **not** reproduce; this shows the bug does not express if the keyboard is not up.
0:00: There are some unusual attempts at the start of the video where I try to bring up the poll but while it is coming up the media pane brings itself up instead. I think that what is happening is the glitched-menu hitbox for "add media" is weird enough it comes up while I am trying to cancel the poll box. I don't have a consistent repro for this.

* * * *
- Tusky Version: 1.7.0 from Google Play Store **HOWEVER** also tested with git `8c6ccf426103b` from develop branch and saw same symptoms.
- Keyboard: Microsoft SwiftKey 8.10.13.2 **HOWEVER** also tested with stock Google GBoard and saw same symptoms.
- Android Version: Android 12, March 1 2022 update vended by Sony, "Build number 61.1.A.6.2 release-keys"
- Android Device: Sony XPeria 5 III (XQ-BQ62)
- Mastodon instance (if applicable): mastodon.social

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

Gryzor:
@mcclure I took a quick look at this, I'm interested in what you (and/or @connyduck / @charlag ) think about it? Perhaps you folks have better ideas.

connyduck:
Can't reproduce anymore. Looks like this was fixed by https://github.com/tuskyapp/Tusky/pull/4667

mcclure:
Glad to hear it. I think filing this bug was literally the first thing I ever did on this project lol

PR #2627, opened 2022-07-27
2527: Collapse the Add Poll menu *after* the dialog returns. (WIP)

Gryzor:
### Objective
* Ensure the bottom Sheet is correctly collapsed when adding a poll.

### Issue
https://github.com/tuskyapp/Tusky/issues/2527

### Description
* This is a proposal, perhaps there are better solutions.
* I thought the problem was due to this: 
https://github.com/tuskyapp/Tusky/blob/1b6a0908f63be393f47fc65505e90ac7cf239462/app/src/main/java/com/keylesspalace/tusky/components/compose/dialog/AddPollDialog.kt#L106-L107

I was wrong.

I tried combinations of:

1. Use `doOnLayout` `doOnNextLayout` to no avail (though happened less)
2. Move the "collapse" outside the coroutine to gain time in the message queue.
3. Force dismiss the keyboard using InputManager at different stages.

None permanently fixed it. 

Until I tried the code in this branch: 

1. Don't immediately collapse it when the user taps "add poll".
2. If the dialog is cancelled or dismissed, collapse the bottom sheet.
3. If the poll is added, collapse the bottom sheet in the callback.

This works for me, what do you think?

### Side-effect
* The obvious visual side effect is that the bottom sheet is "expanded" while the dialog to add a poll is displayed and this may or may not be "acceptable" from a UX PoV. I personally think it's not an issue but that's because I spent 30 minutes trying other things and it didn't work. I would prefer for the bottom sheet to be more consistent.

### DIFF
--- a/app/schemas/com.keylesspalace.tusky.db.AppDatabase/41.json
+++ b/app/schemas/com.keylesspalace.tusky.db.AppDatabase/41.json
@@ -1,0 +1,935 @@
+{
+  "formatVersion": 1,
+  "database": {
+    "version": 41,
+    "identityHash": "1de8f20c7f28e1f11b33e7a55137feef",
+    "entities": [
+      {
+        "tableName": "DraftEntity",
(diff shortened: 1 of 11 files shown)


connyduck:
Maybe the WindowInsets api can help with this? 🤔 
https://developer.android.com/training/system-ui/sw-keyboard

Gryzor:
Closing this, reopening a new PR soon. I screwed the branch ;)

PR #2629, opened 2022-07-28
2527: Use WindowsInset API to account for the ime/keyboard shown/hidden (WIP)

Gryzor:
### Objective
* Ensure the bottom Sheet is correctly collapsed when adding a poll.

### Issue
https://github.com/tuskyapp/Tusky/issues/2527

### Description
* After closing [this draft PR](https://github.com/tuskyapp/Tusky/pull/2627), I tried the [WindowInset](https://developer.android.com/training/system-ui) api as suggested by @connyduck 
* It's a bit tricky and convoluted, but "does the basic job" we need here. 

### Testing
* Please test to see if it makes a difference in your device(s). It did for my FairPhone 3 with LineageOS 18.1. Previously the add poll bottom sheet bar would break somewhere in the middle of the animation (as described in the issue) and now it works fine. 

### Future Improvements
* The WindowInsetAnimations API could be leveraged to make the animation _smoother_. I did a very small test and it appears to be "more complicated" that what we're after at this time.


connyduck:
Code looks good, but does not work on my device (Fairphone 4), bug still occurs 😭 
Additionally, it totally breaks the tablet layout where the whole bottom bar started to float. Damn it.

Gryzor:
Dang. Well back to the drawing board.

connyduck:
Can't reproduce anymore. Looks like this was fixed by https://github.com/tuskyapp/Tusky/pull/4667

PR #2630, opened 2022-07-28
adapt file size error messages to show real instance upload limit

QuirkyPony:
The file size limits depend on the server; change strings to reflect that. Even better would be to automatically display the server's limits, but that would require Actual Effort from someone who knows what they're doing.


connyduck:
Good catch, I'd prefer to have the real limit displayed.

Do you want to try that (I can help you) or should I close this and do it myself?

connyduck:
In `ComposeViewModel` there is `instanceInfo` which stores an `InstanceInfo` object that is already prefilled with everything you need  - either limits from the server, if possible, or defaults.
Delete the three strings and create 2 new ones (video & audio limit are always the same), so translators will be forced to translate them again.

QuirkyPony:
okay first attempt at writing Kotlin went about as expected…

by the way, am I missing something or is the string for the image size limit error never actually used?

connyduck:
> by the way, am I missing something or is the string for the image size limit error never actually used?

Ah yes it is indeed never used. It probably was once, but now we downsize images in case they are over the limit so the error should never occur. That means we need only one error message for when video/audio is over the limit.

QuirkyPony:
…I probably should have actually gotten the android sdk and done all that locally innit

connyduck:
>…I probably should have actually gotten the android sdk and done all that locally innit

Yes you would notice that 'videoSizeLimit' is in bytes

QuirkyPony:
yeah that's definitely my bad >_> "just a quick fix amiright"

in any case I can't seem to get the toolchain working on my end so the code is all yours :)

connyduck:
Thanks! I think it should now work like this.

PR #2638, opened 2022-08-03
update minSdkVersion to 23

### LINKED ISSUE #2606: Stopping support for Android 5.x
Tusky will stop supporting Android 5.x with version 20.
Current usage of Android 5.x according to Google Play is 0,2% (140 devices).

connyduck:
closes #2606

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -22,7 +22,7 @@ android {
     compileSdkVersion 31
     defaultConfig {
         applicationId APP_ID
-        minSdkVersion 21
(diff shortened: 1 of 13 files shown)


charlag (approved):
Yes!
Do we still need to ship Conscrypt for TLS then? I don't remember when it was added.

connyduck:
Didn't we add it because of an Android 7 bug? Anyway I would keep it because it adds TLS 1.3 support on versions below 10.

Issue #2637, opened 2022-08-02
Option to follow hashtags


guerda:
There's a new Mastodon endpoint to follow hashtags. It would be great if tusky allows to follow hashtags after a search.

https://github.com/mastodon/mastodon/issues/1096

connyduck:
https://github.com/tuskyapp/Tusky/pull/2642

PR #2642, opened 2022-08-04
Add support for following hashtags

Tak:
Addresses #2637

Adds a follow/unfollow toolbar item to the hashtag list activity when instance responds to the new api, and gracefully does nothing when an instance doesn't support it yet.

Tested with:
- glitch.taks.garden (supported)
- mastodon.gamedev.place (unsupported)

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/StatusListActivity.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/StatusListActivity.kt
@@ -23,14 +23,14 @@ import android.view.Menu
 import android.view.MenuItem
 import android.widget.Toast
 import androidx.fragment.app.commit
-import androidx.lifecycle.Lifecycle
-import autodispose2.androidx.lifecycle.autoDispose
+import androidx.lifecycle.lifecycleScope
 import com.keylesspalace.tusky.components.timeline.TimelineFragment
 import com.keylesspalace.tusky.components.timeline.viewmodel.TimelineViewModel.Kind
 import com.keylesspalace.tusky.databinding.ActivityStatuslistBinding
 import dagger.android.DispatchingAndroidInjector
 import dagger.android.HasAndroidInjector
-import io.reactivex.rxjava3.android.schedulers.AndroidSchedulers
+import kotlinx.coroutines.launch
+import kotlinx.coroutines.rx3.await
 import javax.inject.Inject
 
 class StatusListActivity : BottomSheetActivity(), HasAndroidInjector {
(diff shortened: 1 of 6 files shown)


connyduck:
Nice!

Can you implement it with suspending functions? I'd like to have no new Rx code.

Tak:
Ok, done

app/src/main/java/com/keylesspalace/tusky/network/MastodonApi.kt
```diff
@@ -656,4 +657,13 @@ interface MastodonApi {
...
     ): NetworkResult<ResponseBody>
+
+    @GET("api/v1/tags/{name}")
+    fun tag(@Path("name") name: String): Single<HashTag>
+
+    @POST("api/v1/tags/{name}/follow")
+    fun followTag(@Path("name") name: String): Single<HashTag>
+
+    @POST("api/v1/tags/{name}/unfollow")
+    fun unfollowTag(@Path("name") name: String): Single<HashTag>
```
connyduck, line 668: make these `suspend fun call(...): NetworkResult<Hashtag>` and then call them like so:
```
mastodonApi.call(...)
    .fold(
        { tag ->
            ...     
        },
       { throwable ->
           ...      
       }
)
```

app/src/main/java/com/keylesspalace/tusky/StatusListActivity.kt
```diff
@@ -67,6 +79,64 @@ class StatusListActivity : BottomSheetActivity(), HasAndroidInjector {
...
+        val tag = hashtag
+        if (tag != null) {
+            val context = this
+            lifecycleScope.launch {
+                try {
+                    mastodonApi.followTag(tag).await()
+                    followTagItem?.isVisible = false
+                    unfollowTagItem?.isVisible = true
+                } catch (_: Exception) {
+                    Toast.makeText(context, "Error following #$tag", Toast.LENGTH_SHORT).show()
```
connyduck, line 114: the error message should be translateable, please additionally log the exception (same for unfollwing ofc)

app/src/main/java/com/keylesspalace/tusky/StatusListActivity.kt
```diff
@@ -67,6 +79,64 @@ class StatusListActivity : BottomSheetActivity(), HasAndroidInjector {
...
+        val tag = hashtag
+        if (tag != null) {
+            val context = this
+            lifecycleScope.launch {
+                try {
+                    mastodonApi.unfollowTag(tag).await()
+                    followTagItem?.isVisible = true
+                    unfollowTagItem?.isVisible = false
+                } catch (_: Exception) {
+                    Toast.makeText(context, "Error unfollowing #$tag", Toast.LENGTH_SHORT).show()
```
connyduck, line 132: Snackbar instead of Toast, please

connyduck (changes requested):
This icon as unfollow icon? https://materialdesignicons.com/icon/account-remove

The icons also need to be tinted properly, you can do this by adding `app:iconTint="?attr/colorOnSurface"` to the items in the menu xml.

app/src/main/java/com/keylesspalace/tusky/network/MastodonApi.kt
Tak: Oh sorry, I looked at the wrong sample

Tak:
Ok, I think I've addressed everything

app/src/main/java/com/keylesspalace/tusky/StatusListActivity.kt
```diff
@@ -67,6 +80,72 @@ class StatusListActivity : BottomSheetActivity(), HasAndroidInjector {
...
+        if (tag != null) {
+            val context = findViewById<View>(R.id.action_unfollow_hashtag)
+            lifecycleScope.launch {
+                mastodonApi.unfollowTag(tag).fold(
+                    {
+                        followTagItem?.isVisible = true
+                        unfollowTagItem?.isVisible = false
+                    },
+                    {
+                        Snackbar.make(context, getString(R.string.error_unfollowing_hashtag_format, tag), Snackbar.LENGTH_SHORT).show()
```
connyduck, line 139: move the `binding` creation from `onCreate` to a class member like so `private val binding by viewBinding(ActivityStatuslistBinding::inflate)` and then you don't need the `findViewById`and can just do
`Snackbar.make(binding.root ...`

Also, it would be easy to add a retry button to the snackbar if you want.

guerda:
Thanks, that is great! :)

PR #2643, opened 2022-08-04
fix caching of instance defaults and emojis

connyduck:
The problem was that the insert  with `OnConflictStrategy.REPLACE` overrode previous values, so only emojis OR instance defaults were cached, never both, depending on which loaded last. Using upsert instead fixes this.


app/src/main/java/com/keylesspalace/tusky/db/InstanceDao.kt
```diff
@@ -20,15 +20,37 @@ import androidx.room.Insert
...
+import androidx.room.Transaction
+import androidx.room.Update
 
 @Dao
 interface InstanceDao {
 
-    @Insert(onConflict = OnConflictStrategy.REPLACE, entity = InstanceEntity::class)
-    suspend fun insertOrReplace(instance: InstanceInfoEntity)
+    @Insert(onConflict = OnConflictStrategy.IGNORE, entity = InstanceEntity::class)
+    suspend fun insert(instance: InstanceInfoEntity): Long
```
charlag, line 30: I would rename them, `IGNORE` can bite us in the butt later

charlag (approved):

Issue #2459, opened 2022-04-26
Support setting toot language


Natureshadow:
It would be great if the toot language could be set when sending a new toot.

The API seems to support this: https://github.com/mastodon/mastodon/issues/8753

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

hex-m:
> Add language dropdown to compose form in web UI

Version [3.5.3](https://github.com/mastodon/mastodon/releases/tag/v3.5.3) of the Mastodon Web-Interface supports this. It will probably be used more regularly in the future.

chrysn:
As #2651 didn't explcitly contain a "Closes: #245", this wasn't closed automatically, but I think it can be now.

mmahmoudian:
Considering that #2651 was merged about 3 months ago and there is a mass migration to mastodon, I would like to suggest releasing the next version. But I wonder if there is already a timeline or is there a defined release cycle?

connyduck:
@mmahmoudian https://chaos.social/@ConnyDuck/109355257388353984
You could also join the dev chat on matrix where we discuss releases

connyduck:
This release was delayed by Google play, we couldn't publish for a while

Natureshadow:
Rumour has it there are other ways to publish software besides Google Play.

charlag:
> Rumour has it there are other ways to publish software besides Google Play.

Excuse me? what is this passive-aggressive entitled statement is supposed to mean? Do you not know we are distributing on F-Droid? Would *you* like to set up a nightly F-Droid build?  (there is is one nightly repo by @nailyk-fr to my knowledge)

Natureshadow:
> Excuse me? what is this passive-aggressive entitled statement is supposed to mean? Do you not know we are distributing on F-Droid? Would *you* like to set up a nightly F-Droid build?  (there is is one nightly repo by @nailyk-fr to my knowledge)

You may want to read my comment again, and also the one I replied to.

The argument was that a release was entirely impossible because issues with Google Play.

So, if you think my remark is wrong, please enlighten me on why Google can prevent a release to F-Droid.

charlag:
Because we are people who do this in our free time and we have limited resources and we doing the release at the same time everywhere. I misunderstood your comment but you tone was still demanding.

nailyk-fr:
Google is not friendly at all and I'd happy to explain why. But this is another topic :slightly_smiling_face: 

[...]

PR #2651, opened 2022-08-05
Add language dropdown to compose view

Tak:
#2459 

- Add dropdown to compose view
  - iso language code on spinner button
  - language name in current language (language name in target language) on dropdown entries
- Language is included when posting statuses
- Language is saved/restored with drafts
- Compose language defaults to the language chosen for Tusky or the system language

Known issues:
- ~~We're tight on UI space in the compose view bottom bar, maybe need to rearrange~~
- ~~At least one of our translations (norsk bokmål) is keyed with a language code that doesn't match the one android gives it, making the defaulting not work (see [matrix message](https://matrix.to/#/!PWkUGGpeMZWDPmdfRE:matrix.org/$1659728795246423pnsUm:matrix.org?via=m.charlag.space&via=matrix.org&via=tchncs.de))~~
- Dropdown doesn't prioritize recently used languages (always ordered by iso language code)
- General UI guidance that I seem to always need :wink:

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/adapter/LocaleAdapter.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/adapter/LocaleAdapter.kt
@@ -1,0 +1,28 @@
+package com.keylesspalace.tusky.adapter
+
+import android.content.Context
+import android.graphics.Typeface
+import android.view.View
+import android.view.ViewGroup
+import android.widget.ArrayAdapter
+import android.widget.TextView
+import com.keylesspalace.tusky.util.ThemeUtils
+import java.util.*
+
+class LocaleAdapter(context: Context, resource: Int, locales: List<Locale>) : ArrayAdapter<Locale>(context, resource, locales) {
+    override fun getView(position: Int, convertView: View?, parent: ViewGroup): View {
+        return (super.getView(position, convertView, parent) as TextView).apply {
+            setTextColor(ThemeUtils.getColor(context, android.R.attr.textColorTertiary))
+            typeface = Typeface.DEFAULT_BOLD
+            text = super.getItem(position)?.language?.uppercase()
+        }
+    }
+
+    override fun getDropDownView(position: Int, convertView: View?, parent: ViewGroup): View {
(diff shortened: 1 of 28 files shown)


connyduck:
Works quite well already, thx. My main problem with it is that we don't have enough space for the button in the bottom bar.
Either we remove the #/@ and put it there, or we move the avatar into the TextField like Twitter and the Mastodon app do.

Tak:
Yeah, same. We could put it *next to* #@ for now?

connyduck:
> Yeah, same. We could put it next to #@ for now?

yes! please do

connyduck:
(sry I did not want to edit your comment)

Tak:
😄

app/src/main/res/layout/activity_compose.xml
```diff
@@ -22,6 +22,19 @@
             tools:ignore="ContentDescription" />
         <!--content description will be set in code -->
 
+        <Spinner
+            android:id="@+id/composePostLanguageButton"
+            style="@style/TuskyImageButton"
+            android:layout_width="44dp"
```
connyduck, line 28: I think it needs to be set to something larger or even `wrap_content`

app/src/main/java/com/keylesspalace/tusky/adapter/LocaleAdapter.kt
```diff
@@ -0,0 +1,28 @@
+package com.keylesspalace.tusky.adapter
```
connyduck, line 1: add the Tusky license header please

connyduck (commented):
Works very well 😍

app/src/main/res/layout/activity_compose.xml
Tak: Argh, what language is that?
The issue with `wrap_content` is that it wants to size itself for the dropdown items instead 😐.
Although if you know of a workaround (or if I'm just Doing It Wrong)…

Tak: Ok, checked this variation against all languages installed in my emulator

connyduck (approved):

PR #2652, opened 2022-08-07
Rename norsk bokmål translation from no_NB to nb_NO

Tak:
Update the language tag for norsk bokmål from `no_NB` to `nb_NO`, and migrate existing users


connyduck:
What happens if we don't do the migration? User that had it selected in app will have to reselect? Asking because such one-time fixes are quite ugly to have in the code and unless it crashes, I'd not do it.

Tak:
Yeah, users that previously had "no-nb" stored will get the default locale after upgrading and have to reset the preference

connyduck:
Ok in that case, please remove the migration.

PR #2664, opened 2022-08-15
Upgrade robolectric

Tak:


app/src/test/java/com/keylesspalace/tusky/db/TimelineDaoTest.kt
```diff
@@ -111,10 +111,10 @@ class TimelineDaoTest {
...
 
         val expectedAccounts = listOf(
-            1L to "3",
             1L to "10",
-            1L to "R10",
             1L to "20",
+            1L to "3",
+            1L to "R10",
             2L to "5"
         )
```
Tak, line 120: After the upgrade, the following check was failing with the lists having the same contents, but differently ordered

PR #2671, opened 2022-08-15
improve sending error notifications

### LINKED ISSUE #2635: “Sending cancelled“ notifications are not dismissable
When I cancel a image uploading, the sending notification becomes "sending cancelled" but it's not dismissable nor clickable and I have to force stop tusky to get rid of the notification. 

* * * *
- Tusky Version: 19.0
- Android Version: 12
- Android Device: Pixel 6 Pro
- Mastodon instance (if applicable): pixelfed.social

connyduck:
Clicking the error or cancel notification will now open drafts, both can be dismissed and the cancel one auto-cancels after 5 sec. Not sure about the auto cancelling, but it was the original behavior that got broken in some refactoring.

Closes #2635


app/src/main/java/com/keylesspalace/tusky/service/SendStatusService.kt
```diff
@@ -259,7 +266,41 @@ class SendStatusService : Service(), Injectable {
...
-        return PendingIntent.getService(this, statusId, intent, NotificationHelper.pendingIntentFlags(false))
+        return PendingIntent.getService(
+            this,
+            statusId,
+            intent,
+            NotificationHelper.pendingIntentFlags(false)
+        )
+    }
+
+    private fun draftNotification(
```
charlag, line 277: I would rename it to `showDraftNotification()`

charlag (approved):
I think cancelling error is really misleading, otherwise should be fine?

app/src/main/java/com/keylesspalace/tusky/service/SendStatusService.kt
connyduck: it is not showing, just building it though

connyduck:
> I think cancelling error is really misleading, otherwise should be fine?

what exactly is misleading?

charlag:
> > I think cancelling error is really misleading, otherwise should be fine?
> 
> what exactly is misleading?

You send a post, throw the phone in your pocket, it fails and shows notifications for 5s, you never see that it fails. Or does it not work like that? I might be confused

connyduck:
> > > I think cancelling error is really misleading, otherwise should be fine?
> > 
> > 
> > what exactly is misleading?
> 
> You send a post, throw the phone in your pocket, it fails and shows notifications for 5s, you never see that it fails. Or does it not work like that? I might be confused

The error notification stays forever until user clicks or dismisses. If you click cancel on an ongoing sending notification, it turns into a cancel notification that stays only for 5 seconds.

charlag:
Ah, sorry, I see now, it's good!

PR #2684, opened 2022-08-21
AccountMediaFragment improvements

connyduck:
- AndroidX paging + coroutines
- use blurhash where possible
- respect sensitive media settings
- show video indicator
- added a bit of spacing between items 
- better preloading for smooth scrolling
- The old implementation openend all loaded attachments in the detail view, which made it necessary to put them all into Intent Extras, which caused `android.os.TransactionTooLargeException` if there were a lot of them. I fixed this by only openening attachments that belong to the same status in the detail view. This behavior also matches the web ui.


charlag (commented):
OMG I was just thinking few weeks ago how we don't hide sensitive media, thank you so much for addressing this.
The code looks great, I have few comments on functionality tho:
 - opening media animation is glitchy, unlike before, it makes the image jump around, especially if the image is narrow
 - audio is not indicated in any way
 - there's no way to know what is behind sensitive media, would be nice if it read the description of media on long press, like in timelines. Would be great to read post CW too?
 - I didn't check it but I think media description is not set on an views either?

connyduck:
All fixed!

charlag (approved):
Awesome!

Issue #2188, opened 2021-06-02
[Feature Request] LibreTranslate integration


PJ-Finlay:
https://github.com/uav4geo/LibreTranslate

adeekshith:
Also see #134

connyduck:
I like this way more than previous suggestions that involved close-source apis.

Does anyone know what they charge for an api key? Or probably better, who can host an instance for us?

a1batross:
@connyduck 
I can help with hosting, though it will be much better to have fallback LibreTranslate servers.

PJ-Finlay:
You can buy API keys from the project which helps support future development! There are also mirrors and self hosting is always a good option.

- https://libretranslate.de/
- https://translate.astian.org/

@pierotofy

connyduck:
If we hardcode an instance into Tusky, we will also need a privacy policy. I think it would be cool if the user can choose their instance in the settings and the translation option is only available when it is configured.

connyduck:
Mastodon added a new api for translation, this will be the way to go
https://github.com/mastodon/mastodon/pull/19434
We can detect the availabilty of the api via /api/v2/instance https://github.com/mastodon/mastodon/pull/19434

PJ-Finlay:
Mastodon also now support LibreTranslate.

nikclayton:
It is also possible to do this on-device with https://developers.google.com/ml-kit/language/translation/android.

- Translation is on-device, no content is sent to a third party (LibreTranslate, DeepL, etc)
- Translation is presumably faster, and more reliable because of that
- Translation works offline

There are a couple of caveats:

- The option to start the translation must be labelled "Translate with Google", per https://cloud.google.com/translate/attribution#mobile_apps
- The translation results must have a "translated by Google" badge at the bottom. There are UI examples of this at https://cloud.google.com/translate/attribution#attribution_and_logos
- There must be a disclaimer in the app (e.g., in the preferences that enable this feature), https://cloud.google.com/translate/attribution#disclaimer
- Per https://developers.google.com/ml-kit/terms#privacy, content to be translated does not leave the device, but "The ML Kit APIs may contact Google servers from time to time in order to receive things like bug fixes, updated models and hardware accelerator compatibility information. The ML Kit APIs also send metrics about the performance and utilization of the APIs in your app to Google. [...] You are responsible for informing users of your app about Google’s processing of ML Kit metrics data as required by applicable law."

[...]

PR #2685, opened 2022-08-24
Added support for status translation with lingva.ml

retiolus:
An option is missing in the settings to let the user choose his Lingva instance.

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/components/preference/PreferencesFragment.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/components/preference/PreferencesFragment.kt
@@ -22,14 +22,7 @@ import com.keylesspalace.tusky.R
 import com.keylesspalace.tusky.db.AccountManager
 import com.keylesspalace.tusky.di.Injectable
 import com.keylesspalace.tusky.entity.Notification
-import com.keylesspalace.tusky.settings.AppTheme
-import com.keylesspalace.tusky.settings.PrefKeys
-import com.keylesspalace.tusky.settings.emojiPreference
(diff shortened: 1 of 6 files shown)


connyduck:
Nice but

- I'd prefer LibreTranslate
- Calling an api and showing the translation in our interface should be the implementation we want. Having the option only on detailed statuses and adding the translation below the original content like Twitter does it maybe.
- we should host a translation service at translate.tusky.app and use it as default instance before making the feature available to users
see also https://github.com/tuskyapp/Tusky/issues/2188

thedaviddelta:
Hello everyone, Lingva creator here!

I'm super glad for this work, and super happy that Lingva is being considered for this key task. But I agree with @connyduck about using LibreTranslate.
I created Lingva as a Free as in Freedom front-end for enable people the use of Google Translate engine without loading any propietary Google code in your browser, just like Invidious or Nitter. But if we want to implement any translation service to other Free projects, we should use Free tools, not just front-end substitutes.

For example, by using Lingva here, we would be sending all the translation traffic directly to Google, even if it's using unofficial ways with less tracking, or even if it's completely anonymized because it would travel only between the Lingva and the Google servers.

The LibreTranslate and ArgosTranslate projects are amazing, and they're creating a completely Free translation engine by themselves. I always try to name and support them in case more privacy-aware or FOSS-aware people prefer to use a completely Free engine, even if currently is not as good as the billion-dollar-Google one.

Once again, thanks for the work with this. Translations have been the missing feature for both Mastodon and Tusky for some time. This work will help to bring them closer, using whatever engine we decide to use.
Regards, David.

connyduck:
we now support them Mastodon translation api

PR #2689, opened 2022-08-28
Fix broken timeline when there are only expired filters

charlag:
The issue happened when the only applicable filters are expired. There
was a check to not produce an empty regex when there are no filters but
it was done before removing expired filters so we would produce an
empty regex that would match (and remove) everything and the timeline
would get stuck.


Tak (approved):
🤦

charlag:
how do I configure idea to pass the bloody lint check?

connyduck:
> how do I configure idea to pass the bloody lint check?

There is a gradle task `ktlintApplyToIdea` that should configure it correctly and after that "reformat code" should format the code so it passes the check. Alternatively run `ktlintFormat` to let ktlint format the code.

charlag:
Thanks, I forgot about it.
I won't be able to do it until the evening today, feel free to fix it if you want to merge it sooner

app/src/main/java/com/keylesspalace/tusky/network/FilterModel.kt
```diff
@@ -54,9 +55,9 @@ class FilterModel @Inject constructor() {
     }
 
     private fun makeFilter(filters: List<Filter>): Pattern? {
-        if (filters.isEmpty()) return null
-        val tokens = filters
-            .filter { it.expiresAt?.before(Date()) != true }
+        val nonExpiredFilters = filters.filter { it.expiresAt?.before(Date()) != true }
```
connyduck, line 58: How about saving some allocations by moving the `Date()` out of the loop?

connyduck (approved):

Issue #2626, opened 2022-07-26
Accessibility: keep draft image descriptions


davidbtc2009:
It is too easy to close the image description box and loose the text you have entered for the description by accidentally tapping outside of the dialog. The current text should be saved unless the "cancel" button is pressed

* * * *
- Tusky Version: 19
- Android Version: 11
- Android Device: Motorola One 5G Ace
- Mastodon instance (if applicable):

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

connyduck:
fixed by https://github.com/tuskyapp/Tusky/pull/2693

PR #2693, opened 2022-09-11
Refactor Caption Dialog to handle screen rotation (#2626)

pt2121:
re #2626

Refactored the dialog to be `CaptionDialog` and save/restore the text input state.
I tried to keep existing logic where I can. The main difference is how the dialog communicates with the Activity.


connyduck (approved):
Very nice, thanks for the help!

PR #2697, opened 2022-09-13
Add display of handle when using multiple accounts

viviicat:
- Shown on the main toolbar (subtitle)
- Shown just above the "replying to" message (even if not replying)

I made this feature for users who frequently post across multiple accounts with the same avatar. Currently there isn't an easy way to know which timeline you're on from the main view without sliding the sidebar open, and when posting (rather important) it's impossible to know who you're posting as, unless you have a unique avatar.

I am hoping these changes are minor enough that they should be OK. Please let me know if any tweaks would improve this.

Note that these altered visuals are only present when more than one account is present in the list.

I decided to change the text of `compose_active_account_description` instead of adding a new entry, because the text seems to only be used for accessibility and it seems like it would be OK to make use of it in this way. Also at least the German translation seems to match the new text, so seems there is some inconsistency anyway.

When replying:
Screenshot_20220912_235329

On the top toolbar (for all tabs):
Screenshot_20220912_235412

When creating a new post:
Screenshot_20220912_235446

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/MainActivity.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/MainActivity.kt
@@ -834,6 +834,9 @@ class MainActivity : BottomSheetActivity(), ActionButtonActivity, HasAndroidInje
         header.clear()
         header.profiles = profiles
         header.setActiveProfile(accountManager.activeAccount!!.id)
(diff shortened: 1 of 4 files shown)


connyduck (approved):
Thank you

Issue #2694, opened 2022-09-12
Indicate that the link is hypertexted or it is the real address that is rendered as a link


mmahmoudian:
## The problem

Considering that there are plenty of services use ActivityPub and many let the user to create hyperlinks for part of the text, it is possible to have something like `<a href="https://malicious.url">https://safe.url</a>` and it would render it as [https://safe.url](https://malicious.url). Therefore, user would not know that what they are clicking on will take them. A less malicious thing would be to hide some tracking code but the text only shows the main website.

## My suggestion

Show an icon at either end of the link to indicate that the link is hyperlinked text and is not a URL that is rendered as a link in Tusky.

* * * *
- Tusky Version: 19.0
- Android Version: -- Not Applicable --
- Android Device: -- Not Applicable --
- Mastodon instance (if applicable): -- Not Applicable --

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

connyduck:
fixed by #2698

PR #2698, opened 2022-09-13
Show target domains for non-mention/non-hashtag links where the target domain is not provided or differs from the domain in the text

Tak:
Addresses #2694

My *instinct* is to spend more time trying not to iterate the content multiple times, but maybe it's ok because many posts don't contain links/tags/mentions, and they only have a few when they do 🤷 

Screen Shot 2022-09-13 at 10 07 17

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/util/LinkHelper.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/util/LinkHelper.kt
@@ -83,7 +83,7 @@ fun markupHiddenUrls(context: Context, content: CharSequence, mentions: List<Men
         val start = spannableContent.getSpanStart(span)
         val end = spannableContent.getSpanEnd(span)
         val originalText = spannableContent.subSequence(start, end)
-        val replacementText = context.getString(R.string.url_domain_notifier).format(originalText, getDomain(span.url))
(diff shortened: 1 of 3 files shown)


connyduck:
Nice, but I am also seeing it in mentions both in profiles and in posts:
Screenshot_20220913_202746

Tak:
🤔 I expect it in the profile, because we don't pass `mentions` there, but I don't understand why you're seeing it in the post below - in fact, I don't see it locally on exactly the same activity/post.

I could also back down on the mention/hashtag verification and just not display it for linked text that starts with `@`/`#`?

Tak:
Screenshot_20220914-102053

connyduck:
In the profile it is because there we have no `mentions` array, and in the post where you mention maloki it is because the url of the mention is https://rage.love/users/maloki but the one of the Span is https://rage.love/@maloki
You probably don't see it because we are on different instances.

> I could also back down on the mention/hashtag verification and just not display it for linked text that starts with @/#?

That is probably the solution, yes

Tak:
> the url of the mention is https://rage.love/users/maloki but the one of the Span is https://rage.love/@maloki
h...uh

So if you tap that mention, it goes through the whole "looking up" thing, like when we see a normal url that looks like it could be a mention?

PR #2700, opened 2022-09-15
Fix the tag span generation for tags with nonascii characters

Tak:
I noticed this while testing https://github.com/tuskyapp/Tusky/pull/2698 - hashtags like #bitsundbäume, #séance21h, and #GroßerGarten weren't getting the custom tag span

Note: there are some inconsistencies with non-mastodon servers: for example, pixelfed normalizes "GroßerGarten" to "grossergarten" and mastodon makes it "großergarten", but I guess our approach is generally to follow mastodon

- Update mention and tag regexes from current mastodon
- Perform unicode => ascii normalization the way that mastodon does
- Add test cases for nonascii tags

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/util/AsciiFolding.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/util/AsciiFolding.kt
@@ -1,0 +1,26 @@
+/* Copyright 2022 Tusky contributors
+ *
+ * This file is a part of Tusky.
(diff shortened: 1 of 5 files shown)


Tak:
Oops, this fixes the spans, but there are more issues - I was too confident with the test coverage 😐

connyduck (approved):
Oh wow, thanks for finding this

PR #2718, opened 2022-10-15
Preference to disable multiple-login usernames (draft)

mcclure:
PR#2697 is a good change and especially useful for people with multiple logins with the same avatar. However all my logins have different avatars so to me it is just clutter and I'd be happier without it. When 2697 was merged it was suggested we could add a preference for it. This patch makes it a preference.

This is not ready to merge because of the problem described below/in Matrix.


app/src/main/java/com/keylesspalace/tusky/components/preference/PreferencesFragment.kt
```diff
@@ -145,6 +145,13 @@ class PreferencesFragment : PreferenceFragmentCompat(), Injectable {
                     isSingleLineTitle = false
                 }
 
+                switchPreference {
```
mcclure, line 148: Something I do not understand:
* Changing this preference has no effect on the main view unless MainActivity.onCreate is re-called
* Changing this preference does not recreate MainActivity
* But changing the App Theme preference above (line 55) does recreate MainActivity (so any change to this preference takes effect)
* How do I make this preference recreate MainActivity when it changes?

app/src/main/res/values/strings.xml
```diff
@@ -617,6 +617,7 @@
     <string name="no_scheduled_posts">You don\'t have any scheduled posts.</string>
     <string name="no_announcements">There are no announcements.</string>
     <string name="warning_scheduling_interval">Mastodon has a minimum scheduling interval of 5 minutes.</string>
+    <string name="pref_title_show_self_username_disambiguation">Show username when multiple accounts are logged in</string>
```
mcclure, line 620: This string is long enough it may be confusing to read. However all simpler strings I can think of are vague. Solutions I can think of:
* "Show username when multiple accounts are in use"?
* We could make this a selector, "Explicitly show username:" "Always" "When multiple accounts are logged in" "Never" this might scan better, but also is complexity I'm not sure anyone needs
Is "username" even the right term here?

app/src/main/java/com/keylesspalace/tusky/components/preference/PreferencesFragment.kt
mcclure: Fixed

mcclure:
Works in testing now, wording question is open but marking as non-draft.

app/src/main/res/values/strings.xml
connyduck: There is also the possibility to use a "summary" to clarify a setting, e.g. like this:

mcclure: Oh, I think that's good. Do you think it's a problem that the label says "main toolbar" but the setting applies to main toolbar AND compose view?

I guess people will figure it out quick...

connyduck: Maybe we are thinking too complicated. Not showing the username when there is only one login made sense without options. But now it could just be "show username in toolbars", default off. Maybe someone even wants to see the username with only one account.

mcclure: I like that wording...

mcclure:
Screenshot_20221016-192322
Screenshot_20221016-192325

mcclure:
I went ahead with the selection approach. I can switch it over to switch+subtitle if you want, but I think it is probably the best way to match user expectation (if we just add an on/off switch, I think most users will simply never notice this was added)

One last question, where should this go in the list? I just realized the place I put it is particularly illogical

I think it should probably go either below or above "hide top toolbar"  
If I don't hear back I'm going to drop one more commit putting it above "hide top toolbar"

connyduck (approved):

PR #2719, opened 2022-10-16
Api 33 support

connyduck:
Pretty basic Api level upgrade, fixing of new deprecations, handling notification permission


app/src/main/java/com/keylesspalace/tusky/MainActivity.kt
```diff
@@ -267,6 +272,33 @@ class MainActivity : BottomSheetActivity(), ActionButtonActivity, HasAndroidInje
...
+                        else -> {
+                            finish()
+                        }
+                    }
+                }
+            }
+        )
+
+        if (ContextCompat.checkSelfPermission(this, Manifest.permission.POST_NOTIFICATIONS) != PackageManager.PERMISSION_GRANTED) {
+            ActivityCompat.requestPermissions(
```
connyduck, line 296: This logic very basic but I think it is enough? Android will not show the dialog if a user denied the Permission more than twice anyway.

connyduck:
Will do the native language switching thing in another pull request, maybe Appcompat 1.6.0 is stable until then

app/src/main/java/com/keylesspalace/tusky/TabPreferenceActivity.kt
```diff
@@ -74,6 +74,12 @@ class TabPreferenceActivity : BaseActivity(), Injectable, ItemInteractionListene
 
     private val hashtagRegex by lazy { Pattern.compile("([\\w_]*[\\p{Alpha}_][\\w_]*)", Pattern.CASE_INSENSITIVE) }
 
+    private val onFabDismissedCallback = object: OnBackPressedCallback(false) {
```
charlag: why `false`?

app/src/main/java/com/keylesspalace/tusky/components/preference/PreferencesActivity.kt
```diff
@@ -148,20 +156,6 @@ class PreferencesActivity :
         overridePendingTransition(R.anim.fade_in, R.anim.fade_out)
     }
 
-    override fun onBackPressed() {
-        /* Switching themes won't actually change the theme of activities on the back stack.
```
charlag, line 152: we've lost this comment

app/src/main/java/com/keylesspalace/tusky/MainActivity.kt
charlag: Seems fine

app/src/main/java/com/keylesspalace/tusky/TabPreferenceActivity.kt
connyduck: So it is inactive. It gets activated when the FAB is extended and on back press collapses it. When the FAB is not extended I want the default behavior on back press = close the activity.

app/src/main/java/com/keylesspalace/tusky/components/preference/PreferencesActivity.kt
connyduck: added it back

K900:
Looks like this change broke picking/saving files on Android 13 (possibly only QPR1 beta? Can't test right now). Attempting to save or post any file just gives the "permission required" snackbar with no OS permission prompt or anything. The stable version built against API 31 works.

connyduck:
@K900 please create an issue so I don't forget to check that

K900:
Filed https://github.com/tuskyapp/Tusky/issues/2801

PR #2722, opened 2022-10-17
Retain instance state in compose activity

fynngodau:
This PR improves the restoration behavior of `ComposeActivity` for the case that it is killed by the Android system and later restored from instance state.

Some information, like post content, is retained correctly by the Android system. Certain information is however currently lost during kill&restore:
* post visibility
* schedule time
* visibility of content warning field (however, not the content warning text itself)

To test:

1. Open developer settings, scroll to "Apps", toggle "Don't keep activities".
2. Open Tusky, compose a new post.
3. Expand content warning field, set a visibility that is different from the default, and a schedule time.
4. Open the app switcher (this kills `ComposeActivity`) and open it again.

Without this PR:

5. Observe that the CW field is collapsed (even though the text is still hidden inside), the default visibility is back and the schedule time has been removed.

With this PR:

5. The information is restored correctly and you can finish composing your post ;)

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/components/compose/ComposeActivity.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/components/compose/ComposeActivity.kt
@@ -281,7 +281,6 @@ class ComposeActivity :
             }
 
             it.getString(SCHEDULED_TIME_KEY)?.let { time ->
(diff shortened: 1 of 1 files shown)


connyduck (approved):
Wow this is nice, thank you

PR #2730, opened 2022-10-28
Use lighter actions buttons / lighter divider

nethad:
Hi!

I've exchanged the toot actions buttons (reply, boost, favorite, bookmark) for lighter versions. I also made the dividers between the toots lighter.

Not sure if this is a welcome change, I played around with it and decided to open a PR and happy to receive some feedback.
Cheers!

Preview of the new buttons:

new_buttons

Old buttons for comparison:

old_buttons


connyduck:
I'm against this because all icons should be from the same icon set.

charlag:
> I'm against this because all icons should be from the same icon set.

Same, good icons but they just don't fit.

nethad:
@connyduck do you mean the ones visible here or the complete app should have the same icons?

connyduck:
Yes the complete app. If there is a good open iconset we can definitely discuss changing all icons. An important criteria is the number of available icons. The material design one used now has so many icons it is never a problem to find a fitting one which is great.

nethad:
@connyduck I see here that the weight is configurable: https://fonts.google.com/icons?icon.query=repeat&icon.platform=android

Would it make sense to replace the icons with a lower weight?

connyduck:
Sure, if it looks nice. Can you show me on a few icons please?

snaggen:
The current action bar was the first thing I noticed about Tusky, it takes a bit to much focus from the actual content. The suggested style would be a big improvement in this regard. 
However, I fully understand the desire for consistency. I hope it is possible to update the full app with a similar icon set.

scottiefalkon:
> Yes the complete app. If there is a good open iconset we can definitely discuss changing all icons. An important criteria is the number of available icons. The material design one used now has so many icons it is never a problem to find a fitting one which is great.

This is a simple and elegant set. I looked through and think it has everything you would need. 

https://feathericons.com/

PR #2745, opened 2022-11-04
Introduce Gradle version catalogs for versioning

colintheshots:
Gradle version catalogs seem to make it much easier to modify versions and to see what 3rd party dependencies are in use at a glance. They'll also auto-complete after switching to build.gradle.kts files. I just like working on a cleaner project.


connyduck:
Looks good! Please resolve the merge conflict and I'll merge it

colintheshots:
This should be good to merge now

PR #2750, opened 2022-11-05
[UX Improvement] Always show the toolbar, but hide section title

evanisnor:
Hi, Tusky! 👋 ❤️ 

Fixing an issue where enabling _**Hide the title of the top toolbar**_ in preferences would hide the entire toolbar. This would make it ~~impossible~~ DIFFICULT/CONFUSING to open the App's navigation drawer, therefore locking the user out of a significant portion of the app, including Preferences. The user would have to delete the App Cache for it to re-appear if they didn't know about long-pressing on the left side of the screen (which they probably don't).

Instead, let's set the toolbar title to the `app_name` string and avoid setting the title with the section names. This avoids putting the user in a bad UX state.

### Summary of changes
Improving intent of the _**Hide the title of the top toolbar**_ preference
- Rename PrefKeys constant `HIDE_TOP_TOOLBAR` to `HIDE_TOP_TOOLBAR_TITLE` 
- During MainActivity `onCreate`, set the title of the toolbar to `Tusky` from `R.string.app_name`
- When a tab is selected, only change the title of the toolbar to `TabData.title` if the `HIDE_TOP_TOOLBAR` preference is enabled. Otherwise, the title will remain `Tusky`
- Always showing the Search menu action on the toolbar, because we want to always show the toolbar!

### Comparison

| Before | After |
|----|----|
| <video src="https://user-images.githubusercontent.com/478342/200122525-358075e3-d85b-4cda-8c11-cc4226fe8648.mp4" width="400" /> | <video src="https://user-images.githubusercontent.com/478342/200122552-0869a462-83e9-4f0f-a97a-7650b6856211.mp4" width="400" /> |

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/MainActivity.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/MainActivity.kt
@@ -90,7 +90,6 @@ import com.keylesspalace.tusky.util.hide
 import com.keylesspalace.tusky.util.show
 import com.keylesspalace.tusky.util.updateShortcut
 import com.keylesspalace.tusky.util.viewBinding
-import com.keylesspalace.tusky.util.visible
 import com.mikepenz.iconics.IconicsDrawable
 import com.mikepenz.iconics.typeface.library.googlematerial.GoogleMaterial
 import com.mikepenz.iconics.utils.colorInt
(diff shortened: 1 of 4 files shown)


connyduck:
We introduced the option to hide the whole bar because people on small screens wanted more space, this doesn't help.
We need to go back to the old Tusky design when the top bar is hidden. It looked like this https://github.com/tuskyapp/Tusky/blob/3fb766fc68cee4c742264c390015a6521c28e9e7/fastlane/metadata/android/en-US/images/phoneScreenshots/01_timeline.png

connyduck:
@evanisnor Are you in the dev chat? We are discussing that exact issue right now. https://riot.im/app/#/room/#Tusky:matrix.org

evanisnor:
> We introduced the option to hide the whole bar because people on small screens wanted more space, this doesn't help.

The toolbar auto-hides when you scroll. Wouldn't that help in that case?

connyduck:
> The toolbar auto-hides when you scroll. Wouldn't that help in that case?

For me that is enough, I don't use the feature. But aparently some people want the bar gone completely. See here for some user opinions https://mastodon.social/@Tusky/109291155735828325

boehs:
I think the solution is just to show a little message that says you can access the sidebar by holding the left

Issue #2753, opened 2022-11-05
Use exoplayer for video playback


evant:
Was looking into addressing https://github.com/tuskyapp/Tusky/issues/2419 and I'm wondering if there'd be any desire to switch from `VideoView` to https://exoplayer.dev/. In particular it has a clean separation between the media state and the ui which makes handling configuration changes much easier. I find the api nicer to work with as well and it has more robust support for various media codecs which may or may not be useful. 

* * * *
- Tusky Version:
- Android Version:
- Android Device:
- Mastodon instance (if applicable):

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

connyduck:
The image viewer/media player has some of the worst code in all of Tusky and needs a complete rewrite. And yes, it should definitely run on [AndroidX media3](https://developer.android.com/jetpack/androidx/releases/media3) (which is the future of exoplayer)

nikclayton:
https://cunnin.me/@josh/109540231998431698 is an example user report where I suspect switching to Exoplayer would fix the problem.

nikclayton:
@mcclure is migrating from Exoplayer something you're going to tackle?

mcclure:
> @mcclure is migrating from Exoplayer something you're going to tackle?

Yes, I have started looking at it and plan to do this in the 22.0 cycle. If it proves too big for me I will report back in Matrix but it seems straightforward.

charlag:
fixed by #3857

PR #2754, opened 2022-11-06
Handle config changes for video playback

### LINKED ISSUE #2419: Video progress resets when the screen is rotated
When watching a video in fullscreen mode, if you rotate the phone (with autorotation mode on) the video restarts.

* * * *
- Tusky Version: 16.0
- Android Version: 10
- Android Device: Samsung

evant:
Fixes #2419

Note: I filled a suggestion here about [switching to exoplayer](https://github.com/tuskyapp/Tusky/issues/2753) which I think is still worth considering but this address the issue simply by handling config changes for `ViewMediaActivity`.

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/view/ExposedPlayPauseVideoView.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/view/ExposedPlayPauseVideoView.kt
@@ -12,6 +12,7 @@ class ExposedPlayPauseVideoView @JvmOverloads constructor(
     VideoView(context, attrs, defStyleAttr) {
 
     private var listener: PlayPauseListener? = null
+    private var playing = false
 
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/keylesspalace/tusky/view/ExposedPlayPauseVideoView.kt
```diff
@@ -12,19 +12,26 @@ class ExposedPlayPauseVideoView @JvmOverloads constructor(
...
+    private var playing = false
 
     fun setPlayPauseListener(listener: PlayPauseListener) {
         this.listener = listener
     }
 
     override fun start() {
         super.start()
-        listener?.onPlay()
+        if (!playing) {
```
evant, line 23: start is called again when the surface size changes. Added the check to not trigger a double-call to the listener.

thebino (approved):

PR #2755, opened 2022-11-06
Show toast if pin fails

### LINKED ISSUE #2229: Error handling for pinning toots
Mastodon has a limit of five toots that can be pinned simultaneously. When another toot is pinned, the web app pops up a toast saying "422 Validation failed: You have already pinned the maximum number of toots". On Tusky, the action fails silently. Simply doing the same thing as the web app would solve this.

* * * *
- Tusky Version: 15.1
- Android Version: 12 beta
- Android Device: Pixel 4A 5G

evant:
Fixes #2229

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/usecase/TimelineCases.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/usecase/TimelineCases.kt
@@ -131,10 +131,10 @@ class TimelineCases @Inject constructor(
     fun pin(statusId: String, pin: Boolean): Single<Status> {
         // Replace with extension method if we use RxKotlin
         return (if (pin) mastodonApi.pinStatus(statusId) else mastodonApi.unpinStatus(statusId))
-            .onErrorResumeNext(::convertError)
             .doOnError { e ->
-                Log.w("Failed to pin", e)
+                Log.w("Failed to change pin state", e)
             }
+            .onErrorResumeNext(::convertError)
             .doAfterSuccess {
                 eventHub.dispatch(PinEvent(statusId, pin))
             }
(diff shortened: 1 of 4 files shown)


thebino (approved):

connyduck:
Can we have a snackbar instead of a toast?

evant:
Done!

connyduck:
Nice! What do you think about a generic "Failed to pin" message when we don't get one from the server? Also the exception should be logged.

evant:
Makes sense. Added a generic error for cases where we don't have a server error.

app/src/main/java/com/keylesspalace/tusky/usecase/TimelineCases.kt
```diff
@@ -130,6 +131,10 @@ class TimelineCases @Inject constructor(
     fun pin(statusId: String, pin: Boolean): Single<Status> {
         // Replace with extension method if we use RxKotlin
         return (if (pin) mastodonApi.pinStatus(statusId) else mastodonApi.unpinStatus(statusId))
+            .onErrorResumeNext(::convertError)
+            .doOnError { e ->
+                Log.w("Failed to pin", e)
```
thebino, line 136: should the log reflect the action (pin/unpin)

connyduck: Not necessarily. How about "Failed to change pin state"

evant: Updated the message, also swapped it around to log the original error instead of the converted one since that's probably more useful.

PR #2757, opened 2022-11-06
Navigate to the Follow Requests page from notification

### LINKED ISSUE #2655: Pressing "<user> requested to follow you" android notification sends user to wrong UI
Pressing "<user> requested to follow you" android notification doesn't present the user with the "Follow Requests" UI. Instead, it presents it with the "Notifications" UI, but I'm not sure this is always the case, it could be another too.

This is really confusing from user point of view because there's nothing to do in the "Notifications" UI. Then the user needs to remember to go press the menu app button, and go to the "Follow Requests" UI, then remember the user which showed up in the notification and press the user in the list to finally be able to accept or not.

When the "<user> requested to follow you" android notification is pressed, the user shoul be presented either with:
A) The "Follow Requests" UI
B) The user's profile, where the "Accept Request" + "Follow" buttons should be available.

* * * *
- Tusky Version: 19.0
- Android Version: 11
- Android Device:

- [X ] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

evant:
Fixes #2655

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/components/notifications/NotificationHelper.java
+++ b/app/src/main/java/com/keylesspalace/tusky/components/notifications/NotificationHelper.java
@@ -84,6 +84,8 @@ public class NotificationHelper {
      */
     public static final String ACCOUNT_ID = "account_id";
 
+    public static final String TYPE = "type";
+
     private static final String TAG = "NotificationHelper";
 
     public static final String REPLY_ACTION = "REPLY_ACTION";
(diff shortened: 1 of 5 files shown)


app/src/main/java/com/keylesspalace/tusky/components/notifications/NotificationHelper.java
```diff
@@ -268,6 +270,7 @@ public static void make(final Context context, Notification body, AccountEntity
     private static NotificationCompat.Builder newNotification(Context context, Notification body, AccountEntity account, boolean summary) {
         Intent summaryResultIntent = new Intent(context, MainActivity.class);
         summaryResultIntent.putExtra(ACCOUNT_ID, account.getId());
+        summaryResultIntent.putExtra(TYPE, body.getType().name());
```
evant, line 273: There's a small risk of using `name` here if the enum name ever changes across versions. It looks like this particular class is excluded from proguard obfuscation so it should be safe but open to suggestions if there's a concern.

PR #2763, opened 2022-11-06
migrate `IOUtils` to Kotlin

wiryadev:


connyduck (changes requested):
How about making them extension functions `fun Closeable?.closeQuietly()` and `fun Uri.copyToFile(contentResolver: ContentResolver, file: File): Boolean`?

Also, please add the license header back in.

wiryadev:
> How about making them extension functions `fun Closeable?.closeQuietly()` and `fun Uri.copyToFile(contentResolver: ContentResolver, file: File): Boolean`?
> 
> Also, please add the license header back in.

converted to ext. function and headers added back

PR #2765, opened 2022-11-06
#2177 enhance 'show more' visibility

### LINKED ISSUE #2177: cosmetic: Make "LOAD MORE" button more visible
99% of the time when reading toots top to bottom (newer to older toots) in my timeline I am unable to realize I passed through the "LOAD MORE" button until a while after when I realize I am looking at toots far older and I have to scroll back up to find the button. That means, I pass the "LOAD MORE" button unnoticed and hence I don't press on it, meaning I loose content while for instance following news feeds.

It's really annoying because I have to really focus on actively looking for the "LOAD MORE" section while scrolling down and reading toots.

The main problem imho is that current UI is not making that "LOAD MORE" section colorful or shiny enough so that it's really easy to lose it while scrolling. Some ideas on how to improve it:
* Use bolder text, probably in red color?
* Make the frame around LOAD MORE larger, so that the empty gap in between news becomes larger and hence more difficult to pass by unnoticed.

* * * *
- Tusky Version: 15.0

- [ X ] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

thebino:
fixes #2177 

adjust text size and style to make the 'show more' placeholder more visible

### DIFF
--- a/app/src/main/res/layout/item_status_placeholder.xml
+++ b/app/src/main/res/layout/item_status_placeholder.xml
@@ -10,7 +10,8 @@
         android:layout_width="match_parent"
         android:layout_height="match_parent"
         android:text="@string/load_more_placeholder_text"
-        android:textSize="?attr/status_text_medium" />
+        android:textStyle="bold"
+        android:textSize="?attr/status_text_large" />
 
     <ProgressBar
         android:id="@+id/progressBar"


connyduck:
I'm not 100% sure about this, it does not look bad, but also not like other buttons we have. Merging to see what users say, we may have to adjust it again.

Issue #2072, opened 2021-02-08
Announcement Reactions limited to 8


connyduck:
There can only be 8 different reactions to 1 Announcement in Mastodon. We should hide the + button if that limit is reached like they do. Currently we show the button, but nothing happens after selecting an emoji.

PR #2766, opened 2022-11-06
#2072: hide add reaction when limit is reached

thebino:
fixes: #2072

hide/show add reaction chip based on the reaction count

[device-2022-11-06-183243.webm](https://user-images.githubusercontent.com/841681/200185881-ffe1cb1d-2832-47e5-a99b-14af62d9c362.webm)

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/components/announcements/AnnouncementAdapter.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/components/announcements/AnnouncementAdapter.kt
@@ -75,7 +75,7 @@ class AnnouncementAdapter(
         }
 
         // hide button if announcement badge limit is already reached
-        addReactionChip.visible(item.reactions.size >= 8)
+        addReactionChip.visible(item.reactions.size < 8)
 
         item.reactions.forEachIndexed { i, reaction ->
             (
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/keylesspalace/tusky/components/announcements/AnnouncementAdapter.kt
```diff
@@ -73,6 +73,13 @@ class AnnouncementAdapter(
             return
         }
 
+        // hide button if announcement badge limit is already reached
+        if (item.reactions.size >= 8) {
+            addReactionChip.visibility = View.GONE
+        } else {
+            addReactionChip.visibility = View.VISIBLE
+        }
```
connyduck, line 81: We have a neat extension function for this, you can write it as `addReactionChip.visible(item.reactions.size < 8)`

thebino: ✅ adjusted

connyduck: You got the condition the wrong way

thebino: sorry, stupid me :)

Issue #1133, opened 2019-03-20
[Feature Request] Notify user that uploaded media will be lost on saving to drafts


MasterGroosha:
A minor issue, but anyway:
1. Start composing a new toot.
2. Add some mediafiles (photo, video)
3. Press back and save toot as draft.

In this case, mediafiles will be lost (by the way, do they keep being stored on servers as loose objects?), and after restoring draft to post, you have to upload media again.

Even though I think this is pretty obvious, I also think that "Save draft" dialog box should also state "Note: you'll have to re-uploaded mediafiles" or "Note: all uploaded media files will be lost!".

* * * *
- Tusky Version: Nightly 5.2-463b008
- Android Version: 9.0
- Android Device: Oneplus 5

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

connyduck:
Saving drafts is a complete mess and needs to be rewritten from scratch 😓
> by the way, do they keep being stored on servers as loose objects?

server will clean them up after a while, no idea how long it takes

LukeLR:
I think, they are stored as loose objects on the instance server, since I seem to have hit my daily upload limit on chaos.social when debugging issue #1367 without having posted a single photo.

TheOtherDoctor:
So drafts are actually a server-side feature?!? This came as a surprise for me, as I am accessing the same instance account with different devices, but the drafts (as shown in tusky) are completely different ones on the devices (running tusky 10). When they are on the server, I expect them to be same...
* Is this what's meant be "loose" - not connected to my account, but somehow directly ID-referenced by the client session?

Having to worry that drafts might be deleted by the server at some random time would be a rather bad thing - if we knew the server timeout, it would be nice to give the user a notifier on that in the drafts list. 
* With an unclear delete threat in the back, maybe going for a (closer) integration with external text editor is the better way... There it is then also clear that media are not covered. ;-)

connyduck:
No drafts are completely client side. But when you compose a toot, the media is uploaded immediately. When you continue on a draft, it reuploads the media, which might not be necessary.

jsoref:
While it might make sense to rewrite drafts, for the time being, a simple warning would be a good start. Ideally it would cover the content in the pending item that will be lost (so if there's media, mention the media, I'd there's alt text, mention that as well).

[...]

PR #2784, opened 2022-11-08
Warn about losing media

jsoref:
Try to address #1133

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -420,6 +420,7 @@
     <string name="lock_account_label">Lock account</string>
     <string name="lock_account_label_description">Requires you to manually approve followers</string>
     <string name="compose_save_draft">Save draft?</string>
+    <string name="compose_save_draft_loses_media">Save draft? (Attachments will be uploaded again when you restore the draft.)</string>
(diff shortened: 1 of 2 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -420,6 +420,7 @@
     <string name="lock_account_label">Lock account</string>
     <string name="lock_account_label_description">Requires you to manually approve followers</string>
     <string name="compose_save_draft">Save draft?</string>
+    <string name="compose_save_draft_loses_media">Save draft? (You will lose all attachments and any captions)</string>
```
connyduck, line 423: Nothing will be lost. It will just be re-uploaded. 
How about "Attachments will be uploaded again when you restore the draft"

jsoref: Hmm, appears that I'm just not yet comfortable w/ the UI... I found the draft whose picture+caption I thought I lost (everything intact). Now I wonder what happened that I ended up elsewhere ... Perhaps I ended up in a new compose window?

Anyway, your text sounds fine.

jsoref:
Tested in 19.0-a9c6f69. It works and looks pretty good!

Issue #2785, opened 2022-11-08
Add a view for followed hashtags


connyduck:
There is an api for that `/api/v1/followed_tags`, but no frontend implements it yet as far as I know.
I think it should go into "Account Preferences" between "Tabs" and "Muted users" and just be a simple list where one can unfollow the hashtags again. It would be cool if we could do feature detection for this, but once most instances are on Mastodon 4.0 it shouldn't be a big problem if we don't-

Tak:
I can pick this up when I get a minute

stephan75:
Fedilab has this new  feature to handle followed hashtags that are tied to your Mastodon account at least since version 3.7.4 ...

But that list feature inside fedilab (at least in 3.8.0) has problems to display German umlaut characters like Ä Ö and Ü

Tak:
Implemented in #2794

PR #2794, opened 2022-11-09
Add view for browsing and unfollowing followed hashtags

Tak:
Implements #2785


connyduck:
Can you make slightly more advanced list items with an x button (like we have in "hidden domains")?
And instead of dialog before unfollow I'd make an "undo" snackbar (also like in "hidden domains").

Tak:
Updated

app/src/main/java/com/keylesspalace/tusky/FollowedTagsActivity.kt
```diff
@@ -0,0 +1,138 @@
...
+                        ),
+                        Snackbar.LENGTH_SHORT
+                    )
+                        .show()
+                }
+            )
+        }
+    }
+
+    private suspend fun refreshDisplay() {
```
connyduck, line 97: doesn't need to be suspending

connyduck (approved):
Will merge after Tusky 20 release

connyduck (changes requested):
Sorry to un-approve this again, but I just found out that this api is paginated, so we need to implement that as well or it won't work when a lot of hashtags are followed https://docs.joinmastodon.org/methods/followed_tags/#get

I'd to it the same as in other places with `RemoteMediator`/`PagingSource`

Tak:
Of course in 4.0.3, I'll rework it

connyduck:
Default page size is 100, how many users follow that many hashtags... I'd be ok with merging it without pagination for now as well.

app/src/main/java/com/keylesspalace/tusky/components/followedtags/FollowedTagsViewModel.kt
```diff
@@ -0,0 +1,33 @@
...
+import androidx.paging.cachedIn
+import com.keylesspalace.tusky.di.Injectable
+import com.keylesspalace.tusky.entity.HashTag
+import com.keylesspalace.tusky.network.MastodonApi
+import javax.inject.Inject
+
+class FollowedTagsViewModel @Inject constructor (
+    api: MastodonApi
+) : ViewModel(), Injectable {
+    val tags: MutableList<HashTag> = emptyList<HashTag>().toMutableList()
```
connyduck, line 17:
```suggestion
    val tags: MutableList<HashTag> = mutableListOf()
```

connyduck (commented):
It now crashes when opening the view while offline. I think you need to extend the try block in `FollowedTagsRemoteMediator`.

connyduck (approved):

PR #2800, opened 2022-11-10
Switch to an unambiguous date format in Absolute Time mode

BenDundon:
Noticed when scrolling notifs that the short date format is ambiguous as to whether it is DD/MM or MM/DD (figured it is the latter), but that's unintuitive to users where that's not the standard date format. This change uses MMM instead of MM to make it more clear (e.g. 10 Nov. vs. 11/10).

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/util/AbsoluteTimeFormatter.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/util/AbsoluteTimeFormatter.kt
@@ -23,7 +23,7 @@ import java.util.TimeZone
 
 class AbsoluteTimeFormatter @JvmOverloads constructor(private val tz: TimeZone = TimeZone.getDefault()) {
     private val sameDaySdf = SimpleDateFormat("HH:mm", Locale.getDefault()).apply { this.timeZone = tz }
-    private val sameYearSdf = SimpleDateFormat("MM-dd HH:mm", Locale.getDefault()).apply { this.timeZone = tz }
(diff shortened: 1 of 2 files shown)


connyduck:
Please adjust the tests accordingly

kyori19:
Sorry for interrupting you but we should consider about locales when we think about date and time formatting.

For example in Japanese, **date never comes before month** and **we don't use names of months**.
(Actually we have classical names of months but this won't be used in general.)

This change looks like this screenshot. This formatting is really weird in Japanese.
"8月" (month 8) stands for August and corresponding date format is "11日" (date 11).
Furthermore, correct order will be "8月11日" - date comes after month.

image

We may have to add an option to choose the format or respect OS settings.

connyduck:
Yes. How can that be done in Java?

kyori19:
I found `DateTimeFormatter.ofLocalized*` methods in Java API.

refs:
* https://www.baeldung.com/java-datetimeformatter
* https://zenn.dev/akuraru/articles/06410805bcf9d8 (Japanese article)

connyduck:
Those are not flexible enough, can't format month and day without year while respecting the locale.
I think we need something like this: https://unicode-org.github.io/icu/userguide/format_parse/datetime/index#datetimepatterngenerator

@BenDundon sorry if this gets out of hand, I'm fine with merging this for now and improving it later

BenDundon:
@connyduck I'd originally hoped to put a format selection in the settings page, but this is my first experience with Kotlin so I wasn't able to figure it out. I'll adjust the checks & keep at it to see if I can get it to do that, but if you'd like to merge now then I'm down, but I think it would be ironic for a fix that was meant to improve regional readability to break it in different regions.

PR #2814, opened 2022-11-12
Fix `markupHiddenUrls` destructs original CharSequence

kyori19:
It seems like URLs could be marked up repeatedly. See screenshot below.

# Steps to reproduce

1. Display a post with URL markup
2. Hide the post by scrolling up or down
3. Display the post again by scrolling
4. Repeat 2 and 3

# Cause of the bug

`SpannableBuilder.valueOf` does not create new instance when passed `CharSequence` is an instance of `SpannableBuilder`.
`markupHiddenUrls` is called every time `StatusBaseViewHolder.setupWithStatus` called, and updates status object.

# Screenshots

Example post from my misskey account: https://misskey.io/notes/97hkq07foc

before after


Tak (approved):
Nice find!

PR #2826, opened 2022-11-14
Add option for default reply privacy set to unlisted by default

### LINKED ISSUE #2798: Replies visible as public
### Steps  to reproduce:

Reply a publicly visible toot.

### What happens:

Visibility of the reply is set to "public"

### Expected result:

Visibility of the reply by default is set to "unlisted". User can change it.

* * * *
- Tusky Version: 19.0

evant:
Fixes #2798

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/components/compose/ComposeViewModel.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/components/compose/ComposeViewModel.kt
@@ -87,13 +87,18 @@ class ComposeViewModel @Inject constructor(
     val markMediaAsSensitive: MutableStateFlow<Boolean> =
         MutableStateFlow(accountManager.activeAccount?.defaultMediaSensitivity ?: false)
 
-    val statusVisibility: MutableStateFlow<Status.Visibility> = MutableStateFlow(Status.Visibility.UNKNOWN)
+    val statusVisibility: MutableStateFlow<Status.Visibility> =
+        MutableStateFlow(Status.Visibility.UNKNOWN)
     val showContentWarning: MutableStateFlow<Boolean> = MutableStateFlow(false)
     val poll: MutableStateFlow<NewPoll?> = MutableStateFlow(null)
     val scheduledAt: MutableStateFlow<String?> = MutableStateFlow(null)
 
     val media: MutableStateFlow<List<QueuedMedia>> = MutableStateFlow(emptyList())
-    val uploadError = MutableSharedFlow<Throwable>(replay = 0, extraBufferCapacity = 1, onBufferOverflow = BufferOverflow.DROP_OLDEST)
+    val uploadError = MutableSharedFlow<Throwable>(
+        replay = 0,
+        extraBufferCapacity = 1,
+        onBufferOverflow = BufferOverflow.DROP_OLDEST
+    )
 
     lateinit var composeKind: ComposeActivity.ComposeKind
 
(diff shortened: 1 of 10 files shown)


app/schemas/com.keylesspalace.tusky.db.AppDatabase/45.json
```diff
@@ -2,7 +2,7 @@
   "formatVersion": 1,
```
Tak, line 2: I guess these old .json should be removed?

app/src/main/java/com/keylesspalace/tusky/components/compose/ComposeViewModel.kt
```diff
@@ -416,15 +443,17 @@ class ComposeViewModel @Inject constructor(
             return
         }
 
-        val preferredVisibility = accountManager.activeAccount!!.defaultPostPrivacy
+        inReplyToId = composeOptions?.inReplyToId
+
+        val activeAccount = accountManager.activeAccount!!
+        val preferredVisibility =
+            if (inReplyToId != null) activeAccount.defaultReplyPrivacy else activeAccount.defaultPostPrivacy
```
Tak, line 450: I guess a multiline block is preferable here

Tak (commented):
Overall, this change looks reasonable to me, provided that this is a feature we want

app/src/main/java/com/keylesspalace/tusky/components/compose/ComposeViewModel.kt
```diff
@@ -89,13 +89,18 @@ class ComposeViewModel @Inject constructor(
     val markMediaAsSensitive: MutableStateFlow<Boolean> =
         MutableStateFlow(accountManager.activeAccount?.defaultMediaSensitivity ?: false)
 
-    val statusVisibility: MutableStateFlow<Status.Visibility> = MutableStateFlow(Status.Visibility.UNKNOWN)
+    val statusVisibility: MutableStateFlow<Status.Visibility> =
```
nikclayton, line 92: Many of the changes here appear to be formatting changes, not functionality changes. That makes it difficult to review the code that's actually making changes. @evant would you mind reverting the formatting changes on existing code, and leaving just the functional changes? Thanks.

app/src/main/java/com/keylesspalace/tusky/components/preference/AccountPreferencesFragment.kt
```diff
@@ -194,6 +194,28 @@ class AccountPreferencesFragment : PreferenceFragmentCompat(), Injectable {
...
 
+                val activeAccount = accountManager.activeAccount
+                if (activeAccount != null) {
+                    listPreference {
+                        setTitle(R.string.pref_default_reply_privacy)
+                        setEntries(R.array.post_privacy_names)
+                        setEntryValues(R.array.post_privacy_values)
+                        key = PrefKeys.DEFAULT_REPLY_PRIVACY
+                        setSummaryProvider { entry }
+                        val visibility = activeAccount.defaultReplyPrivacy
```
nikclayton, line 205: Something like this instead?

```suggestion
                        val visibility = accountManager.activeAccount?.defaultReplyPrivacy ?: accountManager.activeAccount?.defaultPostPrivacy ?: Status.Visibility.PUBLIC
```

That way you can remove the null check on line 198, and this will fallback through their default post privacy if this is not set, or public otherwise (consistent with how the other visibility preference is managed).

nikclayton: ?

evant: It sounded like we didn't want to implement this until it was implemented server-side so I haven't really been focusing on this pr

PR #2828, opened 2022-11-14
[WIP] Add post editing capability

Tak:
This doesn't seem to require a ton of additional work from delete/redraft.

Verified use cases:
- Edit post, send
- Edit post, send fails, resend from draft
- Edit post, save draft, resend from draft

We still don't have a solution for support detection (`status.edited_at` will be null both when the instance doesn't provide it *and* when the instance sends it as null (thanks, json!))

I'm also not sure what to do with the edited date: set it as the date for the post in the timeline? ignore it? something else?

Finally, I haven't thoroughly tested all the edge cases yet. I feel like the fact that we're leveraging the existing draft system etc. means that it should be pretty solid, but there are still a lot of different variables wrt attachments, polls, scheduling, etc., that need to be actually explored.


Tak:
Found two TODOs just now:
- It looks like we have to always reupload media like we do with delete and redraft
- We also need to fix up mentions (and presumably hashtags)

sl1txdvd:
Not a dev so apologies in advance if I state obvious things or ask stupid questions. Not sure if I am supposed to answer here as a user.

> We still don't have a solution for support detection 

Checking the Mastodon version per account (maybe at login) would give you that information and might be useful for future features too. Probably outside of the scope of this feature.

> I'm also not sure what to do with the edited date: set it as the date for the post in the timeline? ignore it? something else?

The edited post should stay at the original place/time in the timeline. But there needs to be a way to open the revisions. In the web client there is an arrow or something next to the date/time when you open an edited post. Thete you can open the revisions.

> there are still a lot of different variables wrt attachments, polls, scheduling, etc.,

Polls can be edited as a whole and start again, but it is not possible to edit the options of a running poll.

connyduck:
I think assuming the server supports it is fine. Most Mastodon instances are on 4.0.

> It looks like we have to always reupload media like we do with delete and redraft

Wait, with delete & redraft we don't reupload but reuse the media id, this should work here as well 🤔

connyduck:
> We also need to fix up mentions (and presumably hashtags)

There is an api where one can get the post source: `/api/v1/statuses/:id/source`

Tak:
Ok, mentions actually *do* work - the problem I saw earlier was either workflow-related, or maybe because I was testing while all instances were heavily loaded with new users

app/src/main/java/com/keylesspalace/tusky/service/SendStatusService.kt
```diff
@@ -138,6 +138,7 @@ class SendStatusService : Service(), Injectable {
                     statusToSend.mediaProcessed.forEachIndexed { index, processed ->
                         if (!processed) {
                             when (mastodonApi.getMedia(statusToSend.mediaIds[index]).code()) {
+                                404, // This can happen when editing posts with media that were uploaded a long time ago
```
Tak, line 141: @connyduck Does this make any sense to you?

connyduck: No, because that codepath should only be hit when reuploading 🤔

Tak:
Ok, I think this is ready for other people to try it.
Tested scenarios:
- Edit the text of a post (+/- media)
- Edit the text of a reply (+/- media)
- Add/remove content warning on a post (+/- media)
- Add/remove media from a post
- Change media sensitivity on a post
- Change language of a post

app/src/main/java/com/keylesspalace/tusky/entity/Status.kt
```diff
@@ -137,7 +137,7 @@ data class Status(
         )
     }
 
-    private fun getEditableText(): String {
+    fun getEditableText(): String {
```
connyduck, line 140: No, don't use this. This is only here to have a fallback for ancient Mastodon versions before deleting returned the status source. To get the status source without deleting a status, call `GET /api/v1/statuses/:id/source`.

connyduck (commented):
Works like a charm! I wonder if we should somehow indicate that one is currently in editing mode?

app/src/main/java/com/keylesspalace/tusky/fragment/SFragment.kt
```diff
@@ -0,0 +1,513 @@
...
+/* Note from Andrew on Jan. 22, 2017: This class is a design problem for me, so I left it with an
+ * awkward name. TimelineFragment and NotificationFragment have significant overlap but the nature
+ * of that is complicated by how they're coupled with Status and Notification and the corresponding
+ * adapters. I feel like the profile pages and thread viewer, which I haven't made yet, will also
+ * overlap functionality. So, I'm momentarily leaving it and hopefully working on those will clear
+ * up what needs to be where. */
+abstract class SFragment : Fragment(), Injectable {
+    protected abstract fun removeItem(position: Int)
+    protected abstract fun onReblog(reblog: Boolean, position: Int)
+    private var bottomSheetActivity: BottomSheetActivity? = null
```
connyduck, line 78:
```suggestion
    private lateinit var bottomSheetActivity: BottomSheetActivity
```

then you don't need any calls with ?

app/src/main/java/com/keylesspalace/tusky/fragment/SFragment.kt
```diff
@@ -0,0 +1,513 @@
...
+
+    protected fun reply(status: Status) {
+        val actionableStatus = status.actionableStatus
+        val account = actionableStatus.account
+        var loggedInUsername: String? = null
+        val activeAccount = accountManager.activeAccount
+        if (activeAccount != null) {
+            loggedInUsername = activeAccount.username
+        }
+        val mentionedUsernames: MutableSet<String?> = LinkedHashSet(
```
connyduck, line 128: `MutableSet<String>`, then you don't need the `map { it!! }` below

app/src/main/java/com/keylesspalace/tusky/fragment/SFragment.kt
```diff
@@ -0,0 +1,513 @@
...
+        val statusIsByCurrentUser = loggedInAccountId != null && loggedInAccountId == accountId
+        if (statusIsByCurrentUser) {
+            popup.inflate(R.menu.status_more_for_user)
+            val menu = popup.menu
+            when (status.visibility) {
+                Status.Visibility.PUBLIC, Status.Visibility.UNLISTED -> {
+                    menu.add(0, R.id.pin, 1, getString(if (status.isPinned()) R.string.unpin_action else R.string.pin_action))
+                }
+                Status.Visibility.PRIVATE -> {
+                    var reblogged = status.reblog?.reblogged ?: status.reblogged
```
connyduck, line 167:
```suggestion
                    val reblogged = status.reblog?.reblogged ?: status.reblogged
```

app/src/main/java/com/keylesspalace/tusky/fragment/SFragment.kt
```diff
@@ -0,0 +1,513 @@
...
+                    menu.findItem(R.id.status_reblog_private).isVisible = !reblogged
+                    menu.findItem(R.id.status_unreblog_private).isVisible = reblogged
+                }
+                else -> {
+                    // TODO: anything?
```
connyduck, line 172: I don't think so 🤔

app/src/main/java/com/keylesspalace/tusky/fragment/SFragment.kt
```diff
@@ -0,0 +1,513 @@
...
+    }
+
+    private fun showConfirmEditDialog(id: String, position: Int, status: Status) {
+        if (activity == null) {
+            return
+        }
+        AlertDialog.Builder(requireActivity())
+            .setMessage(R.string.dialog_redraft_post_warning)
+            .setPositiveButton(android.R.string.ok) { _: DialogInterface?, _: Int ->
+                timelineCases!!.delete(id)
```
connyduck, line 389:
```suggestion
                timelineCases.delete(id)
```

app/src/main/java/com/keylesspalace/tusky/fragment/SFragment.kt
```diff
@@ -0,0 +1,513 @@
...
+                    ) { error: Throwable? ->
+                        Log.w("SFragment", "error deleting status", error)
+                        Toast.makeText(context, R.string.error_generic, Toast.LENGTH_SHORT).show()
+                    }
+            }
+            .setNegativeButton(android.R.string.cancel, null)
+            .show()
+    }
+
+    private fun editStatus(id: String, position: Int, status: Status) {
```
connyduck, line 426: position is unused

app/src/main/java/com/keylesspalace/tusky/fragment/SFragment.kt
```diff
@@ -0,0 +1,513 @@
...
+                requireContext().openLink(attachment.url)
+            }
+        }
+    }
+
+    protected fun viewTag(tag: String) {
+        startActivity(newHashtagIntent(requireContext(), tag))
+    }
+
+    protected fun openReportPage(accountId: String, accountUsername: String, statusId: String) {
```
connyduck, line 357: can be private (showConfirmDeleteDialog as well)

connyduck (commented):
nice!

connyduck (approved):

PR #2829, opened 2022-11-14
support Android 13 per-app languages

connyduck:


app/src/main/java/com/keylesspalace/tusky/BaseActivity.java
```diff
@@ -92,11 +91,6 @@ protected void onCreate(@Nullable Bundle savedInstanceState) {
         requesters = new HashMap<>();
     }
 
-    @Override
-    protected void attachBaseContext(Context base) {
```
charlag, line 96: why do we not need it anymore?

app/src/main/res/xml/locales_config.xml
```diff
@@ -0,0 +1,50 @@
+<?xml version="1.0" encoding="utf-8"?>
+<locale-config xmlns:android="http://schemas.android.com/apk/res/android">
+    <locale android:name="en"/>
+    <locale android:name="ca"/>
+    <locale android:name="cs"/>
+    <locale android:name="cy"/>
+    <locale android:name="de"/>
+    <locale android:name="en-gb"/>
```
charlag, line 8: is it intended that it's `en-gb` here but `en-GB` in `donottranslate.xml`?

charlag (commented):
I know nothing about new api but the code looks reasonable, languages is also a major pain in pref refactoring I've been doing, not sure how it will work now either

app/src/main/java/com/keylesspalace/tusky/BaseActivity.java
connyduck: Because that was our custom hack, now that functionality is inside AndroidX and `AppCompatDelegate.setApplicationLocales` does everything for us

app/src/main/res/xml/locales_config.xml
connyduck: nope should be the same everywhere, thx

Tak (approved):
Makes sense to me? Did you verify that the post language still defaults correctly?

connyduck:
> Makes sense to me? Did you verify that the post language still defaults correctly?

No, will check

connyduck:
It did not, thx @Tak!

PR #2832, opened 2022-11-15
Add option to not crop image previews

### LINKED ISSUE #1995: No cropped images please
Good morning, please is possible to add an option to no cropped images?
98% of my feed images are cropped.
Please.
Thanks for understanding

evant:
Fixes #1995

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/view/MediaPreviewLayout.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/view/MediaPreviewLayout.kt
@@ -1,0 +1,210 @@
+package com.keylesspalace.tusky.view
+
+import android.content.Context
+import android.graphics.Canvas
+import android.util.AttributeSet
+import android.view.View
+import android.view.ViewGroup
+import android.widget.LinearLayout
+import com.keylesspalace.tusky.R
+import kotlin.math.roundToInt
+
+/**
+ * Lays out a set of [MediaPreviewImageView]s keeping their aspect ratios into account.
+ */
+class MediaPreviewLayout(context: Context, attrs: AttributeSet? = null) :
+    ViewGroup(context, attrs) {
+
+    private val spacing = context.resources.getDimensionPixelOffset(R.dimen.preview_image_spacing)
+
+    /**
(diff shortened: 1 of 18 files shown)


connyduck:
I don't like adding more settings and would prefer some kind of layout logic that works in all cases. You need to have some cutoff where you crop anyway: https://chaos.social/web/@ConnyDuck@mastodon.social/109355050185283869

So how about
- if it is one image
  - do not crop aspect ratios between 2:1 and 1:2
  - crop longer or wider ratios to 2:1 or 1:2 respectively
- if it is two images
  - if the first image is higher than wide, display them next to each other, adjusting to aspect ratio within certain bounds, cropping otherwise
  - if the first image is wider than high, display them above each other, adjusting to aspect ratio within certain bounds, cropping otherwise
- if it is three images
  - similar concept as two images, expect image 2+3 share the space that the second image has in the 2 image case
- if it is four images
  - do the same as we do now, maybe adjust to aspect ratio of the first image in certain bounds

evant:
> I don't like adding more settings

The reason I added it as a setting is that web has it 
image

In terms of smarter logic around this, I personally wouldn't want even taller posts to be cropped (think webcomics), your other suggestions seems fine to me though. Idk how others would feel.

evant:
Updated to use a custom layout with smarter layout rules for various image sizes and aspect ratios. May tweak the rules a bit further.

app/src/main/java/com/keylesspalace/tusky/view/MediaPreviewLayout.kt
```diff
@@ -0,0 +1,210 @@
...
+import android.view.View
+import android.view.ViewGroup
+import android.widget.LinearLayout
+import com.keylesspalace.tusky.R
+import kotlin.math.roundToInt
+
+/**
+ * Lays out a set of [MediaPreviewImageView]s keeping their aspect ratios into account.
+ */
+class MediaPreviewLayout(context: Context, attrs: AttributeSet? = null) :
```
charlag, line 15: :heart_eyes:

evant:
It should be bounded to 2:1

Lakoja:
Ok, you're right it seems to be 1:2.

(It looked much higher to me.)

Issue #2801, opened 2022-11-10
Storage permission issues on Android 13 after API 33 update


K900:
When running a build with https://github.com/tuskyapp/Tusky/pull/2719 on Android 13, I am unable to download or upload any files - instead of a permission prompt I immediately get the "permission required" snackbar. This seems to be due to the (severely under documented) scoped storage permission changes in Android 13. The stable version is built with target SDK 31, and thus hits some sort of backwards compatibility path, allowing it to work.

K900:
Possibly relevant change: https://developer.android.com/about/versions/13/behavior-changes-13#granular-media-permissions

PR #2837, opened 2022-11-15
fix media permissions on Android 10+

connyduck:
Turns out the scoped storage thing on Android 10+ allows us to just not ask for the permission and still read/write what we want.

closes https://github.com/tuskyapp/Tusky/issues/2801


connyduck:
I could do some kind of `requestPermissionOnPre10` with params for permission type and callbacks for granted/denied, but I think i prefer some duplication

charlag:
I am okay with duplication too, not like we will get to clean this up anytime soon anyway

PR #2843, opened 2022-11-16
show "now" instead of "in 0s" timestamps

kylegoetz:
Issue is that a very recently created submission will show "in 0s" as timestamp. Preference described as display "now" for some small range around 0. I decided on <500ms, so any time span that would round to 0 would render as "now" instead of a display of seconds.

It appears components use `TimestampUtils` for this reason.

Added unit tests to be sure, but didn't mock a context to separate locale implementation of EN "now" (vs DE `jetzt` or ES `ahora` etc.) so that's something to improve upon. Didn't look like Mockito-Kotlin was a dependency, and that's the only way I know how to mock Context.

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/util/TimestampUtils.java
+++ b/app/src/main/java/com/keylesspalace/tusky/util/TimestampUtils.java
@@ -21,7 +21,6 @@ import com.keylesspalace.tusky.R;
 
 public class TimestampUtils {
 
-    private static final long ZERO_SECOND_THRESHOLD_IN_MILLIS = 500;
     private static final long SECOND_IN_MILLIS = 1000;
     private static final long MINUTE_IN_MILLIS = SECOND_IN_MILLIS * 60;
     private static final long HOUR_IN_MILLIS = MINUTE_IN_MILLIS * 60;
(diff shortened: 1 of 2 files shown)


connyduck:
- still shows 0s for 501-999ms, I don't think we want that
- yes the test should work on all locales please
- please fix kotlin formatting so the CI check passes

kylegoetz:
@connyduck I have updated the code as follows:

- [x] `ktlintCheck` passes
- [x] 501–999ms are treated the same way as 1ms or -1ms in that they're rendered as `R.string.status_created_at_now`, which is "now" for the default
- [x] the unit test for the TimestampUtil should work for all locales since it mocks `getString` as opposed to just hard-coded testing equality with `"now"`
- [ ] I have Spanish, German, and Japanese versions of `R.string.status_created_at_now` as those are also languages I speak, and I compared with what Twitter uses for "now" to double check

connyduck (approved):

nikclayton:
@kylegoetz -- I'm doing a thread of improvements to Tusky in v21, and will mention this. I'll reference the PR, but if you'd like me to mention your Mastodon account as well, please can tell me what it is (here, or @nikclayon@mastodon.social). Cheers.

kylegoetz:
@sturmen it's kylegoetz@mstdn.social

PR #2850, opened 2022-11-18
Prioritize users' default locales in language list

### LINKED ISSUE #2844: Unordered post language selection
The language selection when creating a new post is completely unordered (system language German), so when I want to switch to English or another language (would be great if configurable), I have to search the complete list, to find it there... (see screenshot)

Screenshot_20221116-212526

* * * *
- Tusky Version: Tusky Test 19.0-9f7cd2fa
- Android Version: 13
- Android Device: Pixel 6a
- Mastodon instance (if applicable): phpc.social

Tak:
Closes #2844


connyduck:
That is definitely better than before! Should we add the currently selected app language as well in case it is not also a device preferred one?

Tak:
🤔 I did not consider that that could happen, but I suppose it makes sense

connyduck:
You can get it with `AppCompatDelegate.getApplicationLocales()` but you will need to check for duplicates in the other defaults

connyduck (approved):

PR #2862, opened 2022-11-19
fix buttons and usernames overlapping in follow requests

connyduck:
Before / After


Lakoja:
Observation: names on the left can all be read fully on the right only two.

connyduck:
True. One can click through to the full profile if it is a problem.

Lakoja:
As an idea (or maybe future further step): Shouldn't there be more info directly in this list to facilitate direct decisions.
If so the single entries would be much bigger and have room for a full ine address.

(However I'm not sure which additional information...)

charlag (approved):

connyduck:
The idea is not bad, consider openening an issue @Lakoja

Issue #2857, opened 2022-11-18
Sometimes content of tag timeline/tab shown instead of federated


Lakoja:
I cannot really realiably reproduce this but I can offer some indications.

I have one additional tab for a tag. So tab number 5.
Sometimes when selecting the tab for the public federated timeline (tab 4) it instead shows the contents for that tag but still tab 4 for is highlighted and the title is "Federated".

It happens rather reliably when debugging but not when starting it without debugger (or directly from phone).

When debugging the onCreate() of TimelineFragment is actually called multiple times. First correctly for "federated" and after that for "tag". (And I've seen even "local federated" tab 3 called once.)

What is very realiably happening in any case though: When selecting the fedarated timeline the log says that both content of the federated and the content of the tag is queried from the api.
This does not for example happend for the home timeline (only that content is loaded from the api).

* * * *
- Tusky Version: develop from 16th (https://github.com/tuskyapp/Tusky/commit/b718b80d909c9ce7424520c3848a2e39e19a56dc)

- [X] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

Lakoja:
Avoiding "smoothScroll" on the TabLayoutMediator gets rid of two problems here:

- showing wrong content
- requesting too many api content

Although this is probably meaningful it also masks the real problem here: why can (with several parallel? requests) data be shown in the wrong view fragment?

Also, at least for me another problem seems to become more prominent: sometimes when switching to another timeline only the "there is nothing animal" is shown (there is actually no request for content) only after a manual update something is shown.
(However that is probably a different issue and the fix here not the culprit.)

connyduck:
could this https://github.com/tuskyapp/Tusky/issues/1122 be related?

Lakoja:
> could this #1122 be related?

Not sure. But as that one is quite old... and it doesn't really look alike.

nikclayton:
I might be fixing this as part of fixing https://github.com/tuskyapp/Tusky/issues/3251. I've not been able to reproduce the problem though, so I can't confirm.

I'll be very interested to know if this recurs on any builds that have the https://github.com/tuskyapp/Tusky/issues/3251 fix present.

nikclayton:
https://github.com/tuskyapp/Tusky/pull/3255 is the PR with the fix.

PR #2864, opened 2022-11-19
2857 precise tabbing

Lakoja:
Masks the problem of one timeline (tagline) potentially replacing the contents of another.
https://github.com/tuskyapp/Tusky/issues/2857

And also avoids loading timeline content that is actually not displayed (i. e. needless network traffic).


connyduck:
This removes the scroll animation when clicking on a tab and the swipe animation looks very weird. Would need to disable swiping entirely to make it work and I don't think we want this.

PR #2866, opened 2022-11-19
Replace the default federated timeline tab with the direct messages tab

### LINKED ISSUE #2552: Replace federated timeline with direct messages as default

Tak:
Closes #2552


charlag (approved):

fbievan:
This entirely removes the federated tab with no way to get it back

connyduck:
No it does not affect already logged-in accounts, and you can change the tab setup in Account Preferences -> Tabs @fbievan

PR #2872, opened 2022-11-20
Add a touch delegate to increase action touch targets to 48dp

### LINKED ISSUE #2825: Post interaction buttons could use larger tap targets
I think the tap targets for the buttons to interact with a post (fav, boost, etc.) are too small. I find myself accidentally tapping to expand the post often when I try to use these. It's a common practice in mobile apps to increase the invisible "tap target" or "hit slop" to make buttons more usable. c.f. [Android accessibility guidelines](https://support.google.com/accessibility/android/answer/7101858?hl=en#:~:text=The%20recommended%20target%20size%20for%20touchscreen%20objects%20is%207%2D10mm.).

Please let me know if you would be open to a PR for this issue.

* * * *
- Tusky Version: 19.0
- Android Version: 13
- Android Device: Google Pixel 4a

evant:
Fixes #2825


app/src/main/java/com/keylesspalace/tusky/util/TouchDelegateHelper.kt
```diff
@@ -0,0 +1,61 @@
...
+import android.view.ViewGroup
+import androidx.core.view.doOnLayout
+import at.connyduck.sparkbutton.helpers.Utils
+
+/**
+ * Ensures the given children of the view will have a touch area of at least the given size, using
+ * a [TouchDelegate].
+ */
+@JvmOverloads
+fun ViewGroup.ensureMinTouchSize(childIds: List<Int>, minSize: Int = Utils.dpToPx(context, 48)) {
```
charlag, line 18: we would you pass list of ids instead of the views themselves? We could avoid expensive `findViewById` again

evant: It was an attempt to prevent you passing non-child views but looking back at it not a particularly good one. These days findViewById doesn't really cause a performance problem as if your view hierarchy is deep enough, it'll cause layout slowdowns long before searching gets slow. Don't mind changing it though.

Tak:
Closes #2932

evant:
Note: I'm seeing some cases where CW are used in the touch area becomes wildly off. Suspect it's due to another layout happening. Will try changing doOnLayout to a permanent layout listener when I get a chance.

connyduck:
Why can't we just give the icons more padding or remove clickable area from between buttons or something like that?

evant:
Was avoiding changing the design. Can certainly try that. It'll take up a bit more vertical space if we are ok with that.

connyduck:
> Was avoiding changing the design. Can certainly try that. It'll take up a bit more vertical space if we are ok with that.

Ok, I understand.

I just tried this out but I'm not sure if I can really notice a difference 🤔 
For me the biggest problem is that clicking between buttons opens the post. It should just do nothing there.

evant:
Think expanding the width to fill available space would make sense.

evant:
Updated the layout to adjust the size of the buttons
before:
Screenshot from 2022-12-04 23-16-09
after:
Screenshot from 2022-12-04 23-12-25

Note: kept the touch delegate around so that tapping in between the buttons touches the closest button. Couldn't figure out how to make layout work otherwise, but open to ideas here.

connyduck:
This is a significant improvement, thank you! Should we get rid of the 4dp top/bottom margin on the button row in order to save some space?

evant:
Done

connyduck (approved):

Issue #2875, opened 2022-11-20
Parts of content warning posts still shown


Lakoja:
Regarding profile setting "Always expand posts marked with content warnings":

Even with this setting disabled some contents are shown: all user handles in the text and the images.

* * * *
- Tusky Version: 20.beta.1

- [X] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

charlag:
It works like expected, mimicking web UI. Handles are shown to see that this is a conversation. Images in CWd posts are marked as NSFW my mastodon. Only text is hidden.

GPHemsley:
> It works like expected, mimicking web UI. Handles are shown to see that this is a conversation. Images in CWd posts are marked as NSFW my mastodon. Only text is hidden.

Expected by whom? I don't use the web UI, I only use Tusky. I frequently see posts clearly marked with a content warning, yet the images or videos are still shown without user interaction, despite my settings. I have also seen other apps where this is not the case.

PR #2877, opened 2022-11-20
2875 do not show critical content

Lakoja:
Fixes https://github.com/tuskyapp/Tusky/issues/2875


app/src/main/java/com/keylesspalace/tusky/adapter/StatusBaseViewHolder.java
```diff
@@ -226,8 +226,10 @@ protected void setSpoilerAndContent(boolean expanded,
                 setContentWarningButtonText(!expanded);
 
                 this.setTextVisible(sensitive, !expanded, content, mentions, tags, emojis, poll, statusDisplayOptions, listener);
+                mediaContainer.setVisibility(!expanded ? View.VISIBLE : View.GONE);
```
Lakoja, line 229: Not quite sure though why this (and only this) would need a special handling here.

connyduck:
As @charlag noted in the issue, the current behavior is correct.

Issue #2833, opened 2022-11-15
Swipe down to close video does not work


nikclayton:
Open an image from a post and you can swipe down to dismiss it and return to the post.

E.g, https://fedi.simonwillison.net/@simon/109346111039046677

Open a video from a post, and swipe down does nothing.

E.g., https://mastodon.social/@alexkehayias/109345818257251667

* * * *
- Tusky Version: 19.0
- Android Version: 13
- Android Device: Pixel 4a
- Mastodon instance (if applicable):

- [X] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

PR #2879, opened 2022-11-20
Support a swipe down to dismiss video

nikclayton:
Images can already be dismissed with a swipe, this adds the same functionality to videos.

- Add a VideoActionsListener interface for the hosting activity to dismiss the fragment
- Add a gesture listener for swipes
- Dismiss the fragment if a swipe has a greater Y component than X
- Provide UI feedback by scaling the video viewer as the user swipes

Fixes https://github.com/tuskyapp/Tusky/issues/2833


connyduck:
Very good idea, but this code makes no difference for me though :/

nikclayton:
Huh. Have you got a link to a status with media it doesn't work on? Works for me on the emulator and Pixel 4a.

babogoos:
I test your version, and it is work for me.
But `SimpleOnGestureListener.onFling()` only invoke when the touch track is quite vertical.
Maybe you can use `SimpleOnGestureListener.onScroll()` and calculate the slope ratio of touch track with `distanceX` and `distanceY`.

nikclayton:
You could, but then the calculation would be different from the swipe-down-to-dismiss calculation in ViewImageFragment, and I imagine that both gestures should be identical.

connyduck:
> Huh. Have you got a link to a status with media it doesn't work on? Works for me on the emulator and Pixel 4a.

I tried again and I think I figured it out, it only works when starting the swipe on the black background, and I was trying on a vertical video that filled the whole frame.

nikclayton:
> > Huh. Have you got a link to a status with media it doesn't work on? Works for me on the emulator and Pixel 4a.
> 
> I tried again and I think I figured it out, it only works when starting the swipe on the black background, and I was trying on a vertical video that filled the whole frame.

I can't reproduce that.

I just tried with the full screen video in https://aus.social/@Godfrey642/109449267279590400, playing full screen, and swiping up/down (where the swipe stays entirely within the video, and never touches any black background) dismisses it as expected.

PR #2882, opened 2022-11-20
Add a menu option to mute / filter a hashtag from a status list

nikclayton:
Un/muting uses the "home" filter


connyduck:
Good idea, but it always shows that a hashtag is not muted initially even if I open the timeline of a muted hashtag.

connyduck:
- from a public timeline (it is only muted on home after all)
- via search

nikclayton:
> * from a public timeline (it is only muted on home after all)
> * via search

Thanks. Fixed.

I'm not sold on the icons though. Technically they are the ones for mute / unmute, but I'm not sure they work in context.

connyduck:
I think the icons are fine. 

One thing that is now very weird and irritating: The two icons (mute and follow) appear at different times. Can you do it so the both appear at the same time?

nikclayton:
> I think the icons are fine.
> 
> One thing that is now very weird and irritating: The two icons (mute and follow) appear at different times. Can you do it so the both appear at the same time?

Done -- the icon appears immediately, but is initially disabled, then adjusts when the API call completes.

connyduck:
Nice!

Issue #2874, opened 2022-11-20
[Feature Request] Disable swipe gesture between profile tabs


ivadham:
The “`Enable swipe gesture to switch between tabs`” setting only affects main screen tabs. It might be a good idea to apply the same setting to profile tabs: POSTS | WITH REPLIES | PINNED | MEDIA

* * * *
- Tusky Version: 20.0 beta 1
- Android Version: 10

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

nikclayton:
Also Search (swipe between Posts/Accounts/Hashtags).

PR #2889, opened 2022-11-21
Disable tab swiping by preference in profile and search

nikclayton:
The "Enable swipe gesture to switch between tabs" preference was ignored on the tabs on a profile page ("Posts", "With Replies", "Pinned", "Media"), and search ("Posts", "Accounts", "Hashtags").

Fix this.

While I'm here, replace a string for the preference name in MainActivity.kt with a constant.

Fixes https://github.com/tuskyapp/Tusky/issues/2874.


Tak (approved):
Makes sense to me

Issue #2712, opened 2022-10-02
New handle display feature shows cut-off text on my phone


mcclure:
PR #2697 introduces text displaying the username when you are logged in with more than one account. On my phone, the @ and any descenders get cut off. This happens only in the main post view not in the compose view. From the screenshots in #2697, this did not happen on the submitter's phone.

I attempted to fix this myself without success. [Stack Overflow](https://stackoverflow.com/questions/35451630/toolbar-subtitle-is-cut-it-does-not-fit-in-height) has several questions about a problem like this, many of which suggest it is a phone-specific thing. They suggest several things that do not work:
* Set `android:minHeight="?attr/actionBarSize"` on the toolbar (no effect, it is already that height on my phone)
* Add bottom padding (it adds padding but this just gives cut-off text with a big blank space under this)
* Remove `android:fitsSystemWindows="true"` (this sounds scary and also it does nothing)

* * * *
- Tusky Version: Self built [91d1899](https://github.com/tuskyapp/Tusky/commit/91d18998acac2116010d191922a8d61455243cf7)
- Android Version: Android 12 (61.1.A.11.36 release-keys, August 1 2022 Sony update)
- Android Device: Sony XPeria 5 III
- Mastodon instance (if applicable): mastodon dot social

mcclure:
![Screenshot_20221002-161604](https://user-images.githubusercontent.com/277318/193475275-2838ae9b-1d64-427c-b0df-057ce78e796e.png)
![Screenshot_20221002-162626](https://user-images.githubusercontent.com/277318/193475278-6374190d-9512-47a6-a34a-6bca5558db59.png)

mcclure:
Fixed now

PR #2909, opened 2022-11-22
Use the appcompat style for Toolbar. Fixes subtitle vertical cropping.

Gryzor:
### Objective
* Fix the toolbar subtitle appearing slightly cropped (see #2712).

### Description
* It's an AppCompat.Toolbar, so applying the default theme appears to do the trick.

This is how it looks after the change:

image


connyduck:
I can't reproduce it on any of my devices but seems fine. 

Please reformat the file (style should go after android:id)

Issue #2890, opened 2022-11-21
[Feature request] Highlight image descriptions; esp. when missing composing


Lakoja:
Followup (or refinement) of https://github.com/tuskyapp/Tusky/issues/1049

Missing media descriptions are a nuisance for some and could help a lot of users (I like to see them as well).
One problem with this is that people are not aware that image descriptions are helpful and encouraged. Or that they even exist..

Therefore I'd like to make them a bit more prominent:

- When viewing a status with images mark the ones with description; i. e. with a small "T" in a corner
- When composing add a small warning icon in the top bar as long as there is media without description. Tapping that icon will show a short dialog or toast message.

This is in the range of light nudging and not so obtrusive as a dedicated warning dialog when sending.

- [X] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

charlag:
Some users asked about a warning for image descriptions when composing. So far we have two things:

 * when composing the icon in the corner of the image will switch between black and blue
 * when reading a post one can long-press to show the description as a toash

but we could improve on that

connyduck:
I like the idea of having an option that prevents posting without descriptions when enabled.
Also an "ALT" icon in the corner of images when they have descriptions (like Twitter does it) would be cool.

Lakoja:
Is merged.

PR #2919, opened 2022-11-23
2890 show warning on missing description

Lakoja:
Feature step to show a warning icon on missing image descriptions for 
https://github.com/tuskyapp/Tusky/issues/2890

This is feature-wise complete but regarding visuals it is more of a request for comments:

- How to display that warning dialog when tapping the warning icon? For the buttons at the bottom of the compose view a "BottomSheetBehavior" is used.
- Is the warning icon well placed on the top?
- What color to use? It should be a warning-like color. Currently "textColorHighlight" is used which is the "brightest" I could detect.
- The icon itself is copied from the content warning button. How are those icons normally created (take from where)?


connyduck:
Works & is definitely better than before.

- It needs a different icon though, the one you used is the same used for "Content warning", that is too confusing.
- I'd use tusky_red as the color
- Snackbar is better than Toastbecause it stays local within the Activity.

Lakoja:
Icon changed (new one "missing comment").

There is now only one problem remaining: I cannot position it sensibly (higher than the bottom) and it is shown **below** the toolbar at the bottom.

app/src/main/java/com/keylesspalace/tusky/components/compose/ComposeActivity.kt
```diff
@@ -498,6 +498,11 @@ class ComposeActivity :
         binding.composeScheduleView.setListener(this)
         binding.atButton.setOnClickListener { atButtonClicked() }
         binding.hashButton.setOnClickListener { hashButtonClicked() }
+        binding.descriptionMissingWarningButton.setOnClickListener {
+            Snackbar.make(binding.composeMainScrollView, R.string.hint_media_description_missing, Snackbar.LENGTH_LONG)
+                .setAction(R.string.action_retry) {  }
```
Lakoja, line 503: It is not shown without an action??

connyduck:
Check the other snackbars in ComposeActivity. You need to set some additional things so they are shown correctly.

Lakoja:
The problem with the snackbar being under the linear layout (toolbar) at the bottom is, that that has an "elevation" (12) that is higher than that of a snackbar (6).

It does work to add another CoordinatorLayout as wrapper around the main scroll part so that the snackbar is shown above the toolbar. However that feels like a hack/workaround. Also it would be the only location the app where it is done this way.

The second solution would be to use different elevation (i. e. 5 or none) where currently 12 is used. That elevation is rather old (2018) so I'm not really sure what it is or was for.

As a third alternative I reverted to using a Toast. This does not have this elevation problem and is also used for other explanation messages. For example error_no_custom_emojis.

Lakoja:
Ah found it now: displayTransientError() which explicitely sets an elevation...

Ok, I used that one, refactored it a bit and used it also for the emoji error.

I can also adapt the remaining two locations in ComposeActivity which still use Toast.

connyduck (approved):

PR #2942, opened 2022-11-27
issue 2890: Add an "ALT" sticker to the media preview container

Lakoja:
Closes second half of https://github.com/tuskyapp/Tusky/issues/2890


app/src/main/res/layout/item_media_preview.xml
```diff
@@ -110,6 +110,25 @@
             app:srcCompat="@drawable/ic_eye_24dp"
             app:tint="@color/white" />
 
+        <TextView
```
Lakoja, line 113: I see that other elements here seem to be duplicated in the report view.

I'm not sure if that is desired also for this element?

app/src/main/java/com/keylesspalace/tusky/adapter/StatusBaseViewHolder.java
```diff
@@ -503,6 +507,8 @@ protected void setMediaPreviews(final List<Attachment> attachments, boolean sens
             setAttachmentClickListener(imageView, listener, i, attachment, true);
         }
 
+        mediaDescriptionIndicator.setVisibility(withDescription ? View.VISIBLE : View.GONE);
```
Lakoja, line 510: This only shows a single "ALT", even if there are multiple images.

Although that is consistent with the sensitive media button.
And placing 4 of those stickers on the 4 preview images would be way too much clutter.

(Furthermore this is a bit "trusting" as one single character description is sufficient for it to show up, but I like that.)

app/src/main/res/layout/item_media_preview.xml
```diff
@@ -110,6 +110,25 @@
...
+            android:background="@drawable/media_warning_bg"
+            android:contentDescription="@null"
+            android:paddingLeft="5dp"
+            android:paddingTop="3dp"
+            android:paddingRight="5dp"
+            android:paddingBottom="3dp"
+            android:layout_margin="3dp"
+            android:visibility="gone"
+            android:text="ALT"
+            app:layout_constraintRight_toRightOf="@+id/status_media_preview_container"
```
Lakoja, line 128: Is RTL (reading direction) an issue here?

app/src/main/res/layout/item_media_preview.xml
charlag: I don't think it's that important in the report

app/src/main/res/layout/item_media_preview.xml
charlag: Probably best to do `end` here

app/src/main/java/com/keylesspalace/tusky/adapter/StatusBaseViewHolder.java
charlag: Twitter does it for each media and I think it makes sense?

app/src/main/res/layout/item_media_preview.xml
Lakoja: Ok, also did it accordingly ("start") for status_sensitive_media_button.

app/src/main/java/com/keylesspalace/tusky/adapter/StatusBaseViewHolder.java
Lakoja: 0517e090ac3aed98

Currently it looks like this here.

(It is also somewhat easier to understand in the layout code and code with only one item.)

connyduck:
I also prefer the sticker on every preview.

And super small design nitpick: I think it would look even better with smaller radius so it is parallel to the outline of the media previews.

Lakoja:
Unfortunately at the moment I'm a bit lost how more than one label would fit with the new MediaPreviewLayout.
The MediaPreviewLayout hides now what (esp. where) happens with each preview image.

So there must be some "overlayed extension" of the MediaPreviewImageView which draws the text on top of the respective image. Should /could that be a new view/fragment or the image view class modified somehow?

@evant

evant:
Would recommend either extending `MediaPreviewImageView` to support this or replace them in `MediaPreviewLayout` with a viewgroup that includes `MediaPreviewImageView` and whatever other views need to be added.

app/src/main/res/layout/item_image_preview_overlay.xml
```diff
@@ -0,0 +1,36 @@
+<?xml version="1.0" encoding="utf-8"?>
+<androidx.constraintlayout.widget.ConstraintLayout
```
evant, line 2: We could get away with a FrameLayout here right? Just thing the performance might be a bit better

app/src/main/java/com/keylesspalace/tusky/view/MediaPreviewLayout.kt
```diff
@@ -181,9 +185,13 @@ class MediaPreviewLayout(context: Context, attrs: AttributeSet? = null) :
         }
     }
 
-    inline fun forEachIndexed(action: (Int, MediaPreviewImageView) -> Unit) {
+    inline fun forEachIndexed(action: (Int, MediaPreviewImageView, TextView) -> Unit) {
         for (index in 0 until childCount) {
-            action(index, getChildAt(index) as MediaPreviewImageView)
+            action(
+                index,
+                getChildAt(index).findViewById(R.id.preview_image_view) as MediaPreviewImageView,
```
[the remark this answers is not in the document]
evant: Yeah I wrote this `forEachIndexed` purely for convenience sake. I do worry it has a bit too much knowledge of the layout's details and maybe it would be better to do the looping/indexing outside of this class instead? I think it's only called from one place anyway.

app/src/main/res/layout/item_image_preview_overlay.xml
Lakoja: The placement tags (ie layout_constraintEnd_toEndOf) at least require a ConstraintLayout.

Lakoja:
As for the `forEachIndexed` the part in the Layout class is probably alright.

It enables however much manipulation of class knowledge elsewhere.

Do you have an idea on how to improve it?

app/src/main/java/com/keylesspalace/tusky/view/MediaPreviewLayout.kt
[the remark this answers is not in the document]
Lakoja: The part in the Layout class is probably alright.

It enables however much manipulation of class knowledge elsewhere.

Do you have an idea on how to improve it?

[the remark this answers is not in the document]
Lakoja: Have been thinking about this a bit and it seems to me that `setMediaPreviews` (and `loadImage` and `sensitiveMediaWarning`) should move fully to `MediaPreviewLayout`?

Then only the current data from the current StatusBaseViewHolder.setMediaPreviews must be handed over and this is ok (data models and listeners).

But that would be a medium move (I failed on a ten minute attempt just now).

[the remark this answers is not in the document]
evant: Depends on what we think should be the responsibility of this class. I was originally going to have it encompass most of the work for displaying previews, but I realized that would be a significant change to the existing code so I moved it in the opposite direction, only focusing on layout and nothing more, which in hindsight I don't think is a bad thing because it separates the application logic from the reasonably complex layout code. This could actually be even more generic by passing the the layout id for it's children so it would have no knowledge of what the child views actually contain, but not convinced that level of abstraction is really necessary so eh, what you are currently doing seems fine to me.

> Do you have an idea on how to improve it?

You could try using viewbinding here, ex: hold an array of `ItemImagePreviewOverlayBinding`, would save you some findViewById calls.

app/src/main/java/com/keylesspalace/tusky/adapter/StatusBaseViewHolder.java
```diff
@@ -460,16 +460,16 @@ protected void setMediaPreviews(final List<Attachment> attachments, boolean sens
...
-        mediaPreview.forEachIndexed((i, imageView) -> {
+        mediaPreview.forEachIndexed((i, imageView, descriptionIndicator) -> {
             Attachment attachment = attachments.get(i);
             String previewUrl = attachment.getPreviewUrl();
             String description = attachment.getDescription();
 
             if (TextUtils.isEmpty(description)) {
-                imageView.setContentDescription(imageView.getContext()
-                        .getString(R.string.action_view_media));
+                imageView.setContentDescription(imageView.getContext().getString(R.string.action_view_media));
```
connyduck, line 469: Views get recycled, `descriptionIndicator` needs to be hidden here or after a while of scrolling they will just be shown everywhere

Lakoja: Done

connyduck:
Looks good to me, let's merge?

Tak (approved):
Looks ok to me, seems to work fine

Tak:
🤔 Either this one or something in recent `develop` is causing previews not to get updated for me when untoggling hidden status

connyduck:
> 🤔 Either this one or something in recent develop is causing previews not to get updated for me when untoggling hidden status

Can confirm, and it must be this PR, I can't reproduce it on develop. Something causes the `doOnLayout` in StatusBaseViewHolder.loadImage to not be called anymore, and then the images are not updated 🤔 

Why do we need the `doOnLayout`? It works if I remove it and call the image loading code directly, but then probably something else breaks?

Also, a ViewBinding should definitely be used in place of `findViewById` (I can take that over if you want @Lakoja)

app/src/main/res/layout/item_image_preview_overlay.xml
```diff
@@ -0,0 +1,36 @@
...
+        android:textSize="?attr/status_text_small"
+        android:background="@drawable/media_warning_bg"
+        android:contentDescription="@null"
+        android:paddingLeft="5dp"
+        android:paddingTop="3dp"
+        android:paddingRight="5dp"
+        android:paddingBottom="3dp"
+        android:layout_margin="3dp"
+        android:visibility="gone"
+        android:text="ALT"
```
connyduck, line 31: This should probably be translateable

connyduck:
The alt-labels should be hidden as well or we get overlapping texts.

evant:
@connyduck the doOnLayout was to fix an issue where glide would load the image before layout leading it resize to the wrong size and look low quality. Maybe the changed layout is affecting when glide loading vs the layout pass is happening?

Lakoja:
Have been trying to understand how Glide (and placeholder) and "onContentHiddenChange" works. Where does that end?

Because definitely the information for hiding or showing the image content doesn't reach anymore where it's supposed to go.

@connyduck can you point me to an example how view bindings might be used here?
Because the wrapping layout around image and alt text is so-to-say anonymous. There is no view class for that yet.

Maybe this would also point a bit to what has to be done for the hidden status.

app/src/main/res/layout/item_image_preview_overlay.xml
Lakoja: Done

Lakoja:
> The alt-labels should be hidden as well or we get overlapping texts.

Done

connyduck:
ok I tried it with view binding, and while it works it is also not really nice. Let's leave it like it is? The hide/unhide not working bug is definitely more important.

For the record, here is how to do it:

```
    private val bindingCache = Array(4) {
        ItemImagePreviewOverlayBinding.inflate(LayoutInflater.from(context))
    }
```


```
    private fun attachImageViews() {
        removeAllViews()
        for (i in 0 until aspectRatios.size.coerceAtMost(bindingCache.size)) {
            addView(bindingCache[i].root)
        }
    }
```
etc...
```

    fun forEachIndexed(action: (Int, MediaPreviewImageView, TextView) -> Unit) {
        for (index in 0 until childCount) {
            action(
                index,
                bindingCache[index].previewImageView,
                bindingCache[index].previewMediaDescriptionIndicator
            )
        }
    }
```
forEachIndexed can not be inline as long as the bindingCache is private

connyduck:
Regarding the bug, the solution is to call `doOnLayout` on the wrapping `ConstraintLayout` and not on the `MediaPreviewImageView` because with this change it is the wrapper that changes layout and not the ImageView.

Lakoja:
> Regarding the bug, the solution is to call `doOnLayout` on the wrapping `ConstraintLayout` and not on the `MediaPreviewImageView` because with this change it is the wrapper that changes layout and not the ImageView.

Did that.

Although, I have no real idea what is happening there. :-)

It seems to work now. Albeit it seems to be a bit delayed when switching the visibility - at least while debugging.

Issue #2893, opened 2022-11-21
Missing support for GoToSocial and microblog.pub links in looksLikeMastodonUrl


Razem:
Tusky currently supports automated lookup of posts (i.e., URLs inside other posts) from Mastodon, Pleroma, Friendica, Misskey, and Pixelfed, but it fails when it comes to [GoToSocial](https://gotosocial.org/) or [microblog.pub](https://docs.microblog.pub/). I have absolutely no experience with Kotlin or the workflow here, so I'm not really comfortable with creating a pull request myself. But if someone is willing to do it, I believe the correct regexes are as follows:

* `"^/@[^/]+/statuses/[0-9ABCDEFGHJKMNPQRSTVWXYZabcdefghjkmnpqrstvwxyz]{26}$"` for GoToSocial posts
  * It uses ULID, so it's guaranteed to be 26 symbols in Crockford's Base 32. And it seems to be case-insensitive.
  * See https://github.com/superseriousbusiness/gotosocial/blob/25edd57eaf3eecf0b449a4dabb5b5fc5d6ae7687/internal/id/ulid.go#L35 and https://github.com/oklog/ulid
* `"^/o/[0-9a-f]{32}$"` for microblog.pub posts
  * It uses UUID, so it's guaranteed to be 32 hex symbols. It is case-sensitive (lowercase).
  * See https://git.sr.ht/~tsileo/microblog.pub/tree/v2/item/app/boxes.py#L51
* Optionally `"^/articles/[0-9a-f]{7}/[-\\w]+$"` for microblog.pub articles
  * Articles use a combination of the short id (currently first 7 characters of the actual id) and a slug.
  * Maybe something more general like `"^/articles/[0-9a-f]+/[-\\w]+$"` should be used in case they decide to change the length of the short id.
  * This is not necessary as it would complicate getting to the actual article. Users would be forced to use the "Open in browser" button to see the whole content. (Mastodon shows only the name of the article.)
  * See https://git.sr.ht/~tsileo/microblog.pub/tree/v2/item/app/boxes.py#L658 and https://git.sr.ht/~tsileo/microblog.pub/tree/v2/item/app/utils/text.py#L5

It's supposed to go in Tusky/app/src/main/java/com/keylesspalace/tusky/BottomSheetActivity.kt.

* * * *
- Tusky Version: 19.0
- Android Version: 11
- Android Device: FP4
- Mastodon instance (if applicable):

[...]

PR #2945, opened 2022-11-28
Load goto social and microblog.pub urls in the app

Tak:
- goto social status urls
- goto social profile urls
- microblog.pub status urls

(microblog.pub profile urls are just site urls, so 🤷 )

I also moved the utility function into `LinkHelper`, it felt weird to keep it in `BottomSheetActivity`

 #2893

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/util/LinkHelper.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/util/LinkHelper.kt
@@ -284,6 +284,8 @@ private fun openLinkInCustomTab(uri: Uri, context: Context) {
 // https://pixelfed.social/p/connyduck/391263492998670833
 // https://pixelfed.social/connyduck
 // https://gts.foo.bar/@goblin/statuses/01GH9XANCJ0TA8Y95VE9H3Y0Q2
+// https://gts.foo.bar/@goblin
+// https://foo.microblog.pub/o/5b64045effd24f48a27d7059f6cb38f5
 fun looksLikeMastodonUrl(urlString: String): Boolean {
     val uri: URI
     try {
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/keylesspalace/tusky/util/LinkHelper.kt
```diff
@@ -270,4 +272,49 @@ private fun openLinkInCustomTab(uri: Uri, context: Context) {
...
+            it.matches("^/users/\\w+$".toRegex()) ||
+            it.matches("^/notice/[a-zA-Z0-9]+$".toRegex()) ||
+            it.matches("^/objects/[-a-f0-9]+$".toRegex()) ||
+            it.matches("^/notes/[a-z0-9]+$".toRegex()) ||
+            it.matches("^/display/[-a-f0-9]+$".toRegex()) ||
+            it.matches("^/profile/\\w+$".toRegex()) ||
+            it.matches("^/p/\\w+/\\d+$".toRegex()) ||
+            it.matches("^/\\w+$".toRegex()) ||
+            it.matches("^/@[^/]+/statuses/[a-zA-Z0-9]+$".toRegex()) ||
+            it.matches("^/o/[a-f0-9]+$".toRegex())
```
Lakoja, line 316: How "exact" should this matching be?

Most of the examples look like they have some minimum (or exact length) longer than the "one character" (+) in the expressions.

app/src/main/java/com/keylesspalace/tusky/util/LinkHelper.kt
```diff
@@ -270,4 +272,49 @@ private fun openLinkInCustomTab(uri: Uri, context: Context) {
...
+
+    if (uri.query != null ||
+        uri.fragment != null ||
+        uri.path == null
+    ) {
+        return false
+    }
+
+    return uri.path.let {
+        it.matches("^/@[^/]+$".toRegex()) ||
```
Lakoja, line 305: Suggestion: it could be more readable with the examples above being comments directly after each expression here.

Issue #2740, opened 2022-11-01
If posting with an image takes a long time, the post+image can be lost with user not informed


mcclure:
On 2022-Oct-31 mastodon.social was under heavy load due to the Twitter buyout and attempts to post images could hang for many minutes, or indefinitely. Tusky reacted incorrectly to the long delays.

Repro steps:
- Attach an image to a post
- Immediately tap "Post"
- This screen appears
    [<img width=300 alt="Finishing server upload..." src="https://user-images.githubusercontent.com/277318/199138227-fcc3108e-37db-4320-9cd2-af7d8d1e1e93.png">](https://user-images.githubusercontent.com/277318/199138227-fcc3108e-37db-4320-9cd2-af7d8d1e1e93.png)
- Press the Android "Home" button to go to the home screen. Wait a bit
- [[ In the background: the Tusky client gets oomkilled ]]
- Reopen Tusky

Assuming the server is taking long enough to accept an upload that the upload has still not completed, at this moment the **OBSERVED BEHAVIOR** is that when Tusky reopens, the screen flashes as if the Tusky app is rebooting and the feed is shown. The post does not get posted and in fact is **permanently lost**. There is no way to recover the text and if the image you uploaded was a crop created with "create image", there is no way to recover the image because we only keep our crop images in temporary storage.

**EXPECTED BEHAVIOR**

My **minimum expected behavior** is that when the post gets lost, Tusky notices it, moves the post to "Drafts", and on next boot pops up a warning saying a post failed to post.

A **best case expected behavior** is that Tusky continues attempting to make the post after uploading, and only falls back to "draft + notify" if the post outright fails. It is kind of irritating that you have to keep your phone looking at Tusky's "uploading..." box and not switch to another app, and there is no notification to the user that they need to do this so they are likely to switch away assuming it is safe. @connyduck made the observation in discord: "It would be extremely awesome if we could handoff uploading from that screen to the service when user hits send, but that is probably very complicated". But if this is complicated we should at least warn the user they lost a post.

**CONFIGURATION**

I tested and saw this with both version 19.0 on my main account and develop branch 5887af37721b (20.0 prerelease) on my test account.

[...]

PR #2947, opened 2022-11-28
Handoff media upload

connyduck:
Fixes a crash that can occur on Android 12+ when starting the SendStatusService while the app is in the background.
Stacktrace: https://gist.github.com/mcclure/9fee2ac07dd71c0444e80eab89b8e410

With this approach ComposeActivity no longer blocks the user when upload is still in progress but hands the upload over to SendStatusService, which then either proceeds with sending, or when the upload fails saves a draft and notifies the user.

closes https://github.com/tuskyapp/Tusky/issues/2740


Tak (approved):
Looks sensible, I haven't reproduced the problem here

charlag:
I am afraid I do not understand how it works fully. We start upload and then we pick it up in the service?

From practical:
1. If I start upload and set description and then send it, media description is lost
2. In this case I don't see a foreground notification even if I put the app in background

mcclure:
Test summary of [3d63596](https://github.com/tuskyapp/Tusky/pull/2947/commits/3d63596b0cf9bac82fc0323ca8c6f20fff790606) as posted on Matrix. Tested on Android 12 on Sony XPeria 5 III.

1. 5MB image, hit post during upload then exited. Post went up but WITHOUT AN IMAGE ATTACHED.
2. 5MB image, hit post during upload but did not exit. Post went to drafts (without me being notified) \*\* notice how weird this is, PR appears to have created a new regression
3. 2 smaller images, hit post during upload and attempted to exit. Posted successfully.
4. 5MB image, hit post during upload then exit. Post went to drafts (without me being notified)
5. 3MB image, hit post during upload then FORCE QUIT (ie something that would not have worked in 19.0). Post posted successfully WITH image.
6. 3MB image, hit post during upload then exit. Post went to drafts (without me being notified)
7. 5MB image, let complete upload, then post, then exit. Posted successfully WITH image.

Conclusions

PR fixes issue #2740 (see test 5) and the background crash Conny linked. When it fails, the failure mode is a strict improvement on either #2740 or in the crash Conny found (because a draft was saved except in test 1) . But the PR appears to contain three new regressions

1. Many posts appear to fail and go to drafts that (probably?) would have succeeded under 19.0.
2. When posts fail to post, there is *no* user notification, not even in the "normal" case where we hit post but do not background the app.
3. Test 1 saw the nightmare scenario of a post appearing but not in the form the user asked for. Note this MIGHT not be caused by the PR and it MIGHT even be something that happened on the mastodon dot social side rather than the Tusky side, because I have seen this behavior exactly once when issue #2740 is expressing.

connyduck:
> If I start upload and set the description and then send it, the description is lost

That is not a new bug though, it is in 19.0 already.

connyduck:
> I am afraid I do not understand how it works fully. We start upload and then we pick it up in the service?

Before ComposeActivity/ComposeViewModel were keeping track of uploads. Now it is media uploader, which is now a singleton and is also injected into SendStatusService. This way SendStatusService can get the upload state from MediaUploader, even though the uploads were started somewhere completely different.

charlag:
Ah I keep forgetting that it's the same process by default, it makes sense

connyduck:
Ok, I found another way to improve the code: Since Mastodon 4.0.0 api/v2/media can immediately process media, removing the need to check the state when it returns 200. Media that is still processing will still return 202. https://docs.joinmastodon.org/methods/media/#v2

The problem with no notifications was apparently caused by denied notifications.

I could reproduce posting and losing an attachment and I think it is a Mastodon issue https://github.com/mastodon/mastodon/issues/21864

connyduck:
I checked this again, I'm very sure its fine. The sporadic posting errors also happen in Tusky 20, but this fixes the crash.

mcclure:
Tests on 962ed17743f6f4:
* Posted 1MB image, no text, waited to finish uploading, hit "Post". Success
* Posted 4MB image, no text, hit "Post" before finish uploading. **Failure**. Post went to "Drafts" with no notification.
* Opened drafts, selected post with 5MB image in drafts, post had text, hit "Post" before upload indicator finished. Success.
* Posted 1MB image, hit "Post" before finish uploading, no text. Success (although image did not appear on timeline by itself, I had to pull down for more posts).
* Posted 4MB image, hit "Post" before finish uploading then immediately went into process select and force quit Tusky, post had text. **Failure**, post did **not** go to drafts and was lost completely.
* Posted 4MB image, hit "Post" before finish uploading then immediately went to home screen. Success.
* Posted 4MB image, hit "Post" before finish uploading. Success (image did not appear on timeline by itself)
* Posted 4MB image, hit "Post" before finish uploading then immediately went into process select and force quit Tusky. No text.  **Failure**, post did **not** go to drafts and was lost completely.
* Posted 1MB image, hit "Post" before finish uploading, no text. **Failure**, post went to "Drafts" with no notification.
 
    THIS time I grabbed a logcat:

```
2022-12-28 17:22:27.059 30260-30350/com.keylesspalace.tusky.test I/okhttp.OkHttpClient: --> POST https://mastodon.social/api/v2/media (834717-byte body)
2022-12-28 17:22:27.673 30260-30260/com.keylesspalace.tusky.test D/TimelineF: talkback was enabled: false, now false
2022-12-28 17:22:31.783 30260-30350/com.keylesspalace.tusky.test I/okhttp.OkHttpClient: <-- 200 https://mastodon.social/api/v2/media (4723ms, unknown-length body)
2022-12-28 17:22:31.796 30260-30350/com.keylesspalace.tusky.test I/okhttp.OkHttpClient: --> POST https://mastodon.social/api/v1/statuses (122-byte body)
2022-12-28 17:22:31.990 30260-30350/com.keylesspalace.tusky.test I/okhttp.OkHttpClient: <-- 422 https://mastodon.social/api/v1/statuses (193ms, unknown-length body)
2022-12-28 17:22:31.992 30260-30260/com.keylesspalace.tusky.test W/SendStatusService: failed sending status
    retrofit2.HttpException: HTTP 422 
        at at.connyduck.calladapter.networkresult.NetworkResultCall$enqueue$1.onResponse(NetworkResultCall.kt:55)
        at retrofit2.OkHttpCall$1.onResponse(OkHttpCall.java:161)
        at okhttp3.internal.connection.RealCall$AsyncCall.run(RealCall.kt:519)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1137)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:637)
        at java.lang.Thread.run(Thread.java:1012)
2022-12-28 17:22:32.004 30260-30335/com.keylesspalace.tusky.test D/DraftHelper: saved draft to db
* Brought up 1MB image from previous test in drafts, hit "Post" before upload indicator finished, no text. **Failure**, post  disappeared from drafts on send, but when I looked again. Curiously this time [nothing relevant-looking showed up in the logcat](https://gist.github.com/mcclure/da5de1f0f88aae95ad49f7dfc7b32aae) (only a "skia: --- Failed to create image decoder with message 'unimplemented'" message that appears frequently when using the app normally.
```

**CONCLUSIONS**

* Probably good to merge
* Unlike previous tests on Nov 29, posts nearly always succeeded, so I do not consider this version a regression. HOWEVER
    * Image uploads when hitting "Post" and staying sitting at the timeline seem to me to fail more often than they do in 20.0. Because failures are relatively sporadic, maybe the failure rate is the same and I am only noticing because I upload large images more often more often while I am doing this test, giving mastodon.social more chances to fail/reject an upload (notice the HTTP 422, implying mastosoc really did reject the image)
    * Still bothers me no notification on an upload fail
* In this version in every "drop to home screen" test I performed I got success, so this is an improvement over 20.0 in that respect
* Every single test where I force-quit during a background update failed with no recovery to drafts, but this would have failed in 20.0 as well, so no regression there
* I wish I could think of a way to test "mastodon process crashes during a background upload". This is a more interesting case than "user force quits Tusky", one would THINK they did that on purpose. Is it *expected* that force-quitting in the process viewer will cause the background process to close as well even though it is still working? Is it expected that a Tusky main process crash while the background process is working will force the background process to close?
* I think post-merge improvements would be worth it:
    * I think when background process fails an update and sends to drafts, the background process should notify the main process, or maybe leave a sentinel file, and the main process should bring up a dialog saying "a post upload failed and was saved to drafts" with options "OK" and "See drafts". I can help with this UI maybe?
    * When a background process **succeeds**, we know there is more content on the home feed, so maybe the feed should update to show the new post?

mcclure:
Update, just saw the post-fail-to-drafts in 20.0 with a 1MB image during normal non test use. This seems like evidence there is NO regression and the fail to drafts events I saw above were on the Mastodon dot social side

connyduck:
Thanks for testing!

> I think when background process fails an update and sends to drafts, the background process should notify the main process, or maybe leave a sentinel file, and the main process should bring up a dialog saying "a post upload failed and was saved to drafts" with options "OK" and "See drafts". I can help with this UI maybe?

yes we should do that

> When a background process succeeds, we know there is more content on the home feed, so maybe the feed should update to show the new post?

It actually does that. The problem is that sometimes the status is not yet inserted into the timeline when we refresh it, but for me it works most of the time.

mcclure:
> yes we should do that
> 
> >

OK, do you want to merge this now and then I will file a followup issue?

connyduck:
Yes please!

mcclure:
Wait, I already filed the followup issue. #2953

PR #2950, opened 2022-11-29
wrapper class for API that sanitizes unicode domain names

### LINKED ISSUE #2873: Tusky does not support IDN domains (with special characters)
Tusky does not work well with Mastodon servers with IDN domain names (those with special characters).

It fails right at the "Which instance?" screen when adding an account. If I type "federação.social" I get an error that the authentication with that instance failed.

If I use the punycode version of the same domain ( xn--federao-2wa9a.social ) things do generally work, but we see the punycode version of the domain in the UI (e.g. in the left pane or in local followers).

I have created an account for you to test:
login: debug@zorked.net password: debugpwd
server: federação.social (or xn--federao-2wa9a.social in punycode)

The account has a couple followers.

* * * *
- Tusky Version: 19.0 
- Android Version:  13
- Android Device: Pixel 5
- Mastodon instance (if applicable): this is a general issue with any instance with an IDN domain (but federação.social as an example)

kylegoetz:
fixes #2873 

Originally I wrote an interceptor that would convert unicode domains to ascii domains with the user's input being deemed the normalized form, but the code would still error out because OkHttp does not support unicode headers either, and the app sets the domain header.

So I wrote an implementation of `MastadonApi` that uses the Retrofit-generated version for most API calls but the ones that use the DOMAIN header convert Unicode to ASCII and then directly call the Retrofit-generated version with no other changes. This implementation is what gets injected by Dagger.

- [x] lint passes
- [x] new unit tests written
- [x] all unit tests pass

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/di/NetworkModule.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/di/NetworkModule.kt
@@ -27,6 +27,7 @@ import com.keylesspalace.tusky.json.Rfc3339DateJsonAdapter
 import com.keylesspalace.tusky.network.InstanceSwitchAuthInterceptor
 import com.keylesspalace.tusky.network.MastodonApi
 import com.keylesspalace.tusky.network.MediaUploadApi
+import com.keylesspalace.tusky.network.UnicodeSafeMastodonApi
 import com.keylesspalace.tusky.util.getNonNullString
 import dagger.Module
 import dagger.Provides
@@ -117,7 +118,10 @@ class NetworkModule {
 
     @Provides
     @Singleton
-    fun providesApi(retrofit: Retrofit): MastodonApi = retrofit.create()
+    fun providesApi(retrofit: Retrofit): MastodonApi {
+        val api = retrofit.create<MastodonApi>()
+        return UnicodeSafeMastodonApi(api)
+    }
 
     @Provides
     @Singleton
--- a/app/src/main/java/com/keylesspalace/tusky/network/UnicodeSafeMastodonApi.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/network/UnicodeSafeMastodonApi.kt
@@ -1,0 +1,142 @@
+/* Copyright 2022 kylegoetz
+ *
+ * This file is a part of Tusky.
+ *
+ * This program is free software; you can redistribute it and/or modify it under the terms of the
+ * GNU General Public License as published by the Free Software Foundation; either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * Tusky is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even
+ * the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU General
+ * Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License along with Tusky; if not,
+ * see <http://www.gnu.org/licenses>. */
+
+package com.keylesspalace.tusky.network
+
+import at.connyduck.calladapter.networkresult.NetworkResult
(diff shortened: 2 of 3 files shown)


charlag:
Hey, thanks for the PR!
Question: did it not work with an interceptor? It does seem like the primary use case for OkHttp interceptors?

kylegoetz:
@charlag It was unfortunately not sufficient. Consider this call defined in `MastodonApi`:

```kotlin
@GET("api/v1/notifications")
    fun notificationsWithAuth(
        @Header("Authorization") auth: String,
        @Header(DOMAIN_HEADER) domain: String,
        @Query("since_id") sinceId: String?
    ): Single<List<Notification>>
```

If a user calls this with `notificationsWithAuth(..., "http://federação.social", ...)` that's the final control a developer has over the domain header (second parameter) until the interceptor.

However, the domain header cannot be Unicode, only ASCII. So before the interceptor is even called, OkHttp errors out with an illegal character in the header (the `ç`).

I looked to see if Retrofit had some kind of header transformer (like how Room can take optional transformer decorator that will convert, say, a `Weight` class into a `Long` for DB insertion), and use that to transform Unicode to ASCII for the `domain` header, but didn't look like there are any. Would've been nice to do something like

```kotlin
@GET("api/v1/notifications")
    fun notificationsWithAuth(
        //...
        @Header(DOMAIN_HEADER) @Transformer(UnicodeToASCII) domain: String,
        //...
    ): Single<List<Notification>>
```

but not possible.

So the next best solution I had was to wrap `MastodonApi` in a "unicode-safe" version that, for the current calls that have a domain header, are a short function that converts Unicode to ASCII and then passes that into the original MastodonApi that is generated by Retrofit.

The `UnicodeSafeMastodonApi` is then what's injected by Dagger instead of the Retrofit-generated `MastodonApi`. The `UnicodeSafeMastodonApi` takes `api: MastodonApi` as a constructor param and then implements `MastodonApi by api` and then we just override the `domain` header calls with the Unicode-to-ASCII code. So as far as the rest of the existing code is concerned, no changes are needed, and they are still receiving `@Inject api: MastodonApi` the same as before.

connyduck:
How about using `retrofit2.http.Tag` annotation instead of a custom header? Seems like that would work with the interceptor and we don't need the wrapper.

kylegoetz:
Is the domain header ONLY necessary to do the domain swap in the interceptor? Or is it used anywhere else (like required by Mastodon server?).

I ask because if it's only required for that, then I think the header will cease to be necessary when addressing #2969 (an issue I noticed yesterday while tracing through the proxy code to fix a different issue):

1. don't use the domain header at all
2. move the "fix unicode to ascii" to an interceptor (like the original suggestion)
3. don't inject a singleton `MastodonApi`, 

Basically a singleton is generated once, the first time it's injected, and for the application's lifetime, it never changes. So if proxy settings change, they will not take effect for `MastodonApi` until the application is restarted. And it does not appear there is a way to change them in an interceptor.

So I'm thinking inject `Provider<MastodonApi>`, which lets us generate the api with proxy settings (and, if we want, the active API's domain name/base URL) in an activity/viewmodel instead of configuring it with a dummy URL and then swapping it out using the domain header in an interceptor.

I've never used `Provider<ABC>` before, but [the official docs](https://dagger.dev/api/2.13/dagger/Lazy.html) indicate that's what it's for.

connyduck:
> Is the domain header ONLY necessary to do the domain swap in the interceptor?

yes

> ...

That sounds like a big change, but also like it is worth it. We had some bugs in the past where domains got confused, those would also be avoided. Keep in mind that not in all places the domain is known at injection time (login)

charlag:
I don't really want to duplicate all API methods in the "fixed" implementation tbh, that's what bugs me the most, if we can find some generic solution that would be the best

Issue #2956, opened 2022-12-01
[Feature Request] Improve absolute time format with localized pattern.


babogoos:
Last week I joined into mastodon federation, and found wonderful open source android client Tusky here.

Then I found a [feature request post](https://mastodon.gamedev.place/@aeva/109385239203515343) by federation search that want to show 12 hour time format when using absolute time.
After tracing code, there is a relation PR #2413. 
This PR comment mention that we want to show day + month formatted with a localized way.

So I send a pull request to fulfill those wish. :D 

Please check my pull request when you have time.
If there are something need to changed, please let me know.

### Screenshots

![2022-12-01 15 26 46](https://user-images.githubusercontent.com/1038449/204991759-95b428d5-3d15-4758-907e-ef2b81b59a0e.jpg)

* * * *
- Tusky Version: 20 beta
- Android Version: 13
- Android Device: Pixel 4
- Mastodon instance (if applicable):

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

PR #2957, opened 2022-12-01
Improve absolute time format with localized pattern.

babogoos:
### Target

See Issue #2956.

This pull request do the things as following

* Show the best localized time pattern when using absolute time.
* Use android system setting to show 12/24 hour format.
* Add 12 hour format unit test.

### Something to consider/discuss:

DateFormat.is24HourFormat() need android context.
But I’m not familiar with Dagger. :(
I’m using Koin on my daily project. 
So I add the property without DI.

If there is more clear way to add property with android context, please let me know.

****
### Test on 
- Android Version: 13
- Android Device: Pixel 4


app/src/main/java/com/keylesspalace/tusky/util/AbsoluteTimeFormatter.kt
```diff
@@ -15,17 +15,30 @@
...
+    private val sameDayPattern = DateFormat.getBestDateTimePattern(Locale.getDefault(), sameDaySkeleton) ?: sameDaySkeleton
+    private val sameYearPattern = DateFormat.getBestDateTimePattern(Locale.getDefault(), sameYearSkeleton) ?: sameYearSkeleton
+    private val otherYearPattern = DateFormat.getBestDateTimePattern(Locale.getDefault(), otherYearSkeleton) ?: otherYearSkeleton
+    private val otherYearCompletePattern =
+        DateFormat.getBestDateTimePattern(Locale.getDefault(), otherYearCompleteSkeleton) ?: otherYearCompleteSkeleton
+
+    private val sameDaySdf = SimpleDateFormat(sameDayPattern, Locale.getDefault()).apply { this.timeZone = tz }
+    private val sameYearSdf = SimpleDateFormat(sameYearPattern, Locale.getDefault()).apply { this.timeZone = tz }
+    private val otherYearSdf = SimpleDateFormat(otherYearPattern, Locale.getDefault()).apply { this.timeZone = tz }
+    private val otherYearCompleteSdf = SimpleDateFormat(otherYearCompletePattern, Locale.getDefault()).apply { this.timeZone = tz }
```
Lakoja, line 41: Not sure how, but this actually displays different (correct looking) things for different Locales. :-)

babogoos: Thanks for review. :-D
Just using this api to get the best localize date format pattern. 
[DateFormat.getBestDateTimePattern(Locale locale, String skeleton)](https://developer.android.com/reference/android/text/format/DateFormat.html?is-external=true#getBestDateTimePattern(java.util.Locale,%20java.lang.String))

app/src/main/java/com/keylesspalace/tusky/adapter/NotificationsAdapter.java
```diff
@@ -112,6 +113,7 @@ public NotificationsAdapter(String accountId,
     @Override
     public RecyclerView.ViewHolder onCreateViewHolder(@NonNull ViewGroup parent, int viewType) {
         LayoutInflater inflater = LayoutInflater.from(parent.getContext());
+        absoluteTimeFormatter = new AbsoluteTimeFormatter(DateFormat.is24HourFormat(parent.getContext()));
```
connyduck, line 116: Pass a context into the NotificationsAdapter constructor then you can only create one AbsoluteTimeFormatter instead of one for every ViewHolder.

babogoos: The other ViewHolder does not build in `NotificationsAdapter`.
I think `AbsoluteTimeFormatter` is a UI Util tool, so I use ViewHolder context.

connyduck: How does that make a difference if it is only used to look up the 24h format?

PR #2961, opened 2022-12-01
2952/proxy

### LINKED ISSUE #2952: Does the HTTP Proxy feature even work?
There is an "HTTP proxy" feature at the bottom of the Tusky preferences.

I ran on my PC the https://deelay.me/ proxy, enabled "Enable HTTP proxy" and set "HTTP prox server" to "http://192.168.[redacted]:4567/10000". I did not set the port. I expected this to route all mastodon traffic through the deelay proxy, delaying every API call by 10 seconds.

Instead I was able to browse mastodon normally and nothing unusual happened. I did not see on the deelay proxy window notifications it was getting http requests (which I did when I tested with a PC). I did not get any errors. I did not notice anything in a casual view of the logcat. It continued working even after I shut down the deelay server, implying it was never even trying to contact deelay in the first place.

I tried restarting Tusky and did not get any different results.

Maybe deelay is not a compliant proxy server or the proxy server expects a hostname not a url (deelay does require a url because the delay length is part of the url). But I would expect some sort of error to be reported instead of Tusky just going ahead and contacting the normal Internet (the originally stated use case for the http proxy feature was privacy protection, so you really wouldn't want it to bypass the privacy protection on a failure).

* * * *
- Tusky Version: 20.0, [3d63596](https://github.com/tuskyapp/Tusky/pull/2947/commits/3d63596b0cf9bac82fc0323ca8c6f20fff790606)
- Android Version: 12
- Android Device: Sony XPeria 5 III

kylegoetz:
fixes #2952

Setting now only saves setting if empty or port between min and max values. If invalid, does not change previous value *plus* fires off a Toast message about the valid port range.


mcclure:
I think this is a good change but should it also check whether the host name is nonempty / well formed?

kylegoetz:
I don't think we should check if it's non-empty since non-empty is a valid setting (i.e., "no proxy"). But I don't like the idea of clearing out hostname would also clear out port.

We could have "empty hostname" or "empty port" auto-disable the proxy enable toggle, but the problem there is if someone re-enables proxy then the initial setting will be empty hostname or empty port (or both), which would re-trigger the disabling if they clicked to edit and then canceled. Not sure if that's good UI. I'm a programmer, not a UI/UX guy.

And regarding validating hostname, I think that'd be good, but it's not part of the issue this fixes (which turned out to be a port issue—i.e, not validating port when changing user setting—based on comments below)

mcclure:
Okay. To be clear what I'm trying to prevent was when I tried to use this feature I put an *entire URL*, `https://hostname:port/path` into the box, and that appears to be wrong above and beyond me having left the port blank. If this is wrong the user should be gently prevented from doing it...

app/src/main/java/com/keylesspalace/tusky/util/SettingsUtils.kt
```diff
@@ -0,0 +1,12 @@
+package com.keylesspalace.tusky.util
+
+const val MIN_PROXY_PORT = 0
+const val MAX_PROXY_PORT = 65535
+private val PROXY_RANGE = IntRange(MIN_PROXY_PORT, MAX_PROXY_PORT)
+
+fun isValidProxyPort(value: Any): Boolean = when (value) {
```
connyduck, line 7: Can you put this into ProxyPreferencesFragment please? I'd only create util files when something is used in multiple palces, otherwise it just makes code harder to understand if stuff is all over the place.

app/src/main/java/com/keylesspalace/tusky/components/preference/ProxyPreferencesFragment.kt
```diff
@@ -44,8 +49,23 @@ class ProxyPreferencesFragment : PreferenceFragmentCompat() {
...
+                    R.string.pref_title_http_proxy_port_message,
+                    MIN_PROXY_PORT,
+                    MAX_PROXY_PORT
+                )
+                this.dialogMessage = portMessage
                 setTitle(R.string.pref_title_http_proxy_port)
                 key = PrefKeys.HTTP_PROXY_PORT
+                onPreferenceChangeListener = Preference.OnPreferenceChangeListener { _, newValue ->
+                    isValidProxyPort(newValue).also { isValid ->
+                        if (!isValid) Toast.makeText(
```
connyduck, line 62: Snackbar is preferred to Toast because it stays local to the activity

connyduck (changes requested):
A 💯 solution would be a custom dialog that only allows number input in the range in the first place. 

This definitely improves the situtation of #2952, but I don't think it closes it.

nikclayton:
This should probably be at least two distinct settings.

1. A boolean for whether or not the proxy should be used
2. The proxy configuration

This is so that the user can e.g., temporarily disable the proxy, and re-enable it later without having to reenter the proxy information.

I've attached an image of how Twidere does it to make it clearer.

image

kylegoetz:
@nikclayton Yes, I had considered this, but figured it should be a different PR for a different issue (that doesn't currently exist). But given @connyduck most recent comment on this PR, this is how I would expect to implement a "toggle-off disables the host and port onpress listeners"

app/src/main/java/com/keylesspalace/tusky/util/SettingsUtils.kt
kylegoetz: I think it will be used in two places: 

1. user input validation (ProxyPreferencesFragment)
2. when retrieving values from settings to create the proxy itself. Theoretically the settings should never be invalid, but since we're deserializing data, I personally don't trust it. Also if this PR is accepted, there could be pre-existing invalid settings that we might be loading from a user's computer.

kylegoetz: It is in my local already :) I've got the custom dialog + error message stuff, OK button disabled if the port isn't valid so you can't even save an invalid port, but I need to write tests for it before updating the PR.

kylegoetz:
Updated: 

1. as requested by @nikclayton above,  fixes #2984 toggling "enable" to off will enable disable hostname/port inputs; toggling to "on" will enable them
2. as requested by @connyduck above, has custom inputs for proxy hostname and port. For hostname, "OK" button is disabled if hostname is an invalid hostname. For port, same behavior plus has error tooltip if port is invalid while editing

app/src/main/java/com/keylesspalace/tusky/entity/ProxyConfiguration.kt
```diff
@@ -0,0 +1,32 @@
+package com.keylesspalace.tusky.entity
+
+import java.net.IDN
+
+class ProxyConfiguration private constructor(
```
connyduck, line 5: The entity package is only for models returned by the Mastodon api, this should be placed somewhere else, maybe settings?

app/src/main/java/com/keylesspalace/tusky/components/preference/ProxyPreferencesFragment.kt
```diff
@@ -16,38 +16,58 @@
...
                 setDefaultValue(false)
             }
 
-            editTextPreference {
+            val serverPref = validatedEditTextPreference(null, ProxyConfiguration::isValidHostname) {
                 setTitle(R.string.pref_title_http_proxy_server)
                 key = PrefKeys.HTTP_PROXY_SERVER
                 isIconSpaceReserved = false
                 setSummaryProvider { text }
+                isEnabled = enableToggle.isChecked
```
connyduck, line 46: If you use the preference dependency api you don't need that custom listener.

app/src/main/java/com/keylesspalace/tusky/components/preference/ProxyPreferencesFragment.kt
```diff
@@ -16,38 +16,58 @@
...
         makePreferenceScreen {
-            switchPreference {
+            val enableToggle = switchPreference {
                 setTitle(R.string.pref_title_http_proxy_enable)
                 isIconSpaceReserved = false
                 key = PrefKeys.HTTP_PROXY_ENABLED
                 setDefaultValue(false)
             }
 
-            editTextPreference {
```
connyduck, line 39: `editTextPreference` is now unused and can be deleted.

kylegoetz: Gotcha. I was afraid to be too aggressive with deleting unused code I didn't write!

connyduck (approved):
definitely better than before

mcclure:
Very cool

I will test in nightly, if I can get this to misbehave I might file an additional issue >_>

Issue #2970, opened 2022-12-02
Crash on opening with nightly PlayStore version


Draky50110:
Hello.

When opening "last" Tusky Nightly, it crashes.
Here are errors lines : 

java.lang.NullPointerException: Attempt to invoke virtual method 'int android.app.NotificationChannel.getImportance()' on a null object reference
at f.f0.h(SourceFile:5)
at k0.e.a(SourceFile:5)
at m6.i.o(SourceFile:10)
at x9.a.j(SourceFile:3)
at ma.h0.run(SourceFile:20)
at ra.e.run(SourceFile:2)
at ta.i.run(SourceFile:1)
at ta.a.run(SourceFile:15)
Suppressed: ma.f0: [q1{Cancelling}@ab80b11, Dispatchers.Main.immediate]

* * * *
- Tusky Version: 20.0-cc790ccf
- Android Version: 9 (through Xiaomi MIUI)
- Android Device: Xiaomi Redmi Note 6 Pro
- Mastodon instance (if applicable): mastodon.social + piaille.fr

- [X] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

connyduck:
Unobfuscated stacktrace

```
Exception java.lang.NullPointerException: Attempt to invoke virtual method 'int android.app.NotificationChannel.getImportance()' on a null object reference
  at com.keylesspalace.tusky.components.notifications.NotificationHelper.filterNotification (NotificationHelper.java:549)
  at com.keylesspalace.tusky.components.notifications.PushNotificationHelper.buildSubscriptionData (PushNotificationHelper.java:155)
  at com.keylesspalace.tusky.components.notifications.PushNotificationHelper.access$buildSubscriptionData (PushNotificationHelper.java:1)
  at com.keylesspalace.tusky.components.notifications.PushNotificationHelper$updateUnifiedPushSubscription$2.invokeSuspend (PushNotificationHelper.kt:200)
  at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith (ContinuationImpl.kt:33)
  at kotlinx.coroutines.DispatchedTask.run (DispatchedTask.kt:106)
  at kotlinx.coroutines.internal.LimitedDispatcher.run (LimitedDispatcher.kt:42)
  at kotlinx.coroutines.scheduling.TaskImpl.run (Tasks.kt:95)
[...]

PR #2971, opened 2022-12-02
Nullcheck channels when reading importance

Tak:
Blindfix for #2970 , which I can't reproduce locally

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/components/notifications/NotificationHelper.java
+++ b/app/src/main/java/com/keylesspalace/tusky/components/notifications/NotificationHelper.java
@@ -473,7 +473,7 @@ public class NotificationHelper {
 
             if (notificationManager.areNotificationsEnabled()) {
                 for (NotificationChannel channel : notificationManager.getNotificationChannels()) {
-                    if (channel.getImportance() > NotificationManager.IMPORTANCE_NONE) {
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/keylesspalace/tusky/components/notifications/NotificationHelper.java
```diff
@@ -473,7 +473,7 @@ public static boolean areNotificationsEnabled(@NonNull Context context, @NonNull
 
             if (notificationManager.areNotificationsEnabled()) {
                 for (NotificationChannel channel : notificationManager.getNotificationChannels()) {
-                    if (channel.getImportance() > NotificationManager.IMPORTANCE_NONE) {
+                    if (channel != null && channel.getImportance() > NotificationManager.IMPORTANCE_NONE) {
```
nikclayton, line 476: Wouldn't `getNotificationChannels()` return an empty list, instead of a list with a null entry? Or is this a bug in the SDK/platform?

Tak: I hope? I didn't get a full stack trace for the issue, and I couldn't reproduce locally, so I added guards to both the places we're calling `NotificationChannel.getImportance()`

connyduck (approved):

Issue #2799, opened 2022-11-10
Opening menu / settings is very hard after disabling top-bar


benthillerkus:
Even before disabling the setting I could only open the drawer by clicking on my profile icon, which acted as a hamburger. Swiping left did not work (not sure if it was supposed to.)

Either way, it seems I locked myself out now, as I cannot access the menu to revert the setting anymore.

I had also changed the tab bar to appear at the bottom
![Screenshot_2022-11-10-05-57-36-313_com.keylesspalace.tusky.jpg](https://user-images.githubusercontent.com/29630575/201004485-d41a72b7-f519-44ad-9e44-9b2467798d35.jpg)

Edit: After stumbling upon #2750 I realised the menu could still be opened by sliding the finger vertically on the the display edge ---- I would've never figured that out myself.

PR #2973, opened 2022-12-02
Show Avatar next to tabs when main top bar is disabled

connyduck:
This should help all those users that disable the top bar and then don't find the menu anymore.

The 4 possible states:

   

closes https://github.com/tuskyapp/Tusky/issues/2799
makes https://github.com/tuskyapp/Tusky/pull/2750 obsolete


Gryzor:
Love it

charlag (approved):

PR #2990, opened 2022-12-05
Implement getFilters() without rxjava

nikclayton:


connyduck (changes requested):
Can you add an explicit name for the lambda parameter please? I think it makes the code easier to read.

nikclayton:
I can, but just to double check, that'll make it inconsistent with the surrounding code, which pretty consistently doesn't use named lambda parameters for the error path.

connyduck:
You can change them all if you want! But it is not a big deal, you can also leave all, or name only the changed ones, it's up to you.

nikclayton:
> You can change them all if you want! But it is not a big deal, you can also leave all, or name only the changed ones, it's up to you.

I'd say leave as is at the moment, and I'll make a note / file an issue for a custom ktlint rule to flag all lambdas as needing named parameters if you want that throughout the codebase (it's counter to https://kotlinlang.org/docs/coding-conventions.html#lambda-parameters, but I can see the arguments for wanting it).

nikclayton:
Although having said that, there's 150+ instances of use of `it` in lambdas with unnamed single parameters at the moment in the codebase., so perhaps keeping it is OK?

PR #2993, opened 2022-12-05
Cleanup NotificationFragment to make future conversion to Kotlin easier

nikclayton:
- Convert to use view binding
- Use requireContext() in places a context is required
- Simplify code with .sublist() and .contains()
- Add @nonnull annotations where appropriate
- Use consistent comment styles


connyduck:
Thank you! I will start with the refactoring soon, but I expect it will take at least a month.

PR #2995, opened 2022-12-05
don't include dependency info in apk

connyduck:
Makes the apk a tiny bit smaller.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -90,6 +90,10 @@ android {
     kotlinOptions {
         freeCompilerArgs += "-opt-in=kotlin.RequiresOptIn"
     }
+    dependenciesInfo {
+        includeInApk false
+        includeInBundle false
+    }
 }
 
 // library versions are in PROJECT_ROOT/gradle/libs.versions.toml


Tak (approved):
I don't have insight into the implications, but it makes sense to me

nikclayton:
Reading between the lines, I wonder if this might cause future problems with Play store review?

https://developer.android.com/reference/tools/gradle-api/7.2/com/android/build/api/dsl/DependenciesInfo says:

> Including dependency information in your APK or Bundle allows Google Play to ensure that any third-party software your app uses complies with [Google Play's Developer Program Policies](https://support.google.com/googleplay/android-developer/topic/9858052). For more information, see the Play Console support page [Using third-party SDKs in your app](https://support.google.com/googleplay/android-developer/answer/10358880).

https://support.google.com/googleplay/android-developer/answer/10358880 says:

> To help you ensure that any third-party software your app is using complies with [Google Play’s Developer Program Policies](https://support.google.com/googleplay/android-developer/topic/9858052) and contributes positively to your user experience, Google Play will begin flagging known issues with popular SDKs in Play Console.

Hence my concern that if the info is not there the review process might take longer.

I'm not sure that this will have an impact on the size of the APK that's delivered to the end user either. If I'm reading https://developer.android.com/guide/app-bundle correctly, submitting to Play pushes an App Bundle, and Play generates per-device optimized APKs for the user automatically.

I guess this might have an impact on F-Droid? I'm not as familiar with their practices.

connyduck:
Yes, Google Play could cause problems. As I see it, they use this info just to inform developers and not to enforce anything. But we will know more with the next nightly deployment.

Issue #2938, opened 2022-11-25
"load more" experience


kulimuliF:
If you come back after a while and you have missed a lot of posts, you see the "load more" button, which is loading a bulk of new posts. By clicking on it it, you are at the top or maybe only somewhere at the middle 
of the posts if there is another bulk of messages, which you can load by clicking again on the button.

If you want to continue reading where you have stopped yesterday (old to new), it's awful to find the post where you have paused previously.

Is it possible to leave the focus on the old post, or load the new posts dynamically when you scroll up?

* * * *
- Tusky Version: 19.0, 20beta.1
- Android Version: 12
- Android Device: one plus nird
- Mastodon instance (if applicable): aut.social

- [ ] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

nikclayton:
Duplicate of https://github.com/tuskyapp/Tusky/issues/2788

charlag:
@nikclayton it sorts this way because status IDs are not guaranteed to be numbers, but are sortable. We paginate by ID therefore we must sort by ID.

connyduck:
And they are sorted by length first because when they are numbers "2" must come before "12" (which in alphabetic ordering isn't the case)

nikclayton:
So the Mastodon server guarantees that sorting by status ID gives you a stable ordering of the statuses? Presumably in the order that they were received by the server? Is that a guarantee in the API docs somewhere and I can't find it, or is this an implementation detail that's effectively baked in to the protocol now?

charlag:
They added this info after there was an Incident where we assumed IDs to be a number and in few server impls they aren't. Mastodon predates AP and does not use AP internally nor in the client API (which also predates AP). It has nothing to do with AP objects/activities. I don't know how Mastodon assigns those IDs

nikclayton:
Doing some more experimentation, the `min_id` parameter is the one to use (https://docs.joinmastodon.org/methods/timelines/#home).

Consider statuses with IDs 10, 11, 12, ... 20 (in order oldest to newest).

Requesting /api/v1/timeslines/home?limit=5&since_id=5&max_id=20 will return statuses with IDs 15, 16, 17, 18, 19 (i.e., the most recent 5 statuses that are older than max_id).

Requesting /api/v1/timeslines/home?limit=5&min_id=5&max_id=20 will return statuses with IDs 6, 7, 8, 9, 10 (i.e., the 5 statuses immediately after min_id).

[...]

Issue #2929, opened 2022-11-24
Option for Load More to insert above the current selection point


Clfh47:
Feature request: I prefer to read timeliness from oldest to newest. At present Load More adds the new content below that point. 
It means that one has to try to remember what was below that point and scroll down through all the newly-added posts to get back to the insertion point. 
The requested option would add the newly-downloaded posts above the Load More point, leaving the post below that still visible. 

* * * *
- Tusky Version:19.0
- Android Version:7.0
- Android Device:Acer P027
- Mastodon instance (if applicable): mastodon.green

- [y ] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

Tak:
Requested timeline behavior changes are being gathered in #2788

Clfh47:
Thanks. Have added my Thumb-up to the best-fitting request in that list.

Issue #2788, opened 2022-11-08
Timeline improvements


connyduck:
This issue is for collecting the various wishes users have about timelines. This way we can think about them together. Patching single issues will lead us nowhere.

- For some users the current timeline behavior works 
- Some users would like to read the timeline from oldest to newest without "load more" breaks
- Some users would like to read from newest to oldest without "load more" breaks
- Some users would like to decide if "load more" loads newer or older posts first (Fedilab has this feature afaik)
- Some users would like to customize how many posts are loaded per page.

Anything else?
 
A solution should ideally address all these needs without adding more than one additional setting.

yoshimo:
Make it two options :
- radio button for the sort order, newest or oldest first
- slider how many messages are loaded per refresh with a sane default. Between say 50,100,200,infinite.  Infinite being a gapless automatic  refresh. If it is a finite number have a load more button shown. 

No matte how many toots you load on refresh, keep the old position and last read toot focusses.

Vilinthril:
My desired user experience is:
* I read all toots on my timeline or a selected list, at irregular intervals. Ideally, there's an unread counter that shows how much I am catching up to, maybe that could be hidden in the wellness option and warned against if other people strongly dislike that?
* Either way, ideally, the app would “anchor” my position in the timeline/list at the last point I had read to and would silently load new toots on top of it, without jumping around while doing that.
* The anchor point should (ideally) be saved even between app restarts/Android reboots.

johseg:
+1 for @Vilinthril

jollysea:
just a note on "unread counter": I think I would interpret that as "I got x notifications", as most other apps work that way. So I'd suggest making it opt-in. :)

iGEL:
If you add a counter of unread toots, it could jump to the newest toot if you tab it.

Ganneff:
> This issue is for collecting the various wishes users have about timelines. This way we can think about them together. Patching single issues will lead us nowhere.

I would like to add one wish, if I may:
When I click on "load more", I do want them to get loaded in a way, that does NOT change my relative position.
Currently, when i scroll up to read newer things, and a load more button appears, clicking it means they are added below my current position (into the area I already read), so I have to scroll back down, hopefully remembering the last toot i read, then continue reading there. That is less than useful IMO. It is double worse, as I then have some unread below and above my current place..

l17r:
[...]

PR #3000, opened 2022-12-05
Keep scroll position when loading missing statuses

nikclayton:
Add a preference for user reading behaviour (oldest first or newest first) and honour it when loading more statuses in the middle of a timeline.

"Oldest first" means the user scrolls up, expecting to load new statuses as they go, and when new statuses load the user's reading position is retained at the oldest status they've read.

"Newest first" means the user is scrolling down, expecting to load new statuses as they go, and the reading position is retained at the newest status they've read.

Default is "oldest first", as this was the behaviour before Tusky 17, and was unintentionally changed.

Ensure that a circular progress spinner appears while loading.

Add a "Developer tools" feature that is enabled in debug builds. Use this to provide an option to remove content from the local cached timeline, to make it easy for developers to examine the behaviour.


nikclayton:
I've been keeping notes and questions in https://github.com/tuskyapp/Tusky/issues/2938

Tak:
Are you planning to add a preference for the reading direction? (I definitely would not be in favor of an unconditional direction change…)

nikclayton:
Reading direction doesn't change in this PR. I'll make a separate PR for that.

I'll break this one in to two as well. One to keep the position in the list consistent (no new preference change needed), and one to control whether newest or oldest statuses are loaded (that needs a pref, there are more details about that pref in https://github.com/tuskyapp/Tusky/issues/2788#issuecomment-1324764616)

Tak:
That's what I mean, changing which statuses get loaded effectively changes the reading direction

nikclayton:
I don't think it does.

Here's a slightly simplified example to keep the numbers and diagrams very easy.

Consider a timeline that shows 40 statuses, with IDs 1..40. ID 1 is the oldest (so appears at the bottom of the timeline), 40 is the newest, so appears at the top of the timeline).

Now suppose that some of those statuses are missing, because the app hasn't loaded them. Statuses 1-10 and 30-40 are present, but 11-29 are missing.

And the user's device has enough vertical height to show about 4 statuses at the same time (again, simplified, assume all statuses take the same vertical height).

In the app that looks like this:

```
    [ status 40   ]
    [ status 39   ]
,-  [ ...         ]  -.
|   [ status 31   ]   |
|   [ "Load more" ]   | <-- viewport
|   [ status 10   ]   |
`-  [ status 9    ]  -'
    [ ...         ]
    [ status 1    ]
```

The user scrolls so status 31, "Load more", and status 10 are visible (marked on the diagram above) and clicks "Load more". This is actually the status with ID 30, and it's visible on the screen.

1. N statuses are fetched to fill the gap, "anchored" to the top. If N is 10 then statuses 20..30 will be fetched.
2. The visible status continues to be the one with ID 30. So most of the newly loaded statuses are scrolled off the bottom of the screen. The user has to scroll down to most recent one, and then scroll back up to read them.

I.e., the state immediately after "Load more" finishes is:

```
    [ status 40   ]
    [ status 39   ]
,-  [ ...         ]  -.
|   [ status 31   ]   |
|   [ status 30   ]   | <-- viewport
|   [ status 29   ]   |
`-  [ status 28   ]  -'
    [ ...         ]
    [ status 21   ]
    [ "Load more" ]
    [ status 10   ]
    [ ...         ]
    [ status 1    ]
```

So the user has clicked "Load more" because they want to see more statuses, the statuses have loaded, but the user has to carry out another action (scroll) before they can achieve their goal.

This PR changes that behaviour[1]. After clicking "Load more" in this example the user's view looks like this:

```
    [ status 40   ]
    [ status 39   ]
,-  [ ...         ]  -.
|   [ status 23   ]   |
|   [ status 22   ]   | <-- viewport
|   [ status 21   ]   |
`-  [ "Load more" ]  -'
    [ status 10   ]
    [ ...         ]
    [ status 1    ]
```

So the newly loaded statuses have scrolled in to view, and the user can continue reading their timeline by scrolling up (which is the same direction they normally scroll to read their timeline).

So the reading direction *hasn't* changed, this is just a bug fix.

The reason it's a bug fix and not new behaviour that needs to be behind a preference is because if the user *didn't* want to read statuses 21-30 they could have scrolled past the "Load more" entry and not clicked it. By clicking it they've clearly signaled they do want to read those statuses, and it's a bug to require them to first scroll down.

[1] It will -- at the time I'm writing this it also changes the anchor position so that in this example it's statuses 11..20 that will be loaded, not statuses 21-30. That's new behaviour, which I'll revert from this PR and put behind a preference in a new PR.

nikclayton:
As I say, that's not necessary because this PR doesn't change the reading direction.

And in the master issue, https://github.com/tuskyapp/Tusky/issues/2788#issuecomment-1324764616, I've clearly set out the two new preferences that will be necessary.

But this PR doesn't make those changes yet (or rather, it won't in an hour or so when I send an update to revert part of it).

Tak:
I don't particularly want to argue about it, but this explanation makes even less sense.
If I'm reading newest-oldest (which I do), I want to load 21-30 and keep reading downward from 30.
If I'm reading oldest-newest, I want to load 11-20 and read upward from 11.
In *no* circumstance do I want to load a block of posts and jump to the *far* end of the loaded block.

nikclayton:
> If I'm reading oldest-newest, I want to load 11-20 and read upward from 11.
> In no circumstance do I want to load a block of posts and jump to the far end of the loaded block.

That ("In no circumstance do I want to load a block of posts and jump to the far end of the loaded block) is the current behaviour (i.e., without this PR) if you're reading oldest to newest. That's why people are complaining about it in:

- https://github.com/tuskyapp/Tusky/issues/2938
- https://github.com/tuskyapp/Tusky/issues/2929
- https://github.com/tuskyapp/Tusky/issues/2423 -- notes that the current behaviour is a regression in Tusky 17
- https://github.com/tuskyapp/Tusky/issues/2508
- https://github.com/tuskyapp/Tusky/issues/2496
- https://github.com/tuskyapp/Tusky/issues/1230

This PR fixes that.

> If I'm reading newest-oldest (which I do), I want to load 21-30 and keep reading downward from 30.

Huh. I think you might be the first person in all of these issues that's reported that they do that.

In that case, this could be put behind a preference (which I can do in this PR), but I imagine the default should be this behaviour, since the current behaviour was a regression and has lead to these complaints.

nikclayton:
> That's likely because people who are suited by the current behavior don't have as much motivation to speak up

That is a possibility. But I think that would be more likely to be true if, pre Tusky 17, there were issues filed by people who wanted the newest-oldest behaviour.

But I can't find any. So I think it's reasonable to conclude that the majority want the oldest-newest behaviour.

nikclayton:
Preference added, PR title and description updated to reflect reality.

nikclayton:
Just occurred to me that the reading order preference would affect pull-to-refresh behaviour as well. I'll add another commit so if the user's reading order is set to "newest first" then pull-to-refresh puts them at the start of their timeline.

nikclayton:
> Just occurred to me that the reading order preference would affect pull-to-refresh behaviour as well. I'll add another commit so if the user's reading order is set to "newest first" then pull-to-refresh puts them at the start of their timeline.

Done.

nikclayton:
I've been able to both simplify this a bit and apply some UI polish. I think it's ready for a detailed review.

app/src/main/java/com/keylesspalace/tusky/adapter/PlaceholderViewHolder.kt
```diff
@@ -14,27 +14,55 @@
...
 import com.keylesspalace.tusky.interfaces.StatusActionListener
 
 /**
  * Placeholder for different timelines.
- * Either displays "load more" button or a progress indicator.
- **/
+ *
+ * Displays a "Load more" button for a particular status ID.
+ *
+ * If loading for the placeholder is already active the user cannot interact
```
Lakoja, line 31: Question/observation about the comments in this PR: they seem rather lenghty and mostly explain what also the code should explain?

In this case here for example the class is so short that I would expect the code to speak mostly for itself.

As far as I understand comment lore they should - when necessary at all - explain the why and maybe what but normally not the details.

Having said that I do feel the need for _some_ more comments in the (existing) code to give some hint what is happening there. But a one-liner is normally sufficient for that.

nikclayton: Would you prefer something like:

```
/**
 * Displays a "Load more" button for a particular status ID, or a
 * circular progress wheel if the status' page is being loaded.
 *
 * The user can only have one "Load more" operation in progress at
 * a time (determined by the adapter), so the contents of the view
 * and the enabled state is driven by that.
 */
```

app/src/main/java/com/keylesspalace/tusky/components/preference/PreferencesFragment.kt
```diff
@@ -51,6 +51,26 @@ class PreferencesFragment : PreferenceFragmentCompat(), Injectable {
     private val iconSize by lazy { resources.getDimensionPixelSize(R.dimen.preference_icon_size) }
     private var httpProxyPref: Preference? = null
 
+    enum class ReadingOrder {
+        /** User scrolls up, reading statuses oldest to newest. Default behaviour. */
+        OLDEST_FIRST,
```
connyduck, line 56: NEWEST_FIRST is the current behavior, right? Can we have it as the default please?

connyduck (commented):
Yeah this probably helps users that want to read oldest first a lot.

But it should also work in non-cached timelines.

app/src/main/java/com/keylesspalace/tusky/usecase/DeveloperToolsUseCase.kt
```diff
@@ -0,0 +1,46 @@
...
+import androidx.room.withTransaction
+import com.keylesspalace.tusky.db.AppDatabase
+import com.keylesspalace.tusky.db.TimelineDao
+import javax.inject.Inject
+
+/**
+ * Functionality that is only intended to be used by the "Developer Tools" menu when built
+ * in debug mode.
+ */
+class DeveloperToolsUseCase @Inject constructor(
```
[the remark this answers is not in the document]
nikclayton: ProGuard already removes it for release builds, because in a release build there's no code that calls it.

You can verify this by:

1. Perform a release build
2. Run `grep DeveloperTools app/build/outputs/mapping/blueRelease/usage.txt` or similar

Despite the `usage.txt` name, this file is a list of all the things that ProGuard removed, and you can confirm that the `DeveloperTools` class, the `MainActivity.buildDeveloperToolsDialog()` function, the companion object, the factory class used for injection, etc etc, are all listed in this file, so they're not in the APK.

app/src/main/java/com/keylesspalace/tusky/components/preference/PreferencesFragment.kt
nikclayton: I think NEWEST_FIRST is a bug, that this fixes.

From what I can tell, up to and including v16 beta 1 Tusky  OLDEST_FIRST behaviour. I verified this by checking out 31da851f, adding the DeveloperTools functionality, and experimenting with creating "Load more" gaps of different sizes and seeing how the app behaved when I clicked the button. It kept the position in the list.

I think the NEWEST_FIRST behaviour was introduced in 6d4f5ad0279e5ff51f143742fca060dd04e1c30c with the migration to Paging3.

I can't see anything that suggests that this was a deliberate change, it's just how Paging3 works by default.

This change rapidly started annoying users, and they submitted issues about it. The earliest issue I can see is https://github.com/tuskyapp/Tusky/issues/2423 from Apr 7th.

I can't find any issues from before the change where users were asking for "Load more" to have its current behaviour. Only issues from after the change with users asking for the behaviour to be reverted.

This is why I think the best thing to do for the users is for the default to be the previous behaviour, which fixes the bug introduced in 6d4f5ad.

nikclayton:
Note: I may have figured out a more elegant way to fix this, but it'll take me a couple of days to have the time to try it out. I'll mark this PR as a draft until I've done that.

app/src/main/java/com/keylesspalace/tusky/components/preference/PreferencesFragment.kt
connyduck: But OLDEST_FIRST is loading the oldest statuses in the gap, we never did that?

app/src/main/java/com/keylesspalace/tusky/usecase/DeveloperToolsUseCase.kt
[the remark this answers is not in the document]
connyduck: Ok, I see, thank you

app/src/main/java/com/keylesspalace/tusky/adapter/PlaceholderViewHolder.kt
Lakoja: Yes, that sounds much better.

app/src/main/java/com/keylesspalace/tusky/components/timeline/TimelineFragment.kt
```diff
@@ -201,12 +234,39 @@ class TimelineFragment :
...
+                            // Swiping and prefer oldest-first? Stay at the same position.
+                            // Otherwise jump to the top.
+                            if (isSwipeToRefreshEnabled && readingOrder == ReadingOrder.OLDEST_FIRST) {
                                 binding.recyclerView.scrollBy(0, Utils.dpToPx(requireContext(), -30))
                             } else binding.recyclerView.scrollToPosition(0)
                         }
                     }
                 }
+
+                // If this insert was because the user clicked "Load more" then ensure the status
```
Lakoja, line 246: Regarding comment / readability: all of this could move to a private method; maybe `ensureLastStatusVisible()`.

And there only the comment about the bug needs to remain.

app/src/main/java/com/keylesspalace/tusky/components/preference/PreferencesFragment.kt
nikclayton: If you don't do that (i.e,. use `since_id`), with this you can get the experience where you click "Load more", the statuses load, but they don't reach the bottom of the gap (i.e., the gap is more than LOAD_AT_ONCE statuses in size).

So then there's a new "Load more" button, and you have to click again. And potentially again.

This code, using `min_id` (and with `OLDEST_FIRST`) means the UX is to click "Load more", then scroll up and read the loaded statuses, eventually encounter another "Load more", click that, and repeat, which seems like the right way to do it (and is requested from the filed issues).

nikclayton: Using `min_id` is also consistent with Mastodon's paging in the `link` headers in HTTP responses.

app/src/main/java/com/keylesspalace/tusky/adapter/PlaceholderViewHolder.kt
```diff
@@ -14,27 +14,55 @@
...
-    fun setup(listener: StatusActionListener, progress: Boolean) {
-        loadMoreButton.visibility = if (progress) View.GONE else View.VISIBLE
-        progressBar.visibility = if (progress) View.VISIBLE else View.GONE
-        loadMoreButton.isEnabled = true
-        loadMoreButton.setOnClickListener { v: View? ->
+        // To allow the user to click anywhere in the layout to load more content set the click
+        // listener on the parent layout instead of loadMoreButton.
+        //
+        // See the comments in item_status_placeholder.xml for more details.
+        itemView.setOnClickListener {
```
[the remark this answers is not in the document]
nikclayton: It's disabled in the loading case (lines 45 and 46) so it's not clickable.

app/src/main/java/com/keylesspalace/tusky/components/preference/PreferencesFragment.kt
Tak: The default usage behavior is to read from newest to oldest - that's why the timeline is in *reverse* chronological order.
In that case, we want to load the *newest* statuses in the gap by default, so that the user who reads from newest to oldest gets an uninterrupted stream of posts. It's possible that there was a bug or otherwise incorrect behavior here in the past, but I believe that today's behavior is correct for this use case

I get that we want to have better support for the vocal minority of users who want to read every post on their timelines from oldest to newest, but we should absolutely not change such a fundamental default behavior on a whim or due to selection bias.

Tak: As a side note, I have felt like I have maybe been coming across as adversarial in some of these discussions, and I've been trying to limit my participation for that reason. *However,* I also feel like there's an abnormally high amount of talking past each other going on, and I don't understand whether it's because one or more of us are fundamentally misunderstanding what the others are saying, or there's some hidden assumption, or something else.

Tak:
I'm trying this now.
If I set "newest first" and I tap the "load more" placeholder, and there are also new posts on the server, I get warped to the very top of my timeline.

app/src/main/java/com/keylesspalace/tusky/components/preference/PreferencesFragment.kt
nikclayton: If it helps, I haven't been interpreting your feedback as adversarial. 

To address what I believe are your key points:

## "The default usage behavior is to read from newest to oldest"

That's not always true, the app is inconsistent about this behaviour.

For example:

1. Check out the current tip of `develop` (eb02d0b at the time I'm writing this)
2. Launch on a device or emulator
3. Open the Federated timeline (just because it changes very quickly and you don't have to wait long)
4. Scroll to the very top, and swipe-to-refresh to make sure you have the most recent statuses
5. Wait maybe 10 seconds, so that new statuses are available
6. Swipe-to-refresh to load new statuses

You should observe that your position in the timeline is retained, and the new statuses are loaded above the one you are currently looking at. This is oldest-first behaviour.

This behaviour is the same on at least the Federated, Local, and Home timelines.

If "the default usage behaviour is to read from newest to oldest" was correct, refreshing any of these timelines should take you to the newest of the freshly loaded statuses.

I suspect this is the source of some of the demand for this behaviour from users. They observe that, in some situations the app behaves the way that they expect, but in others it doesn't.

This also gets weird if there were more than LOAD_AT_ONCE statuses loaded and there's also a "Load more" placeholder. Because if the user clicks that then the statuses are loaded below their current position.

So swipe = load above and keep my position, "Load more" = load below and lose my position

## "vocal minority"

I agree about "vocal". There are at least 14 issues that reference this, many of which either have comments from other users saying they also want the oldest-first behaviour, or heart / thumbs-up emoji from people who haven't commented but also indicating that they want the oldest-first behaviour.

I don't think the assertion that it's a minority is knowable.

Creating a Github issue or commenting on it has a fairly high barrier to entry -- you have to know where to file the issue, you have to create a Github account if you don't already have one, and you have to take the time to actually write the issue.

Participation inequality is a thing (https://www.nngroup.com/articles/participation-inequality/), and I think it's reasonable to assume that for every 1 user who has taken the time to report this issue, there are N other users who either have the same problem, or have abandoned using the app in favour of a different one.

Unfortunately, we don't know what N is. Could be 10, could be 100, could be 1,000, or more, or less.

This also extends to Play reviews. It doesn't appear to be possible to link to individual reviews, so, from the last month or so:

"Tusky experience (if you try to read from oldest to newest): Scrolling up. Tap "Load more". Scroll down to find where you stopped before it loaded newest posts and lost scrolling position. Tap "Load more" again, rushing to scroll down so that you don't lose scrolling progress. "Load more". "Load more". Repeat 15 times [...]" 2 stars, "3 people found this review helpful"

"Looks pretty good. Nice and clean view. However the load more button defaults to going to the top of your feed losing position. Can this be switched or an option included?", 3 stars, "4 people found this review helpful"

"This App is okay for the most cases. But the main feature is complety insane. The timeline is always at the latest twee...äh..toot(?). You can't continue reading the toots from the place were you stopped reading. You always have to start at the newest toot and must search were you left", 2 stars, "25 people found this review helpful"

My intuition from all this is that N is a large number, and that changing the default approach to oldest-first is a net-win for the majority of the userbase.

You, reasonably, disagree.

We can't both be right.

## Proposal

The Tusky Mastodon account has about 15K followers.

Create a poll, run it for 7 days, and boost it once per day for those 7 days. Suggested text:

```
We are thinking of changing the default "Load more" behaviour (https://github.com/tuskyapp/Tusky/pull/3000)

Which do you prefer when you press the "Load more" button?

We'll support both, this is about what the *default* should be.

1. Jump to the newest post (current behaviour)
2. Stay at the oldest post
```

Use the result of that to determine the default.

Tak: "There are behavioral inconsistencies" is not the same as "the default is secretly the opposite of what you think".

I agree that when the user *actively* refreshes, the timeline should move to the top (in the *default* newest-first order).

There's no point in having a poll about the load more behavior, because it should be determined by the reading order, where it doesn't "jump" in either case, but remains at the user's last-read (from the *default* top or optional bottom) post.

nikclayton:
> I'm trying this now. If I set "newest first" and I tap the "load more" placeholder, and there are also new posts on the server, I get warped to the very top of my timeline.

That's because of https://github.com/tuskyapp/Tusky/pull/3000#issuecomment-1345286260 (which I didn't get any feedback on) and https://github.com/tuskyapp/Tusky/pull/3000/commits/1480c6aa3ac5c0c2d362fb271f47ea2259ab14e2 which are trivial to revert if that's not the desired behaviour.

app/src/main/java/com/keylesspalace/tusky/components/preference/PreferencesFragment.kt
nikclayton: > "There are behavioral inconsistencies" is not the same as "the default is secretly the opposite of what you think".

That is true. 

But that's not what I'm saying.

I'm saying that the default behaviour is inconsistent and is a source (probably a significant one) of user confusion and dissatisfaction with the app.

By pointing out the inconsistency, I intend to suggest that when you say "I believe that today's behavior is correct for this use case" I think you may not have considered all the ways that the app behaves.

In response, you suggested "I agree that when the user actively refreshes, the timeline should move to the top (in the default newest-first order)."

That is something that we could do (not in this PR, and I'm not suggesting we actually do it, I'm using "could" to mean only "This is a technically possible change").

But you've also said (about default reading order) "we should absolutely not change such a fundamental default behavior on a whim or due to selection bias."

I agree with that too -- that's why I suggested the poll, so that it's not a whim, or selection bias.

But I feel this is also inconsistent. On one hand you are suggesting that we should *not* change fundamental default behaviour like this, on the other hand you suggest we should change the user's reading position when they swipe-to-refresh, which I think would definitely be a change to fundamental behaviour.

I think that may be why some of this discussion is challenging; it feels like you are arguing two different and opposite positions, depending on the behaviour that you want to see, and it's difficult to address your concerns because of that.

---

Aside: As I've already said, I agree with when you say "absolutely not change such a fundamental default behaviour on a whim or due to selection bias". 

And I absolutely do *not* want to discuss the specific decision here and derail the conversation, but I'll note that has absolutely happened in the past, probably most recently with https://github.com/tuskyapp/Tusky/issues/2552. 

---

Finally:

> There's no point in having a poll about the load more behavior, because it should be determined by the reading order, where it doesn't "jump" in either case, but remains at the user's last-read (from the default top or optional bottom) post.

That's exactly what this PR does (once I revert https://github.com/tuskyapp/Tusky/commit/1480c6aa3ac5c0c2d362fb271f47ea2259ab14e2, see my other reply https://github.com/tuskyapp/Tusky/pull/3000#issuecomment-1369733104 about that).

The only thing the poll is for is decide the default value for the preference, `OLDEST_FIRST` or `NEWEST_FIRST`.

nikclayton:
> Yeah this probably helps users that want to read oldest first a lot.
> 
> But it should also work in non-cached timelines.

It does, at least in my testing.

To reproduce I:

1. Took the code in this PR
2. Modified `TimelineViewModel.kt::LOAD_AT_ONCE` and changed it from `30` to `3` to reduce the number of statuses before a "Load more" placeholder is inserted
3. Deployed the app
4. Set the preference to OLDEST_FIRST
5. Opened the Federated timeline, and swiped to refresh from the top to fetch the latest statuses
6. Waited 20 seconds
7. Swiped again to refresh, observed a "Load more" placeholder
8. Clicked it
9. Verified that the new statuses are inserted above my current reading position
10. Changed the preference to NEWEST_FIRST
11. Still in the Federated timeline, swiped again to refresh, observed a "Load more" placeholder
12. Clicked it
13. Verified that the new statuses are inserted below my current reading position

Have you seen different behaviour?

Tak:
> I'm trying this now. If I set "newest first" and I tap the "load more" placeholder, and there are also new posts on the server, I get warped to the very top of my timeline.

Oh, and also whenever I post / reply

nikclayton:
> > I'm trying this now. If I set "newest first" and I tap the "load more" placeholder, and there are also new posts on the server, I get warped to the very top of my timeline.
> 
> Oh, and also whenever I post / reply

Thanks, I'll check that. Which commit hash did you build from?

Tak:
I'm still using 5928159c

nikclayton:
> I'm still using [5928159](https://github.com/tuskyapp/Tusky/commit/5928159ce10f2dca2b47b8519b01d70a3cc3f34e)

I can't reproduce this on https://github.com/tuskyapp/Tusky/pull/3000/commits/3f7db5faebc4e77cb768dcd910be3499ef5e8388

For creating new posts:

1. Set reading preference to NEWEST_FIRST
2. Go to the top of my home timeline
3. Swipe to refresh
4. Scroll down a few screens
5. Click the FAB to create a new post
6. Set the post status to "unlisted"
7. Enter some text
8. Click post
9. I'm returned to the home timeline at the same position I left (i.e., I am not scrolled to the top).

For replies:

1. Set reading preference to NEWEST_FIRST
2. Go to the top of my home timeline
3. Swipe to refresh
4. Scroll down a few screens, find a post to reply to
5. Click the reply button
6. Set the post status to "unlisted"
7. Enter some text
8. Click post
9. I'm returned to the home timeline at the same position I left (i.e., I am not scrolled to the top).

connyduck:
It would be nice to get this merged, I think it is very close.

As far as I can tell without reading all 44 comments, open todos are:

1) Figure out what the default setting should be - Nik suggests to run a poll, Tak and me would prefer if the current behavior was the default.
2) Either make the new setting work for uncached timelines as well or indicate that it is only for the home timeline.

nikclayton:
> 1. Figure out what the default setting should be - Nik suggests to run a poll, Tak and me would prefer if the current behavior was the default.

Up to you. I've provided all the evidence I have that either it should be the default, or a poll is the best way to determine the default.

> 2. Either make the new setting work for uncached timelines as well or indicate that it is only for the home timeline.

It mostly does already (tested with the local and federated timelines).

The difference is whether you (reading oldest first):

- Click "Load more", the oldest of the new statuses are loaded, you can read up until the next "Load more" (home timeline), or
- Click "Load more", the newest of the new statuses are loaded, if there are more to be loaded you have to immediately click "Load more" again (network timelines)

connyduck:
Ok, let's please do the following: 
1) resolve conflicts
2) make newest first the default
3) merge
4) wait for user feedback and maybe make adjustments

foss-:
Does this mean when reading timeline from old to new and tapping load more, users would end up at the newest post in display?

nikclayton:
> Does this mean when reading timeline from old to new and tapping load more, users would end up at the newest post in display?

No. When reading oldest to newest, and clicking "Load more", you remain at the oldest post you've read.

nikclayton:
> Ok, let's please do the following:
> 
> 1. resolve conflicts
> 2. make newest first the default
> 3. merge
> 4. wait for user feedback and maybe make adjustments

1 and 2, done.

connyduck (approved):

Tak:
When I'm reading newest-first and I *pull* to refresh, I think the timeline should scroll to the top

Issue #2983, opened 2022-12-04
Image descriptions are displayed as white text on dark background in light mode


magicalskye:
Like many other people with a variety of visual disabilities, I have to use apps in light mode because I am unable to read light text on dark background. For this same reason I also depend on the ability to read the image description if someone posts a screenshot of text in a dark mode app. Currently tusky offers no way for me to read the image description at all, because I cannot read light text on dark background. 

* * * *
- Tusky Version: 20.0
- Android Version: 12
- Android Device: Galaxy Tab A8
- Mastodon instance (if applicable):

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

nikclayton:
Thanks for the report. PR https://github.com/tuskyapp/Tusky/pull/3003 is in progress to fix this.

PR #3003, opened 2022-12-05
Use light/dark mode colors for image description text

nikclayton:
This is an accessibility issue -- in light mode (which should have dark text on a light background) the text color was hardcoded to light grey and the background color was a semi-transparent black.

Fixes https://github.com/tuskyapp/Tusky/issues/2983.


Tak (approved):

app/src/main/res/drawable/ic_drag_indicator_horiz_24dp.xml
```diff
@@ -4,6 +4,6 @@
     <size
         android:width="8dp"
         android:height="1dp" />
-    <solid android:color="?attr/colorBackgroundAccent" />
+    <solid android:color="?android:textColorPrimary" />
```
connyduck, line 7:
```suggestion
    <solid android:color="?android:textColorTertiary" />
```

nikclayton:
Good to go I think.

I'll take a look at colours in the app separately. A very quick check suggests that not all of the Material3 colour tokens are defined, which is going to lead to weirdness down the line as more widgets are used. In this one, for example, there is a bottom sheet component / style that could be used for the drag handle (for accessibility) and the parent layout (for colours, etc).

Issue #2985, opened 2022-12-05
If screen is short, notification filters cannot be set


mcclure:
Spoke to a user on Mastodon who said she could not change her notification filters in Tusky. She then [posted](https://mastodon.gamedev.place/@aeva/109458598090321229) a screenshot of the filter pane in which the "apply" button did not appear. What seems to be happening is her screen is not tall enough and Apply gets cut off:

![https://cdn.masto.host/mastodongamedevplace/media_attachments/files/109/458/597/352/468/064/original/8DA5A8CBE7CC365D.JPG]

Probably before "edits" was added this would not have been a problem.

When I told her content was cut off she said if she scrolls down so the top bar disappears, the pane is taller and she sees Apply. But she shouldn't have to do that.

I know her phone model but I need to ask if I can post it on Github.

Expected behavior: ???

nikclayton:
This is reproducible in an emulator based on the Nexus 4 template.

mcclure:
Aeva confirms fixed https://mastodon.social/@aeva@mastodon.gamedev.place/109473657444671100

PR #3004, opened 2022-12-05
Ensure the "Apply" button is always visible

nikclayton:
On smaller devices the notification filter listview may be longer than the screen height, and pushes the "Apply" button out of sight.

Fix this by:

- Set the height of the listview to 0dp and its layout_weight to 1
- Set the layout_weight of the button to 0

This ensures the button always appears (because the listview height is 0dp) and the listview then expands to fill any remaining space (because the layout_weight is 1).

Fixes https://github.com/tuskyapp/Tusky/issues/2985

### DIFF
--- a/app/src/main/res/layout/notifications_filter.xml
+++ b/app/src/main/res/layout/notifications_filter.xml
@@ -6,12 +6,14 @@
     <ListView
         android:id="@+id/listView"
         android:layout_width="match_parent"
-        android:layout_height="wrap_content"/>
+        android:layout_height="0dp"
+        android:layout_weight="1" />
     <Button
         android:id="@+id/buttonApply"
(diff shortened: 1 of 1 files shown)


mcclure:
Cool, one question, will this lead to the listview being cut off (IE maybe people will be unable to turn off edits?) it would be very nonobvious to me looking at this view that it is scrollable (if it is scrollable)

nikclayton:
Here's what it looks like in the Nexus 4 emulator immediately after opening the filter list.

Screenshot_1670260283

Like normal Android listviews, a scrollbar appears as soon as you touch the list.

connyduck (approved):
Excellent, a lot of users complained already.

PR #3005, opened 2022-12-05
Remove rxjava from API calls used by AccountListFragment::fetchAccounts()

nikclayton:


connyduck:
As you are already touching this, would you mind refactoring the whole thing to use AndroidX paging? And give it its own package in `components`?

PR #2935, opened 2022-11-25
Status: Display indicators of edited posts 

fruyek:
Screenshot_20221125-181626

PR #3006, opened 2022-12-05
make date converter null safe

connyduck:
This is needed to convert the "editedAt" attribute of ConversationStatusEntity, without it ConversationsFragment shows the sad elephant frend.
(regression from https://github.com/tuskyapp/Tusky/pull/2935)


charlag (approved):

charlag:
This will box mire, maybe a separate one is better but probably doesn't matter

PR #3008, opened 2022-12-05
Remove rxjava from API calls used by AccountViewModel::changeRelationship()

nikclayton:
The affected API functions are also called from

- ReportViewModel.kt
- SearchViewModel.kt
- AccountListFragment.kt
- SFragment.kt
- TimelineCases.kt

so they have also been updated.


nikclayton:
*Important*: I've marked this as a draft because it's a stacked PR that also contains the changes in https://github.com/tuskyapp/Tusky/pull/3005 since this one also affects AccountListFragment.kt.

I'll un-draft it when https://github.com/tuskyapp/Tusky/pull/3005 is approved and submitted.

nikclayton:
So drop the new Coroutine stuff and name the functions like `muteSync` instead of `muteFromJava`?

I'll hold off on bigger refactors until I've got a good grasp of the code. In particular, `SFragment.java` has this comment that's been there for almost 6 years:

```
/* Note from Andrew on Jan. 22, 2017: This class is a design problem for me, so I left it with an
 * awkward name. TimelineFragment and NotificationFragment have significant overlap but the nature
 * of that is complicated by how they're coupled with Status and Notification and the corresponding
 * adapters. I feel like the profile pages and thread viewer, which I haven't made yet, will also
 * overlap functionality. So, I'm momentarily leaving it and hopefully working on those will clear
 * up what needs to be where. */
```

I'm hoping that a lot of these design problems will disappear if converted to a newer Android architecture (UI -> ViewModel -> Repository -> Data Source), but I'm still investigating that.

connyduck:
SFragment needs to go, but first everything implementing it must be Kotlin. 
It will be changed to Kotlin in https://github.com/tuskyapp/Tusky/pull/2828

app/src/main/java/com/keylesspalace/tusky/usecase/TimelineCases.kt
```diff
@@ -95,30 +94,22 @@ class TimelineCases @Inject constructor(
...
-                    Log.w("Failed to mute account", t)
-                }
-            )
-            .addTo(cancelDisposable)
+    suspend fun mute(statusId: String, notifications: Boolean, duration: Int?) {
+        try {
+            mastodonApi.muteAccount(statusId, notifications, duration)
+            eventHub.dispatch(MuteEvent(statusId))
+        } catch (t: Throwable) {
+            Log.w("Failed to mute account", t)
```
nikclayton, line 102: Ha, no. This retains a bug in the original code, the first parameter should be the log tag.

app/src/main/java/com/keylesspalace/tusky/components/account/AccountViewModel.kt
```diff
@@ -97,37 +99,47 @@ class AccountViewModel @Inject constructor(
...
+            }
         }
     }
 
     fun changeBlockState() {
-        if (relationshipData.value?.data?.blocking == true) {
-            changeRelationship(RelationShipAction.UNBLOCK)
-        } else {
-            changeRelationship(RelationShipAction.BLOCK)
+        viewModelScope.launch {
```
connyduck, line 112: moving the `viewModelScope.launch` into `changeRelationship` would save quite some lines

nikclayton: Done.

connyduck (approved):

PR #3016, opened 2022-12-06
fix crash in TouchDelegateHelper when not all views are available

connyduck:
in conversations there is no boost button

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/util/TouchDelegateHelper.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/util/TouchDelegateHelper.kt
@@ -1,3 +1,18 @@
+/* Copyright 2022 Tusky contributors
+
+ * This file is a part of Tusky.
+ *
+ * This program is free software; you can redistribute it and/or modify it under the terms of the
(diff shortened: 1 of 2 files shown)


connyduck:
cc @evant

charlag (approved):

app/src/main/java/com/keylesspalace/tusky/util/TouchDelegateHelper.kt
```diff
@@ -12,25 +27,26 @@ import android.view.ViewGroup
...
-                val left = children.getOrNull(i - 1)
-                if (left != null) {
-                    // extend half-way to previous view
-                    rect.left -= (view.left - left.right) / 2
-                }
-                val right = children.getOrNull(i + 1)
-                if (right != null) {
-                    // extend half-way to next view
-                    rect.right += (right.left - view.right) / 2
+            children.filterNotNull()
```
evant, line 34: Worried this will mess up the before/after indexing, maybe filter before passing them in?

connyduck: yes, thx!

PR #3042, opened 2022-12-12
Remove rxjava from API calls reblog/favourite/bookmark

nikclayton:
This change requires bridging from Java code to Kotlin `suspend` functions,
by creating wrappers for the `mute` and `block` functions that can be
called from Java and create a coroutine scope.

I've deliberately made this fairly ugly so that it sticks out and can be
removed later.


nikclayton:
This is a stacked PR on top of https://github.com/tuskyapp/Tusky/pull/3008. When that has been merged I'll make this as ready for review.

connyduck:
Let's not start calling Coroutines from Java. More libraries are not helpful, we want to get rid of them. I'll convert the NotificationsFragment to Kotlin soon, let's have a look at this when I'm done.

nikclayton:
OK. But this is blocking other PRs. I can also take a run at converting NotificationsFragment

PR #3046, opened 2022-12-13
Convert util/{HttpHeaderLink,PairedList,TimestampUtils,ThemeUtils} to Kotlin

nikclayton:


app/src/main/java/com/keylesspalace/tusky/util/ThemeUtils.kt
```diff
@@ -0,0 +1,74 @@
...
+import android.util.TypedValue
+import androidx.annotation.AttrRes
+import androidx.annotation.ColorInt
+import androidx.appcompat.app.AppCompatDelegate
+
+/**
+ * Provides runtime compatibility to obtain theme information and re-theme views, especially where
+ * the ability to do so is not supported in resource files.
+ */
+object ThemeUtils {
```
connyduck, line 30: `ThemeUtils`/`TimeStampUtils` shouldn't be objects but top-level functions.

`object` in Kotlin is similar to a singleton pattern in Java, the class will be instantiated and kept around at runtime, which makes no sense if it does not hold any state.
Top-level functions on the other hand are like static functions in Java (=what they were before the conversion) and thus have less overhead. With an `@file:JvmName` annotation you can choose the name for the class that holds the static functions and if you set it to `ThemeUtils`/`TimeStampUtils` no Java callsites have to be updated. See e.g. `LinkHelper` for that pattern.

app/src/main/java/com/keylesspalace/tusky/util/HttpHeaderLink.kt
```diff
@@ -0,0 +1,150 @@
...
+class HttpHeaderLink private constructor(uri: String) {
+    data class Parameter(var name: String? = null, var value: String? = null)
+
+    private val parameters: MutableList<Parameter> = ArrayList()
+
+    var uri: Uri
+
+    init {
+        this.uri = Uri.parse(uri)
+    }
```
connyduck, line 29:
```suggestion
    val uri: Uri = uri.toUri()
```

app/src/main/java/com/keylesspalace/tusky/util/HttpHeaderLink.kt
```diff
@@ -0,0 +1,150 @@
...
+            relationType: String
+        ): HttpHeaderLink? {
+            for (link in links) {
+                for (parameter in link.parameters) {
+                    if (parameter.name == "rel" && parameter.value == relationType) {
+                        return link
+                    }
+                }
+            }
+            return null
```
connyduck, line 147: You can write this in nice idiomatic Kotlin like so:

```suggestion
         return links.find { link ->
                link.parameters.any { parameter ->
                    parameter.name == "rel" && parameter.value == relationType
                }
            }
```

app/src/main/java/com/keylesspalace/tusky/util/HttpHeaderLink.kt
```diff
@@ -0,0 +1,150 @@
...
+            }
+            return -1
+        }
+
+        /**
+         * @param line the entire link header, not including the initial "Link:"
+         * @return all links found in the header
+         */
+        fun parse(line: String?): List<HttpHeaderLink> {
+            val linkList: MutableList<HttpHeaderLink> = ArrayList()
```
connyduck, line 110: if it doesn't have to be an `ArrayList` better use `mutableListOf()`

app/src/main/java/com/keylesspalace/tusky/util/HttpHeaderLink.kt
```diff
@@ -0,0 +1,150 @@
...
+
+import android.net.Uri
+
+/**
+ * Represents one link and its parameters from the link header of an HTTP message.
+ *
+ * @see [RFC5988](https://tools.ietf.org/html/rfc5988)
+ */
+class HttpHeaderLink private constructor(uri: String) {
+    data class Parameter(var name: String? = null, var value: String? = null)
```
connyduck, line 21: I am sure you can make these both `val` if you restructure the code a bit

app/src/main/java/com/keylesspalace/tusky/util/HttpHeaderLink.kt
```diff
@@ -0,0 +1,150 @@
+/* Written in 2017 by Andrew Dawson
+ *
+ * To the extent possible under law, the author(s) have dedicated all copyright and related and
+ * neighboring rights to this software to the public domain worldwide. This software is distributed
+ * without any warranty.
+ *
+ * You should have received a copy of the CC0 Public Domain Dedication along with this software.
```
connyduck, line 7: uhm that is not our license

app/src/main/java/com/keylesspalace/tusky/util/HttpHeaderLink.kt
```diff
@@ -0,0 +1,150 @@
...
+            return -1
+        }
+
+        private fun parseValue(line: String, start: Int): ValueResult {
+            val result = ValueResult()
+            val foundIndex = line.indexOfAny(charArrayOf(';', ',', '"'), start, false)
+            if (foundIndex == -1) {
+                result.value = line.substring(start)
+                return result
+            }
```
connyduck, line 64: How about making `value` a constructor param of `ValueResult` and then something like

```suggestion
            val foundIndex = line.indexOfAny(charArrayOf(';', ',', '"'), start, false)
            if (foundIndex == -1) {
                return ValueResult(line.substring(start))
            }
```

app/src/test/java/com/keylesspalace/tusky/util/HttpHeaderLinkTest.kt
```diff
@@ -0,0 +1,83 @@
...
+
+@Config(sdk = [28])
+@RunWith(AndroidJUnit4::class)
+class HttpHeaderLinkTest {
+    data class TestData(val name: String, val input: String, val want: List<HttpHeaderLink>)
+
+    @Test
+    fun shouldParseValidLinks() {
+        // Constructor is private, so call through reflection
+        val constructor = HttpHeaderLink::class.primaryConstructor!!
```
connyduck, line 19: Usually we just make public what we need and annotate it with `@VisibleForTesting`

app/src/main/java/com/keylesspalace/tusky/util/PairedList.kt
```diff
@@ -0,0 +1,78 @@
...
+    fun getPairedItem(index: Int): V {
+        return synced[index]
+    }
+
+    fun getPairedItemOrNull(index: Int): V? {
+        return if (index >= 0 && index < synced.size) {
+            synced[index]
+        } else {
+            null
+        }
```
connyduck, line 45:
```suggestion
        return synced.getOrNull(index)
```

connyduck (changes requested):

app/src/main/java/com/keylesspalace/tusky/util/HttpHeaderLink.kt
nikclayton: It's the license on the most recent version of the code, https://github.com/tuskyapp/Tusky/blob/7a2ee150788ed1e1353214f125c418a9eae6969e/app/src/main/java/com/keylesspalace/tusky/util/HttpHeaderLink.java.

I can change it if you think that's appropriate.

connyduck: Probably an artifact of the time when Vavassor tried to do a license switch, or maybe he copied the code from somewhere else. 
Please change it to 2022 Tusky contributors and GPL

nikclayton:
Thanks for the helpful feedback. Changes made.

I removed `ThemeUtils.getColors` as well, and replaced that with the one that's already in the `MaterialColors` library.

connyduck (approved):

PR #3075, opened 2022-12-19
Remove debug logging of keypresses

nikclayton:
While only present in a Log.d call, which should be stripped out of release builds, this feels like an unnecessary privacy risk.


nikclayton:
If some details of the event are needed they should instead be logged without including the actual characters being typed.

PR #3090, opened 2022-12-22
Handle preference fragments using the framework

nikclayton:
The previous code started new preference "screens" as activities, even though each one hosted a single fragment.

Modify this to use the framework's support for swapping in/out different preference fragments.

PreferencesActivity:
- Remove the code for launching tab and proxy preferences
- Remove the code for setting titles, each fragment is responsible for that
- Implement `OnPreferenceStartFragmentCallback` to swap fragments in/out with the correct animation

PreferencesFragment:
- Use `fragment` property instead of `setOnPreferenceClickListener`
- Set the activity title when resuming

Everything else:
- Set the activity title when resuming


nikclayton:
I've got a WIP that moves the UI preferences to their own fragment and shows a preview of the look and feel of a status as the user makes preference changes.

I did some refactoring to make that work easier, and split the refactoring in to this PR so it can be reviewed without the more complicated work getting in the way.

app/src/main/java/com/keylesspalace/tusky/components/preference/PreferencesActivity.kt
```diff
@@ -81,27 +84,41 @@ class PreferencesActivity :
...
+        pref: Preference
+    ): Boolean {
+        val args = pref.extras
+        val fragment = supportFragmentManager.fragmentFactory.instantiate(
+            classLoader,
+            pref.fragment!!
+        )
+        fragment.arguments = args
+        fragment.setTargetFragment(caller, 0)
+        supportFragmentManager.beginTransaction()
```
connyduck, line 109: You can make this nicer with the `commit` extension function. https://developer.android.com/kotlin/ktx#fragment

nikclayton: Done.

PR #3097, opened 2022-12-23
Revert removal of Federated timeline from defaults

Doomsdayrs:
Due to the removal, people have begun to stop using tusky and or finding another app to use that has a federated timeline. And adding the federated timeline back is extremely difficult and frustrating for uses.

This reverts the removal of the federated timeline, by adding it back to the defaults, resulting in a 5 button navigation bar.


connyduck:
I'd rather not, for the reasons given in https://github.com/tuskyapp/Tusky/issues/2552

> extremely difficult and frustrating

I doubt that

Doomsdayrs:
I shall now post links of users having a difficult time

Doomsdayrs:
https://beige.party/@theropologist/109533331556426553

Issue #2705, opened 2022-09-21
Edit scheduled post then back out = broken UI


mcclure:
Create a scheduled post then edit it. If you click "toot" to save your changes, it will back you out to the Scheduled Posts listing and the pencil and X buttons are again available to edit the post again.

However, if you create a scheduled post then hit "back", it will ask if you want to save your draft (see #2703). Regardless of what you select, you will return to the scheduled posts listing **which will then be partially broken**. Clicking the pencil again will have no effect. (Clicking the "X" will delete the post as expected.)

Expected behavior: Either

1. Pencil icon should work upon backing out of scheduled post edit or
2. Backing out should take you all the way back to the main view and not show the scheduled post listing.

* * * *
- Tusky Version: Self built develop branch 7684f06938
- Android Version: 12
- Android Device: Sony Xperia 5 III USA
- Mastodon instance (if applicable): Mastodon dot social

nikclayton:
Looking in to this.

> Create a scheduled post then edit it. If you click "toot" to save your changes, it will back you out to the Scheduled Posts listing and the pencil and X buttons are again available to edit the post again.

That's not quite right. If you do this enough times (more than 1 at least, but it varies when I try it) the buttons stop working.

No idea why yet, mind, but I'll keep looking.

nikclayton:
Here's the problem (I think, it's late, and I'm writing this down so I can review it later)

When the edit button is clicked, it disables itself (https://github.com/tuskyapp/Tusky/blob/develop/app/src/main/java/com/keylesspalace/tusky/components/scheduled/ScheduledStatusAdapter.kt#L57).

Editing a scheduled status is done by deleting the old status (https://github.com/tuskyapp/Tusky/blob/develop/app/src/main/java/com/keylesspalace/tusky/components/compose/ComposeViewModel.kt#L289) and sending a new one.

The new one gets a new ID, so it's a new entry in the adapter, and its edit button is enabled.

If you press Back this doesn't happen -- the original scheduled status is still there, but its edit button is not reenabled.

PR #3100, opened 2022-12-23
Leave the "edit scheduled status" button enabled after clicking

nikclayton:
If the user submits an edit to the scheduled status then this one will be deleted, the paging source will notice, the adapter will be notified in the normal way, and this binding will be reused.

Or the user backs out of the edit, and this adapter entry is still valid and should remain clickable.

Fixes https://github.com/tuskyapp/Tusky/issues/2705

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/components/scheduled/ScheduledStatusAdapter.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/components/scheduled/ScheduledStatusAdapter.kt
@@ -16,7 +16,6 @@
 package com.keylesspalace.tusky.components.scheduled
 
 import android.view.LayoutInflater
-import android.view.View
 import android.view.ViewGroup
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/keylesspalace/tusky/components/scheduled/ScheduledStatusAdapter.kt
```diff
@@ -54,7 +54,6 @@ class ScheduledStatusAdapter(
             holder.binding.delete.isEnabled = true
             holder.binding.text.text = item.params.text
             holder.binding.edit.setOnClickListener { v: View ->
```
nikclayton, line 56:
```suggestion
            holder.binding.edit.setOnClickListener {
```

Remove unnecessary parameter.

connyduck (approved):

Issue #2704, opened 2022-09-21
Confusing UI wording when editing scheduled post


mcclure:
Tried editing a scheduled post and the language confused me.

**Observed behavior:**

Wrote a post. Tapped "schedule". Scheduled for 1 week. Went into left side bar, scheduled posts, clicked the pencil. Changed the text. Backed out.

[<img width="300" src="https://user-images.githubusercontent.com/277318/191592117-8a27daa1-4682-46ab-9f1f-42eb747ca6e9.png"  />](https://user-images.githubusercontent.com/277318/191592117-8a27daa1-4682-46ab-9f1f-42eb747ca6e9.png) [<img src="https://user-images.githubusercontent.com/277318/191592074-4b8491c5-4fe6-4ca2-8826-78d9a12fed5c.png" width="300" />](https://user-images.githubusercontent.com/277318/191592074-4b8491c5-4fe6-4ca2-8826-78d9a12fed5c.png)

The post button read "toot". The back out dialog asked if I wanted to "Save a draft" and said "Delete" and "Save".

**Problems:**

Minor: Button saying "Toot" would be more accurate saying "Update" or "Save"

Major: Back out dialog is inaccurate. It's really asking if you want to save your *changes* to the scheduled tweet. Moreover, **neither button does as it says.** "Delete" does not delete and "Save" does not save. In either case you return to the scheduled posts listing with your changes discarded.

**Expected behavior:** In addition to maybe altering the "Toot" button, the dialog should reference a scheduled post rather than a draft. The buttons should be changed in any of the following ways:

1. Text should ask if you want to save changes to scheduled post. Buttons should say "Discard" and "Save"; clicking one should discard changes and the other should save them.

2. Text should ask if you want to save changes to your scheduled post. Buttons should allow continuing to edit the post and discarding changes.

3. Buttons can stay the same but the functionality really should either delete or save the post. (This is my least preferred option personally but it would match the behavior with drafts.)

* * * *
- Tusky Version: Self built develop branch 7684f06938
- Android Version: 12
- Android Device: Sony Xperia 5 III USA
- Mastodon instance (if applicable): Mastodon dot social

mcclure:
More small thoughts:
- The "TOOT!" button is maybe correct because if you remove the schedule from the scheduled tweet it does just become a normal toot that you are tooting. But it still confused me a little.
[...]

PR #3103, opened 2022-12-24
Fix saving changes to statuses when editing

nikclayton:
With the previous code backing out of a status editing operation where changes had been made  (whether it was editing an existing status, a scheduled status, or a draft) would prompt the user to save the changes as a new draft.

See https://github.com/tuskyapp/Tusky/issues/2704 and https://github.com/tuskyapp/Tusky/issues/2705 for more detail.

The fix:

- Create an enum to represent the four different kinds of edits that can happen
  - Editing a new status (i.e., composing it for the first time)
  - Editing a posted status
  - Editing a draft
  - Editing a scheduled status

- Store this in ComposeOptions, and set it appropriately everywhere ComposeOptions is created.

- Check the edit kind when backing out of ComposeActivity, and use this to show one of three different dialogs as appropriate so the user can:
  - Save as new draft or discard changes
  - Continue editing or discard changes
  - Update existing draft or discard changes

Also fix ComposeViewModel.didChange(), which erroneously reported false if the old text started with the new text (e.g., if the old text was "hello, world" and it was edited to "hello", didChange() would not consider that to be a change).

Fixes https://github.com/tuskyapp/Tusky/issues/2704, https://github.com/tuskyapp/Tusky/issues/2705


app/src/main/java/com/keylesspalace/tusky/components/compose/ComposeViewModel.kt
```diff
@@ -211,15 +213,8 @@ class ComposeViewModel @Inject constructor(
...
-        val textChanged = !(
-            content.isNullOrEmpty() ||
-                startingText?.startsWith(content.toString()) ?: false
-            )
-
-        val contentWarningChanged = showContentWarning.value &&
-            !contentWarning.isNullOrEmpty() &&
-            !startingContentWarning.startsWith(contentWarning.toString())
+        val textChanged = (content ?: "") != (startingText ?: "")
+        val contentWarningChanged = (contentWarning ?: "") != (startingContentWarning ?: "")
```
connyduck, line 217: There is an extension function `orEmpty()` which is nicer than `?: ""` imho

connyduck (changes requested):
Nice improvement, please resolve the merge conflicts

app/src/main/java/com/keylesspalace/tusky/components/compose/ComposeViewModel.kt
nikclayton: Done.

connyduck (approved):

Issue #134, opened 2017-04-12
[Feature Request] translation feature 


ebrulato:
A button to translate a message in the user's language

alerque:
Personally I find the built in "translate this" on social networks to be less useful than it should be. Even the big name folks like Twitter and Facebook don't use the best engines for each possible language. I find an external app often works better, but to make that a meaningful experience it can't be more that a tap away. For programs that have tap-to-copy features this can be combined with Google Translate's clipboard monitoring and auto translation. This brings a translation in 2 or 3 taps depending on the app. Other apps that can read content in on shares would be 4-5 taps away if #147 were to be fixed.

That's not to say I'd be apposed to an internal translator that used some API and did it in place, but it should be optional and the UI should support getting the text out to external apps as well.

ebrulato:
I agree with you @alerque and the purpose is just to add a button to get a translation :)

nailyk-fr:
Useless comment to subscribe. 
FI some projects use [weblate](https://weblate.org/en/). Fully open source.

shoukolate:
Add an option in settings to auto translate all toots or an option to translate individual toots. You can use google translate API (free for unlimited use AFAIK) from google translate extension from chrome browser (someone posted it on stackoverflow)

API: https://translate.googleapis.com/translate_a/single?client=gtx&sl=auto&tl=en&dt=t&q=こんにちは

You can replace "auto" if with source language if you want to disable auto detection and "en" for choosing a target language. Replace "こんにちは" with text you want to translate.

JSON response:

`[[["Hello","こんにちは",null,null,1]
]
,null,"ja",null,null,null,1.0,[]
,[["ja"]
,null,[1.0]
,["ja"]
]
]
`

suoko:
Now mastodon has added translation as a feature

ShellWen:
I need it😭

ozkanpakdil:
I manage to translate the text in the post but could not find a way to replace the text in the post but open a new dialog to show the translation, you can see it here [video](https://photos.app.goo.gl/TjSd3WHU3eG5sGYS8)

https://user-images.githubusercontent.com/604405/209661508-f699df43-4bb2-4976-9aaf-8f5a82b14308.mp4

s0mt0chukwu:
This needs to be addressed in the next update

alerque:
[...]

PR #3107, opened 2022-12-27
#134 I could not find a way to replace the text in the post but I ope…

ozkanpakdil:
…n an alert dialog to show translation. can someone please help me to set the translation in the main post?


connyduck:
Sorry, but no. We will implement translation via the Mastodon api.

PR #3108, opened 2022-12-27
Provide default text sizes in TuskyBaseTheme

nikclayton:
These aren't necessary for the app, and are overwritten with the actual style in `BaseActivity.onCreate()`.

But if they're missing the Android Studio layout preview renderer crashes.


connyduck (approved):
Ah thats why it crashes, thx 😅

Issue #3086, opened 2022-12-21
No information about loading status from TalkBack


jsoref:
The spinner for when I activate an item on a thread isn't announced and isn't reachable. I have no idea what is happening or why. I also presume I won't be told when the spinner is done. 
 
![Screenshot with TalkBack active (the green box is its focal point) and the TalkBack hover widget on the right side of the screen. The spinner isn't reachable](https://user-images.githubusercontent.com/2119212/208928005-f862ca20-6a79-447f-a7b3-cb0479b3c421.png)

* * * *
- Tusky Version: Tusky Test 20.0-6aed1c68
- Android Version: 12
- Android Device: Pixel 3A
- Mastodon instance (if applicable):

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

PR #3118, opened 2022-12-28
Improve the actual and perceived speed of thread loading

nikclayton:
To improve the actual speed, note that if the user has opened a thread from their home timeline then the initial status is cached in the database. Other statuses in the same thread may be cached as well.

So try and load the initial status from the database, falling back to the network if it's not present (e.g., the user has opened a thread from the local or federated timelines, or a search).

Introduce a new loading state to deal with this case.

In typical cases this allows the UI to display the initial status immediately with no need to show a progress indicator.

To improve the perceived speed, delay showing the initial loading circular progress indicators by 500ms. If loading the initial status completes within that time no spinner is shown and the user will perceive the action as close-to-immediate
(https://www.nngroup.com/articles/response-times-3-important-limits/).

Additionally, introduce an extra indeterminate progress indicator.

The new indicator is linear, anchored to the bottom of the screen, and shows progress loading ancestor/descendant statuses. Like the other indicator is also delayed 500ms from when ancestor/descendant status information is fetched, and if the fetch completes in that time it will not be shown.

While I'm here, add accessiblity info for Talkback, fixes https://github.com/tuskyapp/Tusky/issues/3086


nikclayton:
Here's a before and after. In both videos, a `delay(3000)` has been added to the Mastodon API calls to make the effect more apparent.

Before: https://www.youtube.com/shorts/VriodxHqJpo

After: https://www.youtube.com/watch?v=UVAKla8xejw (first 2 seconds are blank due to emulator weirdness)

nikclayton:
The "Launch a coroutine, wait N ms, show a view, hide the view when the coroutine is cancelled" trick can probably be used everywhere progress indicators are shown. After this is merged I can bring the same functionality everywhere in the app to improve perceived speed.

app/src/main/java/com/keylesspalace/tusky/db/TimelineDao.kt
```diff
@@ -53,6 +53,28 @@ ORDER BY LENGTH(s.serverId) DESC, s.serverId DESC"""
...
+rb.serverId as 'rb_serverId', rb.timelineUserId 'rb_timelineUserId',
+rb.localUsername as 'rb_localUsername', rb.username as 'rb_username',
+rb.displayName as 'rb_displayName', rb.url as 'rb_url', rb.avatar as 'rb_avatar',
+rb.emojis as 'rb_emojis', rb.bot as 'rb_bot'
+FROM TimelineStatusEntity s
+LEFT JOIN TimelineAccountEntity a ON (s.timelineUserId = a.timelineUserId AND s.authorServerId = a.serverId)
+LEFT JOIN TimelineAccountEntity rb ON (s.timelineUserId = rb.timelineUserId AND s.reblogAccountId = rb.serverId)
+WHERE s.serverId = :statusId OR s.reblogServerId = :statusId"""
+    )
+    abstract fun getStatus(statusId: String): TimelineStatusWithAccount?
```
connyduck, line 76: `suspend fun` or it will run on the main thread

app/src/main/java/com/keylesspalace/tusky/components/viewthread/ViewThreadFragment.kt
```diff
@@ -210,6 +237,28 @@ class ViewThreadFragment : SFragment(), OnRefreshListener, StatusActionListener,
...
+     * Create a job to implement a delayed-visible progress bar.
+     *
+     * Delaying the visibility of the progress bar can improve user perception of UI speed because
+     * fewer UI elements are appearing and disappearing.
+     *
+     * When started the job will wait `delayMs` then show `view`. If the job is cancelled at
+     * any time `view` is hidden.
+     */
+    @CheckResult()
+    private fun getProgressBarJob(view: View, delayMs: Long) = viewLifecycleOwner.lifecycleScope.launch(
```
connyduck, line 250: very nice

app/src/main/java/com/keylesspalace/tusky/components/viewthread/ViewThreadViewModel.kt
```diff
@@ -95,36 +98,49 @@ class ViewThreadViewModel @Inject constructor(
...
-            val statusResult = statusCall.await()
+            val timelineStatus = db.timelineDao().getStatus(id)
 
-            val status = statusResult.getOrElse { exception ->
-                _uiState.value = ThreadUiState.Error(exception)
-                return@launch
+            val detailedStatus = if (timelineStatus != null) {
+                Log.d(TAG, "Loaded status from local timeline")
+                val viewDatum = timelineStatus.toViewData(gson) as StatusViewData.Concrete
+                viewDatum.copy(isDetailed = true)
```
connyduck, line 111: instead of copying, pass the isDetailed param down to the constructor?

app/src/main/java/com/keylesspalace/tusky/components/viewthread/ViewThreadViewModel.kt
```diff
@@ -95,36 +98,49 @@ class ViewThreadViewModel @Inject constructor(
...
+            } else {
+                Log.d(TAG, "Loaded status from network")
+                val statusCall = async { api.statusAsync(id) }
+                val statusResult = statusCall.await()
+                val result = statusResult.getOrElse { exception ->
+                    _uiState.value = ThreadUiState.Error(exception)
+                    return@launch
+                }
+                result.toViewData(true)
             }
```
connyduck, line 121: Status from database should only be used to have something to display faster. Even if it is in the database we need to load it from the network to have up to date state (fav/boost count, edited, was it deleted maybe?). After loading it from network it should be saved to the database again so it has the newest state when opening it again afterwards.

connyduck (changes requested):
YES I love the idea 😍

app/src/main/java/com/keylesspalace/tusky/components/viewthread/ViewThreadViewModel.kt
nikclayton: You mean like `timelineStatus.toViewData(gson, isDetailed = true)`, and modify `toViewData` to take the extra parameter?

Or something else?

connyduck: yes exactly, saves one object allocation

app/src/main/java/com/keylesspalace/tusky/db/TimelineDao.kt
nikclayton: Will do.

That's reminded me of something -- `getStatuses` is not marked `suspend`. Is it possible that's the cause of the ANRs on startup you mentioned seeing?

app/src/main/java/com/keylesspalace/tusky/components/viewthread/ViewThreadViewModel.kt
nikclayton: Are you sure?

By analogy, suppose I've refreshed the home timeline and am scrolling to read the newly loaded statuses.

Your suggestion is like saying that when a new status is scrolled in to view there should be a background refresh of its content just in case it's been edited or deleted in the time that elapsed between the initial refresh and the user scrolling to it.

If the user wants a full refresh of a thread they can swipe-to-refresh (I'll add a menu option for that too as part of https://github.com/tuskyapp/Tusky/pull/3121).

So wouldn't it be better to make the choice to fetch from the network even if it's in the database if `_uiState.value == ThreadUiState.Refreshing`? That's true if the user is refreshing, false otherwise.

app/src/main/java/com/keylesspalace/tusky/db/TimelineDao.kt
nikclayton: Done.

app/src/main/java/com/keylesspalace/tusky/components/viewthread/ViewThreadViewModel.kt
nikclayton: Done.

app/src/main/java/com/keylesspalace/tusky/components/viewthread/ViewThreadViewModel.kt
connyduck: Yes I think it should refresh when opening the details.
Mostly because I'm used to seeing the most recent information when opening the details. I'm sure other long term Tusky users are as well. Having to manually refresh when it used to be automatic feels weird, especially if replies are always refreshed, why wouldn't the main content.

> Your suggestion is like saying that when a new status is scrolled in to view there should be a background refresh 

No, scolling past a already loaded status and actively clicking on it are two different things.

connyduck:
There are some very weird change animations going on now. Very irritating jumps when opening or reloading.


https://user-images.githubusercontent.com/10157047/210865457-e007b527-afa7-4676-aee9-1540462f3a06.mp4

nikclayton:
There's one bug:

```
                    val viewData = api.status(id).getOrNull()?.toViewData() ?: return@fold
```

should be

```
                    val viewData = api.status(id).getOrNull()?.toViewData(isDetailed = true) ?: return@fold
```

That fixes part of it. But otherwise, some shifts are to be expected, for example if the status has ancestors it will first appear at the top of the screen, then when the ancestors are loaded it will be moved down the screen.

Others are because there are diffs in the status locally and remote. For example, if the reblog count goes from 6 to 7 this is not visible in the UI, but it means the two objects are not equal, and when the adapter compares them it removes the old status from the list (triggering the animation), and adds the new one (which, to the user, is identical) triggering another animation.

That *might* be fixable in the comparison logic in `ThreadAdapter` and elsewhere. I'll take a look.

It would, of course, be much simpler to drop this requirement.

nikclayton:
That should fix the UI weirdness you saw, and some flickering in the UI from my tests using different threads that I've bookmarked.

connyduck (approved):

Issue #3093, opened 2022-12-22
Feature request: Copy/share user profile/address


mcclure:
Say I find a user I think is pretty neat. I go to their profile and it says:

> **Awesomeuser**
> @awesomeuser@example.com

I want to tell a friend about this user over IM. But if I am in Tusky, there is no way to do any of the following:
- Share a link to the profile / copy a link to the profile into the clipboard
- Copy the username into the clipboard
- Copy the Mastodon address into the clipboard

It should be possible to do all three of these things, probably via a submenu in the ⋮ menu like we have for posts.

* * * *
- Tusky Version: 20.0
- Android Version: 12
- Android Device: Sony XPeria 5 III
- Mastodon instance (if applicable): n/a but mastodon.social

- [X] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

In fact variants HAVE been filed twice before, #2955 and #2824. We can get all three at once

mcclure:
Resolved by PR #3120

PR #3120, opened 2022-12-29
Share and copy menu items for account page

mcclure:
Attempting to address #3093 (and maybe some of the semi-overlap bugs linked at the bottom of 3093). Here's what I have currently:

[](https://files.mastodon.social/media_attachments/files/109/594/372/136/707/170/original/7db50a0dde776e97.png) [](https://files.mastodon.social/media_attachments/files/109/594/373/215/574/008/original/83f72331a773c14c.png)

And what you see works but there are Problems and I need Advice.

1. What should the wording be (IE, should "handle" be "handle"? or "address"? Or "username"? or something else? I think the term we want is "Fully-Qualified Username" but that's no good for). Should it be "Share link to account" or just "Share link"? or "share account link" or… I'm just not sure about any of these strings.

2. Although I like having both "Share" and "Copy" items, this menu is very long now and I think it is just too much. I **think** I'm going to change it so:

    * Only one added item in menu ("Share")
    * Long-press on the username/displayname copies the Fully Qualified Username to the clipboard

    Any thoughts?

3. "Share handle"/"Copy handle" works as expected when copying handles from other instances. However if copying handle from your home instance it will be missing the domain name. In AccountActivity I have access to a struct Account with three fields:

            val localUsername: String,
            val username: String,
            val displayName: String?, 

    displayName is as obvious, localUsername I think (???) always contains just the non-domain part of the username. "username" is the one I'm using, and I think it's the username displayed in the actual GUI. This username does not show the domain name for your home instance, which is correct in that case, but not correct for copying the username to share. I need to do one of the following:

        * Get hold of a AccountEntity from inside AccountActivity
        * Figure out where Account is being constructed and add a fullName field copying AccountEntity.fullName (or a domain field copying AccountEntity.domain, which I can combine with localUsername)

    But I can't figure it out. Help?

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/components/account/AccountActivity.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/components/account/AccountActivity.kt
@@ -16,8 +16,6 @@
 package com.keylesspalace.tusky.components.account
 
 import android.animation.ArgbEvaluator
-import android.content.ClipData
-import android.content.ClipboardManager
 import android.content.Context
 import android.content.Intent
 import android.content.res.ColorStateList
@@ -868,7 +866,7 @@ class AccountActivity : BottomSheetActivity(), ActionButtonActivity, HasAndroidI
                 }
                 return true
             }
-            R.id.action_share_account_handle -> {
+            R.id.action_share_account_username -> {
                 // If the account isn't loaded yet, eat the input.
                 if (loadedAccount != null) {
                     val domain = getDomain(loadedAccount?.url)
@@ -878,27 +876,7 @@ class AccountActivity : BottomSheetActivity(), ActionButtonActivity, HasAndroidI
                     sendIntent.action = Intent.ACTION_SEND
                     sendIntent.putExtra(Intent.EXTRA_TEXT, username)
                     sendIntent.type = "text/plain"
-                    startActivity(Intent.createChooser(sendIntent, resources.getText(R.string.send_account_handle_to)))
-                }
-                return true
-            }
-            R.id.action_copy_account_link -> {
-                // If the account isn't loaded yet, eat the input.
-                if (loadedAccount != null) {
-                    val url = loadedAccount!!.url
-                    val clipboard = getSystemService(Context.CLIPBOARD_SERVICE) as ClipboardManager
-                    clipboard.setPrimaryClip(ClipData.newPlainText(null, url))
-                }
-                return true
-            }
-            R.id.action_copy_account_handle -> {
-                // If the account isn't loaded yet, eat the input.
-                if (loadedAccount != null) {
-                    val domain = getDomain(loadedAccount?.url)
-                    val localUsername = loadedAccount!!.username
-                    val username = "@$localUsername@$domain"
-                    val clipboard = getSystemService(Context.CLIPBOARD_SERVICE) as ClipboardManager
-                    clipboard.setPrimaryClip(ClipData.newPlainText(null, username))
+                    startActivity(Intent.createChooser(sendIntent, resources.getText(R.string.send_account_username_to)))
                 }
                 return true
             }
--- a/app/src/main/java/com/keylesspalace/tusky/components/account/AccountActivity.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/components/account/AccountActivity.kt
@@ -28,7 +28,6 @@ import android.view.Menu
 import android.view.MenuItem
 import android.view.View
 import android.view.ViewGroup
-import android.widget.Toast
 import androidx.activity.viewModels
 import androidx.annotation.ColorInt
 import androidx.annotation.Px
@@ -417,7 +416,8 @@ class AccountActivity : BottomSheetActivity(), ActionButtonActivity, HasAndroidI
                     val fullUsername = loadedAccount!!.fullUsername
                     val clipboard = getSystemService(Context.CLIPBOARD_SERVICE) as ClipboardManager
                     clipboard.setPrimaryClip(ClipData.newPlainText(null, fullUsername))
-                    Toast.makeText(this, getString(R.string.account_username_copied), Toast.LENGTH_LONG).show()
+                    Snackbar.make(binding.root, getString(R.string.account_username_copied), Snackbar.LENGTH_SHORT)
+                        .show()
                 }
                 true
             }
--- a/app/src/main/java/com/keylesspalace/tusky/components/account/AccountActivity.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/components/account/AccountActivity.kt
@@ -16,6 +16,8 @@
 package com.keylesspalace.tusky.components.account
 
 import android.animation.ArgbEvaluator
+import android.content.ClipData
+import android.content.ClipboardManager
 import android.content.Context
 import android.content.Intent
 import android.content.res.ColorStateList
@@ -26,6 +28,7 @@ import android.view.Menu
 import android.view.MenuItem
 import android.view.View
 import android.view.ViewGroup
+import android.widget.Toast
 import androidx.activity.viewModels
 import androidx.annotation.ColorInt
 import androidx.annotation.Px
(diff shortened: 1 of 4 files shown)


nikclayton:
1. Mastodon calls it "username" (e.g., when signing in) so I'd go with that instead of anything else like "handle"
2. Strong agree with reducing number of menu options and copying on long-press
3. I'll take a look later if that helps

nikclayton:
Oh, for 3, maybe look at whatever code populates "mastodon.social" in the "Mute mastodon.social" entry in your screenshot, and do that?

mcclure:
> Oh, for 3, maybe look at whatever code populates "mastodon.social" in the "Mute mastodon.social" entry in your screenshot, and do that?

```
domain = getDomain(loadedAccount?.url)
```

Well, that… doesn't seem perf-optimal, but it *works*!

mcclure:
Probably ready to merge now. Copy items removed from menu, longpress-on-username shortcut works.

The one thing I'm not sure about, it only works if you longpress ON the username. If you longpress on the blank space to the right of the username it does not work. This is usually not a problem but for example my profile is just
```
mcc
@mcc
```
When viewed from mastodon dot social so that's a pretty narrow space.

I think the best way to fix this would be to give the displayname/username labels some invisible margins so their interactive area extends some fixed distance to the right but I do not know where to begin with that.

app/src/main/java/com/keylesspalace/tusky/entity/Account.kt
```diff
@@ -45,6 +46,13 @@ data class Account(
             localUsername
         } else displayName
 
+    val fullUsername: String
+        get() {
+            val domain = getDomain(this.url)
+            val localUsername = this.username
+            return "@$localUsername@$domain"
```
connyduck, line 53: I'm afraid but this doesn't give the correct result. 
`this.username` is the `acct` of the [Mastodon Account object](https://docs.joinmastodon.org/entities/Account/) "The Webfinger account URI. Equal to username for local users, or username@domain for remote users." 
I think you want the `username` of the Mastodon Account ("The username of the account, not including domain.") which is `this.localusername`.

Second, minor problem that we can probably ignore for now is that there are Mastodon instances where the local domain is not the web domain. Check e.g `@Sir_Boops@sergal.org` where the domain is https://mastodon.sergal.org/@Sir_Boops so parsing the domain part out of the url won't work.

app/src/main/java/com/keylesspalace/tusky/components/account/AccountActivity.kt
```diff
@@ -854,6 +870,30 @@ class AccountActivity : BottomSheetActivity(), ActionButtonActivity, HasAndroidI
...
+                    sendIntent.putExtra(Intent.EXTRA_TEXT, url)
+                    sendIntent.type = "text/plain"
+                    startActivity(Intent.createChooser(sendIntent, resources.getText(R.string.send_account_link_to)))
+                }
+                return true
+            }
+            R.id.action_share_account_username -> {
+                // If the account isn't loaded yet, eat the input.
+                if (loadedAccount != null) {
+                    val fullUsername = loadedAccount!!.fullUsername
```
connyduck, line 888: instead of the `!= null` and then `!!` pattern you could 
```
loadedAccount?.let { account ->
// do something with the non-null account
}
```

app/src/main/java/com/keylesspalace/tusky/components/account/AccountActivity.kt
```diff
@@ -407,6 +410,19 @@ class AccountActivity : BottomSheetActivity(), ActionButtonActivity, HasAndroidI
...
         binding.accountDisplayNameTextView.text = account.name.emojify(account.emojis, binding.accountDisplayNameTextView, animateEmojis)
 
+        // Long press on username to copy it to clipboard
+        for (view in listOf(binding.accountUsernameTextView, binding.accountDisplayNameTextView)) {
+            view.setOnLongClickListener {
+                if (loadedAccount != null) {
+                    val fullUsername = loadedAccount!!.fullUsername
+                    val clipboard = getSystemService(Context.CLIPBOARD_SERVICE) as ClipboardManager
+                    clipboard.setPrimaryClip(ClipData.newPlainText(null, fullUsername))
+                    Toast.makeText(this, getString(R.string.account_username_copied), Toast.LENGTH_LONG).show()
```
connyduck, line 420: I'd prefer a SnackBar

app/src/main/java/com/keylesspalace/tusky/entity/Account.kt
connyduck: What I would do is 
- If it is a local account (same instance as logged in user), construct the full name out of acct + active domain 
- If it is a remote user, acct is the full name, just use that

mcclure:
Note to self I forgot to run ktLint

app/src/main/java/com/keylesspalace/tusky/entity/Account.kt
mcclure: You're right, I mean to do localUsername. The other thing is ugly, glad you caught that D:

Unrelated to this issue: We use `getDomain` in the account-pane "block instance" feature. Is this a problem?

app/src/main/java/com/keylesspalace/tusky/components/account/AccountActivity.kt
mcclure: The way I wrote it matches adjacent menu items such as `action_open_in_web`. So I'd prefer to either keep it the way it is or refactor the entire `when` block to use `.let` every time.

app/src/main/java/com/keylesspalace/tusky/entity/Account.kt
mcclure: Wait, how do I tell if it's the same instance of the logged in user or not? I'm not finding a relevant flag in `Account`.

I could check `username` for an internal `@` and assume if there's no @ they're local, but I think this will misbehave if a server ever misreports the `username` field and leaves the domain out.

mcclure:
Fixed the localUsername thing and the toast/snackbar thing, waiting for further changes until I get clarity above (how to detect same-instance user; should I do the .let refactor)

app/src/main/java/com/keylesspalace/tusky/components/account/AccountActivity.kt
connyduck: Its not really important, however you like

app/src/main/java/com/keylesspalace/tusky/entity/Account.kt
connyduck: Yes, check for @. If servers don't comply to the api I'd say it is not our problem.

mcclure:
That should be everything discussed.

app/src/main/java/com/keylesspalace/tusky/components/account/AccountActivity.kt
```diff
@@ -890,6 +930,16 @@ class AccountActivity : BottomSheetActivity(), ActionButtonActivity, HasAndroidI
         } else null
     }
 
+    private fun getFullUsername(account: Account): String {
+        if (account.isRemote()) {
+            return "@" + account.username
+        } else {
+            val localUsername = account.localUsername
+            val domain = accountManager.activeAccount!!.domain
```
mcclure, line 938: !! is safe here?

nikclayton: Can you use the passed in `account: Account` instead of `activeAccount`? That way it's definitely non-null.

mcclure: No. `activeAccount` is an `AccountEntity`, which is why it has the `domain` field `account` does not.

connyduck: Yes I think it is safe.

PR #3127, opened 2023-01-01
Improve muted users list

connyduck:
As discussed in our dev channel, use a switch instead of a button to avoid confusion about the state.

- also migrates the adapter to view binding
- also migrates the layout to a ConstraintLayout

While doing this I found several problems with account lists, like the bot overlay missing and other inconsistencies. I will fix those in a follow up PR.

cc @mal0ki

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -133,7 +133,6 @@
     <string name="action_mute">Mute</string>
     <string name="action_unmute">Unmute</string>
     <string name="action_unmute_desc">Unmute %s</string>
-    <string name="action_mute_notifications_desc">Mute notifications from %s</string>
     <string name="action_mute_domain">Mute %s</string>
     <string name="action_unmute_domain">Unmute %s</string>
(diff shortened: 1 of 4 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -133,7 +133,6 @@
     <string name="action_mute">Mute</string>
     <string name="action_unmute">Unmute</string>
     <string name="action_unmute_desc">Unmute %s</string>
-    <string name="action_unmute_notifications_desc">Unmute notifications from %s</string>
     <string name="action_mute_notifications_desc">Mute notifications from %s</string>
```
Tak, line 136:
```suggestion
```

Tak: Looks like this one is also unused now?

Tak (approved):

app/src/main/res/values/strings.xml
connyduck: yes, thanks!

Issue #2953, opened 2022-11-30
If a post fails to post after leaving the compose pane, there is no in-app UI.


mcclure:
Say you leave the compose pane before your post is fully posted (the easiest way to do this is to install PR #2947, but I think I've seen variations of this in weird forms of #2740).

Say then that the post *fails*. The most likely way to trigger this with PR #2947 is to hit this bug  
https:// github.com /mastodon/mastodon/ issues/ 21864  
(link intentionally broken), but I can make it happen consistently by attaching a particular 1080x2520 video (mastodon.social rejects videos of this resolution).

There is no *in-app* UI to indicate this failure. Under the current behavior of PR #2947, there is a *notification*, and the post is bumped to Drafts. This is good to have but I think it is inadequate because

1. Notifications can be turned off -- in fact I normally run my "Tusky Green" (selfbuilt/test) Tusky apk with notifications off
2. Notifications are a noisy channel -- I get enough notifications I kind of tune them out after a while

My "**expected behavior**" is that the next time the user opens the app (unless it is via the notification) there is a popup saying "Post failed and has been saved as a draft" "OK" "Go to drafts"

(The user should not be able to get into a state where their content "appears" to be lost, but they do not know what is wrong. I ran some tests with notifications off and had no idea that there were dead posts piling up in my drafts.)

* * * *
- Tusky Version: PR #2947
- Android Version: 12
- Android Device: Sony Xperia 5 III
- Mastodon instance (if applicable): mastodon.social

- [X] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

mcclure:
Implementation suggestion now that #2947 is in: When background process fails an update and sends to drafts, the background process should notify the main process, or maybe leave a sentinel file, and the main process should at soonest opportunity (such as next launch) bring up a dialog saying "a post upload failed and was saved to drafts" with options "OK" and "See drafts".

connyduck:
How I would implement this: Add another column to the drafts database to signal that it is a "new" failure. On restarting the app, check if there are any new failures for the current user. If yes, display a Snackbar that takes the user to drafts. On opening drafts, set all drafts of the currrent user to be old failures.

connyduck:
This is fixed now, right? @mcclure

mcclure:
This is resolved by #3135

PR #3135, opened 2023-01-04
Dialog notifying user of failure when media upload fails

mcclure:
**THIS PR MIGRATES YOUR DATABASE**

Addressing #2953

This actually pretty much works but I think there are some issues that need to be ironed out before it is merged

I have several concerns about specific code decisions, I will attach them as comments <s>except one big issue: *Am I supposed to be modifying AppDatabase.java directly*? It seems to be the only place to document the database version or migrations but it's also the only java file I've yet encountered in this program so it smells generated</s> Tak explained this it's fine

**The things holding back merge are**

* See comment in MainActivity.kt about positioning for draftsNeedUserAlert.observe
* **The notification problem**:

    Running with this PR, there are three ways you can get notified of a post failure:

    1. You are inside the app and the box pops up.
    2. You are outside the app, but later return to the app.
    3. You are outside the app and receive a notification.

    I find the notification problematic because (1) it is not possible to enable/disable this notification until after you have received it; whatever it is you do to register with the notification system we don't do it ahead of time. (2) It is "silent" and I don't think it should be. But also **the notification and this PR don't interact at all**. This means you potentially get notified "twice", a notification and then a popup box inside the app itself. How I would fix this:

    1. Make the notification non-silent (so that if the user has left the application they know to come back)
    2. Looking at the drafts pane should kill the popup box, IE, MainActivity should refuse to show the popup if the drafts pane is frontmost or if it was woken up with the Intent to open the drafts pane.
    3. If possible, the popup should suppress the notification. If the app is in a state it can show the popup the notification should not be issued. It would even be nice if looking at the drafts pane / seeing the popup could retroactively kill the notification ([apparently possible](https://stackoverflow.com/questions/3595232/android-remove-notification-from-notification-bar)) This is starting to sound kinda complicated but if we unsilence the notification the user might wind up kind of annoyed by being notified twice literally at once. >_>

    I don't think I know how to do these things ^, though maybe I could figure out iii.

    We could also resolve the conflict by removing the notification, but I think it's worth having, I just don't think we should rely on it exclusively because there are too many ways for people to dismiss/ignore notifications and failure to post content is pretty important.

I think we can and should get some version of this PR  in for 21.0 so that users can have a good first experience with PR#2947


app/src/main/java/com/keylesspalace/tusky/MainActivity.kt
```diff
@@ -165,6 +165,29 @@ class MainActivity : BottomSheetActivity(), ActionButtonActivity, HasAndroidInje
         val activeAccount = accountManager.activeAccount
             ?: return // will be redirected to LoginActivity by BaseActivity
 
+        // As early as possible register to be notified when the service saves a "couldn't upload" draft.
+        // This will be called when the app first comes up, which is desirable if a draft has been waiting awhile.
```
mcclure, line 169: The place this is called is actively Bad. I put it at the very top of MainActivity because it needs to go somewhere that could be activated "anywhere". However it DEFINITELY should go after the intent check, and it needs some kind of awareness of places it "shouldn't" activate— it shouldn't pop up over anything that it would "break" by interrupting (A compose pane?) and it definitely shouldn't activate if the drafts pane is already open. I need guidance on when to call this and how to tell if we're in a "non-interruptable" pane and/or the drafts pane (would having the drafts pane just call clearDraftsAlert on open be enough??).

app/src/main/java/com/keylesspalace/tusky/db/AccountManager.kt
```diff
@@ -42,6 +47,11 @@ class AccountManager @Inject constructor(db: AppDatabase) {
         private set
     private val accountDao: AccountDao = db.accountDao()
 
+    // For tracking when a media upload fails in the service
+    private val draftDao: DraftDao = db.draftDao()
+    private var draftsNeedUserAlertCurrent: LiveData<Int>? = null
+    public val draftsNeedUserAlert = MediatorLiveData<Int>()
```
mcclure, line 53: Using the mediator means that we can switch out the LiveData when we switch out users. In the current version of the feature, if you send a post then switch users very quickly, and the post fails to upload after you switched users, you will not be notified until the next time you switch back. That's sub-optimal but seemed better, in this version of the feature, than trying to negotiate switching users from the "post failed" alert pane (?) or deciding what to do in a nightmare scenario where multiple users have failed posts at the same time.

It's not clear to me AccountManager is the best place to manage the dao object here, BUT it's good in that it means we can get immediate switching to the correct user when the user changes.

app/src/main/java/com/keylesspalace/tusky/db/AccountManager.kt
```diff
@@ -160,6 +170,20 @@ class AccountManager @Inject constructor(db: AppDatabase) {
...
+            // Remove old observer if we just switched accounts
+            draftsNeedUserAlertCurrent?.let {
+                draftsNeedUserAlert.removeSource(it)
+            }
+            // Add new observer
+            val draftsNeedUserAlertCurrent = draftDao.draftsNeedUserAlert(dbId)
+            this.draftsNeedUserAlertCurrent = draftsNeedUserAlertCurrent
+            draftsNeedUserAlert.addSource(draftsNeedUserAlertCurrent, { count ->
+                Log.d(TAG, "draftsNeedUserAlert: account id " + dbId + " has " + count + " notification-worthy drafts")
+                draftsNeedUserAlert.value = count
```
mcclure, line 185: I will remove the debug print before merge. (It's extraneous because this callback is not called if there are no observers.)

app/src/main/java/com/keylesspalace/tusky/MainActivity.kt
mcclure: Also note: The observer callback DOES get called unconditionally on boot / re-entry into the program, which is what we want.

app/src/main/java/com/keylesspalace/tusky/db/AccountManager.kt
```diff
@@ -243,4 +267,14 @@ class AccountManager @Inject constructor(db: AppDatabase) {
...
     }
+
+    /**
+     * Clear drafts alert for current user
+     * Caller's responsibility to ensure there is a current user
+     */
+    fun clearDraftsAlert() {
+        runBlocking {
+            draftDao.draftsClearNeedUserAlert(activeAccount!!.id)
+        }
```
mcclure, line 278: Runblocking might be bad here.

Also, is there a race condition here with activeAccount? IE could we get a post fail notification on one user then somehow the active user changes between draftsNeedUserAlert deciding to signal its observer and the call back to clearDraftsAlert occurring?

app/src/main/java/com/keylesspalace/tusky/db/DraftDao.kt
```diff
@@ -30,6 +31,12 @@ interface DraftDao {
     @Query("SELECT * FROM DraftEntity WHERE accountId = :accountId ORDER BY id ASC")
     fun draftsPagingSource(accountId: Long): PagingSource<Int, DraftEntity>
 
+    @Query ("SELECT COUNT(*) FROM DraftEntity where accountId = :accountId and failedToSendNew=true")
+    fun draftsNeedUserAlert(accountId: Long) : LiveData<Int>
```
mcclure, line 35: COUNT() because we don't need to see the drafts, we just need to know "0, 1, or more than one?"

app/src/main/res/values/strings.xml
```diff
@@ -99,6 +99,11 @@
     <string name="action_login">Login with Mastodon</string>
     <string name="action_logout">Log Out</string>
     <string name="action_logout_confirm">Are you sure you want to log out of the account %1$s?</string>
+    <string name="action_post_failed">Upload failed</string>
+    <string name="action_post_failed_detail">A post failed to upload and has been saved to drafts.</string>
+    <string name="action_post_failed_detail_plural">%1$d posts failed to upload and have been saved to drafts.</string>
```
mcclure, line 104: Very unsure about any of this wording. It does not seem "friendly".

The number of uploaded posts is totally superfluous (and according to Google, including the number will cause [problems for, at least, Lithuanian?](https://developer.android.com/guide/topics/resources/string-resource#Plurals)), but I thought it would be nice to display a different message for one failed post vs more-than-one.

Currently thinking about changing it to "Your post failed to upload and has been saved to drafts." / "Your posts failed to upload and have been saved to drafts." and dropping the number.

Also not sure about the use of "Upload" but I can't think of a better word.

Suggestions welcomed.

mcclure:
Screenshot_20230104-000430

app/src/main/java/com/keylesspalace/tusky/db/DraftDao.kt
mcclure: By the way do we need to add an index for failedToSendNew since we are selecting / UPDATE WHERE-ing on it?

app/src/main/java/com/keylesspalace/tusky/db/DraftDao.kt
```diff
@@ -30,6 +31,12 @@ interface DraftDao {
     @Query("SELECT * FROM DraftEntity WHERE accountId = :accountId ORDER BY id ASC")
     fun draftsPagingSource(accountId: Long): PagingSource<Int, DraftEntity>
 
+    @Query("SELECT COUNT(*) FROM DraftEntity where accountId = :accountId and failedToSendNew=true")
+    fun draftsNeedUserAlert(accountId: Long): LiveData<Int>
```
mcclure, line 35: COUNT() because we don't need to see the drafts, we just need to know "0, 1, or more than one?"

mcclure: By the way do we need to add an index for failedToSendNew since we are selecting / UPDATE WHERE-ing on it?

mcclure:
Test failures:
```
com.keylesspalace.tusky.MainActivityTest > clicking notification of type FOLLOW_REQUEST shows follow requests FAILED
    java.lang.NullPointerException at MainActivityTest.kt:128
com.keylesspalace.tusky.MainActivityTest > clicking notification of type FOLLOW shows notification tab FAILED
    java.lang.NullPointerException at MainActivityTest.kt:128
```
I attempted to test these by hand but was not able to get Tusky Test to emit a notification when I made a follow request. Do I need to ? Is there some way to force it to check the slow way?

Because of the problem mentioned above (registering the popup before checking intent) it's possible to have effects like these, but I don't understand how the observer being registered interrupted either of these tests unless there was a failed image post beforehand. Migration problem [draft getting marked Failed New when it isn't?] DEFAULT 0 seems pretty straightforward and migration seemed to work right in my one test.

mcclure:
Follow request came through, I tried to activate it from lock screen and nothing happened, the notification just highlighted (I think this was just Android being weird), I activated it from notification center and it worked fine.

app/src/main/java/com/keylesspalace/tusky/db/AccountManager.kt
nikclayton: If there is, could you fix that by passing the ID in as a parameter to the function instead?

app/src/main/res/values/strings.xml
nikclayton: "Your Mastodon server temporarily rejected your post. It has been saved as a draft so you can try again."

- Makes it clear that this is the server's fault, not the user's (and not the app's)
- Makes it clear that this is a temporary issue
- Makes it clear they they can (and should) try again

In the actual dialog that pops up you could also include the text of the post, so the user can see which post is being referred to.

If multiple posts fail you can show multiple dialogs. That wouldn't be great UX (the user might have to click through multiple dialogs in the worst cast), but I suspect that multiple posts failing would be a rare occurrence, so probably not a problem (?)

app/src/main/java/com/keylesspalace/tusky/MainActivity.kt
```diff
@@ -165,6 +165,32 @@ class MainActivity : BottomSheetActivity(), ActionButtonActivity, HasAndroidInje
...
+                    .setMessage(
+                        if (count == 1)
+                            getString(R.string.action_post_failed_detail)
+                        else
+                            getString(R.string.action_post_failed_detail_plural, count)
+                    )
+                    .setPositiveButton(R.string.action_post_failed_show_drafts) { _: DialogInterface?, _: Int ->
+                        accountManager.clearDraftsAlert() // User looked at drafts
+
+                        val intent = DraftsActivity.newIntent(this)
```
nikclayton, line 184: Could this be re-worked so that it takes the user to the specific draft that failed, so they only have to press one more button to resend it?

As is, it looks as though they'll need to figure out which draft failed (with nothing in the UI to tell them), open it, and try and re-send it.

E.g., in the DB query instead of returning the count of failed-to-send posts, return a list of the IDs. Then instead of calling `DraftsActivity.newIntent()`, create a `ComposeOptions` with the draft ID, and start `ComposeActivity` instead (see `DraftActivity.onOpenDraft()`, which you might be able to make public)

nikclayton:
I've left a couple of comments specific to the change in this PR, but I also wanted to suggest a different approach that would obsolete this PR and might be better UX.

I'm not sure that using drafts to store posts that failed to send is appropriate, as it conflates two quite different things, which complicates the code, the mental model of the code, and the user's mental model of what's going on.

It also risks a data/privacy failure; the UX might cause the user to send a draft they did not intend to send.

So instead of using drafts for this, how about creating a different table that's specifically for statuses that the user is actively trying to send?

After the user clicks "Post", the app would write a row in this table with the status data.

Elsewhere in the app, a service can watch for new rows in this table. When it sees one it immediately tries to post it (sets the row's state as "in-progress"). If it succeeds the row is deleted (or a column is updated to mark the row's status as "sent").

If it fails, the service updates the row to include:

- The new state (presumably, "temp failure" or similar)
- Any text of the response from the server (so that can be shown to the user)
- The count of sending attempts
- Timestamp of the last attempt to send
- Timestamp of the next scheduled attempt to send (computed from the count of sending attempts)

Periodically (every 5 minutes?) the service checks for any entries in this table where the status is not "in-progress" and the timestamp of the next scheduled attempt is in the past, and tries to send them.

In the UI this could be exposed to the user in a "QueuedStatuses" activity that shows the data from the table; how many statuses are queued, the text, when each one will be retried, a button to "Try again now", maybe a "Convert to draft" affordance, or a "Delete this" affordance.

This QueuedStatuses activity would be visible from the main navigation drawer.

After some number of attempts (or some amount of time has elapsed) a queued status should probably transition to a state where it will not be automatically retried, because enough time has elapsed that the user might not consider the status to be relevant any more (maybe the converstaion has moved on, someone else has made the same point they were making, or the status was time sensitive in some other way).

If that happens, *then* send an Android notification that the user has a queued status they need to look at. And add a "!" badge to the icon for queued statuses in the navigation drawer.

I think this would achieve the following:

- In the common case, there's no change to the user's UX, as statuses are sent effectively immediately
- In the uncommon case, where a status temporarily fails, the user needs to take no action, because the app handles it for them
- In the rare case, where a status repeatedly temp. fails and becomes too old, the user can make an informed decision about what to do next.
- Drafts (which are statuses that the user *does not want to send yet*) and temporarily failed statuses (which the user *does want to send now*) are kept in different parts of the UI and data model, minimising the risk of user and developer confusion, and minimising the risk that user accidentally sends a status they do not want to send.

I appreciate that this is a significantly bigger change, and I'm not sure of the relative urgency of this PR -- if this is fixing an active problem that users are having right now then an incremental improvement is better than waiting for a larger architectural change.

But if this change isn't urgent then I'd like to consider the approach above.

app/src/main/java/com/keylesspalace/tusky/MainActivity.kt
mcclure: They will be told by the large existing banner reading "this status failed to send".

Although it would be nice if this PR could scroll the drafts view to the bottom so they are more likely to see the exact correct draft.

I don't think putting them in compose view is optimal because then they can't get out without making an affirmative decision to send or not send and also media will begin uploading immediately. Going to the drafts list lets them decide to discard without uploading anything if they choose.

[Screenshot_20230104-071742](https://user-images.githubusercontent.com/277318/210553459-83f4ffbc-325f-4988-a982-2d47167f9fe6.png)

app/src/main/java/com/keylesspalace/tusky/db/AccountManager.kt
mcclure: I think I could do this in the mediator callback, yes, but I do not know if there is a race condition.

mcclure:
> Periodically (every 5 minutes?) the service checks for any entries in this table where the status is not "in-progress" and the timestamp of the next scheduled attempt is in the past, and tries to send them.

I don't know if this is necessarily a good idea. The failure might be permanent (my test case for this PR is a video which Mastodon.social rejects 100% of the time because of its format). Post failure of this type most often happens when there are large images or media (or at least that's the assumption of this PR), and if the failure is temporary it might be because the server is under load, in which case automatically trying multiple times to send large media might make things worse. *Some* retry logic might be worth it (do we have retry logic already?). However, we probably still want the ability to save to the drafts eventually, if saving fails ten times for example that implies you're not making progress. So we'll still need this path.

As described in the issue up top, the reason for this PR is to alleviate the effects of PR #2947, which is already in the 21.0 branch. 2947 as far as I know only kicks in when the user impatiently sends before media is done uploading. THAT PR was urgent because it addressed an issue where the user impatiently leaving the app (previously if you tapped post while an image is uploading there would be an "uploading image…" modal) could lead to process termination and loss of the status. So we're trying to fix an urgent UI issue caused by the fix for an urgent bug, and Conny would like to feature freeze 21.0. This PR is already worryingly complex and an actual refactor of the basic functionality, if it turned out to be a good idea, would probably be an even longer delay and should probably wait for post-21.

One thing that might be worth filing as a follow up issue is I think we might know in some cases which failures are permanent and it might make sense to handle them differently. If I wait for my test image to upload instead of posting right away I get a nice human readable message from Mastodon.social explaining the resolution is illegal. The service could possibly save this error message and we could somehow display it, although I'm not sure where or when. I don't think that's urgent tho because the user can see the error if they wait around in the compose box for the upload to finish.

mcclure:
Looks like autogen messed up 46.json on one of my early builds of this PR, so I reverted it in [ae27cf8](https://github.com/tuskyapp/Tusky/pull/3135/commits/ae27cf84d19f23c9aa460ce882e7609efb434602) Note to self now I need to retest both a clean install of 47 and the 46->47 upgrade path

mcclure:
It seems like some of Nik's posts are displaying twice in this view and some of my replies are displaying not at all in the files changed view. Friggin github

mcclure:
`gradlew "testGreenDebugUnitTest" "--no-daemon"` currently succeeds on my system but appears to be failing on the CI server.

mcclure:
Followed my gut and the new strings are

* Your post failed to upload and has been saved to drafts.
* Your posts failed to upload and have been saved to drafts.

connyduck (changes requested):
`updateActiveAccount` is only called when network call succeeded, so you can't rely on it like that. Especially when uploading failed - device might be offline. Here is how I would do it:

Return a `kotlinx.coroutines.flow.Flow` from the Dao and collect it. Will automatically called each time the database changes.

```
fun draftsNeedUserAlert(accountId: Long): Flow<Int>
```

```
lifecycleScope.launch {
    db.draftDao().draftsNeedUserAlert(0).collectLatest {
        // do something, e.g. show dialog
    }
}
```

connyduck:
It can fail both because of network problems (no connection) or server problems, so the wording should indicate that.

mcclure:
> `updateActiveAccount` is only called when network call succeeded, so you can't rely on it like that. Especially when uploading failed - device might be offline. Here is how I would do it:
> 
> Return a `kotlinx.coroutines.flow.Flow` from the Dao and collect it. Will automatically called each time the database changes.
> [...]

That makes sense. But, I need to create the dao somewhere. If not updateActiveAccount, where would you recommend doing it? If the answer is MainActivity, I need to know how to get a handle to the database from inside MainActivity. (The dao tutorials I read suggested doing dao interactions in a model class, and the model classes indeed appear to be the ones that currently have db access handy.) Ideally this place should have access to the activeAccount (IE: occur after activeAccount is known) unless we want to teach the failure notification to switch accounts when "Show Drafts" is clicked.

I *think* the Dao does what a coroutine flow would do automatically without needing another layer of abstraction wrapped around it but maybe I am missing something.

mcclure:
(Also, re: Moving dao setup somewhere other than `updateActiveAccount`: If we are known to be offline, then sending the user to the drafts to re-send something that is just going to fail again is a waste of their time. Should the dialog text reflect that?)

mcclure:
> I'd put it at the end of onCreate

Can do, any thoughts on my [question](https://matrix.to/#/!PWkUGGpeMZWDPmdfRE:matrix.org/$167307463115930UUKmP:matrix.org?via=m.charlag.space&via=matrix.org&via=tchncs.de) from Matrix tho? (EG: How to detect drafts pane is in front when callback occurs and do nothing)

mcclure:
@connyduck Thanks for your responses. This thread is long so to be clear these are the questions I need answers on to complete the PR:

* Where to move Dao creation to ([here](https://github.com/tuskyapp/Tusky/pull/3135#issuecomment-1376385150))
* Drafts in front detection ([here](https://github.com/tuskyapp/Tusky/pull/3135#issuecomment-1376397467)
* Confirm if you're requesting a change to a snackbar [vs just asking for reasoning] ([here](https://github.com/tuskyapp/Tusky/pull/3135#discussion_r1065151408))

I should be able to resolve the other issues you raise on my own.

connyduck:
> Where to move Dao creation to

The ideal place would be a ViewModel, but MainActivity doesn't have one. We should change that but that is out of scope for this PR. So just inject the database directly into MainActivity and get the dao from there.

> Drafts in front detection

That could be tricky. Maybe it works the other way around: Only show the Info when MainActivity is in the front, and make sure to immediately clear the failedToSendNew flag when a draft is displayed in DraftsActivity.

> Confirm if you're requesting a change to a snackbar [vs just asking for reasoning]

I think Dialogs are for situations when the user must make a decision immediately, whereas Snackbars are less intrusive and can be ignored. I have no strong opinion about either just wanted to know if you have considered a Snackbar and why you went for Dialog.

mcclure:
Okay, that's enough for me to move forward although I will be busy during the week this week. Thanks

I did consider Snackbar vs Dialog and went with Dialog. I think it is a 'make immediate decision' kind of thing (IE, if you want to resend it's probably immediately). If a temporary popup was enough we could limit ourselves to making the notification non-hidden. However I am open to being talked into the Snackbar approach and again you have final say.

app/src/main/java/com/keylesspalace/tusky/db/AccountManager.kt
```diff
@@ -243,4 +265,14 @@ class AccountManager @Inject constructor(db: AppDatabase) {
 
         return accounts.size > 1 // "disambiguate"
     }
+
+    /**
+     * Clear drafts alert for current user
+     * Caller's responsibility to ensure there is a current user
+     */
+    fun clearDraftsAlert() {
```
nikclayton, line 273: Instead of using `activeAccount!!.id` a few lines down, if you change this function so that it takes the account ID as a parameter, and don't allow null values for the parameter then it ensures the calling code does a null check.

mcclure:
> It can fail both because of network problems (no connection) or server problems, so the wording should indicate that.

How would you recommend doing this?

Just spell it out?

> Your post failed to upload and has been saved to drafts. Either the server could not be contacted, or it rejected the post.

app/src/main/java/com/keylesspalace/tusky/db/DraftsAlert.kt
```diff
@@ -0,0 +1,126 @@
+/* Copyright 2023 Andi McClure
```
mcclure, line 1: How do we do this? Do I just put my name here because I created the file?

app/src/main/java/com/keylesspalace/tusky/db/AccountManager.kt
```diff
@@ -160,6 +165,9 @@ class AccountManager @Inject constructor(db: AppDatabase) {
 
             Log.d(TAG, "updateActiveAccount: saving account with id " + it.id)
             accountDao.insertOrReplace(it)
+
+            // Drafts alert tracker needs to change its Dao source to current user
+            draftsAlert?.updateActiveAccountId(it.id)
```
mcclure, line 167: Directly binding like this doesn't feel very Kotlin-y, but I don't know a better way (flows were suggested but those require you to have a lifetime, which DraftsAlert doesn't; it connects its data to various activity lifetimes to fire the alert, but *this* method only adjusts routing and could occur before any of those lifetimes have been registered)

app/src/main/java/com/keylesspalace/tusky/db/DraftsAlert.kt
```diff
@@ -0,0 +1,126 @@
...
+            }
+        }
+    }
+
+    /**
+     * Clear drafts alert for current user
+     * Caller's responsibility to ensure there is a current user
+     */
+    fun clearDraftsAlert(id: Long) {
+        GlobalScope.launch {
```
mcclure, line 118: Not crazy about this but not sure how else to do it. Could I get a coroutine scope from a lifecycle? In that case I could pass it in from the calls above, because this is always invoked in the context of the alert box and we always have a particular LifecycleOwner that is called with respect to. Would that allow this coroutine to continue to execute even after that lifecycle suspends (which if we displayed the dialog, will always be right after we do this), though?

What we need is for this to execute before any further draftDao events get routed in to the LiveData mediator.

mcclure:
I think the specific "change requested" by Conny is not relevant in the current code version.
I don't think a Flow is needed here because I'm using LiveData, which is like Flow but the Dao can create it automatically.

This is I think the first version of this PR that could be potentially committed. In my testing it what I'd consider the basic required elements to be user friendly:

- The popup appears at the first opportunity (the user who failed the post is looking at the Main Activity or an account feed)
- If the user opens the drafts view, or is in the drafts view when the failure occurs, they will not get the popup. If the user touches the silent notification currently issued when the draft is created, they will not get the popup.

I wound up creating an injected object named DraftsAlert to manage this.

Things that would be worth considering:

- It would be great if the silent notification could be cleared when the popup is cleared (ie DraftAlert.clearDraftsAlert is called)

Polish to consider for a future version:

- It would be nice if the popup could be displayed regardless of current user, and switch users appropriately.
- It would be nice if using the Show Drafts button (or following the notification) would scroll you to the bottom of the drafts view. (It might even make sense to show the popup if you're in the drafts view but not scrolled to the bottom.)

app/src/main/java/com/keylesspalace/tusky/db/DraftsAlert.kt
connyduck: Best practice is `Copyright 2023 Tusky contributors` but I also don't mind if you put you own name

connyduck:
Daos can also return flows, but LiveData also works. But that is not the important part.

As written earlier:

> updateActiveAccount is only called when network call succeeded, so you can't rely on it like that

Just recreate the Flow or LiveData everytime a MainActivity is created, dispose it when it is destroyed. Pass the current accountId everytime. Most of the logic in DraftsAlert is not needed.

connyduck:
> Just spell it out?

For example. Or use generic "because of a problem" wording.

mcclure:
New dialog text

Screenshot_20230124-011328

mcclure:
Implemented the two asks above. 

Dialog now has a second line (see screenshot). Seems a little wordy but whatever. Maybe in a later version we can teach it to give just one reason (we probably know if the server rejected us or timed out?)

DraftsAlert is now much simpler. The new DraftsAlert requires two assumptions to hold, which the old code would have been resilient to being false:

- MainActivity, DraftsActivity, and AccountActivity are all destroyed when the current account is changed (or at least, a new MainActivity, DraftsActivity, or AccountActivity is created for the new current account).
- MainActivity, DraftsActivity, and AccountActivity are all created (ie onCreate called) *after* current account in Account Manager goes non-nil (should be safe since Account Manager is only null before login).

In my testing, it does work correctly.

mcclure:
When https://github.com/tuskyapp/Tusky/pull/3135/commits/b4a733e454abe793bdfbb121f4be7233a9eb63ad was the top commit bitrise was reporting errors that would have only made sense for https://github.com/tuskyapp/Tusky/pull/3135/commits/750b6798a02d6e99eb4ca1f2f7df29f6238b08d8, and now that https://github.com/tuskyapp/Tusky/pull/3135/commits/50deb1b87e86849d53c8bb4d8d7ae9a8d50823e1 is the top commit bitrise is reporting errors that only make sense for https://github.com/tuskyapp/Tusky/pull/3135/commits/750b6798a02d6e99eb4ca1f2f7df29f6238b08d8. MainActivityTest is succeeding in my local tests. Is it possible bitrise is somehow off by one commit?

app/src/main/java/com/keylesspalace/tusky/db/DraftsAlert.kt
connyduck: Yes this needs to use the scope of the activity. You can get it by calling `lifecycleScope` on an Activity/LifecycleOwner.

mcclure: Made this change, what happens tho if the LifecycleOwner suspends while the launched async task is still unfinished? Like is there a potential for a race condition or are launched jobs allowed to finish?

connyduck (approved):

connyduck (approved):

PR #3140, opened 2023-01-05
Show a close button for hashtag tab chips

nikclayton:
The intent of the previous code seems to be to show an "X" icon on a hashtag chip when two or more chips are present.

This didn't work because the icon was not set as visible.

Fix this. In addition, set this as a "cancel" icon, not the chip's regular icon, so it appears on the right (in LTR locales), as is normal for the close button on chips.

Tinting the icon did nothing, so remove that.


nikclayton:
Before and after screenshots:

connyduck (approved):
nice

Issue #3131, opened 2023-01-03
UX: Use the normal UI text size for items in lists (preferences, menus, etc)


nikclayton:
There's a few places in the UI where the user's preference for post text size is being used, and I suspect it shouldn't be.

[Alternatively, there are a bunch of places in the UI where the user's preference for text size is not being used].

This is most obvious if you set the post text size to the smallest value.

For an example of what I think lists should look like, open "Account Preferences > Filters > Home". If you have any filters set then you get a list of the words that are filtered, the layout is simple_list_item_1.xml, which sets

```
    android:textAppearance="?android:attr/textAppearanceListItemSmall"
```

on the `TextView`.

Or open "Preferences > App Theme", and see the list of themes in the same text size. Same for "Preferences > Language", etc, etc.

Places where this is *not* done are:

- item_followed_hashtag.xml ("Account Preferences > Followed Hashtags")
- item_hashtag.xml (search for a hashtag, and look at the list of results)
- item_list.xml (ListsActivity, from the "Lists" entry on the navigation drawer)
- item_picker_list.xml (ListSelectionAdapter, created from TabPreferenceActivity, "Account Preferences > Tabs")
- item_tab_preference.xml (TabAdapter, also TabPreferenceActivity)
- item_tab_preference_small.xml (TabAdapter, also TabPreferenceActivity)

Changing those to to use

```
    android:textAppearance="?android:attr/textAppearanceListItemSmall"
```

would be straightforward, make those lists consistent with other lists in the app.

As an aside, this would also fix some inconsistencies in the colours used for list items as well.

* * * *
- Tusky Version: 20

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

PR #3142, opened 2023-01-05
Adjust list UX for platform consistency

nikclayton:
Most lists in the app use (explicitly or implicitly) platform metrics for dimensions, text size, colour, and so on, possibly via styles.

A few don't, inadvertently using the user's setting for status text size

Fix these, and simplify code where possible.

- Use android attributes for padding and height, for consistent UX.

- Remove explicit usage of app:tabTextAppearance, rely on the style.

- Remove ListSelectionAdapter and item_picker_list.xml, and adjust TabPreferenceActivity to use an ArrayAdapter with simple_list_item_1.xml

- Simplify item_followed_hashtag.xml, consistent with item_list.xml.

Fixes https://github.com/tuskyapp/Tusky/issues/3131


nikclayton:
Here are some before/after screenshots:

nikclayton:
# Following hashtags

- No dividers (not necessary for single line content)
- List item text size is not affected by status text size

nikclayton:
# Viewing lists

- No dividers (not necessary for single line content)
- List item text size is not affected by status text size

nikclayton:
# Picking a list to add as a tab

- List item text size is not affected by status text size (in the dialog, and the background)

nikclayton:
# Adding a new tab

- List item text is not affected by status text size

nikclayton:
# Searching hashtags

- No dividers (not necessary for single line content)
- List item text size is not affected by status text size
- Tab titles ("POSTS", etc) text size is not affected by status text size

nikclayton:
The account page also benefits from the tab titles text size fix.

connyduck:
I think it is actually a feature that more things are affected by the status text size, the setting is jus tbadly named. But I'd like to hear from someone who actually uses it. https://chaos.social/@ConnyDuck/109677568422013580

nikclayton:
I've put the dividers back for list items with UI affordances at the start/end of the item.

As to what text should change, I think the way to conceptualise this is that Tusky has two types of text

1. Those that are part of the "body" role (https://m3.material.io/styles/typography/overview).
2. Those that are part of the UI chrome

Status text on a timeline should be based on the "body" role (https://m3.material.io/styles/typography/overview). If the user wants to modify that they can, through the Tusky preference.

Everything else -- text on menus, the title/app bar, in lists (whether generated by the app, or e.g., in the preferences fragments), `EditView` when searching, tab titles, etc, etc, is part of the UI chrome). That way the UI is consistent with the type sizes on every other app on the user's device.

If the user wants to change the scale of the UI chrome on their device then Android's global settings let them do that (and Tusky does honour it).

connyduck:
Yes, I agree that is technically the correct way to do things. But there are definitely users with limited eyesight for which the system settings are still not enough and helping them is more important than being technically correct.

The font size preference affects some places without any system behind it, so yes we should clean that up. Either by making it affect everything or by splitting it into two, one for body one for interface.

nikclayton:
https://github.com/tuskyapp/Tusky/pull/3248 addresses the "Users still need a way to adjust the text size in UI chrome" concern.

connyduck (commented):
I still think dividers should be everywhere. E.g. in hashtag search, when the hashtags are very short dividers make it easier to click on the other side of the screen and still hitting the right one.

Also, some lists have a ripple effect when clicked, and some don't. Should we do something about that as well?

nikclayton:
> I still think dividers should be everywhere. E.g. in hashtag search, when the hashtags are very short dividers make it easier to click on the other side of the screen and still hitting the right one.

"everywhere" would encompass places like:

- the preferences screen
- the navigation menu
- creating / editing a filter
- menus when adding a tab
- etc

I've put it back on the hashtag search results.

> Also, some lists have a ripple effect when clicked, and some don't. Should we do something about that as well?

E.g., the "Account preferences > Followed hashtags" doesn't ripple because you can't interact with the whole list item, just the "Unfollow" button at the end of the item.

I added it to item_list.xml, as that is selectable.

connyduck (approved):

PR #3143, opened 2023-01-05
Enable robolectric logging

nikclayton:
Robolectric normally swallows output from `Log.*` calls in tests and the code under test. This can make debugging failing tests more difficult than it needs to be.

Set `robolectric.logging.enabled` to `true` to enable the logs.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -72,6 +72,9 @@ android {
             returnDefaultValues = true
             includeAndroidResources = true
         }
+        unitTests.all {
+            systemProperty 'robolectric.logging.enabled', 'true'
+        }        
     }
     sourceSets {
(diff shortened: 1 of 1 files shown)


app/build.gradle
```diff
@@ -72,6 +72,9 @@ android {
             returnDefaultValues = true
             includeAndroidResources = true
         }
+        unitTests.all {
+            systemProperty 'robolectric.logging.enabled', 'true'
+        }
```
nikclayton, line 77:
```suggestion
            systemProperty 'robolectric.logging', 'stdout'
        }        
```

Note to reviewer: The PR as stands enables Robolectric debug logging. To get the logs from general Log.* calls you need this line as well.

Issue #2109, opened 2021-03-08
Trending Hashtags page


soudk:
Wouldn't it be cool to have a trending page in the sidebar to see what hashtags are trending on local timelines, maybe with a toggle to view trends across Mastodon? 

I think Fedilab has this but I'm not sure if it works.

guerda:
There finally is a trends API, based on the instance you're on that gives you this.
Maybe it can be implemented in the search tab like in the web app?

https://docs.joinmastodon.org/methods/instance/trends/

guerda:
The status trends (called "Explore" in the UI) is behind the instance trends API. It seems to return a simple list of toots, like the search API in the element `statuses`.
Unfortunately, it's not yet documented on docs.joinmastodon.org (see https://github.com/mastodon/documentation/issues/947 for more information).

Here's an example for the statuses result:

    [
      {
          "id": "108521492941340842",
          "created_at": "2022-06-22T14:02:48.000Z",
          "in_reply_to_id": null,
          "in_reply_to_account_id": null,
          "sensitive": false,
          "spoiler_text": "",
          "visibility": "public",
          "language": "de",
          "uri": "https://mastodon.social/users/watch_union/statuses/108521492880858106",
          "url": "https://mastodon.social/@watch_union/108521492880858106",
          "replies_count": 4,
          "reblogs_count": 19,
          "favourites_count": 2,
          "edited_at": null,
          "content": "<p><a href=\"https://mastodon.social/tags/Lindner\" class=\"mention hashtag\" rel=\"nofollow noopener noreferrer\" target=\"_blank\">#<span>Lindner</span></a> will die deutsche Bevölkerung auf \"5 Jahre der Knappheit\" einstellen.</p><p>Währenddessen schütten deutsche Konzerne 70 Mrd. € an Dividenden aus.<br>Da ist nicht knapp, nur scheiße verteilt und der <a href=\"https://mastodon.social/tags/FDP\" class=\"mention hashtag\" rel=\"nofollow noopener noreferrer\" target=\"_blank\">#<span>FDP</span></a> gefällt es.<br>/PM</p>",
          "reblog": null,
          "account": {
              "id": "108238000644624010",
              "username": "watch_union",
              "acct": "watch_union@mastodon.social",
              "display_name": "UnionWatch",
              "locked": false,
              "bot": false,
              "discoverable": true,
              "group": false,
              "created_at": "2022-05-03T00:00:00.000Z",
[...]

Issue #3010, opened 2022-12-05
[Feature Request] make side menu configurable like the tab navigation


Sirs0ri:
Hi, I'd like to be able to configure what navigation items show up in the side menu, like it's already possible with the tab navigation. 

At the moment you have to limit yourself to the 5 tabs available (see #2512), and other parts of the App are not accessible when you run out of tabs. A lot of the items currently available as tabs would also fit into the side menu very well (notifications, DMs, various lists/followed tags, for example), so it would be nice to have control over the items in the menu.

* * * *
- Tusky Version: 20
- Android Version: 12
- Android Device: Galaxy s10e
- Mastodon instance (if applicable): n.a.

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

george-steel:
This would definitely be helpful. Even if it wasn't customizable, showing anything missing from the tabs (or even all of the 5 main panes all the time) would be quite useful.

keelancook:
+1 for this. It's the single biggest thing that keeps me from using Tusky most of the time. I love this app. Looks good. Runs well. However, that 5 tab cap to navigate easily to something becomes a deal breaker.

Another option would be providing the ability to have an overflow button of some form in the tab nav menu as well. Fedilab does this with their pinned lists, and it's a killer feature. You could make one of your five buttons in the nav menu an overflow button that has all lists of pinned tags in it.

At present, I have more lists than I can regularly check in Tusky, because it takes far to many clicks just to switch lists.

Issue #2914, opened 2022-11-23
Migrate to ktlint 0.47.1 or newer


nikclayton:
The org.jlleitschuh.gradle.ktlint plugin (ktlint-gradle) is using an old version of ktlint (0.43.2, per https://github.com/JLLeitschuh/ktlint-gradle/blob/master/plugin/src/main/kotlin/org/jlleitschuh/gradle/ktlint/KtlintExtension.kt#L38).

At the time of writing the most recent version of ktlint is 0.47.1.

The different ktlint versions have different rulesets, with newer versions recognising new issues.

This causes problems if you try and use other tools that incorporate ktlint. For example, editing Tusky code with the IntelliJ ktlint plugin (https://plugins.jetbrains.com/plugin/15057-ktlint-unofficial-) installed reports problems with the code that `gradlew ktlintCheck` does not find.

ktlint-gradle does not work with any version of ktlint > 0.45.x at the moment (https://github.com/JLLeitschuh/ktlint-gradle/issues/589).

Other ktlint + gradle options exist (supporting 0.47.1), including:

- https://github.com/jeremymailen/kotlinter-gradle
- https://github.com/diffplug/spotless/tree/main/plugin-gradle#applying-ktlint-to-kotlin-files
- https://github.com/autostyle/autostyle/tree/master/plugin-gradle#applying-ktlint-to-kotlin-files

I experimented with the first one with [kotlinter-gradle.patch.txt](https://github.com/tuskyapp/Tusky/files/10073685/kotlinter-gradle.patch.txt) (probably needs some work, I did the first thing that works, not necessarily the best thing).

With that, running `gradlew lintKotlin` generates an additional ~ 280 lint warnings, see [lint.txt](https://github.com/tuskyapp/Tusky/files/10073689/lint.txt)

That probably doesn't make sense as a PR, as it touches the entire codebase and would need to be regenerated with every other PR that gets merged.

I think the next steps are;

1. Decide whether or not to fix this
2. If "fix", decide which plugin to use. kotlinter-gradle solves this specific problem, Spotless and Autostyle can format more files (Markdown, SQL, XML, JSON, etc) but require more configuration.
3. In one commit (a) modify the gradle files, (b) apply the style fixes, (c) update any CI commands, (d) update documentation

connyduck:
[...]

PR #3149, opened 2023-01-07
Add trending tags

DavidEdwards:
Related to #2109.

This adds the ability to add a trending tab to the bottom navigation bar.

Includes a new custom view for graphing. This could perhaps be made _nicer_ with an external library.

Only hashtags are supported currently, there are two other trending APIs that could eventually be added.

- [x] Fix failed checks
- [x] Add error handling (see airplane mode)
- [x] Black theme, graph background should be black

image


nikclayton:
I'll send separate feedback on the code, but some high level stuff.

# Appearance

It's probably easier if I show an alternative layout, and describe what I think it improves:

image

(based on https://dribbble.com/shots/2513500-Stats-card/attachments/9339291?mode=media)

- It can scale to the width of the screen, all the space is used to display useful data.

- The date range of the chart is indicated, without needing markers under the X axis

- The high/low numbers are marked (they could also be shown as smaller text, perhaps underneath the larger usage and accounts numbers?). I think they're optional, but probably nice to have.

- This also gives a sense of the scale of the charts, without needing to explicitly label the Y axis.

- Very little text would be repeated across charts (just the date range and labels). Subjectively, this feels less busy than the repeated "X people are talking" text.

(Colours should match the Tusky colours, of course).

More examples for inspiration; https://chartio.com/blog/new-chart-type-sparklines/, https://dribbble.com/shots/2716084-Study-of-sparkline-like-graphics/attachments/9446250?mode=media, https://www.thesmallman.com/excel-sparklines

Also: I know the API only hands back 7 days worth of data, which might not be enough to make a full-width view work (the mockup above has 30 data points).

If that's not enough data, how about a two column layout using `GridLayoutManager`, where each entry in the grid looks like this:

image

?

# General

Since there are other trending options available (statuses, links) it's probably best if the name is more specific in both the UI and the code, so that when those others are added it's not an issue.

Speaking of that, have you given any thought to how those others would be shown in the UI / how it will scale to show them / how the user will switch between the ones they want?

Optional: The list of trending hashtags in the results includes ones that I am actively filtering. That's consistent with what the Mastodon website does, but maybe there's an opportunity to do better than the website, and filter them from this view too?

If the user doesn't add the Trending tab, can it be added to the main navigation drawer as an entry there? For example, if the user disables the top navigation then the Search menu (which is normally there) moves to the Navigation drawer. This could behave the same way, so it's accessible to users that want it semi-frequently, but don't want to give up an entire tab for it.

app/src/main/java/com/keylesspalace/tusky/components/trending/viewmodel/TrendingViewModel.kt
```diff
@@ -0,0 +1,78 @@
...
+            invalidate()
+        }
+    }
+
+    suspend fun trendingTags(): List<TrendingTag> {
+        return trendingCases.trendingTags()
+    }
+
+    /** Triggered when currently displayed data must be reloaded. */
+    suspend fun invalidate() {
```
nikclayton, line 61: This shouldn't be suspend (https://developer.android.com/kotlin/coroutines/coroutines-best-practices#viewmodel-coroutines). This can be

```kotlin
fun invalidate() = viewModelScope.launch {
    ...
}
```

and then the caller doesn't need to worry about the scope.

app/src/main/java/com/keylesspalace/tusky/components/trending/viewmodel/TrendingViewModel.kt
```diff
@@ -0,0 +1,78 @@
...
+
+class TrendingViewModel @Inject constructor(
+    private val trendingCases: TrendingCases,
+    private val accountManager: AccountManager,
+) : ViewModel() {
+    enum class LoadingState {
+        INITIAL, LOADING, LOADED, ERROR_NETWORK, ERROR_OTHER
+    }
+
+    data class State(val trendingViewData: List<TrendingViewData>, val loadingState: LoadingState)
```
nikclayton, line 39: The rest of the code base generally calls these `${prefix}UiState`, where `${prefix}` is the view model name without the `ViewModel` suffix. So `TrendingUiState` would be appropriate here.

app/src/main/java/com/keylesspalace/tusky/components/trending/viewmodel/TrendingViewModel.kt
```diff
@@ -0,0 +1,78 @@
...
+
+    val state: Flow<State> get() = _state
+    private val _state = MutableStateFlow(State(listOf(), LoadingState.INITIAL))
```
nikclayton, line 42: The result of the code generally calls these variables `uiState` and `_uiState` respectively.

app/src/main/java/com/keylesspalace/tusky/components/trending/viewmodel/TrendingViewModel.kt
```diff
@@ -0,0 +1,78 @@
...
+
+    private var alwaysShowSensitiveMedia = false
+    private var alwaysOpenSpoilers = false
+
+    fun init() {
```
nikclayton, line 47: Can this be an `init { ... }` block, so it runs automatically instead of a function that needs to be called?

app/src/main/java/com/keylesspalace/tusky/components/trending/TrendingFragment.kt
```diff
@@ -0,0 +1,364 @@
...
+                                    Utils.dpToPx(requireContext(), -30)
+                                )
+                            } else binding.recyclerView.scrollToPosition(0)
+                        }
+                    }
+                }
+            }
+        })
+
+        if (actionButtonPresent()) {
```
nikclayton, line 165: Should the FAB be present here? I'm not sure it makes sense (unless we want people to be able to create a new status from anywhere in the app).

app/src/main/java/com/keylesspalace/tusky/components/trending/TrendingFragment.kt
```diff
@@ -0,0 +1,364 @@
...
+            TrendingViewModel.LoadingState.LOADED -> applyLoadedState(trendingState.trendingViewData)
+            TrendingViewModel.LoadingState.ERROR_NETWORK -> networkError()
+            TrendingViewModel.LoadingState.ERROR_OTHER -> otherError()
+        }
+    }
+
+    private fun applyLoadedState(viewData: List<TrendingViewData>) {
+        clearLoadingState()
+        adapter.submitList(viewData)
+        Log.v("TESTTEST", "list: ${viewData.map { it.asTagOrNull()?.tag?.name }}")
```
nikclayton, line 252: Holdover, and can be removed? If not, please use the `TAG` const.

app/src/main/java/com/keylesspalace/tusky/components/trending/TrendingPagingAdapter.kt
```diff
@@ -0,0 +1,134 @@
...
+                mediaPreviewEnabled = mediaPreviewEnabled
+            )
+        }
+
+    init {
+        stateRestorationPolicy = StateRestorationPolicy.PREVENT_WHEN_EMPTY
+    }
+
+    override fun onCreateViewHolder(viewGroup: ViewGroup, viewType: Int): RecyclerView.ViewHolder {
+        return when (viewType) {
```
nikclayton, line 48: Why the `when` here? These branches do the same thing. Am I missing something?

app/src/main/java/com/keylesspalace/tusky/adapter/TagViewHolder.kt
```diff
@@ -0,0 +1,75 @@
...
+import android.view.View
+import android.widget.TextView
+import androidx.recyclerview.widget.RecyclerView
+import com.keylesspalace.tusky.R
+import com.keylesspalace.tusky.entity.TrendingTagHistory
+import com.keylesspalace.tusky.interfaces.LinkListener
+import com.keylesspalace.tusky.view.GraphView
+import com.keylesspalace.tusky.viewdata.TrendingViewData
+
+class TagViewHolder(itemView: View) : RecyclerView.ViewHolder(itemView) {
```
nikclayton, line 27: `TrendingTagViewHolder`? To avoid confusion with other view holders that handle hashtags?

app/src/main/java/com/keylesspalace/tusky/components/trending/TrendingPagingAdapter.kt
```diff
@@ -0,0 +1,134 @@
...
+    }
+
+    override fun getItemViewType(position: Int): Int {
+        return if (getItem(position) is TrendingViewData.Tag) {
+            VIEW_TYPE_TAG
+        } else {
+            VIEW_TYPE_TAG
+        }
+    }
+
```
nikclayton, line 103: `override fun getItemViewType(position: Int): Int = VIEW_TYPE_TAG` ?

app/src/main/java/com/keylesspalace/tusky/entity/TrendingTagsResult.kt
```diff
@@ -0,0 +1,29 @@
...
+ * You should have received a copy of the GNU General Public License along with Tusky; if not,
+ * see <http://www.gnu.org/licenses>. */
+
+package com.keylesspalace.tusky.entity
+
+data class TrendingTag(
+    val name: String,
+    val url: String,
+    val history: List<TrendingTagHistory>,
+    val following: Boolean,
```
nikclayton, line 22: Is this an expected property? It's not in the example at https://docs.joinmastodon.org/methods/trends/#tags, and the string "following" doesn't appear on that page.

app/src/main/java/com/keylesspalace/tusky/entity/TrendingTagsResult.kt
```diff
@@ -0,0 +1,29 @@
...
+ * Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License along with Tusky; if not,
+ * see <http://www.gnu.org/licenses>. */
+
+package com.keylesspalace.tusky.entity
+
+data class TrendingTag(
+    val name: String,
+    val url: String,
```
nikclayton, line 20: I know it's inconsistent with the rest of the code, but would you mind adding one/two line Javadoc here explaining the semantics of the different fields. I know it's in the API docs, but it's tedious to have to go and look that up when it could appear inline in Android Studio.

For example:

```kotlin
[...]
/** URL to a page (not an API call) on the server to view statuses with this tag */
var url: String
[...]
```

I want to slowly clean this up in the existing API implementation, so may as well start here.

app/src/main/java/com/keylesspalace/tusky/usecase/TrendingCases.kt
```diff
@@ -0,0 +1,48 @@
...
+import kotlinx.coroutines.withContext
+import javax.inject.Inject
+
+/**
+ * Created by @knossos@fosstodon.org on 2023-01-07.
+ */
+class TrendingCases @Inject constructor(
+    private val mastodonApi: MastodonApi,
+) {
+    suspend fun trendingTags(): List<TrendingTag> {
```
nikclayton, line 31: Why do you have this layer of indrection? The only caller is in TrendingViewModel, so you could `MastodonApi` there, and call the API's `trendingTags()` function directly, in the viewmodel's scope.

Am I missing something subtle?

app/src/main/java/com/keylesspalace/tusky/viewdata/TrendingViewData.kt
```diff
@@ -0,0 +1,37 @@
...
+
+import com.keylesspalace.tusky.entity.TrendingTag
+
+/**
+ * Created by charlag on 11/07/2017.
+ *
+ * Class to represent data required to display either a notification or a placeholder.
+ * It is either a [TrendingViewData.Concrete] or a [TrendingViewData.Placeholder].
+ */
+sealed class TrendingViewData {
```
nikclayton, line 26: This is probably obsolete, right, as there are no placeholders in the trending API?

app/src/main/res/layout/item_trending.xml
```diff
@@ -0,0 +1,66 @@
...
+        android:id="@+id/tag"
+        android:layout_width="0dp"
+        android:layout_height="wrap_content"
+        android:layout_marginStart="14dp"
+        android:layout_marginEnd="14dp"
+        android:ellipsize="end"
+        android:importantForAccessibility="no"
+        android:maxLines="1"
+        android:textColor="?android:textColorPrimary"
+        android:textSize="?attr/status_text_medium"
```
nikclayton, line 40: That's the wrong text size, as it will adjust based on the user's choice of status text size. I've fixed this elsewhere in https://github.com/tuskyapp/Tusky/issues/3131, the diff there should show the values to use here.

app/src/main/res/layout/item_trending.xml
```diff
@@ -0,0 +1,66 @@
...
+        android:id="@+id/text"
+        android:layout_width="0dp"
+        android:layout_height="wrap_content"
+        android:layout_marginEnd="14dp"
+        android:layout_marginBottom="8dp"
+        android:hyphenationFrequency="full"
+        android:importantForAccessibility="no"
+        android:lineSpacingMultiplier="1.1"
+        android:textColor="?android:textColorPrimary"
+        android:textSize="?attr/status_text_medium"
```
nikclayton, line 59: Ditto, re text size.

nikclayton (changes requested):
I hope that's helpful feedback.

DavidEdwards:
Thank you for the very thorough feedback. I'll work through it over the next several days.

You are correct, the 7 days of data is the problem. I like the idea of using a grid. That could also scale nicely for tablets. 

The explicitly provided date range makes sense too.

Is it worth showing both the account and usage graphs? I don't know whether the data is interesting enough to show both. The peaks should be mostly identical. The accounts graph is just more likely to be flatter. I would be interested to hear your perspective.

I completely agree with the repeated text argument. The new design certainly feels like it provides more information to the user.

Regarding the code, a lot of the structures are copy-pasted-adapted from existing code. Hence the java file(s), findByViewId, etc. I feel like I am getting more comfortable with the codebase, so cleaning that up shouldn't be a hassle.

nikclayton:
> Is it worth showing both the account and usage graphs? I don't know whether the data is interesting enough to show both. The peaks should be mostly identical. The accounts graph is just more likely to be flatter. I would be interested to hear your perspective.

Maybe... the info's there, so perhaps it's useful? I don't a very strong opinion, and I can see how it could also make the charts harder to read. I suspect it's a "try it and see" situation.

app/src/main/java/com/keylesspalace/tusky/adapter/TrendingBaseViewHolder.java
```diff
@@ -0,0 +1,1179 @@
...
+
+import java.text.NumberFormat;
+import java.util.Date;
+import java.util.List;
+
+import at.connyduck.sparkbutton.SparkButton;
+import at.connyduck.sparkbutton.helpers.Utils;
+import kotlin.collections.CollectionsKt;
+
+public abstract class TrendingBaseViewHolder extends RecyclerView.ViewHolder {
```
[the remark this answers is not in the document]
DavidEdwards: :+1: Resolved locally. Deleted the class entirely. I copied the structure of the Timeline partially. This wasn't necessary.

app/src/main/java/com/keylesspalace/tusky/components/trending/viewmodel/TrendingViewModel.kt
```diff
@@ -0,0 +1,78 @@
...
+        this.alwaysShowSensitiveMedia = accountManager.activeAccount!!.alwaysShowSensitiveMedia
+        this.alwaysOpenSpoilers = accountManager.activeAccount!!.alwaysOpenSpoiler
+
+        viewModelScope.launch {
+            invalidate()
+        }
+    }
+
+    suspend fun trendingTags(): List<TrendingTag> {
```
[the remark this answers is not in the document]
DavidEdwards: :+1: Resolved locally.

app/src/main/java/com/keylesspalace/tusky/components/trending/viewmodel/TrendingViewModel.kt
DavidEdwards: :+1: Resolved locally. I wasn't aware that was best practices. Thanks for the heads up.

app/src/main/java/com/keylesspalace/tusky/components/trending/viewmodel/TrendingViewModel.kt
DavidEdwards: :+1: Resolved locally.

app/src/main/java/com/keylesspalace/tusky/components/trending/viewmodel/TrendingViewModel.kt
DavidEdwards: :+1: Resolved locally.

app/src/main/java/com/keylesspalace/tusky/components/trending/TrendingFragment.kt
DavidEdwards: I thought it should be. This is a bigger question that me though. It seems to be possible everywhere else? 

My opinion, I don't think we lose anything by keeping it.

app/src/main/java/com/keylesspalace/tusky/components/trending/TrendingFragment.kt
DavidEdwards: :+1: Resolved locally.

app/src/main/java/com/keylesspalace/tusky/components/trending/TrendingPagingAdapter.kt
DavidEdwards: I kept this here as a reminder that other Trending types exist. However, if that isn't clean enough I can remove it. Just let me know.

app/src/main/java/com/keylesspalace/tusky/adapter/TagViewHolder.kt
DavidEdwards: +1 Resolved locally.

app/src/main/java/com/keylesspalace/tusky/components/trending/TrendingPagingAdapter.kt
DavidEdwards: The same as when above. I kept this here as a reminder that other Trending types exist. However, if that isn't clean enough I can remove it. Just let me know.

app/src/main/java/com/keylesspalace/tusky/entity/TrendingTagsResult.kt
DavidEdwards: I believe so. It appears in this instance at least:

https://fosstodon.org/api/v1/trends/tags

app/src/main/java/com/keylesspalace/tusky/entity/TrendingTagsResult.kt
DavidEdwards: No problem at all. That makes complete sense.

:+1: Resolved locally.

app/src/main/java/com/keylesspalace/tusky/viewdata/TrendingViewData.kt
DavidEdwards: Yes, the comment was wrong. However, I was planning on using the TrendingViewData as a parent for the other Trending data too. In order to use the same adapter.

app/src/main/java/com/keylesspalace/tusky/usecase/TrendingCases.kt
DavidEdwards: Having the cases class allows us to replace the functionality for testing purposes. API calls don't belong in the ViewModel ideally.

I would usually implement a Repository (in a similar style to InstanceInfoRepository), but since I was copying the Timeline to a degree, I ended up using the cases class as a style guide (see: TimelineCases).

We can implement the other trending APIs in the same class.

What do you think?

app/src/main/res/layout/item_trending.xml
DavidEdwards: :+1: Resolved locally.

DavidEdwards:
@nikclayton  I have replied to your comments, and here is a visual of the current cells.

image

image

DavidEdwards:
> Since there are other trending options available (statuses, links) it's probably best if the name is more specific in both the UI and the code, so that when those others are added it's not an issue.

I will need to look at this more closely. In the UI, it could be labelled Trending Tags perhaps. In the code, could you give an example? Would `TrendingViewData` be confusing currently?

> Speaking of that, have you given any thought to how those others would be shown in the UI / how it will scale to show them / how the user will switch between the ones they want?

Not yet, but I think that will be my next task. Trending links has a similar structure, I will likely look there.

> Optional: The list of trending hashtags in the results includes ones that I am actively filtering. That's consistent with what the Mastodon website does, but maybe there's an opportunity to do better than the website, and filter them from this view too?

I think that makes sense. I will need to investigate how the filtering happens.

> If the user doesn't add the Trending tab, can it be added to the main navigation drawer as an entry there? For example, if the user disables the top navigation then the Search menu (which is normally there) moves to the Navigation drawer. This could behave the same way, so it's accessible to users that want it semi-frequently, but don't want to give up an entire tab for it.

I see, I think that makes sense.

nikclayton:
That is looking pretty sweet. I haven't looked at the code yet or your other replies, but some quick feedback on the new UI:

1. My bad, there's probably no need to repeat the dates in each chart's view. Could they be added to a single view above the list of charts?

I thought about in the title bar, but there may not be enough room, especially in different locales. So above the list probably makes sense. I'm not sure if it should scroll (so it's the first item in the list), or if it should be fixed (so should be a separate view).

2. The title should probably be "Trending hashtags" ("hashtags" for consistency with how else the term is used in the app)

[Although -- I could imagine that the future presentation of this stuff is in a TabLayout with different tabs for hashtags, statuses, and links. Maybe that's worth doing now, even if there's only one tab?]

3. Can the Y-axis of all the charts be scaled to the same maximum (i.e., whatever the largest values is). At the moment, glancing at the charts it looks as though they're all equally popular, which is not the case.

4. Can a tick mark be added on the X-axis to indicate each day?

5. In the examples in the screenshot they all have non-zero values. But the chart lines all (except for "#exxon" for some reason) go back down to zero. Bug?

6. Not sure the FAB makes sense here. Maybe a small "post" icon button on each hashtag listitem, that would create a new post populated with that tag?

Although since clicking the view takes you to the search page for that hashtag, maybe that would be better implemented as a FAB on the hashtag activity, and avoid further content on each listitem?

DavidEdwards:
>there's probably no need to repeat the dates in each chart's view. 
Sounds good. I added to the start of the list. Since otherwise it would take up too much screen real estate in landscape.

> The title should probably be "Trending hashtags"
I will use this for now. The TabLayout would make sense in the future.

> Can the Y-axis of all the charts be scaled to the same maximum (i.e., whatever the largest values is). At the moment, glancing at the charts it looks as though they're all equally popular, which is not the case.
Absolutely. There is already a `proportionalTrending` boolean for the GraphView. I will activate that.

>Can a tick mark be added on the X-axis to indicate each day?
Sure!

>In the examples in the screenshot they all have non-zero values. But the chart lines all (except for "#exxon" for some reason) go back down to zero. Bug?
I will double check this.

>Not sure the FAB makes sense here. Maybe a small "post" icon button on each hashtag listitem, that would create a new post populated with that tag? Although since clicking the view takes you to the search page for that hashtag, maybe that would be better implemented as a FAB on the hashtag activity, and avoid further content on each listitem?
When you click on the tag, it takes you to the list where you can create a post with the FAB. Perhaps that is enough. I will disable it for now.

DavidEdwards:
>>In the examples in the screenshot they all have non-zero values. But the chart lines all (except for "#exxon" for some reason) go back down to zero. Bug?
>
>I will double check this.

Looks like I found a bug. 

However, there are commonly troughs on the latest day, especially early in the day when the value isn't as populated yet as the previous day.

New screenshots available:

darkmode

lightmode

app/src/main/java/com/keylesspalace/tusky/components/trending/TrendingPagingAdapter.kt
```diff
@@ -0,0 +1,147 @@
...
+                return false // Items are different always. It allows to refresh timestamp on every view holder update
+            }
+
+            override fun getChangePayload(
+                oldItem: TrendingViewData,
+                newItem: TrendingViewData
+            ): Any? {
+                return if (oldItem == newItem) {
+                    // If items are equal - update timestamp only
+                    listOf(StatusBaseViewHolder.Key.KEY_CREATED)
```
connyduck, line 141: There are no timestamps shown here, you can scrap the whole change payload business

app/src/main/java/com/keylesspalace/tusky/components/trending/TrendingPagingAdapter.kt
```diff
@@ -0,0 +1,147 @@
...
+    }
+
+    override fun getItemViewType(position: Int): Int {
+        return if (getItem(position) is TrendingViewData.Tag) {
+            VIEW_TYPE_TAG
+        } else if (getItem(position) is TrendingViewData.Header) {
+            VIEW_TYPE_HEADER
+        } else {
+            throw IllegalStateException("Trending view cannot be retrieved due to viewType mismatch.")
+        }
```
connyduck, line 113:
```suggestion
        return if (getItem(position) is TrendingViewData.Tag) {
            VIEW_TYPE_TAG
        } else {
            VIEW_TYPE_HEADER
        } 
```

app/src/main/java/com/keylesspalace/tusky/components/trending/TrendingPagingAdapter.kt
connyduck: yes please clean it up

connyduck (commented):
Thx you very much!

Cut off hashtags are a no-go here. It is the most relevant info. Also why is a gap between # and name, we don't have that anywhere else.


Eventually we should add a library that can draw really nice graphs, for now this is ok I guess.

app/src/main/java/com/keylesspalace/tusky/usecase/TrendingCases.kt
DavidEdwards: Sure, I'll remove it :+1:

app/build.gradle
```diff
@@ -48,7 +48,9 @@ android {
             shrinkResources true
             proguardFiles 'proguard-rules.pro'
         }
-        debug {}
+        debug {
+            applicationIdSuffix '.debug'
+        }
```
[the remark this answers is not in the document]
DavidEdwards: Sure. That was an unintentional commit.

Do you tend to have the one App for development? I quite like having a live app and the development app (for quickly cross-checking).

app/src/main/java/com/keylesspalace/tusky/MainActivity.kt
```diff
@@ -256,7 +258,11 @@ class MainActivity : BottomSheetActivity(), ActionButtonActivity, HasAndroidInje
 
         binding.viewPager.reduceSwipeSensitivity()
 
-        setupDrawer(savedInstanceState, addSearchButton = hideTopToolbar)
+        setupDrawer(
+            savedInstanceState,
+            addSearchButton = hideTopToolbar,
+            addTrendingButton = !accountManager.activeAccount!!.tabPreferences.hasTab(TRENDING),
```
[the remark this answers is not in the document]
DavidEdwards: OK, I can remove the option again.

app/src/main/java/com/keylesspalace/tusky/MainActivity.kt
```diff
@@ -687,6 +743,20 @@ class MainActivity : BottomSheetActivity(), ActionButtonActivity, HasAndroidInje
         updateProfiles()
     }
 
+    private fun refreshComposeButtonState(adapter: MainPagerAdapter, tabPosition: Int) {
```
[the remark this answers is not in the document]
DavidEdwards: @nikclayton What do you think?

Regarding:
> Not sure the FAB makes sense here. Maybe a small "post" icon button on each hashtag listitem, that would create a new post populated with that tag?

app/build.gradle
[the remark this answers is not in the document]
connyduck: Yes I do have green for development and blue as live app.

app/src/main/java/com/keylesspalace/tusky/MainActivity.kt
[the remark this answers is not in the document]
connyduck: Its fine to remove it, was just wondering why

app/src/main/java/com/keylesspalace/tusky/MainActivity.kt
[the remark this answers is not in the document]
connyduck: We should really implement #3010, doesn't matter which way we'll do it, some users will be unhappy

app/build.gradle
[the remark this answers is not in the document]
DavidEdwards: How are you separating them without committing? Ill do the same thing then.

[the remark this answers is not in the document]
connyduck: Not sure what you mean? I just have Android Studio set to the green variant and thats what I get deployed to my device/emulator when developing.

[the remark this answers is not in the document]
DavidEdwards: :facepalm: OK. I didn't see the variants. That does exactly the same thing. Great, thank you for the explanation!

app/src/main/java/com/keylesspalace/tusky/MainActivity.kt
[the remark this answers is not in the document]
DavidEdwards: I will keep the architecture there, since it can be used for #3010. I will set it to be false explicitly though.

app/src/main/java/com/keylesspalace/tusky/MainActivity.kt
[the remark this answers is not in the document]
DavidEdwards: OK, then I will keep it how it is :+1:

app/src/main/res/layout/fragment_trending.xml
```diff
@@ -0,0 +1,54 @@
...
+        android:visibility="gone"
+        app:layout_constrainedHeight="true"
+        app:layout_constraintBottom_toBottomOf="parent"
+        app:layout_constraintLeft_toLeftOf="parent"
+        app:layout_constraintRight_toRightOf="parent"
+        app:layout_constraintTop_toTopOf="parent"
+        tools:visibility="visible" />
+
+    <androidx.core.widget.ContentLoadingProgressBar
+        android:id="@+id/topProgressBar"
```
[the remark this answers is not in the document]
DavidEdwards: Removed :+1:

app/src/main/java/com/keylesspalace/tusky/components/trending/TrendingFragment.kt
```diff
@@ -0,0 +1,371 @@
...
+    override fun isFabVisible() = false
+
+    companion object {
+        private const val TAG = "TrendingF" // logging tag
+        private const val ARG_ENABLE_SWIPE_TO_REFRESH = "enableSwipeToRefresh"
+
+        fun newInstance(): TrendingFragment {
+            val fragment = TrendingFragment()
+            val arguments = Bundle(1)
+            arguments.putBoolean(ARG_ENABLE_SWIPE_TO_REFRESH, true)
```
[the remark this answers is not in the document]
DavidEdwards: OK, good to know! Removed.

app/src/main/java/com/keylesspalace/tusky/components/trending/TrendingFragment.kt
DavidEdwards: FAB will be hidden

app/src/main/java/com/keylesspalace/tusky/components/trending/TrendingPagingAdapter.kt
DavidEdwards: Cleaned up in the method @connyduck suggests later.

app/src/main/java/com/keylesspalace/tusky/components/trending/TrendingPagingAdapter.kt
DavidEdwards: Now there is the header type, so there is a real usage for it.

nikclayton:
Ultimately, @connyduck has final approval. My arguments for the suggested changes:

# Nav drawer

It should have an entry on the navigation drawer because if it doesn't, the only way the user is going to discover this feature is if they explicitly add a tab for it. That's buried in the UI, and is very non-discoverable.

Out of the box, I think useful features like this should be discoverable so that users have a good first impression with the app and are not left confused.

For an example of how this doesn't work look at how there needs to be a FAQ entry for "How do I see the federated timeline?", and my public timeline of posts is filled with me replying to people sending them a link to the FAQ.

We already have preferences to e.g., hide qualitive stats on posts. If there's some concern that users might never want to see trends an additional pref to show trends could be added, default to "On".

# No FAB

The FAB is for the most common or important action on the screen (https://m3.material.io/components/floating-action-button/overview#13a9b2a5-13f2-4617-a08b-e66576df55b0)

When viewing trends, I don't think "Create a new post" is the most common or important action. I think being able to see the trends is.

The FAB blocks the bottom-right trend, and in some cases there's no way to remove that (e.g., the trends take up just enough space to fill the screen without scrolling).

If the user sees a hashtag and wants to create a post using it they can click through to the hashtag (which should do a search), and then there should be a FAB on that page (aside: at the moment there isn't).

DavidEdwards:
It looks like I pushed at the same time as your commented, the current system removes the FAB and does not add to the nav drawer. I left the drawer code as an option for future tickets though.

app/src/main/java/com/keylesspalace/tusky/components/trending/TrendingFragment.kt
```diff
@@ -0,0 +1,306 @@
...
+    override fun onRefresh() {
+        viewModel.invalidate()
+    }
+
+    override fun onViewUrl(url: String) {
+        bottomSheetActivity.viewUrl(url, PostLookupFallbackBehavior.OPEN_IN_BROWSER)
+    }
+
+    override fun onViewTag(tag: String) {
+        startActivity(StatusListActivity.newHashtagIntent(requireContext(), tag))
```
connyduck, line 188: Open it with a slide-in animation please, it is also closed with one. (check e.g. [here](https://github.com/tuskyapp/Tusky/blob/a6b6a40ba6f940f8d612789c92524d5afeff078b/app/src/main/java/com/keylesspalace/tusky/components/search/fragments/SearchFragment.kt#L123) how to do it)

DavidEdwards: Done :+1:

DavidEdwards:
@connyduck Is there anything missing??

Navigation Drawer: We can solve that in [this PR](https://github.com/tuskyapp/Tusky/pull/3199). Currently it will not be added to the nav bar as you requested.

FAB: This will not be added to the trending page.

connyduck (approved):
I'd have a lot of nitpicks that are not really important (are you interested?), but I think generally this is now ok. Will merge after the 21 release.

DavidEdwards:
> I'd have a lot of nitpicks that are not really important (are you interested?), but I think generally this is now ok. Will merge after the 21 release.

Sounds good. Write down what is bugging you, and I can work on that over the meantime. Perhaps this PR could be frozen, if it is fine for merge? A new PR could resolve outstanding small issues.

nikclayton:
Hi @DavidEdwards -- I've taken another review pass. A few things stuck out at me, some code, some UX, and I figured it was easier to experiment and see if they'd work than put you through another round of reading suggestions and having to go off and implement.

They're on https://github.com/nikclayton/Tusky/commits/add-trending-tags (diff is https://github.com/DavidEdwards/Tusky/compare/add-trending-tags...nikclayton:Tusky:add-trending-tags, I can create a PR to merge them in to your branch if you like), which you should be able to merge in and play around with.

Irrespective of the other changes, https://github.com/nikclayton/Tusky/commit/e31303847e5c829cf326ac8f99b15f3a726367bc is a necessary bug fix, otherwise the line width from the layout file isn't interpreted properly.

The UI changes:

- Increase the width of the chart data lines
- Use platform metrics for padding, list item text height (normal, secondary, and small)
- Remove dividers from the list, add an x-axis to each chart instead, to see if that provides enough visual separation
- Make the date format more explicit (day of week and year)
- Format numbers with locale-aware separators ("1,000" instead of "1000", etc)

Here's the before:



And this is the after:

nikclayton:
I've added a change that skips tags the user is actively filtering from their home timeline. This exposed a crash if the list of trending tags was empty, so I've fixed that too.

nikclayton:
I think I may have been wrong to suggest that the default grid should have two columns. I see

```
#songsormoviesabout
beinglost
```

```
#meteorologicalmelod
ies
```

and 

```
#holocaustmemoriald
ay
```

which isn't a great look.

I've just pushed an alternative to my branch, single column in portrait, 3 column when wider, looks like this:



and this:



WDYT?

DavidEdwards:
> which isn't a great look.

That is true, but there will be a point when that will happen anyway. Though I suppose that if we can ensure the 99% cases don't wrap. That is good. 

 I will take a look over the changes when I get a chance, but one thing from the screenshots stands out:

image

The line chart looks to be the complete cell height. I can't guarantee that isn't my fault thought.

nikclayton:
> > which isn't a great look.
> 
> That is true, but there will be a point when that will happen anyway. Though I suppose that if we can ensure the 99% cases don't wrap. That is good.

Yeah. I tried ellipsize and marquee attributes, to see if the tag could scroll if overlong, but couldn't get them to work (possibly user error on my part).

> Incidentally, why the removal of the bold? I think that is good to highlight the important information.

1. Made the text a bit narrower
2. Consistent with text styling in other Android lists
3. I felt that the difference in text size was sufficient to delineate the info

> I will take a look over the changes when I get a chance, but one thing from the screenshots stands out:
> 
> image
> 
> The line chart looks to be the complete cell height. I can't guarantee that isn't my fault thought.

Not your fault, I experimented with that.

1. It saves some vertical space
2. It clearly ties the text to the chart, which is necessary without dividers between them (otherwise you can't at a glance tell if the text is referring to the chart immediately below or immediately above it.

"Why remove the dividers?" you might be thinking.

I'm a big fan of Tufte's "The Visual Display of Quantitative Information" (https://faculty.salisbury.edu/~jtanderson/teaching/cosc311/fa21/files/tufte.pdf), in which he writes (talking about printed material, but it's just as relevant to the screen):

> A large share of ink on a graphic should present data-information, the ink changing as the data change. Data-ink is the non-erasable core of a graphic, the non-redundant ink arranged in response to variation in the represented. [p93, but the whole section from p84 onwards is really worth a read, you'll never look at a badly put together presentation the same way again :-)]

s/ink/pixels/, and you get the data/pixels ratio, the number of pixels that convey data / the total number of pixels in the graphic.

So with these UI experiments I'm trying to see what's the most useful information we can comfortably convey with the fewest pixels. By, e.g., removing the dividers.

The legend might still need some work if data points go under it -- a transparent white background to provide some visual separation maybe -- but I haven't tried that yet.

DavidEdwards:
> It clearly ties the text to the chart, which is necessary without dividers between them

I meant specifically this chart cell. The line will be underneath the text in some cases. That really don't feel correct.

image

`fun setup(start: Date, end: Date) {`
I am usually a fan of multi-line parameter function definitions. It makes it easier to see changes in diffs. Though, with two small variables this is pretty debatable.

`private val binding: ItemTrendingCellBinding`
I don't think that it is necessary to remove the trailing comma. I explicitly add those so that new changes don't cause an additional line in a diff. 

`KMGTPE`
This confuses me. GTPE are very tech terms, and don't really match accounts / uses. For example 100G could appear under accounts. Giga accounts?

`TrendingPagingAdapter` — absolutely, it used to use the Paging3 system. Paging really makes no sense for this one. 

Filters — This is great, I wanted to add this, so you saved me a lot of fiddling.

Line Width — I didn't realise that sp is used for these dimensions. I thought that sp was only for fonts. TIL.

The graphing X Axis makes sense instead of the dividers.

nikclayton:
> > It clearly ties the text to the chart, which is necessary without dividers between them
> 
> I meant specifically this chart cell. The line will be underneath the text in some cases. That really don't feel correct.
> 
> image

Yes -- that's why I think the text views might need a translucent background, to cover that case and ensure they're legible. Or maybe a white stroke around the letters?

It might be worth constructing some fake pathological data  (say, 5 days of the value being 100, and the last two days being 50), and see the UI looks like in that case. If it's unreadable even with a translucent background when pushing the labels back above the chart is probably the simplest solution.

> `fun setup(start: Date, end: Date) {` I am usually a fan of multi-line parameter function definitions. It makes it easier to see changes in diffs. Though, with two small variables this is pretty debatable.

I agree. And if it's going to grow to need many other parameters then the churn in the function body probably dominates the churn caused by changing the arg list.

> `private val binding: ItemTrendingCellBinding` I don't think that it is necessary to remove the trailing comma. I explicitly add those so that new changes don't cause an additional line in a diff.

Normally I would agree with you here as well. Newer versions of ktlint disagree however, and I've learned that if there's a decent formatter / linter then it's generally much simpler to just go with that.

We're not using those newer versions of ktlint via `gradlew ktlintCheck` yet, but it'll be an issue when we do. And Android Studio ktlint plugins are using the new versions of ktlint now, so I figured it was good to be ahead of the curve, ready for when we do update the ktlint version (https://github.com/tuskyapp/Tusky/issues/2914).

> `KMGTPE` This confuses me. GTPE are very tech terms, and don't really match accounts / uses. For example 100G could appear under accounts. Giga accounts?

Think of it as extreme futureproofing :-) Realistically I can't imagine we need more than "K" at the moment, but I didn't think it hurt to include.

> Line Width — I didn't realise that sp is used for these dimensions. I thought that sp was only for fonts. TIL.

You can use `sp` for anything (I think). The key property is that it will scale if the user changes the font size on their device.

I suspect that if the user's changing their font size then anything else that's thin lines should probably scale as well.

But `dp` would also work, and would scale if the user uses Android settings to scale their display.

DavidEdwards:
> > > It clearly ties the text to the chart, which is necessary without dividers between them
> > 
> > 
> > I meant specifically this chart cell. The line will be underneath the text in some cases. That really don't feel correct.
> > image
> 
> Yes -- that's why I think the text views might need a translucent background, to cover that case and ensure they're legible. Or maybe a white stroke around the letters?
> 
> It might be worth constructing some fake pathological data (say, 5 days of the value being 100, and the last two days being 50), and see the UI looks like in that case. If it's unreadable even with a translucent background when pushing the labels back above the chart is probably the simplest solution.

I see, that might work. 

> > `private val binding: ItemTrendingCellBinding` I don't think that it is necessary to remove the trailing comma. I explicitly add those so that new changes don't cause an additional line in a diff.
> 
> Normally I would agree with you here as well. Newer versions of ktlint disagree however, and I've learned that if there's a decent formatter / linter then it's generally much simpler to just go with that.
> 
> We're not using those newer versions of ktlint via `gradlew ktlintCheck` yet, but it'll be an issue when we do. And Android Studio ktlint plugins are using the new versions of ktlint now, so I figured it was good to be ahead of the curve, ready for when we do update the ktlint version (#2914).

Oh, that is news to me. I had just persuaded my colleagues to start adding trailing commas too. :disappointed: 

> > `KMGTPE` This confuses me. GTPE are very tech terms, and don't really match accounts / uses. For example 100G could appear under accounts. Giga accounts?
> 
> Think of it as extreme futureproofing :-) Realistically I can't imagine we need more than "K" at the moment, but I didn't think it hurt to include.

No I don't mean it is bad. The concept is fine! I don't think the technical terms make sense though. K is often used for thousands, M for millions, B for billions, for example. In terms of generic quantities, I think using G would confuse a lot of people, since we aren't talking about bytes, for example. Am I wrong in this?

> > Line Width — I didn't realise that sp is used for these dimensions. I thought that sp was only for fonts. TIL.
> 
> You can use `sp` for anything (I think). The key property is that it will scale if the user changes the font size on their device.
> 
> I suspect that if the user's changing their font size then anything else that's thin lines should probably scale as well.
> 
> But `dp` would also work, and would scale if the user uses Android settings to scale their display.

Ok, fair enough.

nikclayton:
OK, one more, and then I really am done. The spiky lines bugged me, so I made them bezier's instead for a bit of a curve. WDYT?

nikclayton:
OK, one last set, and then I really, really am done :-) 

- Sort the tags by popularity, most uses first
- Make it clear in the legend that it's totals, not the current value
- Hide FAB
- Show the current values at the end of the chart

I'm not sure about that last one -- sometimes it works quite well (e.g., in the first row), other times it looks a bit clunky, like the second one

nikclayton:
Irrespective of the UI stuff, there's a bug where the menu item is not added to the drawer if trending isn't present as a tab. I'm just taking a look at that now.

nikclayton:
@DavidEdwards Haven't heard from you, and we wanted to move forward with this, so I've rolled everything up in to https://github.com/tuskyapp/Tusky/pull/3315 and in to a fully mergeable state.

If you have the time, or disagree with any of the feedback and want to continue on with this PR please say so.

DavidEdwards:
It has been a busy couple of weeks. I'll make some time Monday evening to look things over.

DavidEdwards:
@nikclayton It looks good. I like how it has turned out, especially the bezier curves. 

Should I close this PR?

nikclayton:
> @nikclayton It looks good. I like how it has turned out, especially the bezier curves.
> 
> Should I close this PR?

Depends how you feel about attribution in GitHub I think.

1. If you merge my branch in to yours (in this PR), and this PR gets submitted that shows as a contribution from you.

2. If 3315 is merged I think it shows as a contribution from me.

My preference is #1, you've done the bulk of the work, I just tidied up a few things.

But if you haven't got time to do the merging, 3315 is ready for merging now.

DavidEdwards:
Then I will merge it into my branch during my lunch today. Should be within the next hour or so.

PR #3151, opened 2023-01-08
Update FilemojiCompat to version 3.2.7

C1710:
- The same version as 3.2.6 on GitHub
- Should fix the NPE on `loadStoredPacks` (again)


connyduck (approved):
Can confirm the fix is now in, thank you very much!

Issue #3146, opened 2023-01-07
Prefer /api/v2/instance


nikclayton:
`/api/v1/instance` is deprecated, and returns a `V1::Instance` (https://docs.joinmastodon.org/methods/instance/#v1).

`/api/v2/instance` is not deprecated, and also returns a `V1::Instance` (https://docs.joinmastodon.org/methods/instance/#v2).

The `getInstance()` API method should probably be split in to a v1 and v2 versions, and `getInstance()` should call the v2 method first, falling back to the v1 method if the v2 one fails.

Not all instances support both v2 and v1. For example:

- https://post.lurk.org/api/v2/instance (works)
- https://post.lurk.org/api/v1/instance (returns a non-JSON error page)

* * * *
- Tusky Version: 20
- Mastodon instance (if applicable): post.lurk.org

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

aaaaargZombies:
Thanks for your help making this issue, I showed it to the admins of the instance. Two of them are using tusky 20 and not experiencing this issue, see thread:

https://post.lurk.org/@az/109647424624580477

very strange

nikclayton:
When the API call succeeds, Tusky caches the result.

If the API call fails, *and* it previously succeeded, *and* the result was cached, the old result was used.

So it sounds like the cache on your device doesn't contain the information. Then Tusky falls back to the 500 character default.

PR #3153, opened 2023-01-08
Support /api/v1/instance and /api/v2/instance

nikclayton:
Start to split the API code in to v1 and v2 interfaces.

Migrate the getInstance() function (rename to instance() for consistency with most other API functions) to v1, and create a v2 equivalent. At the moment they're identical because they return the same type.

Modify InstanceInfoRepository to prefer the v2 call, falling back to v1 if that fails (then the cache, or defaults, per previous behaviour).

Create factory methods for InstanceInfo and InstanceEntity to keep related code together, ditto for defaults.

Update consumers and tests.

Fixes https://github.com/tuskyapp/Tusky/issues/3146

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/network/MastodonApi.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/network/MastodonApi.kt
@@ -26,7 +26,6 @@ import com.keylesspalace.tusky.entity.DeletedStatus
 import com.keylesspalace.tusky.entity.Emoji
 import com.keylesspalace.tusky.entity.Filter
 import com.keylesspalace.tusky.entity.HashTag
-import com.keylesspalace.tusky.entity.Instance
 import com.keylesspalace.tusky.entity.Marker
 import com.keylesspalace.tusky.entity.MastoList
 import com.keylesspalace.tusky.entity.MediaUploadResult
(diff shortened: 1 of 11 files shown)


nikclayton:
Keeping this as a draft, to highlight and discuss a couple of points.

First, obviously having `MastodonApi`, `MastodonApiV1`, and `MastodonApiV2` is a bad idea. For ease of code review I didn't want to migrate everything from `MastodonApi` in to one of the other two until there's a chance to discuss this. So this just deals with the `instance()` method at the moment.

Second, I'm not convinced about the package location.

Instead of `network`, I'm thinking there should be `api.v1` and `api.v2` packages. These packages would contain both the API definitions, and also the classes that encode the bodies of requests and responses.

The `instance()` method was straightforward, as both the v1 and v2 APIs (https://docs.joinmastodon.org/methods/instance/) return a `V1::Instance` (https://docs.joinmastodon.org/entities/Instance/), so I didn't need to create a different response type.

But e.g,. the `filters` API call returns different types for `v1` and `v2` calls, so they'll need to be kept separate.

So perhaps packages:

- `api.v1.request`
- `api.v1.response`
- `api.v2.request`
- `api.v2.response`

would make sense, and move everything under there?

That could be done as a big-bang migration, or on a call-by-call basis (multiple PRs, each PR dealing with a different API call).

app/src/main/java/com/keylesspalace/tusky/network/MastodonApi.kt
```diff
@@ -77,12 +76,10 @@ interface MastodonApi {
         const val PLACEHOLDER_DOMAIN = "dummy.placeholder"
     }
 
+
```
nikclayton, line 79:
```suggestion
```

PR #3160, opened 2023-01-10
Fix media preview bugs

connyduck:
This fixes two issues
- blurhashes sometimes not being scaled to fit the imageview

- ALT labels being displayed when the media preview is hidden

I have also been seeing layout shifts after media was loaded, I can no longer reproduce this though.
Not 100% sure why the `doOnLayout` was added in the first place, it defintely causes more problems than it solves.

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/adapter/StatusBaseViewHolder.java
+++ b/app/src/main/java/com/keylesspalace/tusky/adapter/StatusBaseViewHolder.java
@@ -416,57 +416,56 @@ public abstract class StatusBaseViewHolder extends RecyclerView.ViewHolder {
         return ImageLoadingHelper.decodeBlurHash(this.avatar.getContext(), blurhash);
     }
 
-    private void loadImage(View wrapper,
-                           MediaPreviewImageView imageView,
+    private void loadImage(MediaPreviewImageView imageView,
                            @Nullable String previewUrl,
                            @Nullable MetaData meta,
                            @Nullable String blurhash) {
 
         Drawable placeholder = blurhash != null ? decodeBlurHash(blurhash) : mediaPreviewUnloaded;
 
-        ViewKt.doOnLayout(wrapper, view -> {
-            if (TextUtils.isEmpty(previewUrl)) {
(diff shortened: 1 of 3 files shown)


Tak:
The changes look reasonable, I'll try it tomorrow.
There's no regression in https://github.com/tuskyapp/Tusky/pull/2942#issuecomment-1352175369 ?

nikclayton:
The `doLayout` ref. reminded me of something I looked at the other day -- the logs will occasionally show "requestLayout() improperly called by..." errors.

This happens when a view calls `requestLayout()` in the middle of an existing layout pass. If you dig in to the framework code that logs that (`ViewRootImpl.performLayout()`) you'll see this comment:

```
                // requestLayout() was called during layout.
                // If no layout-request flags are set on the requesting views, there is no problem.
                // If some requests are still pending, then we need to clear those flags and do
                // a full request/measure/layout pass to handle this situation.
```

which explains why this is a bad thing.

I tracked the source of the problem down to `MediaPreviewImageView` and the calls to `setScaleType()` inside `setFocalPoint()` and `removeFocalPoint()`. If you drill down in to the framework source for `setScaleType()`  you see:

```
    public void setScaleType(ScaleType scaleType) {
        if (scaleType == null) {
            throw new NullPointerException();
        }

        if (mScaleType != scaleType) {
            mScaleType = scaleType;

            requestLayout();  // <--- BOOM
            invalidate();
        }
    }
```

Couple that with the `doOnLayout` block which does:

```
        ViewKt.doOnLayout(wrapper, view -> {
            if (TextUtils.isEmpty(previewUrl)) {
                imageView.removeFocalPoint();  // <--- BOOM
        ...
```

there's at least one smoking gun there.

I did very brief testing, changing that to `doOnAttach`, which removed the errors, but I didn't investigate what any other effects of that would be.

connyduck:
> There's no regression in https://github.com/tuskyapp/Tusky/pull/2942#issuecomment-1352175369 ?

Can't rule it out, but I played around a lot on two devices now and seems fine.

Tak (approved):

Issue #3161, opened 2023-01-10
App crashing when opening tabs settings


LuisMayo:
In settings, while trying to open the "tabs" setting in the "filter" category the app crashes to the timeline

* * * *
- Tusky Version: Tusky Test/Nightly 20.0 c650ca93
- Android Version: MIUI 13.2.2 (Android 13 TP1A.220624.014)
- Android Device: Xiaomi Poco X4 GT
- Mastodon instance (if applicable): N/A

- [X] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

connyduck:
Thanks for reporting!
It is a regression from #3090 
Fragment names are hardcoded but obfuscated for release builds, referenced fragments can't be found anymore -> 💥 
cc @nikclayton can you fix this please?

nikclayton:
Looking...

PR #3162, opened 2023-01-10
Keep all subclasses of PreferenceFragmentCompat

nikclayton:
PreferenceFragment references them by string name, which doesn't work after ProGuard has obfuscated the code in release mode. The name is no longer valid and the app crashes.

Fixes https://github.com/tuskyapp/Tusky/issues/3161

### DIFF
--- a/app/proguard-rules.pro
+++ b/app/proguard-rules.pro
@@ -111,3 +111,7 @@
     static void checkReturnedValueIsNotNull(java.lang.Object, java.lang.String, java.lang.String);
     static void throwUninitializedPropertyAccessException(java.lang.String);
 }
+
+# Preference fragments can be referenced by name, ensure they remain
+# https://github.com/tuskyapp/Tusky/issues/3161
+-keep class * extends androidx.preference.PreferenceFragmentCompat
--- a/app/src/main/java/com/keylesspalace/tusky/components/preference/ProxyPreferencesFragment.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/components/preference/ProxyPreferencesFragment.kt
@@ -16,7 +16,6 @@
 package com.keylesspalace.tusky.components.preference
 
 import android.os.Bundle
-import androidx.annotation.Keep
 import androidx.preference.PreferenceFragmentCompat
 import com.keylesspalace.tusky.R
 import com.keylesspalace.tusky.settings.PrefKeys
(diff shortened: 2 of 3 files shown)


nikclayton:
Definitely works, but at the moment I consider this a band aid.

I thought that replacing:

```
            preferenceCategory(R.string.pref_title_timeline_filters) {
                preference {
                    setTitle(R.string.pref_title_post_tabs)
                    fragment = "com.keylesspalace.tusky.components.preference.TabFilterPreferencesFragment"
                }
            }
```

with

```
                    fragment = TabFilterPreferencesFragment::class.java.name
```

might work, but that crashes with variants of:

> androidx.fragment.app.t: Unable to instantiate fragment a4.q: make sure class name exists, is public, and has an empty constructor that is public

I'll dig in some more tomorrow and see if there's a better fix.

One thought -- have you considered disabling the name mangling for release builds? It's not like this is a closed source app where real names might aid someone in reverse engineering the code. As a bonus, I imagine it would make stack traces from release builds that come through the Play Store more readable too (unless that stores a name-demangling table).

connyduck:
If it fixes the problem 🤷 
Do you have a better fix or should I merge?

> One thought -- have you considered disabling the name mangling for release builds? It's not like this is a closed source app where real names might aid someone in reverse engineering the code. As a bonus, I imagine it would make stack traces from release builds that come through the Play Store more readable too (unless that stores a name-demangling table).

The main reason for obfuscation in release builds is that it makes apks smaller, which is a nice feature. I included the mapping file on all recent releases, it is easy to deobfuscate stacktraces with it. The ones I get on Google Play are auto deobfuscated anyway.

nikclayton:
> Do you have a better fix or should I merge?

It's slightly better now -- this way when a new preference fragment is added the author doesn't have to remember to add the `@Keep` annotation.

There's no method to launch a pref fragment by class, or similar, and https://developer.android.com/develop/ui/views/components/settings/organize-your-settings shows using a class-name-as-string, so I think this is the best approach.

app/proguard-rules.pro
```diff
@@ -111,3 +111,7 @@
     static void checkReturnedValueIsNotNull(java.lang.Object, java.lang.String, java.lang.String);
     static void throwUninitializedPropertyAccessException(java.lang.String);
 }
+
+# Preference fragments can be referenced by name, ensure they remain
+# https://github.com/tuskyapp/Tusky/issues/3161
+-keep class * extends androidx.preference.PreferenceFragmentCompat
```
connyduck, line 117: `-keepnames` should be enough

nikclayton: It's not, because the fragments are referenced by string only. If you use `-keepnames` Proguard doesn't realise that the class is actually used, and removes it during the shrinking phase.

connyduck: meh

Issue #2767, opened 2022-11-06
Revert to using browser redirects for auth (or at least provide the option)


virtualdxs:
Asking users to enter third-party credentials inside a webview is bad practice for a few reasons:

- The user has to trust your app not to harvest the credentials (and it teaches users to implicitly trust that with all apps)
- The user has to re-enter their credentials regardless of whether they're logged in in their browser (especially relevant with SSO-integrated instances)
- Webauthn cannot be used, preventing login to any instances/accounts that require it
- At least for me, Android's built-in password manager doesn't work with it
- The user does not have control of the webview, preventing them from working around issues like #2703

The first two were mentioned in the commit message for 4d8289b2. Providing the option to use a browser redirect would resolve some of these, but ideally the webview auth feature is removed entirely as it is considered bad practice for security and UX reasons.

It has also been mentioned that reverting this would help work around issues such as #2462.

Related: PR #2371 which introduced the change

* * * *
- Tusky Version: 19.0
- Android Version: 12
- Android Device: Galaxy S10
- Mastodon instance (if applicable): N/A

- [X] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

connyduck:
Yes we need to do something, logging in with the WebView seems to cause all kind of problems. 
Logging in with the browser was also not perfect, mainly because it only worked with Chrome or Firefox.
So we will probably need some kind of option or automatic detection.

virtualdxs:
Interesting, do you know why it doesn't work with other browsers?

As a stopgap, I'd suggesting adding a menu option to the current webview that redirects to the browser.

connyduck:
They did not redirect back to Tusky

virtualdxs:
Looks like the official app uses [Custom Tabs](https://developer.chrome.com/docs/android/custom-tabs/). Seems like a good solution - wide browser support, callback whenever a navigation event happens, and shares cookies etc. with the user's browser.

connyduck:
We had that as well https://github.com/tuskyapp/Tusky/blob/841373e94700a65594b5f194a550e22949cb8380/app/src/main/java/com/keylesspalace/tusky/LoginActivity.kt#L339

virtualdxs:
If you were using custom tabs, why did you need the browser to redirect back to you? From my understanding as long as they stay in the custom tab you should be able to capture the final redirect event, grab the token from it, and close the custom tab. (Apologies if I'm going off an incorrect assumption; I've never worked with this myself)

[...]

Issue #2703, opened 2022-09-19
Cannot add account


sl1txdvd:
When I try to add another (already existing) account, I enter the domain name and the result is "Die Anmeldeseite konnte nicht geladen werden" (can't load the login or register page). I tried several existing instances.

* * * *
- Tusky Version: 19 (f-droid)
- Android Version: LineageOS 19
- Mastodon instance (if applicable): tried mastodon.social, mastodon.xyz, pixelfed.de

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

retiolus:
I'm having the same issue, can't log in, all instances concerned.
![image](https://user-images.githubusercontent.com/65604466/197570188-02bb5a24-93df-4245-9fa9-faf14663f214.png)

sl1txdvd:
So it worked for me one or two days ago and I thought it got mysteriously fixed without an update, but now it's broken again (for all servers I tried). Is there some central place where the request goes before being sent to the actual instance? This is really weird and annoying.

connyduck:
Do you have some Dns based network filters or something similar in place?

sl1txdvd:
I use Netguard app (VPN) and in Tusky I use a proxy on localhost (Tor/Orbot). But I tried without VPN and proxy, via wifi, via mobile. No difference.

I managed to get screenshots of the error page that shows up in he background for a tiny moment, and it says (translated from German)

Website not available
The website at https://instance.url/oauth/authorize?client_id=xxx[…] could not be loaded because
net::ERR_FAILED

![Screenshot_20221031-151959_Tusky_1_1_1](https://user-images.githubusercontent.com/113895770/199033426-82e0f44d-3b08-4afc-9ba8-fb9480286da7.png)

connyduck:
Yes _something_ causes the page to fail loading and then Tusky closes the window because it can't proceed. We need to find what that something is. It is not a general problem with mastodon.social oder pixelfed.de because they both work for me.

sl1txdvd:
I have no problem accessing the (partial without all parameters) url in my browser (getting redirected to https://instance.url/auth/sign_in for web signin) so it does not seem to be a general problem with my connection.

retiolus:
I fixed the problem by uninstalling the app and reinstalling it...

sl1txdvd:
Thanks for that. I did not want to but now did the same now. I was able to set up all four accounts.

connyduck:
o__0

[...]

Issue #2534, opened 2022-05-16
Cannot authenticate with pawoo.net


meyou69:
When entering the instance name "pawoo.net" and pressing the ”LOGIN WITH MASTODON" button, the application takes me to an empty page afterwards with just the back button in it and then, after like 2 seconds, it turns me back to the "Which instance?" form page. Basically, I cannot login to my pawoo.net account

* * * *
- Tusky Version: 17.0
- Android Version: 11
- Android Device: OnePlus 5T
- Mastodon instance (if applicable): pawoo.net

- [ x ] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

connyduck:
Can you try with 18.0 beta 1 please? We made some improvements to the login that could fix this, at least it should give you an error message now.

meyou69:
I would love to but can you please explain how to install that particular version? Both the Play store and F-Droid only have version 17 available

meyou69:
Never mind, I eventually found the Tusky Test app on the Play store, which, I believe is at version 18.
I installed it but I'm getting a different error now. Please see the attached screenshot
![Screenshot_20220516-162155_Tusky_Test](https://user-images.githubusercontent.com/92801613/168602584-c6916038-2860-4313-a5da-fe138b055c30.png)

lost-geographer:
Hello, I might have a similar problem : I got a "could not load the login page" error, while I was trying to login to the framapiaf.org server. I tried the 18 and 19b version of tusky. I have Android 11.

lost-geographer:
> Hello, I might have a similar problem : I got a "could not load the login page" error, while I was trying to login to the framapiaf.org server. I tried the 18 and 19b version of tusky. I have Android 11.
> 
> Edit : I tried another server like pouet.chapril.org and it works. It might be a server misconfigurarion ?

I found a solution : actually, I use a private ad-blocking dns that blocks the stats.framasoft.org domain (which hosts an instance of the Matomo analytics software), which prevents the login page from loading. Reverting temporarily to the default DNS while connecting to the framapiaf.org server solved the issue. The stats.framasoft.org is blocked also by ad-block lists like  StevenBlack, Energized Protection and oisd full.

connyduck:
> private ad-blocking dns

[...]

Issue #2462, opened 2022-04-27
Tusky doesn't remember SAML account logins


kornysietsma:
If I add an instance with a SAML login, such as toot.thoughtworks.com, Tusky goes through the authentication process but then loads the instance in a browser window not in the app - see screenshot.

![Screenshot_20220427-152316](https://user-images.githubusercontent.com/17765/165541773-013e8aed-77d2-49c4-b3fe-4c79e869d397.png)

The new instance is not added to Tusky, if I press the left arrow, or quit and re-run Tusky, it does not show the new instance as an option.

* * * *
- Tusky Version: 17.0

- Android Version: 11
- Android Device: Pixel 2
- Mastodon instance (if applicable): toot.thoughtworks.com

- [x ] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

connyduck:
This is very weird, thanks for reporting. I think I'll need a SAML login on some instance in order to debug this, what is the easiest way to do that?

jz8132543:
I also find it strange that I can log in probabilistically.

jz8132543:
> This is very weird, thanks for reporting. I think I'll need a SAML login on some instance in order to debug this, what is the easiest way to do that?

I had the same problem, but I'm using OpenID. I can provide you with an account and provide my profile, hopefully the issue will be resolved.Thank you for your attention to this issue!

Steffo99:
I have the same problem, and I can provide a SAML login and example instance to @connyduck if needed.

connyduck:
Yes please tusky@connyduck.at @Steffo99

Steffo99:
@connyduck I've sent you the credentials! Thanks in advance for your time :)

connyduck:
Thanks, I can reproduce it and I think it is an issue with Mastodon itself since it also occurs in other apps https://github.com/mastodon/mastodon/issues/18481

Steffo99:
Strangely enough, I've been able to login with all the other apps, so there must be something strange going on...

Steffo99:
I'm trying to debug the issue as well, and managed to reproduce the problem even on a desktop browser.

I'm trying to understand what the official Mastodon app does during login which Tusky does not.

https://github.com/mastodon/mastodon-android/blob/master/mastodon/src/main/java/org/joinmastodon/android/OAuthActivity.java

So far, I haven't been able to spot any difference...

connyduck:
[...]

PR #3165, opened 2023-01-11
Reinstate optional login via custom browser tab

Tak:
#2767 
Maybe #2703
Maybe #2534
Maybe #2462

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -97,6 +97,7 @@
     <string name="action_more">More</string>
     <string name="action_compose">Compose</string>
     <string name="action_login">Log in with Mastodon</string>
+    <string name="action_browser_login">Login in Browser</string>
     <string name="action_logout">Log out</string>
     <string name="action_logout_confirm">Are you sure you want to log out of the account %1$s?</string>
     <string name="action_follow">Follow</string>
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -4,11 +4,11 @@
     <string name="error_network">A network error occurred! Please check your connection and try again!</string>
     <string name="error_empty">This cannot be empty.</string>
     <string name="error_invalid_domain">Invalid domain entered</string>
-    <string name="error_failed_app_registration">Failed authenticating with that instance.</string>
+    <string name="error_failed_app_registration">Failed authenticating with that instance. If this persists, try "Login in Browser" from the menu.</string>
     <string name="error_no_web_browser_found">Couldn\'t find a web browser to use.</string>
-    <string name="error_authorization_unknown">An unidentified authorization error occurred.</string>
-    <string name="error_authorization_denied">Authorization was denied.</string>
-    <string name="error_retrieving_oauth_token">Failed getting a login token.</string>
+    <string name="error_authorization_unknown">An unidentified authorization error occurred. If this persists, try "Login in Browser" from the menu.</string>
+    <string name="error_authorization_denied">Authorization was denied. If you\'re sure that you supplied the correct credentials, try "Login in Browser" from the menu.</string>
+    <string name="error_retrieving_oauth_token">Failed getting a login token. If this persists, try "Login in Browser" from the menu.</string>
     <string name="error_loading_account_details">Failed loading account details</string>
     <string name="error_could_not_load_login_page">Could not load the login page.</string>
     <string name="error_compose_character_limit">The post is too long!</string>
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -562,6 +562,8 @@
         Poll with choices: %1$s, %2$s, %3$s, %4$s; %5$s
     </string>
     <string name="description_post_language">Post language</string>
+    <string name="description_login">Works in most cases. No data is leaked to other apps.</string>
+    <string name="description_browser_login">May support additional authentication methods, but requires a supported browser.</string>
 
     <string name="hint_list_name">List name</string>
 
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -96,8 +96,8 @@
     <string name="action_unbookmark">Remove bookmark</string>
     <string name="action_more">More</string>
     <string name="action_compose">Compose</string>
-    <string name="action_login">Log in with Mastodon</string>
-    <string name="action_browser_login">Login in Browser</string>
+    <string name="action_login">Login with Tusky</string>
+    <string name="action_browser_login">Login with Browser</string>
     <string name="action_logout">Log out</string>
     <string name="action_logout_confirm">Are you sure you want to log out of the account %1$s?</string>
     <string name="action_follow">Follow</string>
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -38,7 +38,18 @@
         </activity>
         <activity
             android:name=".components.login.LoginActivity"
-            android:windowSoftInputMode="adjustResize">
+            android:windowSoftInputMode="adjustResize"
+            android:exported="true">
+            <intent-filter>
+                <action android:name="android.intent.action.VIEW" />
+
+                <category android:name="android.intent.category.DEFAULT" />
+                <category android:name="android.intent.category.BROWSABLE" />
+
+                <data
+                    android:host="${applicationId}"
+                    android:scheme="@string/oauth_scheme" />
+            </intent-filter>
         </activity>
         <activity android:name=".components.login.LoginWebViewActivity" />
         <activity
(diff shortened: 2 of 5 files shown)


connyduck:
Definitely a workaround for 2462, not sure about the other ones.

Works nicely, thanks.

Only thing I'm not sure about is the wording. The two options are "WebView" and "Browser" and not "Mastodon" and "Browser". It is always Mastodon (or whatever other platform).

Maybe "Login with Tusky" and "Login with Browser" would be better?

app/src/main/res/values/strings.xml
```diff
@@ -97,6 +97,7 @@
     <string name="action_more">More</string>
     <string name="action_compose">Compose</string>
     <string name="action_login">Log in with Mastodon</string>
+    <string name="action_browser_login">Login in Browser</string>
```
nikclayton, line 100:
```suggestion
    <string name="action_browser_login">Log in in Browser</string>
```

nikclayton:
If I'm a user seeing this for the first time, what information do I need to know in order to make an informed choice?

Also, I'm a bit confused after reading https://github.com/tuskyapp/Tusky/issues/2767, which seems to make a strong case that the WebView option should go away entirely. The other issues don't make a case for keeping it either.

If it did go away the user wouldn't need to have a choice. Why is it still needed?

connyduck:
Browser login does not work on all browsers, some users have very old or exotic ones. WebView login does not work for some 2-factor-auth schemes.

nikclayton:
> Browser login does not work on all browsers, some users have very old or exotic ones. WebView login does not work for some 2-factor-auth schemes.

Thanks. @Tak please could that be added as motivating comments in the code.

I guess it's too late in the day to consider https://developer.android.com/training/id-auth/authenticate for this release. Is that on anyone's radar though?

I may have a solution that's workable for this though.

I'm going to guess that if the user has a browser on their device that is new enough to support the Custom Tabs protocol it's new enough to do OAuth properly.

https://developer.chrome.com/docs/android/custom-tabs/integration-guide/#how-can-i-check-whether-the-android-device-has-a-browser-that-supports-custom-tab explains how to check if such a browser is installed.

If it is, I suggest using it, and hiding the webview option if a custom-tabs supporting browser is installed.

This is probably the very common case.

You mentioned browser log in doesn't work on some old or exotic ones. Is there a list (even a partial one) of the package names? If one of those is on the device then the user could be warned appropriately and the webview option presented first.

A handful I checked:

- com.stoutner.privacybrowser.standard (https://github.com/tuskyapp/Tusky/issues/2040)
- Tor Browser, maybe (https://github.com/tuskyapp/Tusky/issues/1609)
- Default browser on Lineage OS (maybe, this is from 2018), https://github.com/tuskyapp/Tusky/issues/682
 
---

Separately, the FAQ should be updated to cover this.

connyduck:
I'm wondering what we could do to better explain the two buttons for users. We could do tooltips on longpress but I don't think that pattern is common enough for users to find it by themselves.

Something like

Login with Tusky - works in most cases, no data leaked to other apps
Login with Browser - supports 2FA with Webauthn but requires compatible browser

nikclayton:
> > I guess it's too late in the day to consider https://developer.android.com/training/id-auth/authenticate for this release. Is that on anyone's radar though?
> 
> Ugh we used to use this at work, it is a weird api. Don't see what the advantage is unless you want single sign on between multiple applications on one device.

As a user, it lets me:

- Easily sync my account info across multiple devices (I use Tusky on phone and tablet)
- Easily see the accounts my device knows about (in the settings), and modify them from one place
- (future/speculative) If other Mastodon Android apps did this I wouldn't need to re-enter my credentials in to each one

I'll think about this a bit, and possibly create a separate issue, to not derail this one any further.

> > ...
> 
> One problem is when we use custom tabs exclusively is that Google Play could block us again when we don't show the server rules during login...

This is https://support.google.com/googleplay/android-developer/answer/9876937 ?

At the moment the user flow for adding an account is:

1. Enter the instance name, click "Log in with Mastodon"
2. Enter the username and password

What if there's an additional step between those two, that displays the rules from /api/v{1,2}/instance, and requires the user to click through them to proceed?

Would that help?

connyduck:
Looking at it now, I don't think we can do it like this.



The screen feels overwhelming now and I don't think the description makes it clear enough 😞 

I wish we wouldn't have that 🤬 rules requirement from Google Play.
Maybe we can make something like detecting available browsers work. If the available browser is not compatible we fall back to WebView login but display a warning that 2fa might not work. But another screen for the rules, really 😞?

charlag:
Put a three dot option to log in via the other option (whatever is the primary). If we see that it didn't work out we can also suggest in the message.

nikclayton:
> > What if there's an additional step between those two, that displays the rules from /api/v{1,2}/instance, and requires the user to click through them to proceed?
> 
> > Would that help?
> 
> yes but it slows signup down

Is that a big deal if (a) it's a one time thing, and (b) the risk of not doing it is the app being rejected from the Play store?

Is there an additional nuance I'm missing?

connyduck:
> Put a three dot option to log in via the other option (whatever is the primary). If we see that it didn't work out we can also suggest in the message.

I like that idea

> Is that a big deal if (a) it's a one time thing, and (b) the risk of not doing it is the app being rejected from the Play store?

It is not that big of a deal, but I think there are better options

> Edit to add: The only argument I can think to make to Google is that Tusky is in the same boat as an IMAP mail client, and they presumably don't require those to show any rules before the user can log in to their account.

Yeah I tried that, it was like they didn't even read my message. Here is my rant from back then: https://chaos.social/@ConnyDuck/109004198955056199

Tak:
Implemented the suggestion from https://github.com/tuskyapp/Tusky/pull/3165#issuecomment-1384584096

nikclayton:
As I understand it, the goal here is to pick the most appropriate initial web login approach, with a user-accessible fallback if necessary.

In addition, some users may have either (a) browsers that just don't work, (b) webview implementations that just don't work. If we can detect a user in either of those states we can prompt them to install another browser, and explain why.

So I think an improved UX flow would be:

**1**: Figure out the log in technology (enum of CUSTOM_TABS, BROWSER, or WEBVIEW)

a. Get the list of installed packages that support custom tabs (see https://developer.chrome.com/docs/android/custom-tabs/integration-guide/#how-can-i-check-whether-the-android-device-has-a-browser-that-supports-custom-tab for code that does this).

If that list is non-empty, then it's CUSTOM_TABS. Go to **2**

b. Get the complete list of installed browsers (the code at the previous link can do that too). Compare the list against either an allow list of browser we know do work, or a deny list of browsers we know don't (https://github.com/tuskyapp/Tusky/pull/3165#issuecomment-1381087194 has an initial deny list).

If the list of browsers, when filtered to either browsers we know work, or excluding browsers we know don't, is non-empty, then the log in technology is BROWSER. Go to **2**

c. Fall back to WEBVIEW.

**2**. Display the log in UI

a. If CUSTOM_TABS, display something like the current UI. Possibly with a "..." menu to allow the user to force switch to another type, but I'd be surprised if that's necessary.

[Extra points: Use the `warmup()` method to preload the browser process so that the UX after the user starts logging in is faster]

b. If BROWSER, display something like the current UI, but with an additional warning, something like "Some browsers do not work with normal Mastodon log in. If logging in does not work you will need to install an additional web browser. The Tusky team knows that both [Firefox] and [Chrome] work.

c. If WEBVIEW, display something like the current UI, but with an additional warning, something like "Logging in normally requires a web browser. The built-in one on your device may be too old. If logging in does not work you will need to install a web browser. The Tusky team knows that both [Firefox] and [Chrome] work

With the [Firefox] and [Chrome] entries launching intents that will install those apps (see https://developer.android.com/distribute/marketing-tools/linking-to-google-play -- I know those URLs are play.google.com in the intent, but if the user has F-Droid installed it will respond to those intents too, see https://gitlab.com/fdroid/fdroidclient/-/blob/master/app/src/main/java/org/fdroid/fdroid/views/main/MainActivity.java#L260 for confirmation).

## Alternatively

As above, but drop WEBVIEW support entirely, and require the user to install a working browser, with a link to the FAQ that explains the security tradeoff.

---

I think that solves the problems -- the common case will work without needing UI changes, and in the BROWSER and WEBVIEW cases the user is given the tools to fix the problem.

We can argue about what should be in the list of browsers to recommend (I know Firefox and Chrome work, I have not been able to find a definitive list of other Android browsers that support custom tabs)

DavidEdwards:
Some additional points:

1. WebView has security implications #2767 ([google link](https://developers.googleblog.com/2021/06/upcoming-security-changes-to-googles-oauth-2.0-authorization-endpoint.html)) — it should ideally not be an option, or as a last resort.
2. Browser can result in a failure in the case parental controls are active. I had an issue with Amazon Tablets being parental locked and users not being able to login with SSO in a Browser.
3. Chrome Tabs do not (if I remember correctly) have access to the saved logins of the user (especially if they don't use Chrome), so that cannot be streamlined.

I have had success in the case of SSO, using the following strategy:

1. Chrome Custom Tabs as the only displayed login strategy.
2. The user can still click to open their browser if they want to, allowing them mobility to use saved logins.

connyduck:
Perfect is the enemy of good - I will merge this now so we have a solution for users affected by the SSO problem.
Please feel free to send PRs to improve on this.

PR #3169, opened 2023-01-12
Upgrade AndroidX dependencies

connyduck:
and use new `@Upsert` from Room


connyduck:
Uh oh this crashes the app on every placeholder in timelines. But nice that it was caught by tests.

nikclayton:
I haven't checked yet, but it would be worth seeing if the recyclerview update fixes the problem with horizontal swipe sensitivity being too high without needing my hack. The release notes mention better nested scrolling support (and ViewPager2 wraps a RecyclerView around a RecyclerView, which is what causes the problem).

connyduck:
Yes, I also didn't notice a difference

charlag (approved):

Goooler:
Can also add Kotlin 1.8.10 update here.

https://github.com/JetBrains/kotlin/releases/tag/v1.8.10

Goooler:
And if we can enable @renovate-bot for this repo?

connyduck:
> And if we can enable @renovate-bot for this repo?

We can try

PR #3170, opened 2023-01-13
Add Belarusian language to the in-app language picker

xzfantom:


connyduck (changes requested):
Thank you! Please also add it to `app/src/main/res/xml/locales_config.xml`.

connyduck (approved):

PR #3185, opened 2023-01-17
Create and use a dedicated notification channel for send-post failures

nikclayton:
Fixes https://github.com/tuskyapp/Tusky/issues/2953

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/service/SendStatusService.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/service/SendStatusService.kt
@@ -82,6 +82,16 @@ class SendStatusService : Service(), Injectable {
             if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
                 val channel = NotificationChannel(CHANNEL_ID, getString(R.string.send_post_notification_channel_name), NotificationManager.IMPORTANCE_LOW)
                 notificationManager.createNotificationChannel(channel)
+
+                val errorChannel = NotificationChannel(
+                    CHANNEL_ID_ERROR,
+                    getString(R.string.notification_send_error_name),
+                    NotificationManager.IMPORTANCE_HIGH
+                )
+                errorChannel.description = getString(R.string.notification_send_error_description)
+                errorChannel.enableVibration(true)
+                errorChannel.setShowBadge(true)
+                notificationManager.createNotificationChannel(errorChannel)
             }
 
             var notificationText = statusToSend.warningText
@@ -271,7 +281,8 @@ class SendStatusService : Service(), Injectable {
                 R.string.send_post_notification_error_title,
                 R.string.send_post_notification_saved_content,
                 failedStatus.accountId,
-                statusId
+                statusId,
+                channelId = CHANNEL_ID_ERROR
             )
 
             notificationManager.cancel(statusId)
@@ -340,7 +351,8 @@ class SendStatusService : Service(), Injectable {
         @StringRes title: Int,
         @StringRes content: Int,
         accountId: Long,
-        statusId: Int
+        statusId: Int,
+        channelId: String = CHANNEL_ID
     ): Notification {
 
         val intent = Intent(this, MainActivity::class.java)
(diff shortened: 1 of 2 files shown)


nikclayton:
Here's a screenshot of the notification:

nikclayton:
Here's a screenshot of the new notification channel:



Like the normal sending channel this is not tied to a particular account.

nikclayton:
A quick way to test this in a local build is to apply the following change to `MastodonApi.kt`, so that all attempts to send a post will hard fail:

```kotlin
//    @POST("api/v1/statuses")
//    suspend fun createStatus(
//        @Header("Authorization") auth: String,
//        @Header(DOMAIN_HEADER) domain: String,
//        @Header("Idempotency-Key") idempotencyKey: String,
//        @Body status: NewStatus
//    ): NetworkResult<Status>
    @POST("api/v1/instance")
    suspend fun createStatus(
        @Header("Authorization") auth: String,
        @Header(DOMAIN_HEADER) domain: String,
        @Header("Idempotency-Key") idempotencyKey: String,
        @Body status: NewStatus
    ): NetworkResult<Status> {
        throw HttpException(Response.error<String>(404, "wtf".toResponseBody()))
    }
```

nikclayton:
If a post fails to send this also adds a badge to the app icon (if the launcher supports it -- in the emulator you need to use an image with the Play libraries installed for this to work).



A long-press on the app icon will also show the notification in this state:

nikclayton:
Here's a complete interaction loop -- https://www.youtube.com/shorts/cGLFsjKDVKM

nikclayton:
Here's the notification on the navigation drawer

mcclure:
Although this is an improvement on the silent notification for sure, it has the same weird property the silent notification has that notification controls in the Android OS do not appear for it until you receive your first post notification failure.

Screenshot_20230121-181233

Screenshot_20230121-181335

Screenshot_20230121-181416

nikclayton:
That's because the channels are only created when there's a failure.

If they should be pre-emptively created then the code needs to move to, IIRC, TuskyApplication.kt.

I agree that that makes sense. @connyduck any reason not to?

mcclure:
Edited the post above after moving from my phone.

One other small note— because this patch does not incorporate the "failure notification seen" tracking from the other patch, if you get a post-fail and it goes to drafts, the "failed" badge will stay up in the menu until the failed-post draft is either sent or actually deleted. This might actually be correct behavior, I don't know. If this PR is merged for 21, maybe we should consider post-21 if the badge should clear once the user looks at the drafts (and one assumes, makes an affirmative decision at that point whether to repost, delete, or just leave it)

nikclayton:
> After receiving 1 failure notification (notice that "sending posts" and "sending errors" are separate channels, IDK if that's intentional)

It is -- "sending posts" is transient, silent, and will not appear on the screen unless the user pulls down the notification shade (an icon will appear on the notification shade).

"sending errors" is permanent (until dismissed), makes a noise by default, and does immediately appear on the screen without the user having to do anything.

nikclayton:
> because this patch does not incorporate the "failure notification seen" tracking from the other patch, if you get a post-fail and it goes to drafts, the "failed" badge will stay up in the menu until the failed-post draft is either sent or actually deleted. This might actually be correct behavior, I don't know. 

I think it's correct.

We don't know the user's intention when they open the Drafts, so clearing the failure notification just because they've looked strikes me as a dangerous thing to do, because we're deleting state from the app when the user has done something ambiguous.

Providing a UI affordance to allow the user to clear the failure state from a draft (separate from either deleting the draft, or trying to resend it), converting a failed draft in to a regular draft is possible, but needs a bit of UI thought. Essentially, a positive action that marks the failure as "read".

nikclayton:
> That's because the channels are only created when there's a failure.
> 
> If they should be pre-emptively created then the code needs to move to, IIRC, TuskyApplication.kt.
> 
> I agree that that makes sense. @connyduck any reason not to?

Done in https://github.com/tuskyapp/Tusky/pull/3206

PR #3186, opened 2023-01-17
Perform preference schema upgrades at startup

nikclayton:
Over time it can be desirable to change how preferences are interpreted.

Preferences might be removed, or renamed. Or the default value for a preference might be changed.

When this happens it's important that users upgrading from one version to the next (or jumping from one version to several versions ahead) get a consistent experience. In particular:

- Preferences that no longer exist should be deleted
- Preferences that have been renamed should have the old preference values copied over
- If the user used the default value for the preference, and the default has changed, the previous default value should be explicitly set as their value for the preference

To support this, store a SCHEMA_VERSION as a preference. This is not exposed to the user, and corresponds to the app's VERSION_CODE.

If the version code does not match the schema version then this is a newer version of the app with older preferences that may need to be changed.

Those changes will be implemented in `upgradeSharedPreferences`.


nikclayton:
https://gitlab.com/nikclayton/opensudoku/-/blob/material3/app/src/main/java/org/moire/opensudoku/OpenSudoku.java for an example of where I've found this useful.

app/src/main/java/com/keylesspalace/tusky/TuskyApplication.kt
```diff
@@ -91,4 +99,18 @@ class TuskyApplication : Application(), HasAndroidInjector {
     }
 
     override fun androidInjector() = androidInjector
+
+    private fun upgradeSharedPreferences(oldVersion: Int, newVersion: Int) {
```
Lakoja, line 103: I was wondering about this task of preference migration handling over there: https://github.com/tuskyapp/Tusky/pull/3205

Especially I was wondering if there is already a sort of migrations (for the database only) and if such a migration should be also located there.

Both task are "update the stored files from one version to the next".

nikclayton: Eventually I'd like anything that looks like a preference (whether that's something that's an actual Android Preference subclass, or some other setting that the user interacts with but is stored differently) to be accessed via a new `PreferenceRepository`.

This would provide type-safe access to the preferences, and all the storage / migration logic can be stored in one place. I go through some of this in https://github.com/tuskyapp/Tusky/pull/2393#issuecomment-1374580034.

I think trying to do all that once is too big a change, so I'm thinking about incremental PRs that move closer to the goal.

connyduck (approved):

app/src/main/java/com/keylesspalace/tusky/TuskyApplication.kt
Lakoja: However I wouldn't see the normal db content (i.e. status) not as a preference.

(Regarding my question if "all migration should be handled similarly".)

app/src/main/java/com/keylesspalace/tusky/TuskyApplication.kt
```diff
@@ -91,4 +99,18 @@ class TuskyApplication : Application(), HasAndroidInjector {
...
+        Log.d(TAG, "Upgrading shared preferences: $oldVersion -> $newVersion")
+        val editor = sharedPreferences.edit()
+
+        // Future upgrade code goes here
```
Lakoja, line 107: How is this meant to be used?

There should be a number of steps - like in migrations?

app/src/main/java/com/keylesspalace/tusky/TuskyApplication.kt
```diff
@@ -65,15 +68,20 @@ class TuskyApplication : Application(), HasAndroidInjector {
 
         AppInjector.init(this)
 
-        val preferences = PreferenceManager.getDefaultSharedPreferences(this)
+        // Migrate shared preference keys and defaults from version to version. The last
+        // version that did not have a SCHEMA_VERSION was 97, so that's the default.
+        val oldVersion = sharedPreferences.getInt(PrefKeys.SCHEMA_VERSION, 97)
```
Lakoja, line 73: (It's actually 100. But I guess that is not so important.)

app/src/main/java/com/keylesspalace/tusky/TuskyApplication.kt
Lakoja: I have added some usage (fleshing out a loop) in my PR:

https://github.com/tuskyapp/Tusky/pull/3205/commits/cfe37d03808bbb6884112082680e108d374d7e1c

nikclayton: Yes, that's a good approach.

Another hypothetical, in case it's useful.

Suppose we decided to change the default colour scheme from "dark" to "follow system", so that the user's first experience with Tusky matches their expressed preferences.

New users should get the new default, old users who didn't change the preference should get the old default. Otherwise old users who hadn't changed the default would see the colour scheme change.

So the migration path would check if the preference schema is being changed (changing the default value for a preference is considered a schema change), and would need to explicitly set an empty preference value to the old default value.

Ultimately all of this should be hidden behind a PreferenceRepository, but baby steps.

app/src/main/java/com/keylesspalace/tusky/TuskyApplication.kt
connyduck: Wait a sec, that is not working. We release the nightly version on every commit to develop, and it has completely different version codes. Needs another solution @nikclayton

Lakoja: We could still "merge" it with the migration code - or at least something similar. It has a manual "counter".

Lakoja: Ok, probably only something similar. As the `AppModule#providesDatabase` does not really have an explicit location where it is called and where we could add a `doPreferenceUpgrade()`.

Lakoja:
Followup is https://github.com/tuskyapp/Tusky/pull/3324

Issue #3189, opened 2023-01-17
Media upload fails for m4a audio file


connyduck:
Original report: https://mastodon.social/@Teratogenese@mamot.fr/109695671921182591
Steps to reproduce: Try to upload [this file](https://connyduck.at/Ring.m4a). Upload fails with "Video has no video stream".

This happens due to `ContentResolver.getType` returning the wrong mime type after the file was copied.
In [`MediaUploader.prepareMedia`](https://github.com/tuskyapp/Tusky/blob/abca91a420939b058d59b8ccb6bb3ca9fdc96fab/app/src/main/java/com/keylesspalace/tusky/components/compose/MediaUploader.kt#L161) the extension is correctly detected as .m4a and the type as `audio/mp4`. Then the file is copied to the cache dir. And later, in [`MediaUploader.upload`](https://github.com/tuskyapp/Tusky/blob/abca91a420939b058d59b8ccb6bb3ca9fdc96fab/app/src/main/java/com/keylesspalace/tusky/components/compose/MediaUploader.kt#L250), the extension is detected as .mp3 and the mime type as `audio/mpeg`. This makes the file upload with the wrong type and the server can't process it. With the right extension and type upload succeeds.

Not sure why this happens and how to properly fix this. One way would be to forward the initially detected type to the upload, but that is quite a big change as it would propagate all the way to drafts.
Ideas and help very much welcome..

* * * *
- Tusky Version: 20.0
- Android Version: Android 11
- Android Device: Fairphone 4
- Mastodon instance (if applicable): mastodon.social

nikclayton:
Hmm. When I right click on https://connyduck.at/Ring.m4a and do "Save link as..." it prompts me to save it as "Ring.mp3", not "Ring.mp4".

That doesn't change anything about the content of the file (verified by fetching it with `curl`, file sizes are the same and `cmp(1)` reports no differences).

`file(1)` reports `ISO Media, MP4 v2 [ISO 14496-14]` in both cases.

nikclayton:
By the time the file is copied to the cache dir it has the .mp3 extension.

I put a breakpoint on `ComposeViewModel.addMediaToQueue`, and after this line:

```
stashMediaItem = mediaItem`
```

the value of `stashMediaItem.uri` is `content://com...../fileprovider/*/randomTemp<digits>.mp3`

nikclayton:
The conversion to `.mp3` extension is happening in `prepareMedia`.

[...]

PR #3190, opened 2023-01-19
Sniff MIME type for audio/video files before upload

nikclayton:
MIME type detection for files based on extensions (the `getType()` method) returns incorrect results from DownloadProvider and FileProvider for (at least) .m4a files.

Investigation details are in https://github.com/tuskyapp/Tusky/issues/3189

Be safe, and use `MediaMetadataRetriever` to sniff the content of the files to determine the correct type.

Fixes https://github.com/tuskyapp/Tusky/issues/3189


connyduck (approved):
Thank you very much 🙏

PR #3203, opened 2023-01-22
Check for status "type" before casting (class cast exception Placeholder)

Lakoja:
Fixes a crash (class cast of a Placeholder) when tapping a reply in a conversation in the notfications and that conversation is then loaded starting with that reply.

I'm not really sure if this is correct/sufficient there. Or if for example that debug output ("Loaded status from network") is still correct.

@nikclayton 

Stems from: https://github.com/tuskyapp/Tusky/pull/3118
(Unfortunately there is no issue attached there.)

That is the exception: StatusViewData$Placeholder cannot be cast to com.keylesspalace.tusky.viewdata.StatusViewData$Concrete

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/components/viewthread/ViewThreadViewModel.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/components/viewthread/ViewThreadViewModel.kt
@@ -107,7 +107,7 @@ class ViewThreadViewModel @Inject constructor(
             val contextCall = async { api.statusContext(id) }
             val timelineStatus = db.timelineDao().getStatus(id)
 
-            var detailedStatus = if (timelineStatus != null) {
+            var detailedStatus = if (timelineStatus != null && !timelineStatus.status.isPlaceholder) {
                 Log.d(TAG, "Loaded status from local timeline")
                 val viewData = timelineStatus.toViewData(
                     gson,
--- a/app/src/main/java/com/keylesspalace/tusky/components/viewthread/ViewThreadViewModel.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/components/viewthread/ViewThreadViewModel.kt
@@ -107,8 +107,7 @@ class ViewThreadViewModel @Inject constructor(
             val contextCall = async { api.statusContext(id) }
             val timelineStatus = db.timelineDao().getStatus(id)
 
-            // TODO: Investigate how this could ever be the ID of a placeholder status
-            var detailedStatus = if (timelineStatus != null && !timelineStatus.status.isPlaceholder) {
+            var detailedStatus = if (timelineStatus != null) {
                 Log.d(TAG, "Loaded status from local timeline")
                 val viewData = timelineStatus.toViewData(
                     gson,
--- a/app/src/main/java/com/keylesspalace/tusky/components/viewthread/ViewThreadViewModel.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/components/viewthread/ViewThreadViewModel.kt
@@ -107,6 +107,7 @@ class ViewThreadViewModel @Inject constructor(
             val contextCall = async { api.statusContext(id) }
             val timelineStatus = db.timelineDao().getStatus(id)
 
+            // TODO: Investigate how this could ever be the ID of a placeholder status
             var detailedStatus = if (timelineStatus != null && !timelineStatus.status.isPlaceholder) {
                 Log.d(TAG, "Loaded status from local timeline")
                 val viewData = timelineStatus.toViewData(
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/keylesspalace/tusky/components/viewthread/ViewThreadViewModel.kt
```diff
@@ -107,7 +107,7 @@ class ViewThreadViewModel @Inject constructor(
             val contextCall = async { api.statusContext(id) }
             val timelineStatus = db.timelineDao().getStatus(id)
 
-            var detailedStatus = if (timelineStatus != null) {
+            var detailedStatus = if (timelineStatus != null && !timelineStatus.status.isPlaceholder) {
```
nikclayton, line 110:
```suggestion
            // TODO: Investigate how this could ever be the ID of a placeholder status
            var detailedStatus = if (timelineStatus != null && !timelineStatus.status.isPlaceholder) {
```

The change LGTM, at least as far as preventing the crash goes.

Can you reproduce this at will? From looking at the code I can't immediately see a path that would ever result in the ID passed to this function being from a placeholder status. AIUI, the user only gets here by some route that involves clicking on a concrete status.

If you can reproduce it, and can provide reproduction instructions, I can also work on a deeper fix (which shouldn't prevent this PR from being merged in the meantime).

Lakoja:
Regarding reproduction: This only happens for one of my statuses (a private reply).

It even is possible to see it correctly in a conversation view (starting with the first status).
But it will crash when actually starting with this particular status as focus.

So I'm not sure how likely this is generally.
It might have to do with me testing the status loading and maybe it crashed there before a placeholder entry was replaced with the actual data?

nikclayton:
Any chance you can grab a stack trace when this happens?

Lakoja:
That won't help you much, I guess.

It looks like this:

```
java.lang.ClassCastException: com.keylesspalace.tusky.viewdata.StatusViewData$Placeholder cannot be cast to com.keylesspalace.tusky.viewdata.StatusViewData$Concrete
at com.keylesspalace.tusky.components.viewthread.ViewThreadViewModel$loadThread$1.invokeSuspend(ViewThreadViewModel.kt:112)
at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:106)
at android.os.Handler.handleCallback(Handler.java:883)
at android.os.Handler.dispatchMessage(Handler.java:100)
at android.os.Looper.loop(Looper.java:241)
at android.app.ActivityThread.main(ActivityThread.java:7617)
```

I understand that the thing got from the database at that location is expected to be "Concrete" but it isn't.
db.timelineDao().getStatus(id)

nikclayton:
Thanks. Are you able to debug this further?

There's only a few places where `loadThread()` is called. I suspect the chain starts at `ViewThreadActivity.onCreate()`.

This calls `ViewThreadFragment.newInstance()`, and it's here that the ID of the status to load is set so that `loadThread()` will eventually receive it.

Could you:

1. Debug with a break point in `newInstance()`, and see what the status ID is.
2. Use the Android Studio Database Explorer, and find out what the state of that entry is in the database.

The query to use is from `getStatus()`, https://github.com/tuskyapp/Tusky/blob/9cf4882f41164d0c9fdd520e0a5feb588d4142f9/app/src/main/java/com/keylesspalace/tusky/db/TimelineDao.kt#L58

That should return exactly one row, the data should correspond to the status you've just clicked on, and `authorServerId` should **not** be null (because if it is null, that's the signal that it's a placeholder, per https://github.com/tuskyapp/Tusky/blob/9cf4882f41164d0c9fdd520e0a5feb588d4142f9/app/src/main/java/com/keylesspalace/tusky/db/TimelineStatusEntity.kt#L88.

Given what you've described, I'm expecting that if you do that database query you'll see the full status, but for some reason `authorServerId` is null.

If it is, there's something odd about your local database.

If it's not then something weirder is going on.

connyduck:
This is logged in Google Play, 11 crashes/5 users, definitely a release blocker.

What I think is happening: Coincidentally, the id of the status that caused the notification is a placeholder in the home timeline. 

So the fix is correct imho. Another possibility would be telling the ViewThreadViewModel that the Status it is about to load is cached (e.g., openend from home timeline) and only then attempt loading it from the database.

connyduck:
Another possibility to fix it would be to change the query to never return placeholders. (I think I prefer that one)

Lakoja:
I used the database and if I read this correctly the status being loaded in my case is a placeholder (all fields null except the serverId).

The code in ViewThreadViewModel with my fix then switches to the api.
And maybe loads that status correctly (but does not put it into the database).
I cannot tell exactly because debugging finishes with `api.status(id).getOrElse` (display is correct then).

connyduck:
@Lakoja can you change the fix to be on Dao level by adding `AND s.authorServerId IS NOT NULL` to the query please?

Lakoja:
Ok.

Is it correct that most status do no come from the DB even if they are displayed before (as part of a conversation view)?

So this is not a cache for everything - but only main timeline status?

connyduck:
yes, correct

Issue #3204, opened 2023-01-22
Profile settings display is sometimes incorrect


Lakoja:
Noticed in https://github.com/tuskyapp/Tusky/issues/3163

All account based settings ("profile settings") might not correspond to their displayed value in the settings screen.

For example when logging out and in the "new" account will get the default settings. However there might already be a duplicate set of "shared preferences" which are used in the settings screen instead.

They will get in sync after the first change to these settings.

*So to reproduce*

- Change the "Always show sensitive content" to on
- Logout and in
- Open the profile settings screen again
Then this settings will be shown as "on" when in fact the actual account setting used is off.

* * * *
- Tusky Version: Any - at least since 2020

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

Lakoja:
Technical explanation (of the bug):

All preferences use by default the "shared preferences" as the data storage.
`AccountPreferencesFragment would like to use the account db object instead and `tries to initialize the (displayed) settings' value with `isChecked`.
This would actually also set the value in the shared preferences. However during construction the preference manager is not yet set and thus nothing is done there.
There is then afterwards the attaching of the preference manager which triggers `TwoStatePreference#onSetInitialValue` which loads and sets the value from the shared preferences.

connyduck:
Nice find. Also, your solution seems like a good idea.

PR #3205, opened 2023-01-22
3204: Add an account based preference store

Lakoja:
Fixes https://github.com/tuskyapp/Tusky/issues/3204

Everything coming from the account db object must be handled specially (and not as normal app settings).


app/src/main/java/com/keylesspalace/tusky/components/preference/AccountPreferencesFragment.kt
```diff
@@ -253,7 +250,8 @@ class AccountPreferencesFragment : PreferenceFragmentCompat(), Injectable {
                     key = PrefKeys.ALWAYS_SHOW_SENSITIVE_MEDIA
                     setTitle(R.string.pref_title_alway_show_sensitive_media)
                     isSingleLineTitle = false
-                    isChecked = accountManager.activeAccount?.alwaysShowSensitiveMedia ?: false
+                    val x = accountManager.activeAccount?.alwaysShowSensitiveMedia ?: false
+                    preferenceDataStore = AccountPreferenceDataStore(accountManager.activeAccount!!)
                     setOnPreferenceChangeListener { _, newValue ->
                         updateAccount { it.alwaysShowSensitiveMedia = newValue as Boolean }
```
Lakoja, line 256: Updating the account object (and maybe saving it) could also move to that DataStore class.

app/src/main/java/com/keylesspalace/tusky/util/AccountPreferenceDataStore.kt
```diff
@@ -0,0 +1,23 @@
+package com.keylesspalace.tusky.util
+
+import androidx.preference.PreferenceDataStore
+import com.keylesspalace.tusky.db.AccountEntity
+import com.keylesspalace.tusky.settings.PrefKeys
+
+class AccountPreferenceDataStore(private val account: AccountEntity): PreferenceDataStore() {
+    override fun getBoolean(key: String?, defValue: Boolean): Boolean {
+        return when(key) {
+            PrefKeys.ALWAYS_SHOW_SENSITIVE_MEDIA -> account.alwaysShowSensitiveMedia
```
Lakoja, line 10: This is not yet an exhaustive list.

But in the end it should/must handle all account settings.

app/src/main/java/com/keylesspalace/tusky/util/AccountPreferenceDataStore.kt
```diff
@@ -0,0 +1,23 @@
+package com.keylesspalace.tusky.util
+
+import androidx.preference.PreferenceDataStore
+import com.keylesspalace.tusky.db.AccountEntity
+import com.keylesspalace.tusky.settings.PrefKeys
+
+class AccountPreferenceDataStore(private val account: AccountEntity): PreferenceDataStore() {
```
Lakoja, line 7: Once such a solution is introduced it might probably be sensible to remove the respective settings from the shared preferences.

There they will be confusing now and have no meaning in the future.

Lakoja: Also see #3186 for this.

app/src/main/java/com/keylesspalace/tusky/components/preference/AccountPreferencesFragment.kt
```diff
@@ -84,7 +84,7 @@ class AccountPreferencesFragment : PreferenceFragmentCompat(), Injectable {
                     colorInt = MaterialColors.getColor(context, R.attr.iconColor, Color.BLACK)
                 }
                 setOnPreferenceClickListener {
-                    openNotificationPrefs()
+                    openNotificationSystemPrefs()
```
Lakoja, line 87: Changed the name to make it more clear that this leaves the "scope of the app code".

app/src/main/java/com/keylesspalace/tusky/settings/AccountPreferenceHandler.kt
```diff
@@ -0,0 +1,47 @@
...
+
+        when (key) {
+            PrefKeys.ALWAYS_SHOW_SENSITIVE_MEDIA -> account.alwaysShowSensitiveMedia = value
+            PrefKeys.ALWAYS_OPEN_SPOILER -> account.alwaysOpenSpoiler = value
+            PrefKeys.MEDIA_PREVIEW_ENABLED -> account.mediaPreviewEnabled = value
+        }
+
+        accountManager.saveAccount(account)
+
+        eventHub.dispatch(PreferenceChangedEvent(key))
```
Lakoja, line 39: It doesn't really strike me as an appropriate job for this class. However it might be better than having it as double code in the change listener in AccountPreferencesFragment.

(That location is also too early as the listener is called before storage / change.)

app/src/main/java/com/keylesspalace/tusky/settings/AccountPreferenceHandler.kt
```diff
@@ -0,0 +1,47 @@
+package com.keylesspalace.tusky.settings
+
+import androidx.preference.PreferenceDataStore
+import com.keylesspalace.tusky.appstore.EventHub
+import com.keylesspalace.tusky.appstore.PreferenceChangedEvent
+import com.keylesspalace.tusky.db.AccountEntity
+import com.keylesspalace.tusky.db.AccountManager
+
+// TODO this must be possible with DI / @Inject somehow?
```
Lakoja, line 9: The problem here is that this class a) is dependent on injected classes and thus must be injectable and injected itself (?) and b) depends on a configurable account instance.

Lakoja: That means it always needs to work with the active account?

(Which would be somewhat hidden and tighter coupling than a handler that can work with any account.)

app/src/main/java/com/keylesspalace/tusky/settings/AccountPreferenceHandler.kt
Lakoja: It is a preference, however (account preference). Other classes wait for them to change (e. g. TimelineViewModel#onPreferenceChanged). And there it probably does not matter where a specific preference is stored - just that it change is important.

app/src/main/java/com/keylesspalace/tusky/settings/AccountPreferenceHandler.kt
```diff
@@ -0,0 +1,47 @@
...
+    override fun getBoolean(key: String?, defValue: Boolean): Boolean {
+        return when (key) {
+            PrefKeys.ALWAYS_SHOW_SENSITIVE_MEDIA -> account.alwaysShowSensitiveMedia
+            PrefKeys.ALWAYS_OPEN_SPOILER -> account.alwaysOpenSpoiler
+            PrefKeys.MEDIA_PREVIEW_ENABLED -> account.mediaPreviewEnabled
+            else -> defValue
+        }
+    }
+
+    override fun putBoolean(key: String?, value: Boolean) {
```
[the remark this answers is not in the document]
Lakoja: Correct. (Changed it)

app/src/main/java/com/keylesspalace/tusky/util/AccountPreferenceDataStore.kt
Lakoja: Updated that migration code to the new settings migration mechanism.

app/src/main/java/com/keylesspalace/tusky/settings/AccountPreferenceHandler.kt
connyduck: You can't inject the active account (maybe it is possible with some assisted injection but I doubt that makes anything easier), so you either stick to your current solution, or apply my suggestion. I'm fine with both.

app/src/main/java/com/keylesspalace/tusky/settings/AccountPreferenceHandler.kt
```diff
@@ -0,0 +1,43 @@
...
+        accountManager.saveAccount(account)
+
+        eventHub.dispatch(PreferenceChangedEvent(key))
+    }
+
+//    companion object {
+//        fun newInstance(accountE: AccountEntity): AccountPreferenceDataStore {
+//            return AccountPreferenceDataStore().apply { account = accountE }
+//        }
+//    }
```
connyduck, line 42: please remove

Lakoja: Done

app/src/main/java/com/keylesspalace/tusky/settings/AccountPreferenceHandler.kt
Lakoja: Sticking (DI wishes removed).

connyduck (approved):

PR #3206, opened 2023-01-22
Create send and error notification channels when the app starts

nikclayton:
The previous code creates the channels when a message is sent, or encounters
a failure to send.

This means that the channel does not show up in the Android list of channels
that the user can interact with, until there's been at least one sent message
or failure.

Instead, create the channels early in the application life cycle, so the
user can see and interact with them immediately.


nikclayton:
This PR is stacked on top of https://github.com/tuskyapp/Tusky/pull/3185, that one should be merged first, then this one.

Issue #3211, opened 2023-01-24
Support editing media descriptions


connyduck:
https://github.com/mastodon/mastodon/pull/20878

PR #3215, opened 2023-01-25
Add support for updating media description and focus point when editing statuses

### LINKED ISSUE #3273: Can't edit caption during "delete and re-draft" anymore
- Make a post, attach an image, do not add a caption
- "Delete and re-draft"
- **In the delete and re-draft pane you will not be given the option to edit caption, edit focus, or edit image**. The only menu item will be "Delete".

According to Maloki this worked before 21 (I'm 90% certain I remember it working before). Maloki thinks when we (correctly) set the edit pane to not allow caption edit, we (accidentally) removed it from delete+redraft as well.

Interestingly if at this point you save your re-draft to drafts, then re-open it, you will be able to edit captions again.

Practical impact is we should make sure to re-test this use case when #3215 merges

* * * *
- Tusky Version: 21.0 (plus a couple manually applied patches, not relevant)
- Android Version: 12
- Android Device: Xperia 5 III
- Mastodon instance (if applicable): mastodon dot social

Tak:
Completes #3211 

We could also add support for viewing them in the edit history view, but I think that will maybe not be super useful since that view is mainly for seeing the history of *other* people's statuses

Also fixes #3273

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/components/compose/ComposeViewModel.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/components/compose/ComposeViewModel.kt
@@ -334,17 +334,20 @@ class ComposeViewModel @Inject constructor(
             }
         }
 
-        val updatedItem = newMediaList.find { it.localId == localId }
-        if (updatedItem?.id != null) {
-            val focus = updatedItem.focus
-            val focusString = if (focus != null) "${focus.x},${focus.y}" else null
-            return api.updateMedia(updatedItem.id, updatedItem.description, focusString)
-                .fold({
-                    true
-                }, { throwable ->
-                    Log.w(TAG, "failed to update media", throwable)
-                    false
-                })
+        if (!editing) {
+            // Updates to media for already-published statuses need to go through the status edit api
+            val updatedItem = newMediaList.find { it.localId == localId }
+            if (updatedItem?.id != null) {
+                val focus = updatedItem.focus
+                val focusString = if (focus != null) "${focus.x},${focus.y}" else null
+                return api.updateMedia(updatedItem.id, updatedItem.description, focusString)
+                    .fold({
+                        true
+                    }, { throwable ->
+                        Log.w(TAG, "failed to update media", throwable)
+                        false
+                    })
+            }
         }
         return true
     }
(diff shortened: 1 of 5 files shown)


mcclure:
This is currently not even exposed on Mastodon Web. Does this feature exist in the API currently?

I would LOVE this if it's possible because currently if I forget a caption I can't add one after-the-fact.

Tak:
Yeah, it's in 4.1.0rc1 (see the mastodon PR linked from the linked issue) - I updated my glitch server to a corresponding version to work on this

sl1txdvd:
I tried this with my Mastodon 4.1.0-rc1 instance and it seems to work. However it shows a message saying there was an error when saving the description ("Fehler beim Speichern der Beschreibung"). As far as I can tell there is no actual error.

Tak:
That message means that the server reported an error when Tusky sent the update.
Can you describe what you did to get the error?
I've tested:
- Updating the description on an image
- Updating the focus point on an image
- Updating both description and focus point on an image
- Updating the description on a video

sl1txdvd:
-Start new post, enter text, add image, press "Toot"
-Edit post:
--"Set caption", enter text, press "ok": error says "Failed to set caption"
-"Set focus point", move circle, press "ok": error says "Failed to set focus point"
-Press "Toot"
-Reload post: caption and focus point are the way I set them despite the error messages

Maybe I will find something in the server logs.

Tak:
Ohh, the message is when setting them *locally*, that gives me something to look into

Tak:
Ok, Tusky was attempting to update the media attributes via the normal path as soon as they were accepted, but the server doesn't like that when the ~~post~~ toot has already been ~~published~~ tooted.
I've added a fix, thanks!

sl1txdvd:
Works nicely now, thanks!

mcclure:
Tested [49d5a89](https://github.com/tuskyapp/Tusky/pull/3215/commits/49d5a89527b1d120a0412067e7bf0e6d25ef73f3) on mastodon.social now that it is on v4.1.0rc3

Can confirm it fixes #3273 and fixes the inappropriate "Edit Image" offer. Both focus change and caption change work.

mcclure:
…even weirder, Tusky 49d5a89 is able to edit focus point on images even though [Mastodon Web itself cannot](https://github.com/mastodon/mastodon/issues/23533)

connyduck (approved):

PR #3221, opened 2023-01-26
Preference to disable "ALT" badge on captioned images

mcclure:
I would like this option. Uncertain how many people would toggle it.

Defaults to on (IE merging this is a no-op).

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/components/conversation/ConversationsFragment.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/components/conversation/ConversationsFragment.kt
@@ -90,6 +90,7 @@ class ConversationsFragment : SFragment(), StatusActionListener, Injectable, Res
             useAbsoluteTime = preferences.getBoolean("absoluteTimeView", false),
             showBotOverlay = preferences.getBoolean("showBotOverlay", true),
             useBlurhash = preferences.getBoolean("useBlurhash", true),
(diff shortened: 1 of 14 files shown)


app/src/main/java/com/keylesspalace/tusky/components/conversation/ConversationsFragment.kt
```diff
@@ -90,6 +90,7 @@ class ConversationsFragment : SFragment(), StatusActionListener, Injectable, Res
             useAbsoluteTime = preferences.getBoolean("absoluteTimeView", false),
             showBotOverlay = preferences.getBoolean("showBotOverlay", true),
             useBlurhash = preferences.getBoolean("useBlurhash", true),
+            showDescriptionBadge = preferences.getBoolean("showDescriptionBadge", true),
```
mcclure, line 93: Not related to the patch, but is it weird that this very similar block of code to create a StatusDisplayOptions is duplicated so many times? Is it weird that some of the instances use PrefKeys and some use literal strings?

nikclayton: I start to tackle this as part of https://github.com/tuskyapp/Tusky/pull/3159/files#diff-795e9e3b0efed324958f1c9b55242bf882405d00c87a7634ce8d57b60f8cc002.

sl1txdvd:
I tried it and I will be using it. Thank you.

PR #3222, opened 2023-01-27
Swipe sensitivity factor preference

mcclure:
Uncertain about this patch, there's more than one problem with it currently. It might be more useful for testing different sensitivity factor values than for shipping.

This puts a slider in the prefs screen with 1.0 at left end and 5.0 at right end. This is the number the sensitivity is divided by. 1.0 is the "Android default" (20.0) sensitivity. Each notch in the slider goes up by 0.5. The default is currently 4.0 (the number from Nik's patch). After some testing I think I am happiest at 1.5 but that might be low for many people.

The freedom of the slider is very nice but it is awkward for various reasons. It currently does not gray out when swipe is disabled, and it doesn't display any visual feedback of your current value. (I'm not sure how to do either of these things currently). If we ship a version of this maybe it would make more sense to make it a picker with values like "Low", Middle", "High".


nikclayton:
I'll do some testing today to determine the normal sensitivity value.

Just to note a misconception though, when you write "1.0 is the "Android default"".

It's not. I mean, I guess it is, in the sense that it's what you get, but it's not the intended default, as ViewPager2 has a much lower sensitivity if no descendants are RecyclerView.

But ViewPager2 has a bug, that's triggered if a descendant view is a RecyclerView (https://issuetracker.google.com/issues/139867645?pli=1) that makes it a lot more sensitive than the intended default.

ViewPager doesn't have this bug.

mcclure:
Oh no :(

connyduck:
Please not another preference 🥺 
Things like this should just work without the user thinking about it

Lakoja:
What does this "sensitivity" influence?

mcclure:
> What does this "sensitivity" influence?

The sensitivity of "side swiping". At the moment this value is (see above) probably too high, and people often accidentally swipe to another tap while trying to scroll.

In the current beta the side swiping is 1/4 as sensitive.

Lakoja:
> > What does this "sensitivity" influence?
> 
> The sensitivity of "side swiping". At the moment this value is (see above) probably too high, and people often accidentally swipe to another tap while trying to scroll.
> 
> In the current beta the side swiping is 1/4 as sensitive.

Ah, I see - I accidentally do something like that a lot...

Might this also be related to the (my) problems with the image view after tapping an image in a status?
I always need 2 or 3 tries to pinch zoom. Before that the app tries to swipe somewhere..

mcclure:
> Might this also be related to the (my) problems with the image view after tapping an image in a status? I always need 2 or 3 tries to pinch zoom. Before that the app tries to swipe somewhere..

Could you install 21.0 beta 1 (see [here](https://mastodon.social/@Tusky/109765527776096625 or [here](https://play.google.com/store/apps/details?id=com.keylesspalace.tusky.test&hl=en_CA&gl=US)) and see if you still reproduce this problem? (It does not have this sensitivity preference but it does have the 25% sensitivity patch.)

Lakoja:
> > Might this also be related to the (my) problems with the image view after tapping an image in a status? I always need 2 or 3 tries to pinch zoom. Before that the app tries to swipe somewhere..
> 
> Could you install 21.0 beta 1 (see [here](https://mastodon.social/@Tusky/109765527776096625 or [here](https://play.google.com/store/apps/details?id=com.keylesspalace.tusky.test&hl=en_CA&gl=US)) and see if you still reproduce this problem? (It does not have this sensitivity preference but it does have the 25% sensitivity patch.)

(Always use newest develop.)

Thanks.
It seems to be better.
There is still an odd behavior at the edges when being zoomed (once a swipe is detected - without something to swipe to - you cannot move the section again) but I guess that that is not sensitivity related.

mcclure:
> There is still an odd behavior at the edges when being zoomed (once a swipe is detected - without something to swipe to - you cannot move the section again) but I guess that that is not sensitivity related.

I'm not sure I understand what you mean, could you file a new issue for this? A video would be great if your phone can take them.

app/src/main/java/com/keylesspalace/tusky/components/account/AccountActivity.kt
```diff
@@ -236,7 +236,7 @@ class AccountActivity : BottomSheetActivity(), ActionButtonActivity, HasAndroidI
         // Setup the tabs and timeline pager.
         adapter = AccountPagerAdapter(this, viewModel.accountId)
 
-        binding.accountFragmentViewPager.reduceSwipeSensitivity(preferences.getInt(PrefKeys.DECREASE_SWIPE_SENSITIVITY, 6))
+        binding.accountFragmentViewPager.reduceSwipeSensitivity(preferences.getInt(PrefKeys.DECREASE_SWIPE_SENSITIVITY, -4))
```
Lakoja, line 239: "reduce -4" looks like an "increase 4" (which would be easier to read)

mcclure:
I don't think this patch is going to be merged (maybe it should be closed?) but if it were to be merged, it would need to be changed so that the display is separate from the value (-1 being reformatted as "80%", for example). The reason for the (highly dubious) negative integer value is because the only easy way I know to change the displayed value range of a slider is to change the actual value range.

mcclure:
New version of patch has an acceptable UI (uses sliderPreference and the slider shows a % value while you drag it) so I am marking it as no longer a draft. Still does not gray out when swiping is disabled. My understanding is the project is not interested in merging this so I am only really maintaining this for my personal/"hacks" fork.

PR #3223, opened 2023-01-27
Use HOME as the default tab if it's not recognised

nikclayton:


nikclayton:
Discovered this when I was checking the PR to add the trending tab, then switched to another branch to work on something else. The code on the other branch tried to restore the trending tab (which didn't exist) and promptly crashed.

nikclayton:
Note that this will add a **second** HOME tab in the situation above. This is something only developers should encounter, so I think that bit of UI oddness is OK.

connyduck:
No, it really should crash in this case. The app is in an invalid state, its a major bug if that would happen, I want to notice it immediately. Showing another tab would hide the problem and lead to more problems down the line (two homes operating on the same cache, I'm not sure that works).

connyduck:
o_0 that should not be possible and should be fixed, yes

PR #3224, opened 2023-01-27
Leave comments on PRs with fixes for lint errors

nikclayton:
Runs `ktlintFormat`, and adds comments to the PR if that generates any diffs. The comments include the fix, which can be accepted immediately through the GitHub UI.


nikclayton:
I've set up a test repository with a PR so you can see what this looks like on PRs after it's activated. Take a look at:

https://github.com/nikclayton/kotlin-workflow-test/pull/1

This PR makes this opt-in on PRs (with me opted-in). It could stay that way, or the early "if" check can be removed and it will run on all PRs.

app/build.gradle
```diff
@@ -65,6 +65,13 @@ android {
         baseline = file("lint-baseline.xml")
     }
 
+    ktlint {
+        reporters {
+            reporter "plain"
+            reporter "checkstyle"
+        }
+    }
+
```
nikclayton, line 74:
```suggestion
```

Gryzor (approved):
LGTM

PR #3226, opened 2023-01-27
don't show sending failed dialog when sending was cancelled

connyduck:


app/src/main/java/com/keylesspalace/tusky/service/SendStatusService.kt
```diff
@@ -265,7 +265,7 @@ class SendStatusService : Service(), Injectable {
 
             mediaUploader.cancelUploadScope(*failedStatus.media.map { it.localId }.toIntArray())
 
-            saveStatusToDrafts(failedStatus)
+            saveStatusToDrafts(failedStatus, true)
```
charlag, line 268: I would use label here and below for the boolean argument

charlag (approved):

Tak (approved):

connyduck:
cc @mcclure

Issue #2823, opened 2022-11-13
Post translation option in post menu.


TurekCom:
Hello to you.
 It annoys me that when I browse posts on foreign servers, there is no option to translate the post in tusky itself.
 When I want to translate a post or read a local timeline that is in a different language, I have to open the chrome browser and select the translate to Polish option there.
 In Tusky, there could be an option to translate individual posts in the menu of each post, and there could be an option to translate all posts.  This would make it much easier to view mastodon in other languages ​​you don't understand.

Android 12
motorola edge 30 pro
tusky 19

ShellWen:
I need this feat too. As you can see my english is not good. 😭

charlag:
Please stop nagging, we are working on this in our free time. Contributions welcome, in any form.

MinekPo1:
Also note: as a temporary workaround, you can use the share content feature
to send the toot text to a translator app.

MinekPo1:
Also duplicate of #134

jsoref:
This is a duplicate of #134

connyduck:
https://github.com/tuskyapp/Tusky/pull/4307

PR #3228, opened 2023-01-27
translations via mastodon api

### LINKED ISSUE #134: [Feature Request] translation feature

Joshix-1:
Me trying to fix #134

I'm trying, I won't guarantee anything. This is just to show you guys, what I'm doing. If you don't see any activity for more than two days you can assume I lost interest or gave up. If someone who knows what they are doing wants to overtake, please feel free.


Joshix-1:
It kind of works :)))
https://user-images.githubusercontent.com/57299889/215277398-a21b390e-8cbf-4030-bd66-89da29d4dd8e.mp4

connyduck:
Thanks for taking a shot at this.

I'd make the interface similar to Twitter where translation is only available on posts in detail view, but one can see the original and translation side by side:

Joshix-1:
I like the Mastodon UI more.

image

image

Joshix-1:
As a side note: As far I know to detect whether or not the instance supports translation Tusky needs to support the /api/v2/instance endpoint (#3153)

app/src/main/java/com/keylesspalace/tusky/components/viewthread/ViewThreadViewModel.kt
```diff
@@ -199,6 +201,19 @@ class ViewThreadViewModel @Inject constructor(
...
 
+    fun translate(alreadyTranslated: Boolean, status: StatusViewData.Concrete): Job = viewModelScope.launch {
+        try {
+            if (alreadyTranslated)
+                timelineCases.dispatchNullTranslation(status.actionableId)
+            else
+                timelineCases.translate(status.actionableId).await()
+        } catch (t: Exception) {
+            ifExpected(t) {
+                Log.d(TAG, "Failed to reblog status " + status.actionableId, t)
```
nikclayton, line 212:
```suggestion
                Log.d(TAG, "Failed to translate status " + status.actionableId, t)
```

nikclayton:
I like this, @Joshix-1 . 

What's the rationale for caching the translation result locally? I suspect this isn't necessary, as I would expect that the vast majority of translations would be a one-time thing as the user is scrolling through their timeline. Am I missing something?

Joshix-1:
> So this is abandoned then? 
For now sadly yes, I currently have neither time nor motivation, sorry

timnolte:
What were the blockers for implementing this change? It seemed as though things were working from what I say in the sample videos.

After recent issues with the beta crashing Tusky I gave the Megalodon client a try, which is a fork of the official client and it has the translation support that my instance is providing, I'm finding it hard to justify going back to Tusky and losing the translation capability.

timnolte:
This would also fix #2823

timnolte:
This would also fix #2188

Tak:
imo this looks pretty good already.
I would agree that caching the translation is unnecessary (and contributes a lot of the complexity of the change!).

Joshix-1:
I don't really know the Tusky code-base so I don't know. I just looked at the implementation for boosting and did the same for translations.

I'm currently running the changes and they mostly work, but there are some annoying bugs. Sometimes Tusky just crashes, I can't translate in the notification view.
And the ui is bad. the feature, shouldn't be in the 3-dot-menu, it should only show if the post isn't already in the language of the users. The text after the translation informing users about it being translated should look better.

connyduck:
> As a side note: As far I know to detect whether or not the instance supports translation Tusky needs to support the /api/v2/instance endpoint (https://github.com/tuskyapp/Tusky/pull/3153)

Yes. I don't think we should add translation without detecting if the feature is supported by the instance. Not all bigger instances have it and users will be very confused if there is a button but it doesn't work.

> >I'd make the interface similar to Twitter where translation is only available on posts in detail view

>Why this limitation? I find that annoying, it is a few extra clicks

To keep the implementation simpler. Putting the translations into the database is overkill for this feature imho, and the overflow menu already pretty overloaded.

Issue #2982, opened 2022-12-04
Increasing the app font size does not increase menu font size


magicalskye:
The text of the menu is still too small to read and increasing app font size does not affect it at all 

* * * *
- Tusky Version: 20.0
- Android Version: 12
- Android Device: Galaxy Tab A8
- Mastodon instance (if applicable):

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

nikclayton:
Which menu do you mean?

Lakoja:
Can confirm: Any context menu (app, conversation) and the side menu keep the same font.
Only the conversation / status text and account name is increased.

connyduck:
The setting in Tusky was originally introduced because the system settings did not make the text big enough for some people

Issue #2461, opened 2022-04-26
Font size too small


Akeel1994:
Hey, I've a low vision and even the biggest available font size is slightly too small to read the feed. Please add more font sizes. Suggestion: maybe you could let the user choose the size between e.g. 10 and 40. So everyone could find the right size. thank you!

Gryzor:
@Akeel1994 What device do you have? (I am not a dev on this project) but from what I see, there's nothing special about Tusky when it comes to font sizes other than them being correctly defined in `sp` (instead of `dp`). Then each device can/will (through the Accessibility panel in Settings) increase (or decrease) the "normal" font size. 

Some devices are notably limited in this, offering merely two to four values, whereas Samsung devices, tend to offer much more granularity and font sizes.

Akeel1994:
> @Akeel1994 What device do you have? (I am not a dev on this project) but from what I see, there's nothing special about Tusky when it comes to font sizes other than them being correctly defined in `sp` (instead of `dp`). Then each device can/will (through the Accessibility panel in Settings) increase (or decrease) the "normal" font size. 
> 
> Some devices are notably limited in this, offering merely two to four values, whereas Samsung devices, tend to offer much more granularity and font sizes. 

Hey, I'm using a Pixel 6 Pro. On this device it's not just big enough even with the biggest setting. So more settings would be appreciated.

Gryzor:
@Akeel1994 Understood. Well... to be honest the developers would have no way of making this happen _short of re-redesigning_ most of the screens, or with quite a bit of work, but I'm more than happy to stand corrected. 

I just looked at a OnePlus 6 and it only offers 4 values: Small, Normal(default), Large, and Largest; I believe this is the Android default (as evidenced by your Pixel 6 Pro running stock Android). 
On a Samsung S20 there are eight values to chose from! But this is probably a Samsung customization. 

Anyway, I think you're absolutely correct that a bigger font would be nice, but I don't know how the project maintainers/devs will want to approach something like this.

PR #3248, opened 2023-01-31
Provide a preference to scale all UI text

nikclayton:
Font scaling is applied in addition to any scaling set in Android system preferences. So if the user set the Android font size to largest (a 1.3x increase) and then sets the preference to 120%, the total change is 1.56x.

Create SliderPreference to adjust the preference.

- Use Slider, which supports float values and step sizes > 1
- Display the selected value in the preference's summary
- Provide buttons to increment / decrement the value

Restart the activity if the preference changes so that the user sees the impact of the change immediately. Fix a bug in PreferencesActivity where the "EXTRA_RESTART_ON_BACK" intent was never processed. Fix this to ensure that other activities are restarted so the new font scale takes effect.

Implement the scaling in BaseActivity by overriding onAttachBaseContext, and providing a wrapped context with the font scaling applied.

Fixes https://github.com/tuskyapp/Tusky/issues/2982, https://github.com/tuskyapp/Tusky/issues/2461


app/src/main/java/com/keylesspalace/tusky/util/FontScaleContextWrapper.kt
```diff
@@ -0,0 +1,43 @@
...
+
+            // Can't simply adjust the `fontScale` in `context`, because that will contain the
+            // result of previous adjustments. E.g., going from 100% to 80% to 100% does not return
+            // you to the original 100%, it leaves it at 80%.
+            //
+            // The application context is unaffected by changes to the base context, so always use
+            // the `fontScale` from that as the initial value to multiply. This will also contain
+            // any changes to the font scale from "Settings > Display > Font size" in the device
+            // settings.
+            val appConfiguration = context.applicationContext.resources.configuration
```
Lakoja, line 24: This line does not use "fontScale"?

(How does this comment relate to the code?)

app/src/main/java/com/keylesspalace/tusky/view/SliderPreference.kt
```diff
@@ -0,0 +1,185 @@
...
+     * Function that will be used to format the summary. The default formatter formats using the
+     * value of the [SliderPreference.format] property.
+     */
+    var formatter: (Float) -> String = { format.format(it) }
+
+    /**
+     * Optional icon to show in a button at the start of the slide. If non-null the button is
+     * shown. Clicking the button decrements the value by one step.
+     */
+    var decrementIcon: Drawable? = null
```
Lakoja, line 82: Does this need to be handled specially here?

(Why not an icon of the button simply in the resources?)

app/src/main/java/com/keylesspalace/tusky/view/SliderPreference.kt
```diff
@@ -0,0 +1,185 @@
...
+
+    /**
+     * Optional icon to show in a button at the end of the slider. If non-null the button is
+     * shown. Clicking the button increments the value by one step.
+     */
+    var incrementIcon: Drawable? = null
```
Lakoja, line 88: As for the comment: Can the icon be clicked?

app/src/main/java/com/keylesspalace/tusky/BaseActivity.java
```diff
@@ -93,6 +97,19 @@ protected void onCreate(@Nullable Bundle savedInstanceState) {
...
 
+    @Override
+    protected void attachBaseContext(Context newBase) {
+        Log.d(TAG, "attachBaseContext()");
+        SharedPreferences preferences = PreferenceManager.getDefaultSharedPreferences(newBase);
+
+        float uiScaleRatio = preferences.getFloat(PrefKeys.UI_SCALE_RATIO, 100F);
+        Context context = FontScaleContextWrapper.Companion.wrap(
+                newBase,
+                uiScaleRatio / 100.f
```
Lakoja, line 108: Curiosity: This takes care of "all" ui texts?

nikclayton: I believe so. The Context carries the font scale ratio, and that's used to compute the text size in pixels from dp, sp, etc.

This is the same mechanism Android uses if you use the system settings to adjust the font scale device wide.

app/src/main/java/com/keylesspalace/tusky/util/FontScaleContextWrapper.kt
nikclayton: It's explaining the use of `appConfiguration.fontScale` a few lines later.

This could also be:

```
val appFontScale = context.applicationContext.resources.configuration.fontScale
[...]
configuration.fontScale = appFontScale * fontScaleRatio
```

It's a simple change, if you think it makes the code clearer.

app/src/main/java/com/keylesspalace/tusky/view/SliderPreference.kt
nikclayton: It's pulled from the resources, if they're specified.

But Tusky uses PreferencesDSL, so there's no preferences resource file to specify it.

app/src/main/java/com/keylesspalace/tusky/view/SliderPreference.kt
nikclayton: Yes -- this is the icon that's shown on the button. If there's no icon then the button is hidden.

app/src/main/res/layout/pref_slider.xml
```diff
@@ -0,0 +1,122 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!--
+  ~ Copyright (C) 2018 The Android Open Source Project
```
Lakoja, line 3: Is this code simply copied from somewhere?

The comment above "better than SeekBar" sounds different?

(Is the license necessary?)

app/src/main/java/com/keylesspalace/tusky/util/FontScaleContextWrapper.kt
Lakoja: I would probably combine the two comments and shorten them to something like

// Always adjust based on the initial value of fontScale; this contains the device settings and also avoids problems with an unaffected application context.
//
// NOTE one could also try to change the densityDpi (see commented code below). However this has further problems (e. g. dialogs being clipped) for larger size.

app/src/main/res/layout/pref_slider.xml
nikclayton: Yes. As noted on line 18.

The source is https://android.googlesource.com/platform/frameworks/support/+/refs/heads/androidx-main/preference/preference/res/layout/preference_widget_seekbar_material.xml, from AOSP. This just has minimal changes to swap in Slider instead of Seekbar, and provide the optional increment/decrement buttons.

app/src/main/java/com/keylesspalace/tusky/components/preference/PreferencesActivity.kt
```diff
@@ -172,7 +178,7 @@ class PreferencesActivity :
     override fun androidInjector() = androidInjector
 
     companion object {
-
+        const val TAG = "PreferencesActivity"
```
connyduck, line 181: unused

app/src/main/java/com/keylesspalace/tusky/BaseActivity.java
```diff
@@ -93,6 +97,19 @@ protected void onCreate(@Nullable Bundle savedInstanceState) {
         requesters = new HashMap<>();
     }
 
+    @Override
+    protected void attachBaseContext(Context newBase) {
+        Log.d(TAG, "attachBaseContext()");
```
connyduck, line 102: I'd prefer to not have these tracing logs in the code

nikclayton:
Sample user request: https://mastodon.social/@KingShawn/109882568882452842

app/src/main/java/com/keylesspalace/tusky/BaseActivity.java
nikclayton: Done.

app/src/main/java/com/keylesspalace/tusky/BaseActivity.java
```diff
@@ -93,6 +97,18 @@ protected void onCreate(@Nullable Bundle savedInstanceState) {
         requesters = new HashMap<>();
     }
 
+    @Override
+    protected void attachBaseContext(Context newBase) {
+        SharedPreferences preferences = PreferenceManager.getDefaultSharedPreferences(newBase);
+
+        float uiScaleRatio = preferences.getFloat(PrefKeys.UI_SCALE_RATIO, 100F);
+        Context context = FontScaleContextWrapper.Companion.wrap(
```
connyduck, line 105: When you annotate a method with `@JvmStatic` you can call it from Java without `Companion`

app/src/main/java/com/keylesspalace/tusky/util/FontScaleContextWrapper.kt
```diff
@@ -0,0 +1,43 @@
+package com.keylesspalace.tusky.util
+
+import android.content.Context
+import android.content.ContextWrapper
+
+class FontScaleContextWrapper(base: Context?) : ContextWrapper(base) {
```
connyduck, line 6: This class is not actually doing anything, could simplify by just having a top level function `wrapWithFontScaling` that just returns a `ContextWrapper`. 
(Also I don't think the `[base](base: Context?)` argument can be nullable.)

app/src/main/java/com/keylesspalace/tusky/view/SliderPreference.kt
```diff
@@ -0,0 +1,185 @@
...
+        value = slider.value
+    }
+
+    companion object {
+        const val TAG = "FloatSeekBarPreference"
+        const val defaultValueFrom = 0F
+        const val defaultValueTo = 1F
+        const val defaultValue = 0.5F
+        const val defaultStepSize = 0.1F
+        const val defaultFormat = "%3.1f"
```
connyduck, line 183: These should probably all be private. And `TAG` is both unused & has the wrong value.

connyduck (approved):

app/src/main/java/com/keylesspalace/tusky/components/preference/PreferencesActivity.kt
nikclayton: Done.

Issue #2723, opened 2022-10-18
Three-dots-menu (hamburger menu) not easily visible depending on background pic of account


Maltimore:
I just tried to block an annoying account. From the account page of that account, I couldn't find a way to do this for some minutes. Then I realized the three-dots-menu is hard to see with their background picture. In the screenshot below it somehow looks more visible than in reality. At least if you know it's there.
![screenshot](https://user-images.githubusercontent.com/10359335/196495923-a3ba5095-5d89-42ac-9a5f-51bb01d5891f.png)

* * * *
- Tusky Version: 19.0
- Android Version: 12
- Android Device: Pixel 6
- Mastodon instance (if applicable): irrelevant

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

PR #3272, opened 2023-02-06
Make "Up" and "Overflow" menu icons more visible in AccountProfile

nikclayton:
The toolbar in AccountProfile is transparent, so any profile image the user has chosen is shown under it.

This makes the "Up" and "Overflow" menu icons also have transparent backgrounds.

Consequently, they can be hard to spot, or possibly invisible, on backgrounds that are very dark or very light.

Fix this by compositing the icons in a LayerDrawable, with a circular background identical to the surface colour. This ensures they stand out against the background image, and blend in when the user scrolls.

Fixes https://github.com/tuskyapp/Tusky/issues/2723


app/src/main/java/com/keylesspalace/tusky/components/account/AccountActivity.kt
```diff
@@ -298,6 +300,21 @@ class AccountActivity : BottomSheetActivity(), ActionButtonActivity, HasAndroidI
         toolbarBackground.fillColor = ColorStateList.valueOf(Color.TRANSPARENT)
         binding.accountToolbar.background = toolbarBackground
 
+        // Provide a non-transparent background to the navigation and overflow icons to ensure
+        // they remain visible over whatever the profile background image might be.
+        binding.accountToolbar.navigationIcon = LayerDrawable(
+            arrayOf(
+                AppCompatResources.getDrawable(this, R.drawable.background_circle),
```
connyduck, line 307: Maybe load the drawable only once and reuse it?

connyduck (commented):
Nice! Would it be possible to have a bit of transparency on the background?

nikclayton:
> Nice! Would it be possible to have a bit of transparency on the background?

A bit, but not much. It's very easy to go too transparent and then on pathological backgrounds everything ends up camouflaged again.

PR #3276, opened 2023-02-07
Use `unsafeLazy` to simplify thread unsafe lazy initializations

Goooler:


app/src/main/java/com/keylesspalace/tusky/util/Lazy.kt
```diff
@@ -0,0 +1,5 @@
+package com.keylesspalace.tusky.util
+
+@Suppress("NOTHING_TO_INLINE")
+inline fun <T : Any> unsafeLazy(noinline initializer: () -> T): Lazy<T> =
```
Lakoja, line 4: Is it better to have it unsafe?

Goooler: Yep

connyduck (approved):

Issue #3267, opened 2023-02-05
Feature: Convert selected text to hashtags


nikclayton:
If you're composing / editing a status and you select some of the text a menu pops up (copy, paste, etc).

That can be customised.

Might be worthwhile adding a "Hashtag" option to that. Selecting it would prepend a "#" to the selected text.

It gets interesting if the selected text is multiple words. Would this:

a) Prepend the "#" to all the words
b) Remove the spaces to create a SingleCamelCasedWord and then prepend the "#"
c) Not show the option at all

?

https://medium.com/androiddevelopers/custom-text-selection-actions-with-action-process-text-191f792d2999 explains how to add an item to this menu, and return modified text.

* * * *
- Tusky Version: 21

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

Lakoja:
That wouldn't really tie in with the feature that you type a hash first and the a special hashtag proposal search starts. (?)

nikclayton:
Yes, I think they're orthogonal use cases.

Tak:
You can do this today by selecting some text and pressing the hash button in the header

nikclayton:
> You can do this today by selecting some text and pressing the hash button in the header

Yes. Except that when you select the text the pop up menu can obscure that.

<img src="https://user-images.githubusercontent.com/773100/217224553-87320303-be4a-457b-9dc5-a3c4068dcfea.png" width="45%">

Lakoja:
> > You can do this today by selecting some text and pressing the hash button in the header
> 
> Yes. Except that when you select the text the pop up menu can obscure that.

Is that a problem?

If so I would probably try to tackle that problem - but not with a whole new feature.

nikclayton:
> Is that a problem?

It's a small annoyance, and I'm all about filing off rough edges wherever I see them.

> If so I would probably try to tackle that problem - but not with a whole new feature.

You can't control where the menu appears unfortunately, so that would require moving the UI that's at the top to the bottom. where it's already quite crowded.

And it's not a big feature -- ~ 100 self-contained lines, in https://github.com/tuskyapp/Tusky/pull/3278

PR #3278, opened 2023-02-07
Handle Intent.ACTION_PROCESS_TEXT and convert to hashtags

nikclayton:
Allows the user to select editable text like "foo bar", choose "As hashtags" from the popup menu (next to Copy, Paste, etc), and get back "#foo #bar".

Fixes https://github.com/tuskyapp/Tusky/issues/3267


connyduck:
This is a quite elaborate workaround for such a tiny inconvenience. I'd rather not add it.

PR #3281, opened 2023-02-07
Show the follower's bio/note in a "followed you" notification

nikclayton:
This makes the notification view for a follow request contain more info about the new follower, and makes the layout (of their name / username) consistent with other notifications that show names/usernames.


nikclayton:
Note: Stacked PR on top of https://github.com/tuskyapp/Tusky/pull/3280

This makes the notification view for a follow request contain more info about the new follower, and makes the layout (of their name / username) consistent with other notifications that show names/usernames.

Does increase the size of `TimelineAccount`. And maybe this would be more useful if there was a "Follow" button right in the notification as well?

Here's what it looks like in context (red highlight is just for the screenshot, obviously that's not present in the app):

mcclure:
Does this patch also add the bio/note to the "follow requests" box?

Speaking as a user, I am unsure whether I'd find this useful, but in the Follow Request pane it would be much more useful because that's a pane where I'm specifically trying to make a decision about the mystery names.

nikclayton:
> Does this patch also add the bio/note to the "follow requests" box?
> 
> Speaking as a user, I am unsure whether I'd find this useful, but in the Follow Request pane it would be much more useful because that's a pane where I'm specifically trying to make a decision about the mystery names.

This one doesn't, but https://github.com/tuskyapp/Tusky/pull/3284 does. Try that and let me know what you think.

connyduck (approved):

PR #3287, opened 2023-02-09
Generate library license list/display with AboutLibraries

nikclayton:
- Ensures that no licenses/libraries are missed


nikclayton:
This may need more work before merging. At the moment it will try and fetch the license information if it doesn't already have it, which means a network request. That's probably not what we want on CI.

The better approach is probably to set `offlineMode` in the `aboutLibraries` configuration block, and add something like this to the "Preparing a release" doc.

```
./gradlew app:exportLibraryDefinitions -PaboutLibraries.exportPath=src/main/res/raw/ -PaboutLibraries.exportVariant=release
```

Or maybe only set offline mode in the CI environment?

connyduck:
I don't like this change, the current list is way more concise and understandable.

Lakoja:
The list now seems to be largely of limited usefullness.

It is bascially (the quite technical) build.gradle dependencies as list.

There are for example some without name and license. And there are pages upon pages of basically "Android base libraries".

grafik

I guess a curated list would be preferrable.

nikclayton:
> I guess a curated list would be preferrable.

Curated how?

More specifically -- we already have a curated list, and it's out of date and not being maintained.

Lakoja:
> > I guess a curated list would be preferrable.
> 
> Curated how?
> 
> More specifically -- we already have a curated list, and it's out of date and not being maintained.

An alternative solution might be to maintain that.

But maybe do "both"?
Use aboutlibraries but with a manual list?

nikclayton:
> An alternative solution might be to maintain that.

Nobody's doing that, and realistically, nobody is going to step up to do that. And even if they did, what do we do if they decide to stop doing it?

And even if all of what weren't true, what if they make mistakes? I wrote, and you approved https://github.com/tuskyapp/Tusky/pull/3894, and neither of us thought to update the licenses list, even though it's now out of date because of that PR.

Maintaining a manual list of libraries to show here is a classic example of toil (https://sre.google/sre-book/eliminating-toil/):

- Repetitive
- Manual
- Automatable
- Has no enduring value
- Grows without end

(the language list is another example, hence https://github.com/tuskyapp/Tusky/pull/3621)

To put it another way -- the amount of work to do on a project like this will grow without end, and probably faster than contributors can keep up with it.

We either automate what we can, or drown under the work.

PR #3288, opened 2023-02-09
Support additional fonts for accessibility

nikclayton:
Provide alternative fonts for use throughout the entire app, based on feedback from users.

- Atkinson Hyperlegible, from the Braille Institute, claims to improve readability for low-vision users.
- Comic Neue "aspires to be the casual script choice for everyone including the typographically savvy."
- Estedad claimes to more readable in Persian script
- Lexend, devloped by an educational therapist, shows demonstrated improvements in reading efficiency
- Luciole, "is a new typeface developed explicitly for visually impaired people"
- OpenDyslexic claims to improve readability for people with dyslexia.

All fonts are available under a permissive license that allows bundling with the app. Include the license information so it shows up in LicenseActivity.

Add a preference to allow the user to switch between the fonts. The preference previews the font in the list so the user can see the difference.


nikclayton:
Note: Stacked on top of https://github.com/tuskyapp/Tusky/pull/3287

nikclayton:
I'm not 100% sold on the idea yet; I'm not sure it's a good idea to go down the route of allowing the user to use any font. But:

a) Maybe this is a good compromise on useful ones?
b) Maybe this could be used to start a discussion on Mastodon and get feedback from users, and narrow it down to a list of two or three additional fonts that would be useful to include?

nikclayton:
New font dialog in preferences:

nikclayton:
Timeline with the two different fonts:

nikclayton:
The creator of OpenDyslexic is also on Mastodon, @abbie@better.boston.

nikclayton:
**Notes from feedback from users**

From responses to https://mastodon.social/@Tusky/110741534444074976

OpenDyslexia

- Some notes that there's no good research that shows formal benefits from this font (e.g., the thread at https://github.com/google/fonts/issues/558)
- Multiple reports of individual users who report that it is useful for them in other apps
- It was added to https://wikisource.org/wiki/Wikisource:WS_Export (Wiki epub project) in response to user requests

Atkinson Hyperlegible

- Multiple reports of individual users who report that it is useful for them in other apps
- A note that it is no longer under active development, with a link to a fork at https://github.com/edent/extended-hyperlegible

Other specific font suggestions

- https://fonts.google.com/specimen/Vazirmatn
- https://github.com/aminabedi68/Estedad
- ABeeZee
- Andika
- Kant
- Inconstant Regular
- Comic Sans
- EasyReading Pro (https://www.easyreading.it/en/) -- note, not freely licensed
- Lexend (https://fonts.google.com/specimen/Lexend/about, https://www.lexend.com/)

Suggestions to download arbitrary fonts

- Multiple suggested supporting downloading arbitrary fonts.
- Android supports this (from O onwards), but I think that's suitable as a second phase.

nikclayton:
New font preferences with the additional fonts:

Lakoja:
Haven't researched this, yet, but maybe you know: Why is this no system setting / feature?

(Packaging fonts with your app sounds a bit like 20 years ago.)

nikclayton:
> Haven't researched this, yet, but maybe you know: Why is this no system setting / feature?

Some manufacturers do -- for example, you can change the font system-wide on some Samsung devices (https://www.howtogeek.com/726320/how-to-change-font-on-a-samsung-galaxy-phone/).

But that's manufacturer specific, and not something in the base OS.

> (Packaging fonts with your app sounds a bit like 20 years ago.)

Yes. A future change could use the font download support in Android, but even with that, some of the fonts are not available through that route, so would need to be packaged as well.

Lakoja:
> Provide alternative fonts for use throughout the entire app, based on feedback from users.
> 
>     * Atkinson Hyperlegible, from the Braille Institute, claims to improve readability for low-vision users.
> 
>     * Comic Neue "aspires to be the casual script choice for everyone including the typographically savvy."
> 
>     * Estedad claimes to more readable in Persian script
> 
>     * Lexend, devloped by an educational therapist, shows demonstrated improvements in reading efficiency
> 
>     * Luciole, "is a new typeface developed explicitly for visually impaired people"
> 
>     * OpenDyslexic claims to improve readability for people with dyslexia.

How about starting with the two most popular / requested and see how that is received and if more are requested?

(The list seems a bit long and then only contains "1 normal and 6 specials"; specials with mostly unknown advantages.)

Lakoja:
(Small side-note: For all these fonts the character rendering seems to be more sub-optimal than with the standard font?

For me - in the emulator - they all are rendered too low. In my screenshot the "Home" is not top-aligned with the avatar and the address is cut at the bottom.

Maybe a fixable bug?)

nikclayton:
> How about starting with the two most popular / requested and see how that is received and if more are requested?

I started with two. Then I posted https://mastodon.social/@Tusky/110741534444074976 to get feedback, and reviewed the responses to determine which other ones to add.

This is all explained in https://github.com/tuskyapp/Tusky/pull/3288#issuecomment-1642474747

nikclayton:
> (Small side-note: For all these fonts the character rendering seems to be more sub-optimal than with the standard font?
> 
> For me - in the emulator - they all are rendered too low. In my screenshot the "Home" is not top-aligned with the avatar and the address is cut at the bottom.

Baseline variation is to expected. I can reproduce the problem in your screenshot with Luciole, but not the others.

I suspect that

```
android:minHeight="?attr/actionBarSize"
```

in toolbar definition in `activity_main.xml` is the fix, I'll try that in a bit.

nikclayton:
It was a tiny bit more involved than that. I also noticed that with Luciole different textviews can have different baselines depending on the text, so I've ensured that they layout and constrain to the baseline rather than the top or bottom of another view.

mcclure:
Does this patch add twemoji or only the license for twemoji? We should keep in mind that twemoji appears to be an abandoned project, so it may not continue to receive new emoji in when unicode updates in future.

nikclayton:
> Does this patch add twemoji or only the license for twemoji? We should keep in mind that twemoji appears to be an abandoned project, so it may not continue to receive new emoji in when unicode updates in future.

It's only the license.

That change is from https://github.com/tuskyapp/Tusky/pull/3287, which this PR is stacked on top of.

PR #3289, opened 2023-02-09
Use "System default" as name of default emoji style

nikclayton:
Other preferences use "System default" (lowercase "d"). This one uses "System Default" because of the default strings in the library. But it can be overridden here for consistency.


connyduck:
Update it in the library instead?

nikclayton:
Done that too now. This override can still go in until the library updates.

Issue #3270, opened 2023-02-06
24h time picker


steelman:
Google [failed](https://github.com/material-components/material-components-android/issues/1450) miserably at the attempt to redesign 24 h time picker. If I understand correctly double-ring design has been re-instantiated. Alas, Tusky is still using the single ring abomination for toot scheduling.

Please investigate possibilities of bringing back double-ring 24 h time picker.

* * * *
- Tusky Version: 20 (F-Droid)
- Android Version: LineageOS 
- Android Device: Moto G7 (river)

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

Lakoja:
I also wondered today what kind of time selector that should be (with half the hours "missing").

connyduck:
The platform picker looks hideous on older Android versions though
![Screenshot_20230225_210318](https://user-images.githubusercontent.com/10157047/221377366-44afb1f8-a887-44b8-9fc7-872e0ca638bc.png)

(this is Android 6, I think 5 is worse)

Lakoja:
Looks ok. Not much different to the new one.
But it is understandable/usable. :-)

connyduck:
Material library has been updated. 24h picker now looks like this on `develop`:
![Screenshot_20230228_221318](https://user-images.githubusercontent.com/10157047/221981444-ce4fa72d-4d25-48cb-9cc2-277bb83759a3.png)

Is that the double ring display you want @steelman?
Honestly I think the Material design people know how do design these pickers better than we do, I'd just trust them.

Lakoja:
Yes, that looks correct.

So that is a complete redesign. :-)

steelman:
> Is that the double ring display you want @steelman?

Yes! Awesometacular! Thank you.

PR #3290, opened 2023-02-09
3270: Use standard time picker with the better 24h display

Lakoja:
Closes https://github.com/tuskyapp/Tusky/issues/3270


Lakoja:
Is undesired.
Fixed by Material update?

Issue #3277, opened 2023-02-07
Can't post in languages not in the locale list (e.g. toki pona) when replying to post in another language or when language is not account's default


nehemiagurl:
(related to #2912)

One can post in a language not in the locale list (e.g. toki pona, Scots) if that language is set as the account default language:
![Screenshot showing language selection options: tok, English, Hebrew](https://user-images.githubusercontent.com/109424320/217243416-e12092ea-35b1-46cb-b1cc-16b5c99e8534.jpg)

However, in any situation other than "new post with default language", those languages are unselectable, meaning you can't reply to a post marked in English in toki pona, or write a post in Scots if your default language is German.

Reply example from the same account with *tok* default replying to post in English:
![Screenshot showing language selection options: English, Hebrew, Afrikaans](https://user-images.githubusercontent.com/109424320/217244095-bce20635-49ba-4973-ae8d-752753a1dc8c.jpg)
New post with account without *tok* as default:
![Screenshot showing language selection options: toki pona is nowhere to be found](https://user-images.githubusercontent.com/109424320/217244825-10ae5de6-e71a-439d-97ce-26c5fd077d74.jpg)

* * * *
- Tusky Version: 21.0
- 
- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.
no

nehemiagurl:
just wanted to confirm that the issue was indeed partially fixed by the commit in question. this is what it looks like when you reply to a post in English while your default language is toki pona:
![signal-2023-09-18-175541](https://github.com/tuskyapp/Tusky/assets/109424320/a38e8b7e-1c4a-4f44-8d20-dd6712b43246)

PR #3293, opened 2023-02-10
Improve language list prioritization

Tak:
Ensures that all of (language of the post being replied to, default post language, configured app languages, configured system languages) are prioritized when constructing the language list for composing, regardless of whether any of those are languages android knows about.

Partially addresses #3277

### DIFF
--- a/app/src/test/java/com/keylesspalace/tusky/util/LocaleUtilsTest.kt
+++ b/app/src/test/java/com/keylesspalace/tusky/util/LocaleUtilsTest.kt
@@ -1,0 +1,82 @@
+package com.keylesspalace.tusky.util
+
+import androidx.appcompat.app.AppCompatDelegate
+import androidx.core.os.LocaleListCompat
+import androidx.test.ext.junit.runners.AndroidJUnit4
+import com.keylesspalace.tusky.db.AccountEntity
+import org.junit.Assert
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.mockito.Mockito
+import org.robolectric.annotation.Config
+
+@Config(sdk = [28])
+@RunWith(AndroidJUnit4::class)
+class LocaleUtilsTest {
+    @Test
+    fun initialLanguagesContainReplySelectedAppAndSystem() {
+        val expectedLanguages = arrayOf<String?>("yi", "tok", "da", "fr", "sv", "kab")
+        val languages = getMockedInitialLanguages(expectedLanguages)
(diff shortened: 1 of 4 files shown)


app/src/test/java/com/keylesspalace/tusky/util/LocaleUtilsTest.kt
```diff
@@ -0,0 +1,82 @@
...
+        )
+        Assert.assertEquals(
+            defaultLanguage,
+            getMockedInitialLanguages(arrayOf(null, defaultLanguage, "fr", defaultLanguage, "kab", defaultLanguage))[0]
+        )
+    }
+
+    @Test
+    fun emptyInitialLanguagesAreDropped() {
+        val languages = getMockedInitialLanguages(arrayOf("", "", "fr", "", "kab", ""))
```
Lakoja, line 53: (curiosity)

Why are there empty languages?

Or I have seen null checks for locales in lists?

app/src/test/java/com/keylesspalace/tusky/util/LocaleUtilsTest.kt
```diff
@@ -0,0 +1,82 @@
...
+import org.mockito.Mockito
+import org.robolectric.annotation.Config
+
+@Config(sdk = [28])
+@RunWith(AndroidJUnit4::class)
+class LocaleUtilsTest {
+    @Test
+    fun initialLanguagesContainReplySelectedAppAndSystem() {
+        val expectedLanguages = arrayOf<String?>("yi", "tok", "da", "fr", "sv", "kab")
+        val languages = getMockedInitialLanguages(expectedLanguages)
```
Lakoja, line 19: Shouldn't a test simply call the normal method with some test data?

(One wonders what is tested about a mocked list.)

app/src/test/java/com/keylesspalace/tusky/util/LocaleUtilsTest.kt
Tak: The default post language and reply language can both be empty in normal usage

app/src/test/java/com/keylesspalace/tusky/util/LocaleUtilsTest.kt
Tak: This function is a wrapper that mocks the response for querying the configured app and system languages, so that:
- the tests don't accidentally rely on whatever values robolectric returns
- we can verify the behavior of the code when it returns certain types of values

connyduck (approved):

PR #3296, opened 2023-02-10
Fix non-git build

mcclure:
This worked in 21.0 but was broken by the gradle modernization patch (PR#3171)

Building with this patch I get the warning

> Values of variant API AnnotationProcessorOptions.arguments are queried and may return non final values, this is unsupported

I don't know if this is because I broke something or if it was printing that warning already.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -5,9 +5,19 @@ plugins {
     alias(libs.plugins.kotlin.parcelize)
 }
 
-final def gitSha = providers.exec {
-    commandLine('git', 'rev-parse', '--short=7', 'HEAD')
-}.standardOutput.asText.get().trim()
+// For constructing gitSha
+def getGitSha = {
+    def stdout = new ByteArrayOutputStream()
+    try {
+        providers.exec {
+            commandLine 'git', 'rev-parse', '--short=7', 'HEAD'
+        }.standardOutput.asText.get().trim()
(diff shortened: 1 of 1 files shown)


app/build.gradle
```diff
@@ -5,9 +5,19 @@ plugins {
     alias(libs.plugins.kotlin.parcelize)
 }
 
-final def gitSha = providers.exec {
-    commandLine('git', 'rev-parse', '--short=7', 'HEAD')
```
mcclure, line 9: By the way, why did we shorten the hash from 8 to 7 characters?

connyduck: No idea, feel free to switch back to 8

app/build.gradle
```diff
@@ -5,9 +5,19 @@ plugins {
...
-}.standardOutput.asText.get().trim()
+// For constructing gitSha
+def getGitSha = {
+    def stdout = new ByteArrayOutputStream()
+    try {
+        providers.exec {
+            commandLine 'git', 'rev-parse', '--short=7', 'HEAD'
+        }.standardOutput.asText.get().trim()
+    } catch (Exception e) {
+        "unknown"
```
connyduck, line 15: add a comment so we (I 😅 ) remember non-git builds break otherwise

connyduck (approved):

PR #3299, opened 2023-02-10
Fix non-git build (for real this time)

mcclure:
Turns out org.gradle.process.internal.execexception is an Error and not an Exception. My last PR was wrong. Sorry!


mcclure:
This is wrong. The problem was Android Studio somehow running an old version of build.gradle even after a sync and a clean build.

PR #3300, opened 2023-02-11
Show categories in emoji picker

d3d9:
[Categories can be defined](https://docs.joinmastodon.org/entities/CustomEmoji/) for custom emojis in Mastodon.  
For example on [zug.network](https://zug.network), there are over 1800 custom emojis and at least some categories have been introduced, which can be seen in the web interface like in the next screenshot. Meanwhile in Tusky the emojis can only be browsed alphabetically without categories.

In this PR I tried to group custom emojis by category and only sort alphabetically within the group, furthermore there's a kind of header for the emoji picker containing all the categories and highlighting the current one.

image

image

When scrolling the emoji grid, the header gets updated, however the code marked with TODO needs to be improved to suppress this during the smooth scrolling that happens when clicking on a category.

Besides that, I'm sure some refactoring has to be done since I probably placed some of the code in inappropriate places. The idea of using empty/fake grid items might also be controversial?

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/adapter/EmojiAdapter.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/adapter/EmojiAdapter.kt
@@ -33,9 +33,11 @@ class EmojiAdapter(
 ) : RecyclerView.Adapter<BindingHolder<ItemEmojiButtonBinding>>() {
 
     private val trueEmojiList: List<Emoji> = emojiList.filter { emoji -> emoji.visibleInPicker == null || emoji.visibleInPicker }
-        .sortedWith(compareBy<Emoji, String?>(nullsLast()) { it.category?.lowercase(Locale.ROOT) }.thenBy { it.shortcode.lowercase(Locale.ROOT) })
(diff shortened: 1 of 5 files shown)


Lakoja:
Why does it need fake grid items (what is missing otherwise or in existing code)?

d3d9:
> Why does it need fake grid items (what is missing otherwise or in existing code)?

I tried using RecyclerView.ItemDecoration with method getItemOffsets and modifying the outRect, but I couldn't get that to work well.  
I should probably have also tried to override the span count for specific items (make the last ones of each category take up all remaining spans), now that I think of it.  
Instead, I quickly changed it to add 3 to 5 spacer grid items for now to get the space between categories.

connyduck (changes requested):
I like the feature, but I'm sorry the code is way to messy to be merged :/

Here are some ideas to improve it:

- do the titles with a second recyclerview, xml layouts and a adapter
- don't touch ComposeActivity at all, it is already doing too much
- move the logic into the one of the other classes, probably EmojiPicker

PR #3303, opened 2023-02-11
Update Contributing.md

connyduck:


Tak (approved):

mcclure:
Conny I notice you're removing the gradle-version and blue/green information additions just one day after merging them. Is this on purpose.

It looks like this change removes ALL build help and makes the contributing document solely about "policy". That makes sense, but if we do that, can we add a "BUILDING" document? Some of the information removed is very basic, but as the discussion in #3310 shows some things can be less obvious than we think, especially in Android studio— when I first started on this project I was an old hand to OSS dev, but I was totally new to Kotlin development, and I will not be the only one.

connyduck:
Yeah I was working on this already before I saw your PR. 
Sorry, I think I don't understand the gradle version thing. You check out the repo, open it in Android studio and everything else happens automatically? Why is the version relevant? Even when I build from command line I never need to specify it anywhere? 

I put the green/blue at the bottom

mcclure:
> You check out the repo, open it in Android studio and everything else happens automatically?

You need the correct version of Android Studio.

The only reason to include the gradle version separately is some people might try to compile Tusky without Android Studio. I wouldn't do this for Tusky but I do it for my NDK projects. [You can get the command line tools package on its own outside Studio](https://developer.android.com/studio/command-line/sdkmanager), for server/CI builds this might even be normal.

Maybe bothering to talk about this in our docs is a stretch, since someone who is doing this can look up SDK/tooling versions corresponding to Android Studio. But I think we should at least specify which Android Studio to use.

mcclure:
Also yeah I see blue/green at the bottom now, sorry I missed that

connyduck:
I also put in "Tusky should be built with the newest version of Android Studio". Hesitating to put in an actual version because I know it will be outdated very quickly. Do you think that is enough or should I phrase it differently or elaborate?

Maybe "Tusky should be built with the newest stable version of Android Studio"
"When building from the command line, the newest build tools must be installed"

charlag:
Aren't all build tools and also gradle version already mentioned in gradle-wrapper properties and such? I also think it will get outdated otherwise

charlag (approved):

mcclure:
> I also put in "Tusky should be built with the newest version of Android Studio". Hesitating to put in an actual version because I know it will be outdated very quickly. Do you think that is enough or should I phrase it differently or elaborate?

As mentioned before I misread the PR, sorry. I think "newest version of Android Studio" does help. I don't think "…and also the command line build tools" would help any more than that.

In the past, with other projects, the reason I have found "newest version" inadequate is sometimes I find myself, for whatever terrible reason, building a three-year-old version of the code. In that case, "newest version" doesn't help unless I can somehow figure out what the newest version was at some arbitrary point in the past. I think it would help if, at least, tagged release versions document what toolchain version they were "intended" to be built with (and specifying an Android Studio version and not worrying about the command line is probably adequate).

If you think this would do more harm than good because you're afraid of it getting out of date, that's reasonable. If there's some automatically-generated file that documents the gradle version, mentioning that in the README would be great (I don't think I could find that file on my own). One way to keep it up-to-date at least for tagged release versions would be to put "update contributing.md with blessed Android Studio version" in the release process.

connyduck:
Here would be the agp/gradle compatibility list https://developer.android.com/studio/releases/gradle-plugin#updating-gradle

PR #3310, opened 2023-02-11
Elaborate in CONTRIBUTING.md/Building on difference between Green/Blue

mcclure:
@nikclayton mentioned he didn't understand why "greenDebug is recommended for development builds", so I elaborated a little. Does this help?


Lakoja:
(to add)

How can I install the green one?

When simply using "develop" branch that is the blue one? (Shouldn't that be the green one then?)

Lakoja:
Will this conflict with 
https://github.com/tuskyapp/Tusky/pull/3303?

mcclure:
@Lakoja Build Variants are a gradle/Android Studio feature. You choose your variant using the Build Variants pane in View->Build Variants. Every branch has both blue and green.

3303 would conflict. I will ask about that

connyduck:
We should set green debug as default https://proandroiddev.com/android-set-a-default-build-type-flavor-a1cdcd1c15fa

mcclure:
#3303 went a different way

PR #3312, opened 2023-02-12
2528: Do not remove notifications on general resume

### LINKED ISSUE #2528: Automatically remove Android notifications only when read
At the moment, Android notifications are automatically removed when the Tusky app is open(ed). Instead, they should only be removed, when the corresponding Mastadon notification has been seen.

Lakoja:
Largely closes #2528

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/MainActivity.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/MainActivity.kt
@@ -666,10 +666,6 @@ class MainActivity : BottomSheetActivity(), ActionButtonActivity, HasAndroidInje
 
         onTabSelectedListener = object : OnTabSelectedListener {
             override fun onTabSelected(tab: TabLayout.Tab) {
-                if (tab.position == notificationTabPosition) {
-                    NotificationHelper.clearNotificationsForActiveAccount(this@MainActivity, accountManager)
-                }
-
                 binding.mainToolbar.title = tabs[tab.position].title(this@MainActivity)
             }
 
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/keylesspalace/tusky/receiver/SendStatusBroadcastReceiver.kt
```diff
@@ -117,7 +117,8 @@ class SendStatusBroadcastReceiver : BroadcastReceiver() {
                 builder.setCategory(NotificationCompat.CATEGORY_SOCIAL)
                 builder.setOnlyAlertOnce(true)
 
-                notificationManager.notify(notificationId, builder.build())
+                // There is a separate "I am sending" notification, so simply remove the handled one.
+                notificationManager.cancel(notificationId)
```
Lakoja, line 121: I wonder if there are more locations (special notifications) like this where notifications would not be removed now.

Upon creation (`NotificationHelper#make`) there are however only two special cases (compose and quick reply).

app/src/main/java/com/keylesspalace/tusky/MainActivity.kt
```diff
@@ -667,10 +666,6 @@ class MainActivity : BottomSheetActivity(), ActionButtonActivity, HasAndroidInje
 
         onTabSelectedListener = object : OnTabSelectedListener {
             override fun onTabSelected(tab: TabLayout.Tab) {
-                if (tab.position == notificationTabPosition) {
```
Lakoja, line 670: Moved to NotificationsFragment.

connyduck (approved):

Issue #1122, opened 2019-03-12
Timeline layout gets messy


espectalll:
For some reason, some toots in Tusky 5.2 are displaced and overlaying each other. I'll just show some screenshots, it's self-descriptive.

![](https://i.imgur.com/8oXPxRZ.png)
![](https://i.imgur.com/mPqIkCQ.png)
![](https://i.imgur.com/lZgUYEo.png)
![](https://i.imgur.com/fp82Q58.png)

espectalll:
(in case you wonder: this is stock Android 9 Pie with February patches on a Sony Xperia XZ2 Compact – no custom ROM, no root)

charlag:
I think I've seen this bug at least once. Does it reproduce reliably for you?

connyduck:
I never had the bug on one of my devices, but saw screenshots. Would be very interested in how to reproduce this so I can fix it.

espectalll:
I'm trying but still no luck, it's probably some rare race condition?

charlag:
Shouldn't be race, rather a bug in one of the libraries. Let's see if it happens often.

charlag:
@espectalll does this still happen to you? I didn't see this in a while

espectalll:
@charlag yes, but it seems completely unpredictable, happens rarely and I can't properly test or debug it

PR #3319, opened 2023-02-14
Fix lifecycle handling bug

nikclayton:
Fragments can go `onCreate` -> `onCreateView` -> `onViewCreated` -> `onDestroyView` without transitioning through `onStart`.

The previous code assumed `onStart` was always called. 

Se https://itnext.io/an-update-to-the-fragmentviewbindingdelegate-the-bug-weve-inherited-from-autoclearedvalue-7fc0a89fcae1


connyduck:
Ok but how does this bug manifest for a user / what do I do to trigger it?

connyduck:
Ohhhh could it be something like this? https://github.com/tuskyapp/Tusky/issues/1122
The bug is older than our use of viewbinding though, but I know it still exists.

nikclayton:
> Ohhhh could it be something like this? #1122 The bug is older than our use of viewbinding though, but I know it still exists.

Maybe. Although if it predates the use of viewbinding and the bug still exists that's extremely weird behaviour.

connyduck (approved):

PR #3324, opened 2023-02-16
Use an explicit SCHEMA_VERSION instead of BuildConfig.VERSION_CODE

nikclayton:
Every nightly release has a new BuildConfig.VERSION_CODE, so the previous code would not do the right thing.

Require the schema version to be explicitly set. While I'm here, provide a clear set of guidelines as to what has to happen when the schema changes.

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/TuskyApplication.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/TuskyApplication.kt
@@ -23,6 +23,7 @@ import autodispose2.AutoDisposePlugins
 import com.keylesspalace.tusky.components.notifications.NotificationWorkerFactory
 import com.keylesspalace.tusky.di.AppInjector
 import com.keylesspalace.tusky.settings.PrefKeys
+import com.keylesspalace.tusky.settings.SCHEMA_VERSION
 import com.keylesspalace.tusky.util.APP_THEME_DEFAULT
 import com.keylesspalace.tusky.util.LocaleManager
 import com.keylesspalace.tusky.util.setAppNightMode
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/keylesspalace/tusky/TuskyApplication.kt
```diff
@@ -69,10 +70,10 @@ class TuskyApplication : Application(), HasAndroidInjector {
         AppInjector.init(this)
 
         // Migrate shared preference keys and defaults from version to version. The last
-        // version that did not have a SCHEMA_VERSION was 97, so that's the default.
-        val oldVersion = sharedPreferences.getInt(PrefKeys.SCHEMA_VERSION, 97)
-        if (oldVersion != BuildConfig.VERSION_CODE) {
-            upgradeSharedPreferences(oldVersion, BuildConfig.VERSION_CODE)
+        // version that did not have a SCHEMA_VERSION was 100, so that's the default.
+        val oldVersion = sharedPreferences.getInt(PrefKeys.SCHEMA_VERSION, 100)
+        if (oldVersion != SCHEMA_VERSION) {
```
Lakoja, line 75: Does this support a step-by-step upgrade? (So a full range of versions inbetween can be enumerated.)

Otherwise you wouldn't be able to do more than one upgrade in total.

(See https://github.com/tuskyapp/Tusky/pull/3205)

app/src/main/java/com/keylesspalace/tusky/TuskyApplication.kt
```diff
@@ -69,10 +70,10 @@ class TuskyApplication : Application(), HasAndroidInjector {
...
+            upgradeSharedPreferences(oldVersion, SCHEMA_VERSION)
```
Lakoja, line 76: I would probably call it "SETTINGS_VERSION" or similar (schema is more headed for apis and databases).

nikclayton: This is a schema though -- it defines the allowed keys, the data types of their values, and how those values are to be interpreted.

Implementation details (are they stored in a KV store, protobufs, an SQL database...?) are orthogonal to the fact that underlying data absolutely has a schema that governs how the data should be interpreted.

app/src/main/java/com/keylesspalace/tusky/TuskyApplication.kt
Lakoja: After thinking about this for a bit: I think I like it.

Especially the expressiveness of the number (giving the date/an age without further effort).

I would get rid of the counter though for readability.

(It is highly unlikely to have two at the same day. And/or on fixing the inevitable merge conflict it is probably already a day later.)

nikclayton: https://wiki.lspace.org/Million-to-one_chance :-)  I think trading off two extra digits is worthwhile.

app/src/main/java/com/keylesspalace/tusky/TuskyApplication.kt
nikclayton: It's already in `SettingsConstants.kt`, and its use is limited to this file. I'm not sure this warrants an additional namespace.

connyduck (approved):

Issue #3294, opened 2023-02-10
Profile text displayed right to left but correct on web


nyankoframe:
The profile text for https://mastodon.social/dungeons changes based on the current character. At the moment the profile in website looks like this:

![Screenshot_20230210_101352](https://user-images.githubusercontent.com/69328423/218166141-04c98a19-7294-4f54-bb06-002c37072345.jpg)

But it appears in Tusky like this:

![Screenshot_20230210_101059](https://user-images.githubusercontent.com/69328423/218166003-b7e3ad61-f1da-4f25-b578-1188e75aff3b.jpg)

The timeline posts mentioning this character also appear using right to left text:

![Screenshot_20230210_101437](https://user-images.githubusercontent.com/69328423/218166370-2846b6dc-ae08-47b0-b165-89b310c9a657.jpg)

* * * *
- Tusky Version: 21.0
- Android Version: 12
- Android Device: Galaxy S10
- Mastodon instance (if applicable):

- [X] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

nikclayton:
The account is https://mastodon.social/api/v1/accounts/109354069933492315, if you pull that and look at the bytes for the profile info you get:

```
00000000  3c 70 3e e2 9a 94 ef b8  8f 20 e1 8e af ef ac a1  |<p>...... ......|
00000010  f0 9d 98 ad ea ad bf ef  ac a1 2c 20 74 68 65 20  |.........., the |
00000020  44 72 61 67 6f 6e 62 6f  72 6e 20 46 69 67 68 74  |Dragonborn Fight|
00000030  65 72 3c 62 72 20 2f 3e  0a                       |er<br />.|
```

If you paste the text in to an editor that supports bidirectionality (e.g., Android Studio) and run the cursor through the name text you'll see a marker at the top of the cursor swap direction as you move through the text.

The sword emoji is `e2 9a 94 ef b8 8f`, `e1 8e af ef ac a1 f0 9d 98 ad ea ad bf ef  ac a1` is the name. 

I thought this might be https://www.w3.org/International/articles/inline-bidi-markup/#nomarkup, but it doesn't look like there's embedded unicode to change the directionality of the text, it's the individual "letters" themselves.

https://www.babelstone.co.uk/Unicode/whatisit.html seems to confirm that, reporting:

```
U+13AF : CHEROKEE LETTER HI
U+FB21 : HEBREW LETTER WIDE ALEF
[...]

PR #3328, opened 2023-02-17
Be explicit about the `firstStrong` text direction on UGC

nikclayton:
`activity_account` sets the root text direction to `anyRtl`.

This is OK for UI elements, but can be a problem for user generated content (UGC) that may contain bidirectional text.

Fix this by explicitly restoring the default behaviour, `firstStrong`, on views in `activity_account` that can show UGC.

Fixes https://github.com/tuskyapp/Tusky/issues/3294


nikclayton:
I think this is correct (my analysis is in the linked PR, with links to other PRs that have dealt with this), but someone more knowledgeable than me about bidi text handling should take a look.

If you want to compare before/after, my profile bio (@nikclayton@mastodon.social) currently contains a mix of LTR and RTL text that did not render correctly before this PR, and now renders correctly with it.

Lakoja:
Also status are user-generated content.

So to be consistent with the Mastodon Web gui it should (at least) also be set for item_conversation.xml#status_content and item_status.xml#status_content.

(Note: In the Web gui it doesn't work for the username. The display of that seems to be a bit random: different in text field than actual display.)

Generally I wonder if it is possible to set an application-wide default for textDirection. (I'd expect a consistent behavior for _everything_, I guess.)

Lakoja:
Maybe it would also be sensible to remove the (3) text directions "any rtl" for a few activities.

nikclayton:
@Lakoja I don't think that's necessary, per https://github.com/tuskyapp/Tusky/issues/3294#issuecomment-1434681929. To summarise:

1. The Android default is `firstStrong`. This is why `item_conversation` and `item_status` work correctly before this PR.
2. This was overridden for some wrapper views in `activity_account.xml` in https://github.com/tuskyapp/Tusky/pull/964 (the discussion in the PR has screenshots showing why `anyRTL` makes sense in a few places)
3. This PR restores the default for some children of the wrapper views that were unintentionally affected by the change in (2)

nikclayton:
I checked this against the profile of https://mastodon.social/@nasser@merveilles.town who runs https://notarabic.com.

Without this PR the "pronouns" section of his account profile **does not** display correctly, appearing as "هو/he/him/".

This is (a) different from the web UI (which shows "he/him/هو") and (b) is also different from what you get if you copy that text out of Tusky and paste it in to an app which does handle this correctly. Tusky is showing the Arabic text in the wrong place (on the left) and has the spare "/" on the right.

So I'm now confident that this is the correct fix.

Issue #3323, opened 2023-02-16
Back button to open drawer/menu before closing app


0ddfactory:
Many apps with slide out menus were met with compatibility issues once gesture navigation rolled out. The best solution I've seen in apps for this is that when the back button is pressed instead of instantly closing the app it would first open the menu from the left as if you had swiped and then if the back button is pressed a second time it closes the app. Please consider adding this feature if it fits, thank you!

nikclayton:
I was going to suggest the same thing. I use "Feedly" on Android, which has this feature and I'm very used to it. Something about Tusky's behaviour here has felt "off" to me for ages, but I couldn't put my finger on it until this morning.

connyduck:
I'd rather not do that.

1) Opening the drawer (press the edge and pull) and back (swipe from edge) are two different gestures
2) The drawer can also be opened by clicking the avatars
3) Changing the behavior of the back button breaks existing habits

PR #3332, opened 2023-02-18
Allow "Back" to open the nav drawer

nikclayton:
Open the nav drawer if the user is on the first tab and presses "Back",

If the nav drawer is open, *and* it was opened by pressing "Back" then pressing "Back" again will finish() the activity.

Otherwise, pressing "Back" with the nav drawer open will close the drawer, as before.

Fixes https://github.com/tuskyapp/Tusky/issues/3323

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/MainActivity.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/MainActivity.kt
@@ -177,6 +177,9 @@ class MainActivity : BottomSheetActivity(), ActionButtonActivity, HasAndroidInje
     /** Adapter for the different timeline tabs */
     private lateinit var tabAdapter: MainPagerAdapter
 
+    /** True if the navigation drawer was opened by pressing "Back" */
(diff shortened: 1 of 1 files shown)


mcclure:
Hm, I feel like if I saw this behavior in an app (enter app, press back, it takes me to a new third pane instead of taking me back where I was) I would be extremely surprised.

nikclayton:
> Hm, I feel like if I saw this behavior in an app (enter app, press back, it takes me to a new third pane instead of taking me back where I was) I would be extremely surprised.

1. Have you actually tried it?
2. It doesn't take you to a new third pane. It pops out the menu (if appropriate), or exits the app (as before)
3. "extremely surprised" is not necessarily a bad thing. For example:

> I pressed 'back' out of habit in Tusky, and was extremely surprised that it anticipated that I wanted to open the menu, because other apps that I use also do that, and I am annoyed by Tusky's previous behaviour where I would keep having to restart the app because I accidentally quit it.

PR #3341, opened 2023-02-20
When looking up fediverse urls, verify that account results returned match the input query

### LINKED ISSUE #2804: Link lookup ends up with wrong link
I have been using Tusky for a good week now and noticed that every now and then clicking on links in posts take me to the wrong internet address.

E.g.:
1. Click a link in a post, which should go to www.patrion.com/…
2. Land on the Mastodon profile of the European Commision?
A bit like chat roulette I guess :-)

Here's a link to a raw video: https://cloud.zell-mbc.com/s/QbQ2xFzY2TajKgZ
Most links work, some don't.

Edit: Forgot to mention that when I click that same link from a different client (browser, FediLab, etc.) the correct page opens.

* * * *
- Tusky Version: 19.0
- Android Version: 10
- Android Device: Stoneage Samsung tablet, SM-T830
- Mastodon instance (if applicable): Doesn't matter

- [x ] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

Tak:
Fixes #2804

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/BottomSheetActivity.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/BottomSheetActivity.kt
@@ -86,8 +86,11 @@ abstract class BottomSheetActivity : BaseActivity() {
                     if (statuses.isNotEmpty()) {
                         viewThread(statuses[0].id, statuses[0].url)
                         return@subscribe
-                    } else if (accounts.isNotEmpty()) {
-                        viewAccount(accounts[0].id)
+                    }
+                    accounts.firstOrNull { it.url == url }?.let { account ->
+                        // Some servers return (unrelated) accounts for url searches (#2804)
+                        // Verify that the account's url matches the query
+                        viewAccount(account.id)
                         return@subscribe
                     }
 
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/keylesspalace/tusky/BottomSheetActivity.kt
```diff
@@ -86,8 +86,11 @@ abstract class BottomSheetActivity : BaseActivity() {
                     if (statuses.isNotEmpty()) {
                         viewThread(statuses[0].id, statuses[0].url)
                         return@subscribe
-                    } else if (accounts.isNotEmpty()) {
-                        viewAccount(accounts[0].id)
+                    }
+                    accounts.firstOrNull { it.url == url }?.let { account ->
```
Lakoja, line 90: (Just to make sure)

Does this url always _exactly_ match / need to?

E. g. maybe a surplus /. Or upper or lower case?

Tak: Mmm, it *should* always match for a pasted account url. And this is not used for mentions etc., but for instances where somebody explicitly pastes `https://some.instance/@User` into a status instead of just mentioning `@User`.

In the worst case, it just opens in the browser instead of in Tusky, which is much better than the current case of "when there's no exact search result for a url, you sometimes get back random accounts and then we open one of those in Tusky instead of the url you tapped"

connyduck (approved):

PR #3349, opened 2023-02-20
Use TypedArray.use to obtain attrs

Goooler:
https://github.com/androidx/androidx/blob/17346638fffb173a69801ee4c9ea293588800214/core/core-ktx/src/main/java/androidx/core/content/res/TypedArray.kt#L227-L236


connyduck (approved):
nice

Issue #3329, opened 2023-02-18
Ogg videos display broken


mcclure:
Got hold of a .ogg video file:

https://mastodon.social/@mcc_test_tusky/109882851902777246

mastodon.social web can't play the video at all in Chrome

Interestingly, Tusky 21 *can* play it, with an unusual problem: The video is displayed 1 pixel high. It animates and everything, it's just only 1 vertical pixel worth of content is shown.

[Exoplayer fork](https://github.com/mcclure/Tusky/tree/exoplayer) exhibits same problem, but displays exactly 1 frame of the full image.

* * * *
- Tusky Version: 21
- Android Version: 12
- Android Device: Xperia X 5 III
- Mastodon instance (if applicable): m.s

- [X] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

nikclayton:
The post with that video doesn't seem to be available any more. Do you have another one?

mcclure:
> The post with that video doesn't seem to be available any more. Do you have another one?

Looks like I accidentally posted it as "followers only". I have spotty internet access this weekend and it isn't letting me change the visibility level in an edit. If you follow mcc_test_tusky does it show you the post?

nikclayton:
I've sent a follow request from the Tusky account

mcclure:
follow accepted. sorry about that

nikclayton:
That works to see the post. Doesn't work in your fork or mine (https://github.com/tuskyapp/Tusky/pull/3857). I'll try with the Exoplayer demo app and see if that fails as well.

PR #3352, opened 2023-02-20
Switch video player to exoplayer

mcclure:
We currently use android.widget video players. The New Hotness is "exoplayer", which is now in androidx as "media3". Switching to this will get us future bugfixes, features and codecs (although I've yet to find a video exoplayer can view but videoview couldn't) and is probably a prerequisite to further work on the video player. It also gives us subtitle support if Mastodon ever adds that, and you can watch videos at 2x or slow speed now.

The patch itself is currently a hatchet job, I am following a [tutorial](https://developer.android.com/codelabs/exoplayer-intro#1) and have very little idea what I am doing. However it works at a basic level, you can view videos.

Things that need to be improved include:

- The nice "disappear player automatically" feature does not work yet (I commented it out). Ditto the "always show player for audio" feature
- The caption now appears and disappears at random, sometime flickering, I don't know why
- ~When the controls are visible, the entire video is dimmed dark. I don't like it and I don't know if this is supposed to happen or if we broke something. [I asked about this on Stack Overflow.](https://stackoverflow.com/questions/75514491/why-do-exoplayer-media3-controls-dim-the-video).~
- ~The controls are drawn in the center of the screen right over the video rather than aligned to the bottom like in the old view, it's very annoying. [I asked about this on Stack Overflow](https://stackoverflow.com/questions/75515091/how-to-reposition-exoplayer-media3-control-buttons).~
- Tutorial gives confusing advice about Android 24 and multiple window mode which I have not attempted to follow.
- Much code is now either dead code or has been commented out by me, often because I didn't know what it does. For example, what did ExposedPlayPauseVideoView do? I'm not sure but I think we can remove it now.
- Media3 Exoplayer is still in beta. This is out of our control and we can probably ignore it (the media3 "beta" corresponds to a release version of exoplayer).

Things I would like to address as long as we're enhancing the video view, but which can be deferred to another PR:

- I would like to give the media progress a bar of some sort
- [Ogg video problem](https://github.com/tuskyapp/Tusky/issues/3329)
- Thumbnail images on audio files (exoplayer has artwork/defaultArtwork properties good for this)


mcclure:
It looks like this

[](https://user-images.githubusercontent.com/277318/220211113-6db7431d-efc7-4e37-9da6-48fb8a9ca52c.png)

mcclure:
Exoplayer docs, for my benefit if no one else's:

[Conceptual docs](https://exoplayer.dev/listening-to-player-events.html)  
[Media3 including](https://developer.android.com/jetpack/androidx/releases/media3)  
[Player reference](https://developer.android.com/reference/androidx/media3/common/Player)  
[PlayerControlView reference](https://developer.android.com/reference/androidx/media3/ui/PlayerControlView)  
[PlayerView reference](https://developer.android.com/reference/androidx/media3/ui/PlayerView)

gradle/libs.versions.toml
```diff
@@ -130,6 +131,8 @@ rxjava3-core = { module = "io.reactivex.rxjava3:rxjava", version.ref = "rxjava3"
 rxjava3-kotlin = { module = "io.reactivex.rxjava3:rxkotlin", version.ref = "rxkotlin3" }
 sparkbutton = { module = "com.github.connyduck:sparkbutton", version.ref = "sparkbutton" }
 unified-push = { module = "com.github.UnifiedPush:android-connector", version.ref = "unified-push" }
+exoplayer-core = { module = "com.google.android.exoplayer:exoplayer-core", version.ref = "exoplayer" }
+exoplayer-ui = { module = "com.google.android.exoplayer:exoplayer-ui", version.ref = "exoplayer" }
```
mcclure, line 135: I don't understand why we need the changes in this file when we already include this information in build.gradle. Also, this seems to be including exoplayer whereas build.gradle is including media3. That is almost certainly wrong. However without these lines the fails with a [url=https://matrix.to/#/!PWkUGGpeMZWDPmdfRE:matrix.org/$1676600712263985pxBEA:matrix.org?via=m.charlag.space&via=matrix.org&via=tchncs.de]mysterious nullPointerException[/url]

app/src/main/java/com/keylesspalace/tusky/fragment/ViewVideoFragment.kt
```diff
@@ -105,8 +107,59 @@ class ViewVideoFragment : ViewMediaFragment() {
...
+        //player.playWhenReady = playWhenReady
+        //player.seekTo(currentItem, playbackPosition)
+        player.prepare()
+
+        mediaPlayerListener = object : Player.Listener {
+            override fun onEvents(player: Player, events: Player.Events) {
+                if (events.contains(Player.EVENT_PLAYBACK_STATE_CHANGED)) {
+                    if (player.getPlaybackState() == Player.STATE_READY) {
+/*
+                        // FIXME This causes the controls bar to move up. Why is it here?
```
mcclure, line 128: What *was* this supposed to do?

mcclure:
After appearance fixes

Screenshot_20230220-213114

Screenshot_20230220-220535

mcclure:
Copypasted from the Matrix:

>How is androidx.media3.ui.PlayerView different than using com.google.android.exoplayer2.ui.StyledPlayerView?

Apparently previously Google had both media2 and Exoplayer and they partially, but not completely, overlapped. Media3 merges media2 with Exoplayer2. 

I interpret this page https://developer.android.com/guide/topics/media/media3/exoplayer/mappings combined with this page https://developer.android.com/guide/topics/media/media3/getting-started/migration-guide to mean that media3 PlayerView is a drop in replacement for Exoplayer StyledPlayerView, and that Exoplayer PlayerView (deprecated) is not present in media3. 

There are API differences between original Exoplayer2 and media3 Exoplayer, which are negligible and seem to be well documented. Everything I've read implies they're supposed to be functionally identical, but at minimum the layout XMLs are not the same.

We appear to be importing media3 Exoplayer in build.gradle and Exoplayer2 in versions.toml. This should not work and seems to be breaking my Android Studio slightly. However if we remove Eric's versions.toml changes the entire app starts failing to boot…

nikclayton:
Not sure if you noticed (there are a bunch of different documentation pages for media3, and they don't all make it clear), but media3 is still in beta, and explicitly not recommended for use, per https://developer.android.com/guide/topics/media/media3.

The recommended APIs (which include pre-media3 Exoplayer) are described at https://developer.android.com/guide/topics/media, which includes a link to https://developer.android.com/guide/topics/media/exoplayer, which in turn links to the Exoplayer documentation at https://exoplayer.dev/. That last one is the link you want, I think.

Mixing and matching the two versions (per your preceding comment) would explain the breakages you've mentioned.

The problems you mentioned with the tutorial / codelab are probably addressed by the more comprehensive tutorials starting at https://exoplayer.dev/hello-world.html.

The specific changes I think you need to make are:

1. In `libs.versions.toml`, replace the two lines for `exoplayer-core` and `exoplayer-ui` with a single:

```
exoplayer = { module = "com.google.android.exoplayer:exoplayer", version.ref = "exoplayer" }
```

2. In `build.gradle`, revert the current changes, and add an:

```
implementation libs.exoplayer
```

entry.

3. Adjust the imports and API calls in the new code as appropriate, following the tutorials at exoplayer.dev.

app/src/main/res/layout/exo_player_control_view.xml
```diff
@@ -0,0 +1,155 @@
...
+     limitations under the License.
+-->
+<merge xmlns:android="http://schemas.android.com/apk/res/android">
+
+    <!-- 0dp dimensions are used to prevent this view from influencing the size of
+         the parent view if it uses "wrap_content". It is expanded to occupy the
+         entirety of the parent in code, after the parent's size has been
+         determined. See: https://github.com/google/ExoPlayer/issues/8726.
+    -->
+    <!-- Change from media3: Make transparent -->
```
nikclayton, line 23: Please don't make these completely transparent, it makes the controls impossible to see in some cases.

I just fixed something similar when viewing profiles.

mcclure: They are completely transparent in the current (Tusky 21) implementation. With our idiom of the controls appearing and disappearing at different moments, a full-screen curtain means a really unpleasant full-screen blinking effect… at minimum, I really want to find some way to avoid that.

Would it help if we used icons with internal backgrounds? For example a black play symbol inset in a white square.

(There's still a background on the play bar— just not the icons— and the play bar background is not unpleasant because it covers a small part of the screen and potentially doesn't cover the video at all.)

nikclayton: > Would it help if we used icons with internal backgrounds? For example a black play symbol inset in a white square.

Yes, that, but it should probably be a circle, it should use the Tusky colours, and there can be some transparency.

Take a look at https://github.com/tuskyapp/Tusky/pull/3272 - the trick is to use a `LayerDrawable` to compose the background drawable and the drawable with the icon.

mcclure:
> Not sure if you noticed…

@nikclayton , thanks , this is very helpful. I do see media3 is in beta, but do we know when it is likely to go stable? It moved from alpha to beta just while I've been poking at this patch in the last few weeks. The exoplayer and media3 interfaces are pretty similar, but it would still be nice if we could avoid switching to exoplayer just to have to switch again to media3 when the summer comes…

Perhaps I will do the patch for v22 with exoplayer but a media3-style layout xml, and maintain a floating PR to replace it with media3 once that leaves beta. Of course then I still have to figure out how to get media3 to compile…

mcclure:
…wait, actually as of last week there is a release candidate. https://developer.android.com/jetpack/androidx/releases/media3#1.0.0-rc01

nikclayton:
No idea when it's going to leave beta. At the moment the media3 API could change from release to release, so then you'd be left chasing a fluctuating media3 API.

I think it's better to go:

```
[where we are] --[this PR]--> [exoplayer] --[future PR]--> [media3 release]
```

rather than

```
[where we are] --[this PR]--> [media3 beta API] --[future PR]--> [media3 another beta API] --[future PR]--> [... more media3 APIs...] --[future PR]--> [media3 release]
```

since it bounds the number of potential API transitions.

mcclure:
Force pushed, this branch now uses "pure" exoplayer, no media3. I will make a separate PR for the media3 version.

mcclure:
BTW I had a minor nightmare around adapting exo_player_control_view into the pure-exoplayer branch:

We are now using exoplayer r2.18.3. If I pull out the exo_player_control_view from this version:

https://github.com/google/ExoPlayer/blob/r2.18.3/library/ui/src/main/res/layout/exo_player_control_view.xml git 4580b52

…and apply that, it doesn't work right, instead of a toggleable play/pause control there are adjacent play and pause controls and neither work. However if I pull the media3 version of the xml and massage it slightly, that works?? … except that now we have a mixture of the media3 play/pause control with the old-style ff/rew buttons and I don't think it looks very good. Either button set is ok by itself but the mix is a little strange. I will try to figure this out later.

mcclure:
Screenshot_20230222-161022

nikclayton:
Looking...

nikclayton:
Steps to troubleshoot.

- Find a status with a video to test, I used https://mastodon.social/@charlienorth/109910547950693062

- Verify the problem exists as described (it does, there's two separate play/pause buttons, and neither work).

- Make sure it's not a code problem, by renaming res/layout/exo_player_control_view.xml to res/layout/exo_player_control_view2.xml. If the problem persists then there's an issue with the code that needs to be fixed first.

- The problem persists; the stock control view has separate play/pause buttons, and they're not reacting.

Looking some more...

nikclayton:
Steps to troubleshoot.

- Find a status with a video to test, I used https://mastodon.social/@charlienorth/109910547950693062

- Verify the problem exists as described (it does, there's two separate play/pause buttons, and neither work).

- Make sure it's not a code problem, by renaming `res/layout/exo_player_control_view.xml` to `res/layout/exo_player_control_view2.xml`. If the problem persists then there's an issue with the code that needs to be fixed first.

- The problem persists; the stock control view has separate play/pause buttons, and they're not reacting.

- Setup in build.gradle / libs.versions.toml appears to be correct.

- Looking at ViewVideoFragment, Android Studio reports that `import com.google.android.exoplayer2.ExoPlayer` is unused. That's weird. When I have AS optimize the imports it replaces it with `import androidx.media3.exoplayer.ExoPlayer`. That smells suspicious, and warrants further investigation.

- "File > Invalidate Caches..." in AS fixed that, and allowed me to ensure the import is correct.

... continuing on ...

nikclayton:
- `fragment_view_video.xml` has this:

```
    <com.google.android.exoplayer2.ui.StyledPlayerView
        android:id="@+id/videoView"
        android:layout_width="wrap_content"
        android:layout_height="match_parent"
        app:layout_constraintBottom_toBottomOf="parent"
        app:layout_constraintLeft_toLeftOf="parent"
        app:layout_constraintRight_toRightOf="parent"
        app:layout_constraintTop_toTopOf="parent"
        app:use_controller="false"
        app:show_previous_button="false"
        app:show_next_button="false"
        app:controller_layout_id="@layout/exo_player_control_view" />
```

What's `app:use_controller` for?

- https://exoplayer.dev/doc/reference/com/google/android/exoplayer2/ui/StyledPlayerView.html says "Whether the playback controls can be shown.", and the default is `true`. Why is that disabled?

- Looks like it's enabled later in code when the player is in STATE_READY. When the does the player transition to STATE_READY?

- https://exoplayer.dev/listening-to-player-events.html says the initial state is STATE_IDLE, then transitions to STATE_READY when the player can play from current position. Setting a breakpoint on the `setUseController(true)` call confirms that it does fire.

- Video auto-plays (as expected). This is controlled by `.playWhenReady`. The documentation for this says

> This method must only be called if COMMAND_PLAY_PAUSE is available.

That's suspicious, and implies that play/pause is sometimes not available, which matches what's being seen. Under what circumstances is COMMAND_PLAY_PAUSE not available?

nikclayton:
- Hmm. `player.isCommandAvailable(COMMAND_PLAY_PAUSE)` returns `true` at important points.

- Is the player setup happening at the correct point in the fragment lifecycle? Current call stack is:
  - ViewVideoFragment.onViewCreated() calls...
  - ViewMediaFragment.finalizeViewSetup() calls...
  - ViewVideoFragment.setupMediaView(), which sets up the player.

- ExoPlayer documentation isn't specific. The sample app (https://github.com/google/ExoPlayer/blob/release-v2/demos/main/src/main/java/com/google/android/exoplayer2/demo/PlayerActivity.java) creates the player in `onStart()`. This is after `onViewCreated()`. What happens if the player initialisation is moved to `onStart()`?

- Appears not to make a difference.

- Are the play/pause buttons being obscured by another transparent view? Not obviously, at least, not by anything that's also not overlaying the rewind and fast forward buttons, and those buttons work.

- What else in the code is looking for touches? Swipe down to dismiss? Disabling that doesn't change things.

- Oh!

mcclure:
> What's app:use_controller for?

It is used to hide the controls until loading is done.

nikclayton:
https://github.com/google/ExoPlayer/blob/r2.18.3/library/ui/src/main/res/layout/exo_styled_player_control_view.xml looks like the correct layout to use. Why isn't it being picked up automatically?

Because it's explicitly overridden in the layout, 

```
    <com.google.android.exoplayer2.ui.StyledPlayerView
        android:id="@+id/videoView"
        android:layout_width="wrap_content"
        android:layout_height="match_parent"
        app:layout_constraintBottom_toBottomOf="parent"
        app:layout_constraintLeft_toLeftOf="parent"
        app:layout_constraintRight_toRightOf="parent"
        app:layout_constraintTop_toTopOf="parent"
        app:use_controller="false"
        app:show_previous_button="false"
        app:show_next_button="false"
        app:controller_layout_id="@layout/exo_player_control_view" />
```

Delete `app:controller_layout_id` and it works as expected.

nikclayton:
So any modifications to the controls need to use the `..styled..` XML files, not the ones without `...styled...` in the name.

mcclure:
> So any modifications to the controls need to use the `..styled..` XML files, not the ones without `...styled...` in the name.

Thanks, this is what I was missing. The xml file is named differently between media3 and exoplayer :( This branch now looks like the media3 one again and future changes will be easier to do in lockstep

[](https://user-images.githubusercontent.com/277318/220239887-09a72ddb-a1f3-4c38-85ae-01dbe184d11e.png)

Screenshot_20230220-220535

nikclayton:
Just poking through the exoplayer docs I noticed a number of features / options that I recommend using or setting:

- [ ] Show metadata in the UI (optional displayTitle and artist, from https://exoplayer.dev/doc/reference/com/google/android/exoplayer2/MediaMetadata.html and https://exoplayer.dev/retrieving-metadata.html)
- [ ] Explicitly use OkHttp (https://exoplayer.dev/network-stacks.html)
- [ ] Debug logging (https://exoplayer.dev/debug-logging.html, if `BuildConfig.DEBUG` is true)
- [ ] Display errors to the user (https://exoplayer.dev/listening-to-player-events.html -- this is important, if we don't do this and a user says "This video doesn't work", and we can't reproduce it, we have no way of knowing what the underlying problem is)
- [ ] If used for audio playback, enable audio offload to reduce the battery impact (https://exoplayer.dev/battery-consumption.html#audio-playback)

mcclure:
media3-exoplayer is now stable, so we should work on #3375 only going forward.

PR #3363, opened 2023-02-21
Enforce lint cleanliness when building

nikclayton:
The volume of lintable issues is growing. To stem the tide:

1. Add `app/lint-baseline.xml`, which contains the current set of lint issues. Any issue appearing here will not cause the build to fail.

2. Move lint configuration settings in to `lint.xml`.

3. Update the lint configuration so that any issue (i.e., any issue not in lint-baseline.xml) causes a build failure.

4. Add the lint tasks as depedency when assembling an APK, to ensure the lint checks are run.


nikclayton:
Running this as part of the normal `assemble${flavour}Debug` tasks is probably not tenable, a build with a one character change to resources (`gradlew assembleGreenDebug`) takes 15s without lint, 1m18s with lint.

Running this as part of the `primary` task in `bitrise.yml` is still feasible, and/or using a GitHub action to run it on PRs to catch early problems.

nikclayton:
The PNG -> WEBP conversion in this PR is one of the issues that lint flags.

On one hand, the WEBP images are smaller.

On the other hand, GitHub does not support previewing / showing diffs of changes to WEBP images, so it might be better to keep them as PNG at the moment and disable the lint check.

connyduck:
Nice. Which gradle command should I add to bitrise after merging? Just `lint`?

nikclayton:
> Nice. Which gradle command should I add to bitrise after merging? Just `lint`?

Yes. Although this could also easily be run as a GitHub action which might be accessibly (unless the Bitrise config is in the repo somewhere and I can't find it).

Either way, it can be run in parallel to the current tests.

PR #3375, opened 2023-02-22
Upgrade exoplayer to media3 exoplayer

mcclure:
Extension of https://github.com/tuskyapp/Tusky/pull/3352 .

media3 exoplayer is an *almost* API-identical version of exoplayer. It is where future patches will land. It is what we want to be using. It is not officially, stably releasedyet. It is in release candidate 1, which is pretty good, but [at least one Google page doesn't recommend using it until it's stable](https://developer.android.com/guide/topics/media/media3).

If we decide rc1 is good enough we can merge this PR, otherwise we can merge the other PR and switch to this one when media3 stabilizes.


mcclure:
Looks like this

[](https://user-images.githubusercontent.com/277318/220239887-09a72ddb-a1f3-4c38-85ae-01dbe184d11e.png)

Screenshot_20230220-220535

mcclure:
Merged with current develop, upgraded to 1.0.1 (stable). This can release as soon as there are no performance/behavior regressions vs main tree.

I am looking for someone to take over this PR.

In current test, the caption flickering is not as bad. It does not occur during playback, only when you are visible-ing and de-visible-ing the captions it flickers a few times before appearing/disappearing. It does seem to be erratic whether tapping causes it to visible or de-visible (that is sometimes when invisible you tap, it flickers but stays invisible, you tap again, it flickers and becomes visible. not sure). Test video: https://mastodon.social/@mcc_test_tusky/110340735175845640

1.0.1 seems to blast a LOT of logcat spam while playing video.

nikclayton:
Closing this PR, as the work continues in https://github.com/tuskyapp/Tusky/pull/3857

Issue #3311, opened 2023-02-11
Make the "load more" break stand out more


grrrrr:
Currently in the subscriber feed tab  you will get to a point going back in history that you are presented with a "load more" button/option.

This "load more" button tends to blend into the toot feed making it easy to scroll past the option to load more toots.

It would be useful if the words "load more" could be presented in a different way to make it more obvious.   My first thought would be having the words "load more" in a stronger color such as red,   though I am sure consideration in color accessibility should be taken into account

* * * *
- Tusky Version: <=21.0
- Android Version: Android 13
- Android Device: Pixel 6
- Mastodon instance (if applicable):

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

Lakoja:
I second this: I often have a hard time spotting the "Load more". I only notice because the status below seem to be old and familiar.

connyduck:
We already made it bigger once, I think it is good now. https://github.com/tuskyapp/Tusky/pull/2765

grrrrr:
Hi @connyduck  I am a little confused  in the pull request you link to you say you want feedback and will revisit if necessary

> merging to see what users say, we may have to adjust it again.

But are closing this discussion as if you don't want feedback,  despite myself and @Lakoja saying there are usability issues with how the "load more" button is post merge

Did I miss something?

connyduck:
Yeah that was meant if someone thinks it is too big now. Personally I think so and it trips me up every time I see it, but I guess if someone wants to send another pull request I can have a look at it again 🤷‍♂️

Lakoja:
(I'd prefer to get rid of it anyway...)

Lakoja:
But will create a proposal. :-)

Lakoja:
Using the divider color as background works well, I guess. It is also ok from a concept point: as that element is a divider...

However it does not work for the dark theme. So I guess it must be a themable color?

![fff513f52f53d4ff](https://user-images.githubusercontent.com/1618905/218851311-57abd2c0-a34b-47ec-be8c-8894cee20667.png)

connyduck:
I like the different color. Could work well with this zig zag tear some other apps show
![Screenshot_20230216-204143](https://user-images.githubusercontent.com/10157047/219470333-449b2568-d11e-4406-924b-7c370ab7953e.png)

Lakoja:
I'll see what I can do.

Lakoja:
That is what I can come up with.
[...]

PR #3376, opened 2023-02-23
3311 load more more prominent

Lakoja:
Fixes https://github.com/tuskyapp/Tusky/issues/3311

I went for a gradient (instead of a zig-zag line) because that matches the other status separators better. And especially it integrates the existing separators on top and bottom of such a placeholder.

2b641fa65a85f8a8_kl


connyduck:
Ok, if you don't want zig zag, then please do just the darker background, I think gradient is the worst of the options.

Lakoja:
> Ok, if you don't want zig zag, then please do just the darker background, I think gradient is the worst of the options.

I want it but can't do it. :-)

So, ok, I'm using only another color now.

connyduck (approved):

PR #3377, opened 2023-02-23
Remove stale pre-edit statuses from the thread view

### LINKED ISSUE #3366: Editing a reply in thread view duplicates that reply
Reported by a user via email, reproduces both in current nightly and Tusky 21.0.

Steps to reproduce:
- Open a post you replied to in thread view.
- Edit the reply
- The reply is now shown twice, once with the old unedited content, and once with the new content

Screenshot_20230221-200553

Tak:
Fixes #3366

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/appstore/Events.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/appstore/Events.kt
@@ -15,6 +15,7 @@ data class MuteEvent(val accountId: String) : Dispatchable
 data class StatusDeletedEvent(val statusId: String) : Dispatchable
 data class StatusComposedEvent(val status: Status) : Dispatchable
 data class StatusScheduledEvent(val status: Status) : Dispatchable
+data class StatusEditedEvent(val originalId: String, val status: Status) : Dispatchable
 data class ProfileEditedEvent(val newProfileData: Account) : Dispatchable
 data class PreferenceChangedEvent(val preferenceKey: String) : Dispatchable
 data class MainTabsChangedEvent(val newTabs: List<TabData>) : Dispatchable
--- a/app/src/main/java/com/keylesspalace/tusky/components/timeline/TimelineFragment.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/components/timeline/TimelineFragment.kt
@@ -45,6 +45,7 @@ import com.keylesspalace.tusky.adapter.StatusBaseViewHolder
 import com.keylesspalace.tusky.appstore.EventHub
 import com.keylesspalace.tusky.appstore.PreferenceChangedEvent
 import com.keylesspalace.tusky.appstore.StatusComposedEvent
+import com.keylesspalace.tusky.appstore.StatusEditedEvent
 import com.keylesspalace.tusky.components.accountlist.AccountListActivity
 import com.keylesspalace.tusky.components.accountlist.AccountListActivity.Companion.newIntent
 import com.keylesspalace.tusky.components.preference.PreferencesFragment.ReadingOrder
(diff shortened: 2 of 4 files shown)


app/src/main/java/com/keylesspalace/tusky/fragment/SFragment.kt
```diff
@@ -85,6 +87,10 @@ abstract class SFragment : Fragment(), Injectable {
     @Inject
     lateinit var timelineCases: TimelineCases
 
+    private val composeLauncher = registerForActivityResult(ComposeResultContract()) { result ->
```
Lakoja, line 90: "result" is the status id here?

app/src/main/java/com/keylesspalace/tusky/components/compose/ComposeViewModel.kt
```diff
@@ -71,13 +71,13 @@ class ComposeViewModel @Inject constructor(
...
     private var startingContentWarning: String = ""
     private var inReplyToId: String? = null
-    private var originalStatusId: String? = null
     private var startingVisibility: Status.Visibility = Status.Visibility.UNKNOWN
 
     private var contentWarningStateChanged: Boolean = false
     private var modifiedInitialState: Boolean = false
     private var hasScheduledTimeChanged: Boolean = false
 
+    var originalStatusId: String? = null
```
Lakoja, line 80: This exposes the variable now?

app/src/main/java/com/keylesspalace/tusky/components/compose/ComposeResultContract.kt
```diff
@@ -0,0 +1,34 @@
...
+ * You should have received a copy of the GNU General Public License along with Tusky; if not,
+ * see <http://www.gnu.org/licenses>. */
+
+package com.keylesspalace.tusky.components.compose
+
+import android.content.Context
+import android.content.Intent
+import androidx.activity.result.contract.ActivityResultContract
+
+class ComposeResultContract : ActivityResultContract<ComposeActivity.ComposeOptions, String>() {
```
Lakoja, line 22: Technical curiosity: Does this need a contract?

(The only other two usages I see are "take picture" and "login with browser" - so external stuff.)

Tak: It looked to me like it did? As in, android studio yelled at me that the old `startActivityForResult` was deprecated, and the docs led me to `ActivityResultContract` 🤷

app/src/main/java/com/keylesspalace/tusky/fragment/SFragment.kt
Tak: yes

app/src/main/java/com/keylesspalace/tusky/components/compose/ComposeViewModel.kt
Tak: Yes, so the activity can put it into the result

app/src/main/java/com/keylesspalace/tusky/fragment/SFragment.kt
Lakoja: Then how about "statusId"?

app/src/main/java/com/keylesspalace/tusky/components/compose/ComposeViewModel.kt
Lakoja: I see. That is done for other variables here as well...

connyduck:
Wouldn't it be better to send a `StatusEditedEvent` similar to a `StatusComposedEvent`, instead of introducing a new mechanism? cc @nikclayton

Tak:
Good point, I've reworked it to use the event hub mechanism

connyduck (approved):

Issue #1567, opened 2019-11-21
copy and paste of a toot containing link adds garbage to links


novadeviator:
When copying the whole toot that contains a link and pasting the text into new toot, Tusky adds "%E2%80%8B" at the end of the actual link. The text in the published toot is the same and looks ok, but actual link points to original URL with added garbage which then prevents from visiting that link (usually 404).

STEPS TO REPRODUCE:

* find any public toot with link in it
* tap on it to open it as separate toot
* long press on it to copy to clipboard
* compose new toot
* long press into it and choose paste, and send
* tap the link in the new sent toot
* it will end up in 404 or not where original links was pointing

an example of link with added characters:
http://scholar.harvard.edu/files/mickens/files/towashitallaway.pdf%E2%80%8B

* * * *
- Tusky Version: 9.1
- Android Version: 7.0
- Android Device: HUAWEI VNS-L21

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

JuanjoSalvador:
Seems related to HTML sanitization.

connyduck:
https://github.com/tuskyapp/Tusky/blob/78b5aa8baf0c34fb5ed00742ec3174dc5bb33fa3/app/src/main/java/com/keylesspalace/tusky/util/LinkHelper.java#L121

charlag:
Maybe we should copy original text somehow? Or it will be HTML and not very nice?

charlag:
@connyduck did we fix this?

modelmat:
https://github.com/tuskyapp/Tusky/blob/ad001d044d9f6b354c948758abe9d6fb2e8036f1/app/src/main/java/com/keylesspalace/tusky/util/LinkHelper.kt#L92-L97

Still an issue in the app too

connyduck:
Maybe there is another way to prevent the large link hitboxes than to add this special character?

charlag:
That would be the best

nikclayton:
A full reproduction recipe is:

1. Bookmark https://chaos.social/@netzpolitik_feed/109365395616419194 (so it's easy to get back to if you're testing fixes)
2. Open the status
3. Long-pressed the link at the bottom of the status (which displays as `https://netpolitik.org/2022/twitter-exodus-wie-behoerden-medien-und-ngos-mastodon-fuer-sich-entdecken/`)
4. Copy the link
5. Click back twice to go Bookmarks -> Timeline
6. Click the compose FAB
[...]

PR #3380, opened 2023-02-24
Ignore clicks outside the start/end of a line

nikclayton:
`LinkMovementMethod` has a bug in its calculation of the clickable width of a span on a line. If the span is the last thing on the line the clickable area extends to the end of the view. So the user can tap what appears to be whitespace and open a link.

Previous code tried to fix this by adding a zero-width space after the link so that `LinkMovementMethod` wouldn't consider it empty. However the ZWS was selected by copy/paste operations, resulting in junk results if users tried to copy the link.

Fix this by subclassing `LinkMovementMethod` and fixing the click detection code to ignore clicks that are outside the bounds of the line that was clicked on.

Remove the code that adds the ZWS.

Fixes https://github.com/tuskyapp/Tusky/issues/1567

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/util/LinkHelper.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/util/LinkHelper.kt
@@ -347,7 +347,5 @@ object NoTrailingSpaceLinkMovementMethod : LinkMovementMethod() {
         return super.onTouchEvent(widget, buffer, event)
     }
 
-    fun getInstance(): NoTrailingSpaceLinkMovementMethod {
-        return NoTrailingSpaceLinkMovementMethod
-    }
+    fun getInstance() = NoTrailingSpaceLinkMovementMethod
 }
--- a/app/src/main/java/com/keylesspalace/tusky/util/LinkHelper.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/util/LinkHelper.kt
@@ -330,7 +330,7 @@ private const val TAG = "LinkHelper"
  *
  * See https://github.com/tuskyapp/Tusky/issues/1567.
  */
-class NoTrailingSpaceLinkMovementMethod : LinkMovementMethod() {
+object NoTrailingSpaceLinkMovementMethod : LinkMovementMethod() {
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/keylesspalace/tusky/util/LinkHelper.kt
```diff
@@ -322,3 +317,46 @@ fun looksLikeMastodonUrl(urlString: String): Boolean {
...
+        private var instance: NoTrailingSpaceLinkMovementMethod? = null
+
+        fun getInstance(): NoTrailingSpaceLinkMovementMethod {
+            if (instance == null) {
+                instance = NoTrailingSpaceLinkMovementMethod()
+            }
+
+            return instance!!
+        }
+    }
```
connyduck, line 361: make `NoTrailingSpaceLinkMovementMethod` a Kotlin `object` then this whole business is not needed

app/src/main/java/com/keylesspalace/tusky/util/LinkHelper.kt
```diff
@@ -322,3 +317,46 @@ fun looksLikeMastodonUrl(urlString: String): Boolean {
...
+ * the span is the last thing on the line the clickable area extends to the end of the view. So the
+ * user can tap what appears to be whitespace and open a link.
+ *
+ * Fix this by overriding ACTION_UP touch events and calculating the true start and end of the
+ * content on the line that was tapped. Then ignore clicks that are outside this area.
+ *
+ * See https://github.com/tuskyapp/Tusky/issues/1567.
+ */
+class NoTrailingSpaceLinkMovementMethod : LinkMovementMethod() {
+    override fun onTouchEvent(widget: TextView?, buffer: Spannable?, event: MotionEvent?): Boolean {
```
connyduck, line 334: I think all three parameters are never null. If they were, super would crash.

connyduck (commented):
😍

app/src/main/java/com/keylesspalace/tusky/util/LinkHelper.kt
nikclayton: Good idea. I've kept the `getInstance()` method for consistency with `LinkMovementMethod`.

connyduck (approved):

PR #3382, opened 2023-02-25
Meet accessibility guidelines for clickable spans

nikclayton:
Clickable spans in textviews do not normally meet the Android accessibility guidelines of a minimum 48dp square touch target.

This can't be fixed with a `TouchDelegate`, as the span is not a separate view to which the delegate can be attached.

Add `ClickableSpanTextView`. If used instead of a `TextView` any spans from `ClickableSpan` will have their touchable area extended to meet the 48dp minimum.

The touchable area is still bounded by the size of the view.

If two spans are closer together than 48dp then the closest span to the touch wins.


nikclayton:
Couple of things, to head off some potential review comments.

1. Yes, there is some debug code in here. It's extremely useful, and is normally compiled out (in both release and debug mode). It needs to be explicitly enabled in the code.

nikclayton:
2. The code that finds the best rectangle if they overlap. There may be an algorithm that gives better results than that, but I couldn't find anything -- lots of literature on hit-testing two or more rectangles, no literature on which is the "closest" to the touch. It's not a difficult problem, but I could imagine that the mathematically "best" algorithm might not give a result that users would expect.

nikclayton:
3. In pathological cases this can result in a status that is almost entirely clickable. For example, this screenshot:



However, the left hand edge of the status is still clickable to go to the detailed view, and that screenshot is with the text size set to the smallest. At the largest you get this:

connyduck:
I'm a bit concerned that this might have a performance impact and it is so much rather complicated code for something I'd expect the platform to handle, but well...

PR #3388, opened 2023-02-26
Convert build scripts to kts

Goooler:
https://docs.gradle.org/current/userguide/migrating_from_groovy_to_kotlin_dsl.html


nikclayton:
Please, not without benchmarks that show this does not affect build times for typical developer operations.

See https://github.com/gradle/gradle/issues/15886 for ongoing issues with build speed and the Kotlin DSL.

mcclure:
I don't want the builds to be any slower than they already are either, BUT I think this would have made my life easier when I was working on the Gradle recently.

nikclayton:
I've got gradle-profiler installed, so I'll profile this, comparing build times with this branch and the commit it's branched from, 000681c7023a742ad4e3612624541dc016efda56

nikclayton:
Partial results are in https://docs.google.com/spreadsheets/d/1UEdp85whNk0xkMLNw2X8EW9oEeDH5R0TRGCQPQ_aG9s/edit?usp=sharing

### Summary

- Completely clean builds are ~ 22% faster with this PR (at p75)
- Incremental builds are ~ 10% slower with this PR (at p75)
- Unit tests are TBD, I'll profile those tomorrow.

### Benchmarks

Benchmarks done with `gradle-profiler`, configured to do 6 warm up builds (to get everything in the cache, JIT'd, etc), then 10 builds. It's the last 10 that are measured here.

### Clean builds

I tried benchmarking a completely clean build, which is equivalent to running `./gradlew clean assembleBlueDebug` with `org.gradle.caching=false` so the build cache is ignored (and `clean` is actually effective).

Unexpectedly, that was quite a bit faster with the PR.

Without the PR builds were taking ~ 1m50 to complete (p75), with the PR 1m28 (p75), and a smaller standard deviation without the PR.

This did not match my expectations given the gradle issue I linked to above, so I re-ran the benchmark without the PR several times, in case there was something odd happening on my computer that affected the first run. But I can reliably reproduce the result, so I think it's sound.

The scenario config is:

```
assemble {
    title = "assembleBlueDebug"
    tasks = ["assembleBlueDebug"]
    cleanup-tasks = ["clean"]
    system-properties {
        "org.gradle.caching" = false
    }
}
```

### Incremental builds

I had the profiler simulate adding a new public method to class, modifying a string resource, and adding a new string resource, to check incremental builds. These were done with the build cache enabled.

With those the performance is closer. On 8 of the runs it's faster **without** this PR. P75 build time is 31s without this PR, 34s with it.

So this PR is about 10% slower for common developer operations.

The scenario config is:

```
incremental_build_abi_resources_change_noclean {
    title = "Incremental build with ABI and resources change"
    tasks = [":app:assembleBlueDebug"]

    apply-abi-change-to = "app/src/main/java/com/keylesspalace/tusky/fragment/NotificationsFragment.java"
    apply-android-resource-change-to = "app/src/main/res/values/strings.xml"
    apply-android-resource-value-change-to = "app/src/main/res/values/strings.xml"
}
```

connyduck:
thx @nikclayton

mcclure:
For the record, I'm not sure that these results should disqualify the patch. I feel like 31 -> 34 seconds is not really that big a gap (I might even consider saving 30 seconds on clean build to lose 3 seconds on incremental build a win, even if incremental build is much more common).

I have a suspicion our build could be optimized more than it now is, as most of the time seems to come from running base Android tools such as the dex tool. Those optimizations might actually be easier to implement coming from a kotlin-buildscript base…

connyduck:
🤷‍♂️

nikclayton:
> as most of the time seems to come from running base Android tools such as the dex tool.

That's not at all what I see with build analyzer. On a recent incremental build (just changed some resources), the build took 41.2s.

Of that time:

11.6s (28.1%) - :app:kaptGenerateStubsBlueDebugKotlin
9.2s (22.3%) - :app:compileBlueDebugKotlin
7.9s (19.3%) - :app:compileBlueDebugJavaWithJavac
3.6s (8.8%) - :app:processBlueDebugResources
3.0s (7.3%) - :app:dexBuilderBlueDebug

kapt is super-slow. It's also in maintenance mode (https://kotlinlang.org/docs/kapt.html), but we can't switch to the replacement (KSP), because KSP doesn't support Dagger  yet (https://kotlinlang.org/docs/ksp-overview.html#supported-libraries).

The DI library it does support (KOIN) isn't appropriate because KOIN achieves its build speedups by operating at run time rather than compile time. So a mistake that would be a compile time error for a single developer becomes a run time error that can affect 1000s of users.

I've experimented with the recommendations at https://developer.android.com/studio/build/optimize-your-build over the last few days. We already do most of it, and I've sent PRs that do some more (parallel GC).

But the biggest impact is likely to come from https://developer.android.com/studio/build/optimize-your-build#create_libraries, because that would turn huge swathes of the build in to no-ops much of the time.

I haven't done any experiments with that yet because it requires a bit of time to think about the right way to carve up the current code in to additional modules, and I've got a couple of large PRs in flight at the moment, so the amount of headspace I've got to consider another large refactoring is currently limited. But I intend to look at this in more detail when the Notifications refactoring PR is merged.

Goooler:
The biggest blocker is replacing Kapt with Ksp, you can try out [Napt](https://github.com/sergei-lapin/napt) to kill the Kapt.

mcclure:
> > as most of the time seems to come from running base Android tools such as the dex tool.
> 
> That's not at all what I see with build analyzer. 

That's very interesting, your results are very different from mine. To prevent derailing the thread I created a separate issue (#3412)

nikclayton:
> The biggest blocker is replacing Kapt with Ksp, you can try out [Napt](https://github.com/sergei-lapin/napt) to kill the Kapt.

Cheers, I'll give that a try.

Goooler:
Easier to land this after #3984 is merged.

mcclure:
I still think this was a good idea, pity we couldn't come to a conclusion about how to go about it.

Goooler:
Some conflicts need to be resolved, I can resent this when we are ready.

Issue #3301, opened 2023-02-11
Prevent users from adding multiple instances of the same timeline


nikclayton:
With the current code the user can go to "Account preferences > tabs" and add a second HOME tab (and all the other "fixed" types, like notifications, etc).

That shouldn't be possible.

PR #3390, opened 2023-02-27
prevent adding multiple tabs of the same type

connyduck:
(unless the tab is configurable, like hastag or list)
The code for this [was always in place](https://github.com/tuskyapp/Tusky/blob/6cc79c8d752925eb57d5ae693b0e4396ad794271/app/src/main/java/com/keylesspalace/tusky/TabPreferenceActivity.kt#L301-L324), but it did not work because TabData.equals did not compare correctly.

fixes https://github.com/tuskyapp/Tusky/issues/3301

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/TabData.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/TabData.kt
@@ -43,63 +43,81 @@ data class TabData(
     val fragment: (List<String>) -> Fragment,
     val arguments: List<String> = emptyList(),
     val title: (Context) -> String = { context -> context.getString(text) }
-)
+) {
+    override fun equals(other: Any?): Boolean {
+        if (this === other) return true
+        if (javaClass != other?.javaClass) return false
+
+        other as TabData
+
+        if (id != other.id) return false
+        if (arguments != other.arguments) return false
+
+        return true
+    }
+
+    override fun hashCode(): Int {
+        var result = id.hashCode()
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/keylesspalace/tusky/TabData.kt
```diff
@@ -43,63 +43,81 @@ data class TabData(
...
+
+        other as TabData
+
+        if (id != other.id) return false
+        if (arguments != other.arguments) return false
+
+        return true
+    }
+
+    override fun hashCode(): Int {
```
Lakoja, line 59: Is it used as hash key somewhere?

Tak (approved):

app/src/main/java/com/keylesspalace/tusky/TabData.kt
nikclayton: Doesn't matter. If you override `equals()` you must override `hashcode()`, it's part of the contract between the two methods.

app/src/main/java/com/keylesspalace/tusky/TabData.kt
```diff
@@ -43,63 +43,81 @@ data class TabData(
...
+        var result = id.hashCode()
+        result = 31 * result + arguments.hashCode()
+        return result
+    }
+}
```
nikclayton, line 64:
```suggestion
    override fun hashCode() = Objects.hash(id, arguments)
```

PR #3394, opened 2023-02-27
include card and collapsed state in instant expanded change

connyduck:
When the spoiler warning is toggled, we update the ViewHolder immediately, but until now the "show more/less" button of collapsible posts and the link preview card were not included in this immediate update. The subsequent update through the rebinding of the ViewHolder sets their state correctly but a weird two step animation occurs, were first the post content is shown/hidden, and in a second step the card. This always felt extremely glitchy to me, I'm glad I could fix it.

Here you can see the glitch but it happens very quickly:
https://user-images.githubusercontent.com/10157047/221639655-03a56165-69b2-4dd9-b15c-a74643e0828b.mp4

I also experimented with removing the immediate updates completly and relying on the adapter rebind entirely, but that makes it feel a bit laggy so I went this way.


Tak (approved):
Looks ok to me, but there are a lot of tricky boolean negations…

Issue #3217, opened 2023-01-26
Sometimes the tab title is really tiny


mcclure:
Left: Normal
Right: Sometimes it does this and I don't know why
Have seen in both Home and Notifications tabs

[<img width=250 src="https://user-images.githubusercontent.com/277318/214759191-740004fc-9788-485d-9ff2-8e4453cc8f14.png">](https://user-images.githubusercontent.com/277318/214759191-740004fc-9788-485d-9ff2-8e4453cc8f14.png) [<img width=250 src="https://user-images.githubusercontent.com/277318/214759264-1f96f635-199d-42ce-984d-43a93f25cc47.png">](https://user-images.githubusercontent.com/277318/214759264-1f96f635-199d-42ce-984d-43a93f25cc47.png)

* * * *
- Tusky Version: 21.0. I am logged in on 2 accounts but I have the "don't show account name" option selected
- Android Version: 12
- Android Device: Sony XPeria 5 III
- Mastodon instance (if applicable): mastodon dot social

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

connyduck:
That is because Android makes it smaller when you start the Activity in landscape mode. We don't recreate Activities on rotation so it stays this way when switching to portrait (and vice versa). The solution is to use a ViewModel that holds the state and recreate the MainActivity on rotation. Forcing it  to be the same size is probably also possible but I guess there is a reason it is smaller in landscape.

nikclayton:
Note to self: This is:

```
        <activity
            android:name=".MainActivity"
            android:configChanges="orientation|screenSize|keyboardHidden|screenLayout|smallestScreenSize"
            android:exported="true">
```

and similar in AndroidManifest.xml.

## MainActivity

TabData.kt defines the kinds of tabs that are supported and the Fragments they host:

They are `TimelineFragment`, `NotificationsFragment`, and `ConversationsFragment`.

Of those, only `NotificationsFragment` doesn't have a view model. But I'm fixing that in https://github.com/tuskyapp/Tusky/pull/3159. When that's merged this will be fixed.

## ViewThreadActivity

`ViewThreadFragment` has a view model. The activity doesn't have any additional UI.

## ViewMediaActivity

Does not host a fragment, and does not have a view model.

## AccountActivity

Has a view model.

---

So I think only `ViewMediaActivity` has to be converted.

mcclure:
[...]

PR #3396, opened 2023-02-28
3335: Use different bounding aspect ratios for landscape

### LINKED ISSUE #3335: problem with image-preview in timeline (new in v21)
I'm using tusky on a tablet, horizontally.
now, vertical images in the timeline are displayed bigger than the viewport. this is very inconvenient. imho the previous behavior was perfect. (also on the vertical phone!) i want to see the posts, and enlarge images if interested. i believe, that one should be able to see a posts header and footer on screen by default. (that's why expanding longer posts on demand is great.)
while i appreciate the approach of having as least options in settings as possible, i think, if you want too keep that behavior, you need to add an option for the previews image-previews for timelines - since it was way more useful/meaningful, imho.

* * * *
- Tusky Version: 21

- [y ] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

Lakoja:
Improves / fixes #3335


app/src/main/java/com/keylesspalace/tusky/util/AttachmentHelper.kt
```diff
@@ -27,10 +28,16 @@ private fun formatDuration(durationInSeconds: Double): String {
 }
 
 fun List<Attachment>.aspectRatios(): List<Double> {
+    val displayMetrics = Resources.getSystem().displayMetrics
```
Lakoja, line 31: Not sure how to cache this (or if necessary) in this helper method.

On could move "deviceIsHigher" outside - there are two calling locations.

This should maybe be a dependency ("service", singleton)? But that would need to have state...

app/src/main/java/com/keylesspalace/tusky/util/AttachmentHelper.kt
```diff
@@ -27,10 +28,16 @@ private fun formatDuration(durationInSeconds: Double): String {
...
+    val deviceIsHigher = displayMetrics.heightPixels > displayMetrics.widthPixels
+
+    val minAspect = if (deviceIsHigher) 0.5 else 1.2
+    val maxAspect = if (deviceIsHigher) 2.0 else 3.0
```
Tak, line 35: Why specifically these constants for landscape?

Lakoja: I tried to reverse "the idea of 0.5 to 2 for a portrait phone". However the literal 1 to 4 is probably a bit too much.

So these are "best and simple guesses for a wider range of landscape".

Generally one could discuss if this should be relative to the actual aspect ratio value of the device. (After all simply "height > width ?" is somewhat generic)

However I would venture that this is a simple solution.

(1.2 and not 1 because at least phones are nowadays much much higher than wide.)

app/src/main/java/com/keylesspalace/tusky/util/AttachmentHelper.kt
nikclayton: You can get the device orientation from https://developer.android.com/reference/android/content/res/Configuration.html#orientation.

So maybe take that as a parameter to this function, and let the calling function cache it.

Be careful though -- at the moment activities are *not* recreated on orientation changes (e.g., see https://github.com/tuskyapp/Tusky/blob/develop/app/src/main/AndroidManifest.xml#L57) so if this is cached by an activity or fragment it needs to be updated when it changes.

The activities handling config changes is a source of another bug (https://github.com/tuskyapp/Tusky/issues/3217, which I've just extracted in to https://github.com/tuskyapp/Tusky/issues/3398).

Lakoja: Hm, I didn't find an easy (maintainable) solution for moving this outward and/or cache.

getSystem() has a measurable time consumption (13 micro seconds for me) but it is only called once new images are displayed (few times per second for very fast scrolling).

So I'd probably leave it be.

nikclayton: Instead of `Resources.getSystem().displayMetrics` you could use `Resources.getConfiguration().orientation` -- that more accurately captures the intent of what you're doing.

Either way, passing the value in as a parameter to this function means that you can write unit tests for this function, which feels important.

However...

Here's an alternative to all of this is which is probably better. Instead of this, what if the `minAspect` and `maxAspect` values were stored as named resources, but in resource directories with different qualifiers?

E.g., (https://developer.android.com/guide/topics/resources/providing-resources#QualifierRules):

- `values`
- `values-land` (device is in landscape orientation)
- `values-sw600dp` (shortest screen dimension is at least 600dp)
- `values-sw600dp-land` (shortest screen dimension is at least 600dp *and* device is in landscape orientation)
- etc

Then let the resource lookup system handle finding the best value? The values should still be passed in as parameters (to make testing easy), but this would move this kind of configuration detail out of code and in to resources, where it arguably belongs.

Lakoja: Changed to resource values.

That is very slick and could support different configs in future.

For the moment it is still simply "portrait" and "landscape".

nikclayton:
Do you have some before/after screenshots of this?

Lakoja:
So far I was only looking at phone resolution (in landscape). But there I think Tusky is generally unusable, regardless of used image crop aspect ratios (see first screenshots here).

It looks ok on a tablet because that does not have a full width display.

So I guess this could be closed.

grafik

digitalcircuit:
Small counterpoint to @Lakoja 's remark - I actually do use Tusky on my phone in landscape, or, well, used to before the recent aspect ratio update.

There's times it's more convenient to have my phone (Pixel 4 XL) in landscape orientation (e.g. propped up on a charging mount, so I'd need to pick it up and hold it instead), and I'd prefer to have smaller thumbnails for landscape even if it's not as usable as it would be on a tablet.

That said, feel free to do what you wish with this change.

Lakoja:
> There's times it's more convenient to have my phone (Pixel 4 XL) in landscape orientation (e.g. propped up on a charging mount, so I'd need to pick it up and hold it instead), and I'd prefer to have smaller thumbnails for landscape even if it's not as usable as it would be on a tablet.

But but but...

Do you have a screenshot for me?
When I do this with my phone (which isn't especially narrow) the usable vertical space is so tiny that I dismissed this usage right away. :-)

digitalcircuit:
@Lakoja
> Do you have a screenshot for me?

Ack, pardon, I missed the "wanting a screenshot" part.

Here's an example of being able to read a full text post in landscape comfortably with the current stable Tusky release:
Screenshot of Tusky in landscape orientation showing a complete post from @MastodonMigration@mastodon.online, including showing the reply/boost/favorite/bookmark buttons

I don't have the old version of Tusky archived in order to get a screenshot of how images looked.

(In my case, I'm using Android's Gesture Navigation with the navigation hint hidden via LineageOS settings.  There's a Magisk mod to do that on stock Android builds, too.  Even without that, it only would've partially cropped out the interaction buttons, none of the actual post.)

connyduck:
Why is this still draft? Its definitely an improvement, would merge.

Lakoja:
> Why is this still draft? Its definitely an improvement, would merge.

Not quite sure if it is an improvement for landscape tablet resolutions (which have quite ok-ish height).
I noticed for example that many images - at least at that time in my feed - have a 1:1 ratio.
Going with 1.2 as min aspect ratio would have cut all those noticeably (that's why I changed it to 1.0 as minimum).

But I un-drafted it now. Maybe somebody (more than one) would like to test it and if necessary we can fine-tune then.

digitalcircuit:
To confirm, what kind of testing in landscape mode would be helpful to get this merged?

I have a Pixel 4 XL and some older, rather slow devices (Nexus 6 on Android 8.1, Galaxy Nexus on Android 6.0).  To get an APK, I assume I'd just need to follow the usual Tusky development build steps with this PR as the base.

Lakoja:
> Tusky development build steps with this PR as the base.

Hi. Thank's for your offer.
You can compile it and install it as a development version from this branch, yes.
If you're somewhat familiar with this, it's ok.

However, I think this might be (discussed internally and) merged, so it can be tested more easily from the nightly version.
I'm not sure however when that might be.

Tak:
Do we need more testing on this?
Should we just merge it and see what nightly users say?

Lakoja:
> Do we need more testing on this? Should we just merge it and see what nightly users say?

Probably not (merge it).

I tested this again and see three classes of devices:
- The standard case: Normal phones (or also tablets) in portrait mode
- Normal phones in landscape mode: This is (mostly) discussed and tested here
- Tablets in landscape mode: Tusky seems to have a width limit for the content and thus the area of the timeline is also more high than wide

So the largest portion of devices including all tablet orientation do not have a problem.

Only the landscape phones would benefit here.
But the current improvement (this PR) targets _all_ landscape devices.

So, if we want to merge this we would first need to specifically only target landscape phones.

PR #3400, opened 2023-02-28
Use Sequences on joinToString

Goooler:
https://kotlinlang.org/docs/sequences.html


connyduck:
I'd rather stay with collections because less lines of code = easier to comprehend. Unless we have a performance problem somewhere?

app/src/main/java/com/keylesspalace/tusky/network/FilterModel.kt
```diff
@@ -59,10 +58,11 @@ class FilterModel @Inject constructor() {
         val now = Date()
         val nonExpiredFilters = filters.filter { it.expiresAt?.before(now) != true }
         if (nonExpiredFilters.isEmpty()) return null
-        val tokens = nonExpiredFilters
+        val tokens = nonExpiredFilters.asSequence()
             .map { filterToRegexToken(it) }
+            .joinToString("|")
 
-        return Pattern.compile(TextUtils.join("|", tokens), Pattern.CASE_INSENSITIVE)
+        return Pattern.compile(tokens, Pattern.CASE_INSENSITIVE)
```
Goooler, line 65: At least we can keep line like this?

Goooler:
> I'd rather stay with collections because less lines of code = easier to comprehend. Unless we have a performance problem somewhere?

Either way.

PR #3404, opened 2023-03-01
Enable parallel GC for Gradle builds

nikclayton:
Per https://developer.android.com/studio/build/optimize-your-build#experiment-with-the-jvm-parallel-garbage-collector

I benchmarked this, and p75 incremental build time dropped from 33s to 30s.

https://github.com/gradle/gradle/issues/19750 means that if `org.gradle.jvmargs` is set any unchanged default values are lost, so include those too.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,5 +1,6 @@
 org.gradle.caching=true
-org.gradle.jvmargs=-Xmx4g -Dfile.encoding=UTF-8
+# If jvmargs is changed then the default values must also be included, see https://github.com/gradle/gradle/issues/19750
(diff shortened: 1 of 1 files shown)


gradle.properties
```diff
@@ -1,5 +1,6 @@
 org.gradle.caching=true
-org.gradle.jvmargs=-Xmx4g -Dfile.encoding=UTF-8
+# If jvmargs is changed then the default values must also be included, see https://github.com/gradle/gradle/issues/19750
+org.gradle.jvmargs=XX:+UseParallelGC -Xmx4g -Dfile.encoding=UTF-8 -XX:MaxMetaspaceSize=2g -XX:+HeapDumpOnOutOfMemoryError -Xms256m
```
nikclayton, line 3:
```suggestion
org.gradle.jvmargs=-XX:+UseParallelGC -Xmx4g -Dfile.encoding=UTF-8 -XX:MaxMetaspaceSize=2g -XX:+HeapDumpOnOutOfMemoryError -Xms256m
```

connyduck (approved):

PR #3406, opened 2023-03-01
Make text selectable when viewing edited statuses

nikclayton:
Allows the user to copy the content from any version of a status.


Goooler (approved):

connyduck:
With #3314 it is no longer possible to copy a version of a status, as the diff is copied as well. Not sure if anyone would want that, but I don't think there is something that can be done about it?

nikclayton:
Closing -- as noted, with the edit diff display this isn't useful as is.

PR #3410, opened 2023-03-02
Extract database code to a separate module

nikclayton:
Jump to https://github.com/tuskyapp/Tusky/pull/3410#issuecomment-1456046457 to see the most recent state, and benchmark results. tl;dr, 14% reduction in build times.


connyduck:
but why

nikclayton:
This is a WIP to look at the impact of extracting code out in to separate modules, and possible pitfalls.

At the time of writing this PR creates a new module, `:core:database`, moves the `Attachment.kt` entity file in to that module, and adjusts the calling code as necessary.

The beneficial impact on build times is pretty big.

Before this PR (develop @ a792d2c0d6b94d25277d6698d5e9fed57fc14351) and using `gradle-profiler` I checked two scenarios. Both scenarios are incremental recompiles of the code after making a change to `Attachment.kt`.

The change is either an ABI change (i.e., a new public method is introduced to the file) or a non-ABI change (the body of an existing method is modified).

The benchmark performs 6 warmup runs, then 10 measured runs.

```
                   Incremental build time (75th percentile)
               
                     ABI change      Non-ABI change    
                  ---------------------------------
Without this PR     8.9 seconds         6.8 seconds
With this PR        7.0 seconds         2.1 seconds
```

The conversion is reasonably smooth, however, there are a couple of things I've noticed so far.

1. Nullable properties in types imported from modules generally can't be smart cast to non-null. There's a detailed discussion of why this restriction exists at https://discuss.kotlinlang.org/t/what-is-the-reason-behind-smart-cast-being-impossible-to-perform-when-referenced-class-is-in-another-module/2201.

   You can see an example of code rewritten to avoid this in https://github.com/tuskyapp/Tusky/pull/3410/files#diff-11b44ae8dd2d90d9e4ea45a8ca43079b3a0773c4cb42eda514d491e6c24b5884R113.

2. The existing chain of dependencies between some types is problematic. For example, we have code attached to a "lower" type, like an entity, that converts that entity to a "higher" type, like a *ViewData.

   This is a problem, because now the entity's type has a dependency on the *ViewData type, and if you're not careful you end up with circular dependencies between modules.

   The fix for this is relatively straightforward -- instead of modelling this as `Entity.to(ViewData)` model it as `ViewData.from(Entity)`. The conversion is easy to make, but can potentially change many call sites.

### How to convert?

I can't imagine that trying to convert a bunch of code in to a new module at once is going to be tenable. This PR moves one file and there are knock on effects on ~ 50 other files. More complex conversions (e.g., changing `to` methods to `from` methods) would be even bigger.

A large PR that tried to migrate a lot of code would also be difficult to keep up to date as other PRs land around it.

So if this is something to be done --- and it probably should be as the build speed improvements by (a) unlocking parallel builds across different modules, and (b) not having to rebuild *every* module on a change are significant developer experience benefits -- it probably needs to be done as a long running refactoring where a PR moves a file (or small number) of files at a time.

### Why start with the database code?

I picked a database entity because:

1. It's low down the dependency tree, so the minimum amount of code is affected by the move
2. Database code doesn't change too often, so getting it in to its own module and not needing to rebuild it frequently is an immediate developer experience benefit
3. Room uses annotation processors in lots of places and that's a large fraction of the build time, so the potential benefits are also large. Much larger than, say, starting with a "compose" module and moving status composition code in to there.

Hopefully this helps start an informed discussion.

nikclayton:
If this refactoring happens it might also be an opportune time to use `app.tusky` as the package prefix for most of the code (not the published package, that can't change).

nikclayton:
Here's more of a WIP, to demonstrate the type of work that needs to be done.

By the end of this chain of commits (which are designed to be looked at one after another, rather than as a single PR), most of the entity and related classes have been moved to a new `:core:database` module.

Along the way I've also created `:core:text` and `:core:network` to hold classes or functions that the database code also depends on.

> These names are provisional, and may change.

> Not everything in `:core:database` belongs there. I suspect a lower level `:core:types` or similar module is appropriate to hold, well, core types of the app that are used by everything else.

There are three types of commit so far:

1. The commit message is `model/<something>`. This is a simple move of the file from one module to another, and a fixup of the import paths everywhere that uses it. These are very simple to review, as there is no real code change.

2. As above, but the commit message also mentions `smart cast`. These are moves, but where the calling code needed to be modified because the compiler can no longer confirm a value is non-null.

   These are relatively easy to review, you just need to verify that the new calling code has the same functionality as the old calling code.

3. Everything else. Right now this is stuff around `ConversationEntity`. That's a little more involved because of the need to convert from the `to` idiom to the `from` idiom, to break the dependency cycle.

In case I wasn't clear enough about that earlier, consider two types, `A` and `B`.

Right now we have code that does the equivalent of:

```
// In A

import B

class A() {
    // ...
    fun toB() : B { /* return a new B() */ }
}
```

and

```
// In B

import A

class B() {
    // ...
    fun toA() : A { /* return a new A() */ }
}
```

This is a problem because now you can't put A and B in different modules without introducing mutual dependency between those modules, because they import each other's types.

To break that, pick one type as being "higher" than the other in the dependency tree. Right now I think that tree of types looks roughly like:

```
ViewData -> ViewModels -> Database -> Network -> "Core" types

<-- higher ---------------------------------------- lower -->
```

In the A/B example, suppose `A` is the "higher" type. You break the cycle by doing this:

```
// In A

import B

class A() {
    fun toB() : B { /* as before */ }

    companion object {
        fun from(b: B) = A(...)
    }
}
```

and remove the `fun toA()` from `B`. Now `A` imports types from the "lower" `B`, and `B` has no knowledge of any "higher" types.

### Current state

This doesn't demonstrate additional compile time speed increases yet because the stuff that needs annotation processing to run hasn't been moved. This is all ground work to get to that point.

I've paused at `TabData`.

This is problematic because it's got types from all over -- database entities, fragments, resources, and so on.

This works at the moment because everything's one big module, but it doesn't work if things move into smaller modules.

I'm currently thinking I'll split it in to `TabData` and `TabViewData` to be similar to the other types, and marshall between them, but I'm going to think about that over the weekend.

nikclayton:
Commits up to this point complete the move, to the point where it can be benchmarked.

### Is it faster?

Yes, by about 15 seconds (p75). Typical fresh builds are ~ 2m6s, with this PR they're ~ 1m50s, **That's a ~ 14% reduction in build time**.

https://scans.gradle.com/s/fp5ps355h4252/timeline is a typical build scan for `develop` when this branch was made. You can see that the start of the build is dominated by `:app:mergeExtDexBlueDebug` and `:app:mergeBlueDebugResources`, and nothing much else happens in this phase.

https://scans.gradle.com/s/nar23wy6oxzog/timeline is a typical build scan for this PR at e766e4b. While the two `merge...` tasks are running Gradle is also compiling `:core:network` and `:core:database` so it gets a head start on that code.

### Developer experience

Apart from being faster to build the developer experience in Android Studio (at least as I use it) is the same. Searching across the whole project works fine, as does jumping between a type and it's definition.

For example, I can be in a file in `:core:database`, hit Ctrl-B on a type and be shown a list of places the type is used anywhere in the project (not just `:core:database`). Similarly I can be in, say, `MainActivity` and hit Ctrl-B on `TabViewData` and be taken to the file, Ctrl-Shift-N to search for a symbol works across the project, etc.

So if you're working on existing code you don't really need to care about which module it's in.

### Reviewing

Probably not as difficult as I had originally feared. Although a lot of files have changed, for the majority of them it's just import paths.

The changes that have a bigger impact are:

- Migrating `Status.Visibility` to `StatusVisibility`
- Splitting `TabData` in to `TabData` and `TabViewData`
- The small handful of changes needed because smart casting to non-null doesn't work across modules

### Still to do

Before this is a merge candidate:

- Rename the library packages to `app.tusky.core....` ?
- Extract the common configuration (e.g., min/target SDK version, Java language target) out of the different build.gradle files in to a common location

### Other benefits

This has highlighted some problematic design choices and made fixing them pretty easy (`TabData` / `TabViewData`).

It's shown that the code that remains in the `db/` directory probably doesn't belong there, as it's nothing to do with the database per se.

### Probable next steps (after this PR)

- Move the remaining network-related code in to `:core:network` (as one PR)
- Move custom views and related resources to `:core:designsystem` (see https://www.nngroup.com/articles/design-systems-101/ for the name) (as one PR)
- Move individual activities and related classes in to their own modules (as individual PRs)

At each stop assess the build time impact. I imagine there's a sweet spot where the build graph is sufficiently wide that there's lots of benefits from parallel builds, but not so deep that that that benefit is lost.

### Probable complications

- If string resources get broken out in to the modules that use them then that would probably require a change to the Weblate configs. Maybe all string resources end up in their own module for simplicity?
- Some resources will definitely be common across multiple modules (e.g., dimensions used in layouts) and will require some thought.

charlag:
I generally like the idea of modularization but I tend to think of it on a feature-level rather than layer level. This change introduces some complexity in both writing code and building it so we should really ask the hard question of whether it is worth it.

I would say that important criteria is "warm" build and not a "cold" build. As far as I can see "cold" build improved by 14% which is pretty cool but maybe not enough on it's own.  if that's all we will get. Do you think we will get similar results with even more modules? It might be justified then.

Another point is (sorry if I missed that) that Kotlin does only parallelize by module by default but can be instructed to parallelize inside modules to as far as I know. Did you look into that? Will kapt prevent any improvement there?

nikclayton:
> I generally like the idea of modularization but I tend to think of it on a feature-level rather than layer level. This change introduces some complexity in both writing code and building it so we should really ask the hard question of whether it is worth it.

Yes -- that's the point of this PR; to provide hard data so we can decide if it's worth it.
 
> I would say that important criteria is "warm" build and not a "cold" build. As far as I can see "cold" build improved by 14% which is pretty cool but maybe not enough on it's own. if that's all we will get.

From the **very first update to this PR**:

```
                   Incremental build time (75th percentile)
               
                     ABI change      Non-ABI change    
                  ---------------------------------
Without this PR     8.9 seconds         6.8 seconds
With this PR        7.0 seconds         2.1 seconds
```

> Do you think we will get similar results with even more modules? It might be justified then.

My intuition is that we'll get a benefit as long as the machine doing the building has spare cores to do more builds.

Looking at one of my build scans, https://scans.gradle.com/s/nar23wy6oxzog/timeline, there's the `:app:kaptBlueDebugKotlin` followed by `:app:compileBlueDebugKotlin` that's on the critical path, currently runs in series, and takes about 55 seconds to run.

With appropriate modularisation I think we can get down to sub-1 minute clean builds, and sub 10 second builds for changes that touch a handful of modules (assuming hardware equivalent to mine, which is a ~ 4 years old Dell XPS laptop, so by no means a fast machine nowadays).

This also improves the developer experience on CI, as you'll get CI results faster.

> Another point is (sorry if I missed that) that Kotlin does only parallelize by module by default but can be instructed to parallelize inside modules to as far as I know. Did you look into that? Will kapt prevent any improvement there?

I'm not aware of that being a thing, for example, it's not mentioned at https://docs.gradle.org/current/userguide/performance.html. If you can provide more details about how to enable that I'll take a look.

nikclayton:
> I generally like the idea of modularization but I tend to think of it on a feature-level rather than layer level. 

I think it's mix. Under this model we would have some core modules; at least `types`, `database`, `network`, `designsystem`, `service` (?), maybe one or two more.

And then -- probably, maybe, I think this needs some experimentation -- the rest of the modules are organised around screens or complete features that the user uses.

So there'd be a hypotetical `:screen:trendingTags` module that contains all the activity / fragment / viewmodel / resources / etc necessary to display the "Trending Tags" screen, and that has dependencies on the relevant `:core` modules (e.g., that feature doesn't use the database, so it would have a `:core:network` dependency but not a `:core:database` dependency).

Thinking about these as `screens` or `features` mean that we don't fall in to the trap of trying to organise everything to mirror Android's class hierarchy.

> This change introduces some complexity in both writing code

Can you articulate what that is? I've already noted that, in a very small number of cases, nullable properties can't be smart cast to non-null. The additional complexity to deal with that is vanishingly close to zero.

> and building it

Can you articulate what that is? The user experience in Android Studio is **identical** as far as I can tell, and if you're building from the command line then `.\gradlew appAssembleGreenDebug` works exactly as it does before.

I would like to address genuine concerns but I can't do that if they're not concrete.

charlag:
> If you can provide more details about how to enable that I'll take a look.

I meant this:
https://blog.jetbrains.com/kotlin/2022/04/kotlin-1-6-20-released/#support-for-parallel-compilation-of-a-single-module-in-the-JVM-backend

but they mention that

> It doesn’t work with [kapt](https://kotlinlang.org/docs/kapt.html) because kapt disables the IR backend.

I don't know if they fixed the kapt situation since then

charlag:
I only could provide some very quick feedback before, I am sorry for not being articulate enough in a few cases.

From the example you mentioned (trending tags) I am now a bit cautious that modules won't get *too* small, we don't want one per each fragment, do we? Some bigger things like Compose screen are perhaps good candidates together with what goes into them. I think database, network and some more you outlined are actually good candidates.

re: code complexity I think we will run into more situations but my general gut feeling is that you need to be constantly aware where you are and what you can depend on - which is not necessarily a bad thing! explicit dependencies are great - but adds to the things you need to keep in mind. My experience here is introducing [similar rules](https://github.com/tutao/tutanota/blob/0283c38e8d35f544970668aa1988523caa2e2057/buildSrc/RollupConfig.js#L20) at my day job. I can say that it can improve code organization but also complicate things.

re: building complexity I don't mean running Gradle, I mean maintaining the build code for it (which as I checked now is just adding more `build.gradle` files). It is not critical but it is something to keep in mind.

It's hard for me to asses it but if we can really reach big gains (like, idk, 40%) it might be very well worth it.

nikclayton:
> I meant this:
https://blog.jetbrains.com/kotlin/2022/04/kotlin-1-6-20-released/#support-for-parallel-compilation-of-a-single-module-in-the-JVM-backend

Interesting, thanks for the link. As you note, doesn't work with `kapt`. Unfortuately, `kapt` is end-of-life, but we can't move to the replacement, `ksp`, because Dagger and Hilt aren't compatible with `ksp` yet. @Goooler mentioned migrating to `napt` (https://github.com/sergei-lapin/napt) which claims to be quite a bit faster, but I haven't had time to experiment with that yet.

> From the example you mentioned (trending tags) I am now a bit cautious that modules won't get _too_ small

I agree. There is a sweet spot where the module tree is of sufficient breadth and depth.

Too narrow and deep and we end up having to recompile too many things on every change. Too wide and shallow and too many unnecessary things get recompiled when a single module changes. Either way we lose the benefits of this sort of change.

> , we don't want one per each fragment, do we?

Almost certainly not. A good rule of thumb is probably "What's the largest unit of UI that the user interacts with".

For example, there's currently `AboutActivity` and `LicenseActivity`. I could imagine a `:screen:about` module that contains both of those.

Or search -- that's multiple fragments (and an activity and viewmodel) -- but that would be in a `:screen:search` module, **not** three different modules like `:screen:searchHashtag`, `:screen:searchAccounts`, `:screen:searchStatuses`.

> re: code complexity I think we will run into more situations but my general gut feeling is that you need to be constantly aware where you are and what you can depend on - which is not necessarily a bad thing! explicit dependencies are great - but adds to the things you need to keep in mind. My experience here is introducing [similar rules](https://github.com/tutao/tutanota/blob/0283c38e8d35f544970668aa1988523caa2e2057/buildSrc/RollupConfig.js#L20) at my day job. I can say that it can improve code organization but also complicate things.

I see what you mean.

On balance, I think this is a positive. Complexity is generally not eliminated, it's just moved around.

Consider the TabData / TabViewData changes I've had to make in this PR (https://github.com/tuskyapp/Tusky/pull/3410/commits/e5eb5e1b6703d40d3cc5d04664343b50ef7a4b90).

The previous code was "complex" along multiple different axes. For example:

- It's hard to test, because `TabData` depends on fragments, view models, and resources
- It relies on strings as identifiers instead of properly typed data
- It used non-class functions to create instances of `TabData`
 
It was simple along one axis -- it was all in one file.

The new code is less complex along those axes, but there is now the separation of concerns between `TabData` and `TabViewData` 

As I say, I think the benefits of the cleaner code, type safety, and testability outweigh the small difference in code organisation

> re: building complexity I don't mean running Gradle, I mean maintaining the build code for it (which as I checked now is just adding more `build.gradle` files). It is not critical but it is something to keep in mind.

I see what you mean. Yes, that's true -- there is a small amount of additional complexity (not implemented in this PR yet) to have a single source of truth for things like the min SDK or compile SDK versions to ensure that all modules are built with the same values.

connyduck:
Having worked on projects with a lot of modules where some are larger than Tusky, I don't think we need to split it up.
I'd rather have a bit more buildtime than code scattered around multiple modules. CI run times of like 10 mins are ok.

nikclayton:
> Having worked on projects with a lot of modules where some are larger than Tusky, I don't think we need to split it up. I'd rather have a bit more buildtime than code scattered around multiple modules. CI run times of like 10 mins are ok.

The CI speed up is a nice side effect, but not the initial reason for experimenting with this.

The initial prompt was @mcclure reporting slow build times, https://github.com/tuskyapp/Tusky/issues/3412, and this provides significant wins.

```
                   Incremental build time (75th percentile)
               
                     ABI change      Non-ABI change    
                  ---------------------------------
Without this PR     8.9 seconds         6.8 seconds
With this PR        7.0 seconds         2.1 seconds
```

As I said earlier, I think we can get from multi-minute builds when switching branches / release types to sub 1-minute builds, which is a significant developer experience improvement.

The day-to-day developer experience in Android Studio appears to be the same -- I didn't notice any problems navigating around the code base, using features like "Jump to declaration or usages", etc.

And it highlights (and lets us fix) problematic code structure (I don't mean the layout of the files, I mean the dependency relationships between different types) that make the code more confusing.

Issue #2486, opened 2022-04-30
[feature request] add stats in home timeline


chrisschm:
[add stats in home timeline].

I am missing (while browsing the timelines) to see if any comment, share or fav is made on each toot. Would be nice to have those stats without the need going inside a toot.

* * * *
- Tusky Version: 17.0
- Android Version: 12
- Android Device: Google Pixel 5
- Mastodon instance (if applicable): norden.social

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

davidbtc2009:
This looks to have been solved in tusky 19. Have you checked it out and is that what you were after?

chrisschm:
> 

I'm sorry. I've closed my mastodon account and can not check anymore if that is solved.

davidbtc2009:
> > 
> 
> I'm sorry. I've closed my mastodon account and can not check anymore if that is solved.

Closing your mastodon account is sad. Here is a screenshot
![Screenshot_20220723-102858_Tusky](https://user-images.githubusercontent.com/67274571/180609507-36db77a0-168f-4e3e-a06b-10429a8ed7b1.png)

chrisschm:
Than it's not fixed. At the first post, it is a boost and so i would think there should be at minimum a 1 close to the boost button.

But the 0 close to the answer buttons seems to be the right direction.

davidbtc2009:
> Than it's not fixed. At the first post, it is a boost and so i would think there should be at minimum a 1 close to the boost button.
> 
> But the 0 close to the answer buttons seems to be the right direction.

Ah, i see what you mean

domenukk:
For me, the information how many people interacted with a toot is interesting metadata, and I don't want to click on everything just to see that.
Would love to see this feature implemented (as option)

PR #3413, opened 2023-03-03
Show toot stat inline

zikasak:
Resolves https://github.com/tuskyapp/Tusky/issues/2486


connyduck:
- Needs some kind of clever number formatting to make sure the labels don't get too long, e.g. instead of 123456 display 123k. 
- Please add a setting to turn this on and off, default to off

zikasak:
I've added formatting and a setting

app/src/main/res/layout/item_status.xml
```diff
@@ -353,6 +353,17 @@
         sparkbutton:primaryColor="@color/tusky_blue"
         sparkbutton:secondaryColor="@color/tusky_blue_light" />
 
+    <TextView
```
Lakoja, line 356: How does this look? Can you add a screenshot?

app/src/main/res/values/strings.xml
```diff
@@ -643,6 +643,7 @@
 
     <string name="pref_title_show_notifications_filter">Show Notifications filter</string>
     <string name="pref_title_enable_swipe_for_tabs">Enable swipe gesture to switch between tabs</string>
+    <string name="pref_title_show_stat_inline">Show toot statistics</string>
```
Lakoja, line 646: You can use "post" instead of "toot".

Maybe word it "Show post statistics in timeline"?

app/src/main/java/com/keylesspalace/tusky/util/NumberUtils.kt
```diff
@@ -0,0 +1,17 @@
+package com.keylesspalace.tusky.util
+
+import java.text.DecimalFormat
+import kotlin.math.floor
+import kotlin.math.log10
+import kotlin.math.pow
+
+fun shortNumber(number: Number): String {
+    val array = arrayOf(' ', 'k', 'M', 'B', 'T', 'P', 'E')
```
Lakoja, line 9: k also large like the others letters?

zikasak:
Screenshot_20230310_144057_Tusky Test copy

Here is an example. The reply counter was mocked to show how it looks on big numbers

Lakoja:
Could probably do with a bit more margin to the images.

Otherwise it looks fine.

zikasak:
Screenshot_20230310_151116_Tusky Test
Is it ok with such a margin?

Lakoja:
That looks good.

Lakoja:
(Old PR that can be closed then: https://github.com/tuskyapp/Tusky/pull/1369)

connyduck (commented):
I like how it works now, code needs a bit of improvement then I'll merge it

connyduck (approved):

scollovati:
May I suggest to remove the 0 (zero) indicator? Adding the number only when different from zero increases readability.

PR #3421, opened 2023-03-07
3419 alt text in any case

### LINKED ISSUE #3419: Slow image upload: alt text will be deleted
When composing:
If you add an image you can edit the description (alt text) while it uploads. If you finish your alt text before upload is finished, then this text will not be part of the published status.

Everything else seems to be correct however (for example the small description icon in the lower right of the preview image is blue).

Not sure if this is recent - maybe upload being slow is not so common that this problem could be present for a longer time.
But there are a number of issues that changed image upload code in the near past: For example #2479 or #2626.

* * * *
- Tusky Version: Recent develop (ce6a350)

Technical help:
"Slow" upload speed can easily be simulated by adding a wait at the end of the transfer here:

ProgressRequestBody#writeTo
`try { Thread.sleep(13000); } catch (InterruptedException ex) {}`

Lakoja:
Fixes #3419

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/components/compose/ComposeActivity.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/components/compose/ComposeActivity.kt
@@ -227,13 +227,13 @@ class ComposeActivity :
         val mediaAdapter = MediaPreviewAdapter(
             this,
             onAddCaption = { item ->
-                CaptionDialog.newInstance(item.localId, item.description, item.uri)
-                    .show(supportFragmentManager, "caption_dialog")
+                CaptionDialog.newInstance(item.localId, item.description, item.uri).show(supportFragmentManager, "caption_dialog")
             },
             onAddFocus = { item ->
                 makeFocusDialog(item.focus, item.uri) { newFocus ->
                     viewModel.updateFocus(item.localId, newFocus)
                 }
+                // TODO this is inconsistent to CaptionDialog (device rotation)?
             },
             onEditImage = this::editImageInQueue,
             onRemove = this::removeMediaFromQueue
@@ -1266,11 +1266,7 @@ class ComposeActivity :
     }
 
     override fun onUpdateDescription(localId: Int, description: String) {
-        lifecycleScope.launch {
-            if (!viewModel.updateDescription(localId, description)) {
-                Toast.makeText(this@ComposeActivity, R.string.error_failed_set_caption, Toast.LENGTH_SHORT).show()
-            }
-        }
+        viewModel.updateDescription(localId, description)
     }
 
     /**
--- a/app/src/main/java/com/keylesspalace/tusky/components/compose/ComposeViewModel.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/components/compose/ComposeViewModel.kt
@@ -48,7 +48,6 @@ import kotlinx.coroutines.flow.asFlow
 import kotlinx.coroutines.flow.first
 import kotlinx.coroutines.flow.shareIn
 import kotlinx.coroutines.flow.update
-import kotlinx.coroutines.flow.updateAndGet
 import kotlinx.coroutines.launch
 import kotlinx.coroutines.withContext
 import javax.inject.Inject
@@ -130,7 +129,7 @@ class ComposeViewModel @Inject constructor(
     ): QueuedMedia {
         var stashMediaItem: QueuedMedia? = null
 
-        media.updateAndGet { mediaValue ->
+        media.update { mediaList ->
             val mediaItem = QueuedMedia(
                 localId = mediaUploader.getNewLocalMediaId(),
                 uri = uri,
@@ -144,11 +143,11 @@ class ComposeViewModel @Inject constructor(
 
             if (replaceItem != null) {
                 mediaUploader.cancelUploadScope(replaceItem.localId)
-                mediaValue.map {
+                mediaList.map {
                     if (it.localId == replaceItem.localId) mediaItem else it
                 }
             } else { // Append
-                mediaValue + mediaItem
+                mediaList + mediaItem
             }
         }
         val mediaItem = stashMediaItem!! // stashMediaItem is always non-null and uncaptured at this point, but Kotlin doesn't know that
@@ -169,13 +168,13 @@ class ComposeViewModel @Inject constructor(
                                 state = if (event.processed) { QueuedMedia.State.PROCESSED } else { QueuedMedia.State.UNPROCESSED }
                             )
                         is UploadEvent.ErrorEvent -> {
-                            media.update { mediaValue -> mediaValue.filter { it.localId != mediaItem.localId } }
+                            media.update { mediaList -> mediaList.filter { it.localId != mediaItem.localId } }
                             uploadError.emit(event.error)
                             return@collect
                         }
                     }
-                    media.update { mediaValue ->
-                        mediaValue.map { mediaItem ->
+                    media.update { mediaList ->
+                        mediaList.map { mediaItem ->
                             if (mediaItem.localId == newMediaItem.localId) {
                                 newMediaItem
                             } else {
@@ -189,7 +188,7 @@ class ComposeViewModel @Inject constructor(
     }
 
     private fun addUploadedMedia(id: String, type: QueuedMedia.Type, uri: Uri, description: String?, focus: Attachment.Focus?) {
-        media.update { mediaValue ->
+        media.update { mediaList ->
             val mediaItem = QueuedMedia(
                 localId = mediaUploader.getNewLocalMediaId(),
                 uri = uri,
@@ -201,13 +200,13 @@ class ComposeViewModel @Inject constructor(
                 focus = focus,
                 state = QueuedMedia.State.PUBLISHED
             )
-            mediaValue + mediaItem
+            mediaList + mediaItem
         }
     }
 
     fun removeMediaFromQueue(item: QueuedMedia) {
         mediaUploader.cancelUploadScope(item.localId)
-        media.update { mediaValue -> mediaValue.filter { it.localId != item.localId } }
+        media.update { mediaList -> mediaList.filter { it.localId != item.localId } }
     }
 
     fun toggleMarkSensitive() {
@@ -323,8 +322,8 @@ class ComposeViewModel @Inject constructor(
     }
 
     private fun updateMediaItem(localId: Int, mutator: (QueuedMedia) -> QueuedMedia) {
-        media.updateAndGet { mediaValue ->
-            mediaValue.map { mediaItem ->
+        media.update { mediaList ->
+            mediaList.map { mediaItem ->
                 if (mediaItem.localId == localId) {
                     mutator(mediaItem)
                 } else {
--- a/app/src/main/java/com/keylesspalace/tusky/components/compose/ComposeViewModel.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/components/compose/ComposeViewModel.kt
@@ -322,9 +322,8 @@ class ComposeViewModel @Inject constructor(
         serviceClient.sendToot(tootToSend)
     }
 
-    // Updates a QueuedMedia item arbitrarily, then sends description and focus to server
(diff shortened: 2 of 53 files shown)


app/src/main/java/com/keylesspalace/tusky/components/compose/ComposeViewModel.kt
```diff
@@ -323,9 +323,8 @@ class ComposeViewModel @Inject constructor(
         serviceClient.sendToot(tootToSend)
     }
 
-    // Updates a QueuedMedia item arbitrarily, then sends description and focus to server
-    private suspend fun updateMediaItem(localId: Int, mutator: (QueuedMedia) -> QueuedMedia): Boolean {
-        val newMediaList = media.updateAndGet { mediaValue ->
+    private fun updateMediaItem(localId: Int, mutator: (QueuedMedia) -> QueuedMedia) {
+        media.updateAndGet { mediaValue ->
```
Lakoja, line 327: I have the feeling that this could be simplified even much further? Get rid of updateMediaItem and "updateAndGet"?

(Simply set the respective field - with a simple "find" before?)

app/src/main/java/com/keylesspalace/tusky/components/compose/ComposeViewModel.kt
```diff
@@ -334,33 +333,16 @@ class ComposeViewModel @Inject constructor(
                 }
             }
         }
-
-        if (!editing) {
-            // Updates to media for already-published statuses need to go through the status edit api
```
Lakoja, line 339: I wonder if this comment was true even before?

With this new code (in StatusSendService) it also works when editing a status.

(maybe @mcclure ?)

app/src/main/java/com/keylesspalace/tusky/network/ProgressRequestBody.java
```diff
@@ -67,8 +67,10 @@ public void writeTo(@NonNull BufferedSink sink) throws IOException {
                 uploaded += read;
                 sink.write(buffer, 0, read);
             }
+
+            uploadListener.onProgressUpdate((int)(100 * uploaded / contentLength));
```
Lakoja, line 71: This fills the upload progress circle.

(But maybe the incomplete circle was intended as there seems to be some processing after it before the blue circle actually disappears?)

app/src/main/java/com/keylesspalace/tusky/service/SendStatusService.kt
```diff
@@ -181,6 +182,20 @@ class SendStatusService : Service(), Injectable {
...
 
+            media.forEach { mediaItem ->
+                if (mediaItem.processed) {
+                    val focus = mediaItem.focus
+                    val focusString = if (focus != null) "${focus.x},${focus.y}" else null
+                    Log.w(TAG, "updating media description on status send "+mediaItem.description )
+                    mastodonApi.updateMedia(mediaItem.id!!, mediaItem.description, focusString)
+                        .fold({
+                        }, { throwable ->
+                            Toast.makeText(applicationContext, R.string.error_failed_set_caption, Toast.LENGTH_SHORT).show()
```
Lakoja, line 193: The message/case now leaves out the "focus failed" message.

I guess the one here should be a more generic one (and the old two removed).

app/src/main/java/com/keylesspalace/tusky/components/compose/ComposeViewModel.kt
mcclure: I don't recognize this code, sorry. The only advice I have is to make sure your change works when editing (cropping) an image (because that at some level deletes the image and replaces it with a new one)

Tak: It was _absolutely_ true before - without this codepath, when editing the caption of an attachment to an already-published status (post status with media attachment, edit status, edit caption of attachment), the user would get a popup about setting the caption failing

Tak: https://github.com/tuskyapp/Tusky/pull/3215#issuecomment-1407476422

app/src/main/java/com/keylesspalace/tusky/service/SendStatusService.kt
Tak: I believe that we prefer snackbars for these

Tak: This needs to be guarded with `!editing` like the previous block in `updateMediaItem` was, for the same reason. Testing locally, I hit the failure block here when editing the caption and focus for an attachment to a published status.

app/src/main/java/com/keylesspalace/tusky/network/ProgressRequestBody.java
Tak: Maybe move this unrelated change to a separate PR?

Tak (changes requested):

app/src/main/java/com/keylesspalace/tusky/components/compose/ComposeViewModel.kt
Tak: The description/focus *do* get updated, but because of the subsequent call to `editStatus`, not the `updateMedia` call here, which fails.

Lakoja: Very odd that the call to update media is then necessary initially.

But you are right.

Added "isEditing".

app/src/main/java/com/keylesspalace/tusky/network/ProgressRequestBody.java
Lakoja: It's such a small collateral fix. :-)

I could remove it.

app/src/main/java/com/keylesspalace/tusky/service/SendStatusService.kt
Lakoja: Re snackbar: That absolutely needs a view.

However this is a service so more or less the opposite.

Other errors are reported as actual notifications here.

So I'm not sure it should even show a Toast (but it probably doesn't hurt).

Tak: Ah right, good point @ view

app/src/main/java/com/keylesspalace/tusky/service/SendStatusService.kt
```diff
@@ -181,6 +182,25 @@ class SendStatusService : Service(), Injectable {
                 return@launch
             }
 
+            val isEditing = statusToSend.statusId != null
```
Tak, line 185: You could refactor the `createStatus`/`editStatus` block below to reuse this variable 🙂

app/src/main/java/com/keylesspalace/tusky/components/compose/ComposeViewModel.kt
Lakoja: Probably not: This flow (its list) is observed by the activity.

However I made it a bit more clear: "AndGet" was never necessary here and "mediaValue" is actually a "mediaList".

app/src/main/java/com/keylesspalace/tusky/components/compose/ComposeActivity.kt
```diff
@@ -224,13 +224,13 @@ class ComposeActivity :
...
             onAddCaption = { item ->
-                CaptionDialog.newInstance(item.localId, item.description, item.uri)
-                    .show(supportFragmentManager, "caption_dialog")
+                CaptionDialog.newInstance(item.localId, item.description, item.uri).show(supportFragmentManager, "caption_dialog")
             },
             onAddFocus = { item ->
                 makeFocusDialog(item.focus, item.uri) { newFocus ->
                     viewModel.updateFocus(item.localId, newFocus)
                 }
+                // TODO this is inconsistent to CaptionDialog (device rotation)?
```
connyduck, line 233: ?

app/src/main/java/com/keylesspalace/tusky/service/SendStatusService.kt
```diff
@@ -181,6 +182,25 @@ class SendStatusService : Service(), Injectable {
...
+                media.forEach { mediaItem ->
+                    if (mediaItem.processed) {
+                        mastodonApi.updateMedia(mediaItem.id!!, mediaItem.description, mediaItem.focus?.toMastodonApiString())
+                            .fold({
+                            }, { throwable ->
+                                Toast.makeText(
+                                    applicationContext,
+                                    R.string.error_failed_set_caption,
+                                    Toast.LENGTH_SHORT
+                                ).show()
```
connyduck, line 197: Shouldn't it retry like it is done for all other failures? Or at least fail hard with a notification? Just showing a toast but continuing with posting the status is not what I would expect as a user

app/src/main/java/com/keylesspalace/tusky/components/compose/ComposeActivity.kt
Lakoja: There was a change/issue to make the text in the alt text dialog survive a device rotation.

This was not done to the focus dialog.

So one could change this as well (which is this todo).

app/src/main/java/com/keylesspalace/tusky/service/SendStatusService.kt
Lakoja: Hm, it was like this before. (which of course might not have been optimal, but probably outside of the scope of this issue?)

Also I thought it was ok because alt text and focus point are not the main concerns on a status with an image.

But I can of course give retry + fail a look.

connyduck: Yes please have a look, maybe it is easy to do. If not, this is still better than before and we can improve it later.

Lakoja: It uses now the same fail or retry logic as failures on the status send itself.

app/src/main/java/com/keylesspalace/tusky/service/SendStatusService.kt
```diff
@@ -290,6 +310,9 @@ class SendStatusService : Service(), Injectable {
             notificationManager.cancel(statusId)
             notificationManager.notify(errorNotificationId++, notification)
         }
+
+        // NOTE only this removes the "Sending..." notification (added with startForeground() above)
```
Lakoja, line 314: Another collateral fix. Otherwise the "Sending..." notification cannot be remove upon errors.

app/src/main/java/com/keylesspalace/tusky/service/SendStatusService.kt
```diff
@@ -244,18 +260,22 @@ class SendStatusService : Service(), Injectable {
...
         }
     }
 
+    private suspend fun failOrRetry(throwable: Throwable, statusId: Int) {
+        if (throwable is HttpException) {
+            // the server refused to accept, save status & show error message
+            failSending(statusId)
+        } else {
+            // a network problem occurred, let's retry sending the status
+            retrySending(statusId)
```
Lakoja, line 275: This would retry indefinitely? (retries are counted but I don't see a cancel?)

What kind of exceptions are network problems?

connyduck: Yes this retries until it is cancelled, the whole idea behind this construct is that status gets send once the device is connected again.

Network problem is everything that causes the call to fail before a server response (successfull or unsuccessfull) is received.

Lakoja:
Regarding the current linter error:

> Error: "error_failed_set_caption" is translated here but not found in default locale

That is correct as I removed it.

Should/must I remove it from all others as well?
(Or is that done "automatically" once any translation is updated by a translator?)

connyduck:
Once this is merged, Weblate should send a PR to remove it from everywhere, yes. But I can't merge because lint is complaining 🙄 
I could bypass the merge block, but I'd prefer it if you removed it everywhere.

Lakoja:
> Should be good to go if you resolve the conflicts

Done.

It doesn't seem to be a good idea to remove / change things in other files than the default strings.xml.
That was a lot of work now.  And lots of conflicts repeatedly...

PR #3426, opened 2023-03-09
3416: Call the list activity when list list empty

### LINKED ISSUE #3416: Add list tab: insufficient gui with no lists
When adding a list tab the list of lists is quite empty when you don't have lists.

(Account preferences - Tabs - Add "List")

I'd expect a possibility to create a new list here. That could also be present if there are already some lists.
Or at least a short explanation on what to do.

grafik

Lakoja:
Fixes #3416.

grafik

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/TabPreferenceActivity.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/TabPreferenceActivity.kt
@@ -293,7 +293,7 @@ class TabPreferenceActivity : BaseActivity(), Injectable, ItemInteractionListene
 
         val dialogBuilder = AlertDialog.Builder(this)
             .setTitle(R.string.select_list_title)
-            .setNeutralButton(R.string.select_list_create) { _, _ ->
+            .setNeutralButton(R.string.select_list_manage) { _, _ ->
                 val listIntent = Intent(applicationContext, ListsActivity::class.java)
                 startActivity(listIntent)
             }
@@ -318,10 +318,7 @@ class TabPreferenceActivity : BaseActivity(), Injectable, ItemInteractionListene
                 { lists ->
                     showProgressBarJob.cancel()
                     adapter.addAll(lists)
-                    if (lists.isNotEmpty()) {
-                        dialog.getButton(DialogInterface.BUTTON_NEUTRAL).hide()
-                        dialog.setTitle(R.string.select_list_title)
-                    } else {
+                    if (lists.isEmpty()) {
                         noListsText.show()
                     }
                 },
--- a/app/src/main/java/com/keylesspalace/tusky/TabPreferenceActivity.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/TabPreferenceActivity.kt
@@ -336,9 +336,11 @@ class TabPreferenceActivity : BaseActivity(), Injectable, ItemInteractionListene
 
     private fun getDimensionFromAttribute(context: Context, attr: Int): Int {
         val typedValue = TypedValue()
-        return if (context.theme.resolveAttribute(attr, typedValue, true))
+        return if (context.theme.resolveAttribute(attr, typedValue, true)) {
             TypedValue.complexToDimensionPixelSize(typedValue.data, context.resources.displayMetrics)
-        else 0
+        } else {
+            0
+        }
     }
 
     // TODO this should be made general somewhere? (it is copied from ViewThreadFragment)
--- a/app/src/main/java/com/keylesspalace/tusky/TabPreferenceActivity.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/TabPreferenceActivity.kt
@@ -24,6 +24,9 @@ import android.util.Log
 import android.util.TypedValue
 import android.view.View
 import android.widget.FrameLayout
+import android.widget.LinearLayout
+import android.widget.ProgressBar
+import android.widget.TextView
 import androidx.activity.OnBackPressedCallback
 import androidx.appcompat.app.AlertDialog
 import androidx.appcompat.widget.AppCompatEditText
@@ -37,7 +40,6 @@ import androidx.recyclerview.widget.RecyclerView
 import androidx.transition.TransitionManager
 import at.connyduck.calladapter.networkresult.fold
 import at.connyduck.sparkbutton.helpers.Utils
-import com.google.android.material.progressindicator.LinearProgressIndicator
 import com.google.android.material.snackbar.Snackbar
 import com.google.android.material.transition.MaterialArcMotion
 import com.google.android.material.transition.MaterialContainerTransform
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/keylesspalace/tusky/TabPreferenceActivity.kt
```diff
@@ -287,7 +278,33 @@ class TabPreferenceActivity : BaseActivity(), Injectable, ItemInteractionListene
...
+        }
+
+        val dialog = dialogBuilder.show()
+
+        lifecycleScope.launch {
+            mastodonApi.getLists().fold(
+                { lists ->
+                    adapter.addAll(lists)
+                    if (lists.isNotEmpty()) {
+                        dialog.getButton(DialogInterface.BUTTON_NEUTRAL).hide()
```
Lakoja, line 297: Not really sure this is a good workflow: For slow loads the gui will change noticably.
For very slow loads it will be wrong.

However also a "still loading information" progress animation is missing here.

Tak: Why not put *all* the dialog code into a function that gets called from the success block of `fold`?

Lakoja: What would happen inbetween? (just to clarify) So you click "add", nothing happens (loading api items in the background) and then the dialog appears?

Lakoja: (added now a loading progress indicator)

app/src/main/java/com/keylesspalace/tusky/TabPreferenceActivity.kt
```diff
@@ -261,28 +272,78 @@ class TabPreferenceActivity : BaseActivity(), Injectable, ItemInteractionListene
 
     private fun showSelectListDialog() {
         val adapter = ListSelectionAdapter(this)
+
+        val progress = LinearProgressIndicator(this)
```
Lakoja, line 276: However this thing does not animate (cycle colors?).

app/src/main/java/com/keylesspalace/tusky/TabPreferenceActivity.kt
```diff
@@ -261,28 +272,78 @@ class TabPreferenceActivity : BaseActivity(), Injectable, ItemInteractionListene
...
-            .show()
+    private fun getDimensionFromAttribute(context: Context, attr: Int): Int {
+        val typedValue = TypedValue()
+        return if (context.theme.resolveAttribute(attr, typedValue, true))
+            TypedValue.complexToDimensionPixelSize(typedValue.data, context.resources.displayMetrics)
+        else 0
+    }
+
+    // TODO this should be made general somewhere? (it is copied from ViewThreadFragment)
+    //   also: maybe it should have a callable instead of simply view.show()?
```
Lakoja, line 335: Tried this now (two functions as parameter to the `getProgressBarJob`).

However, this is very weird. Here especially as there is a clash between dialogBuilder and dialog. dialog is not known when getProgressBarJob is called.

It is probably also weird when dialog titles change with a delay.

app/src/main/java/com/keylesspalace/tusky/TabPreferenceActivity.kt
```diff
@@ -261,28 +275,83 @@ class TabPreferenceActivity : BaseActivity(), Injectable, ItemInteractionListene
...
+        noListsText.setPadding(preferredPadding, 0, preferredPadding, 0)
+        noListsText.text = getText(R.string.select_list_empty)
+        noListsText.visible(false)
+
+        statusLayout.addView(progress)
+        statusLayout.addView(noListsText)
+
+        val dialogBuilder = AlertDialog.Builder(this)
+            .setTitle(R.string.select_list_title)
+            .setNeutralButton(R.string.select_list_create) { _, _ ->
```
Lakoja, line 296: This should probably be visible all the time? (currently it is hidden below)

Maybe called "Manage lists"?

connyduck:
Can you also add this to the dialog on profiles please?

app/src/main/java/com/keylesspalace/tusky/TabPreferenceActivity.kt
connyduck: yes, good idea

app/src/main/java/com/keylesspalace/tusky/TabPreferenceActivity.kt
```diff
@@ -261,28 +275,85 @@ class TabPreferenceActivity : BaseActivity(), Injectable, ItemInteractionListene
...
-            .setAdapter(adapter) { _, position ->
-                val list = adapter.getItem(position)
-                val newTab = createTabDataFromId(LIST, listOf(list!!.id, list.title))
-                currentTabs.add(newTab)
-                currentTabsAdapter.notifyItemInserted(currentTabs.size - 1)
-                updateAvailableTabs()
-                saveTabs()
-            }
-            .show()
+    private fun getDimensionFromAttribute(context: Context, attr: Int): Int {
```
connyduck, line 337: should probably be in `com.keylesspalace.tusky.util.ThemeUtils` where we have similar utils? Also it could be an extension function on `Context`

connyduck: Wait, isn't `getDimension` in `ThemeUtils` doing the same already?

app/src/main/java/com/keylesspalace/tusky/TabPreferenceActivity.kt
```diff
@@ -261,28 +275,85 @@ class TabPreferenceActivity : BaseActivity(), Injectable, ItemInteractionListene
...
+        val typedValue = TypedValue()
+        return if (context.theme.resolveAttribute(attr, typedValue, true)) {
+            TypedValue.complexToDimensionPixelSize(typedValue.data, context.resources.displayMetrics)
+        } else {
+            0
+        }
+    }
+
+    // TODO this should be made general somewhere? (it is copied from ViewThreadFragment)
```
connyduck, line 346: yes probably but it is also fine if it stays here for now

app/src/main/java/com/keylesspalace/tusky/TabPreferenceActivity.kt
Lakoja: Done.

app/src/main/java/com/keylesspalace/tusky/TabPreferenceActivity.kt
Lakoja: Ok, used that.

(Didn't find that as it uses completely different objects/methods.)

app/src/main/java/com/keylesspalace/tusky/TabPreferenceActivity.kt
Lakoja: (comment removed)

connyduck:
Meh now there is a merge conflict in the german translation (this is why translations should only be updated through weblate). Can you fix it please?

connyduck (approved):

Issue #3325, opened 2023-02-16
Can't delete header profile image


nikclayton:
It doesn't appear to be possible to delete your profile header image from Tusky, you have to use your instance's web interface.

* * * *
- Tusky Version: 21

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

Lakoja:
Same thing for the avatar image.

Further problem here: Pressing the "add image" button on either opens a file selector (ok) and simply going back from there will show "Upload has failed" message.

nikclayton:
More fun. The "public" API doesn't actually support deleting these, you have to use the "internal" API.

See https://github.com/mastodon/mastodon/issues/18921

nikclayton:
And the "internal" API doesn't appear to support the same auth token that the public API does (or I'm doing something else wrong).

But this:

```
DELETE https://mstdn.ca/settings/profile/pictures/avatar
User-Agent: Tusky/21.0 Android/11 OkHttp/4.10.0
Authorization: Bearer [my auth token]
```

returns a 422.

PR #3427, opened 2023-03-09
Don't display error message if user cancels picking an image

nikclayton:
Partial fix for #3325


app/src/main/java/com/keylesspalace/tusky/EditProfileActivity.kt
```diff
@@ -34,7 +34,9 @@ import com.bumptech.glide.Glide
 import com.bumptech.glide.load.engine.DiskCacheStrategy
 import com.bumptech.glide.load.resource.bitmap.FitCenter
 import com.bumptech.glide.load.resource.bitmap.RoundedCorners
+import com.canhub.cropper.CropImage
 import com.canhub.cropper.CropImageContract
+import com.canhub.cropper.CropImageView
```
nikclayton, line 39:
```suggestion
```

PR #3431, opened 2023-03-11
Remove ReplacementSpan, display diffs using CharacterStyle

nikclayton:
Remove the use of `ReplacementSpan`. It turns out this span type is incompatible with spans that occupy more than one line, and the result is that a longer diff can run off the end of the screen. The alternative means that the diff'd text doesn't have additional padding and rounded corners, but it's better than not being visible.

Display the most recent version of the status with larger text. Again, consistent with the thread view.

Display the avatar, name, and username of the poster in a pinned header at the top of the screen, instead of duplicating the information on every edit. This reduces the amount of redundant information on the screen.


app/src/main/java/com/keylesspalace/tusky/components/viewthread/edits/ViewEditsViewModel.kt
```diff
@@ -85,23 +84,20 @@ class ViewEditsViewModel @Inject constructor(private val api: MastodonApi) : Vie
...
 
-                                if (i < sortedEdits.size - 1) {
+                                if (i > 1) {
                                     currentContent = previousContent
                                     previousContent = loader.load(
-                                        sortedEdits[i + 1].content.replace(
-                                            "<br>",
-                                            "<br/>"
-                                        )
+                                        sortedEdits[i - 2].content.replace("<br>","<br/>")
```
nikclayton, line 100:
```suggestion
                                        sortedEdits[i - 2].content.replace("<br>", "<br/>")
```

Tak:
History is not the same as a conversation thread.
IMO the order should be newest to oldest, just like a version control log.

connyduck:
> History is not the same as a conversation thread.
IMO the order should be newest to oldest, just like a version control log.

yes please change

nikclayton:
> > History is not the same as a conversation thread.
> > IMO the order should be newest to oldest, just like a version control log.
> 
> yes please change

Done.

connyduck (approved):

Issue #3433, opened 2023-03-11
Migrate TimelineFragment to https://github.com/tuskyapp/Tusky/blob/develop/doc/ViewModelInterface.md


nikclayton:
Child of https://github.com/tuskyapp/Tusky/issues/3432

This is probably going to be a large PR, as `TimelineFragment` interacts with `CachedTimelineViewModel` and `NetworkTimelineViewModel`, so everything under that is going to change as well.

Issue #3320, opened 2023-02-15
Load more missing


slowcar:
I like to start my morning with a bit of scrolling in Mastodon, usually until i reach the point where i left off before.
Tusky seems not to load posts that are more than about 4h old and does not offer the "load more" element, even though it is available in between.
I am not sure if it is due to the instance not sending a "has more" flag correctly or the like, but the browser based feed is shown correctly.
I follow a few "animal every hour" bots, so even if there is nothing else, there should be the pictures of those bots.

Sample screenshot from Tusky: https://imgur.com/a/rRBln5r (5h gap to the next post)
Same post in firefox, with a different one below: https://imgur.com/a/wXxxqlJ

I am comfortable with installing a debug build or participating in analyzing logs etc, would look myself but i haven't made the jump from Android to Kotlin yet, so not really sure where to set breakpoints etc.

* * * *
- Tusky Version: 21.0
- Android Version: 13
- Android Device: Pixel 6 pro
- Mastodon instance (if applicable): social.tchncs.de

Lakoja:
How many status would you say you saw before there is "nothing more"?

(Or how often did you press Load More?)

There is - I think - a limit of posts (400, 1000?) that are provided by a server.

slowcar:
This morning i clicked "load more" 4 (or 5?) times. Didn't use the app over the day yesterday and posts cut off at ~9h, posts below that were 19h old.
I counted the total posts from there and got 147

keefmarshall:
Came here to raise the same bug.. I love Tusky but this bug means I have to go back to the web or use another client (e.g. the official one, which doesn't have this issue..) in order to see posts from more than a few hours ago.

dHeinemann:
Just wanted to add that I often encounter this bug too. First on Tusky 21 (or maybe even Tusky 20, I don't remember). I'm now on 22.0 beta 1.

Seeing as there aren't very many people reporting this bug, maybe it's relatively uncommon, or only associated with certain devices or Mastodon instances?

I see it roughly once per day. It most commonly occurs when checking my home timeline for the first time in 6+ hours. Tusky will load the posts, but as I scroll through, there will occasionally be a gap of 1+ hours worth of missing posts and no Load More button to retrieve them.

I don't recall ever seeing it on other timelines - just my home timeline. But that's also the one that I use the most.

* Tusky Version: 22.0  beta 1
* Android Version: 13
* Android Device: Samsung Galaxy A52s 5G
[...]

Issue #3317, opened 2023-02-13
Local feed has to be refreshed (again) after viewing a toot


6b6279:
When I press on a toot in the "Local" tab to view it full-screen, and then return to the "Local" tab by pressing the Back button, all the loaded toots disappear and I get an empty feed with "There is nothing here." and the confused Tusky mascot.
* * * *
- Tusky Version: 21.0
- Android Version: Android 11
- Android Device: Samsung M30s
- Mastodon instance (if applicable): tilde.zone (running glitch-soc)

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

Lakoja:
Does that always happen?

It doesn't for me at the moment.

However I _sometimes_ see a problem like this _somewhere_ (any timeline) where content should be loaded/present but instead "There is nothing" is shown.

6b6279:
I can confirm that the problem is not limited to the local feed. For me, it happens every single time on the local feed and on profiles.

Lakoja:
But it does work on your personal (home) timeline or the federated one?

6b6279:
Same issue in the federated timeline. Doesn’t occur in my personal timeline, though.

6b6279:
This somehow got fixed going forward, closing.

PR #3436, opened 2023-03-11
Migrate TimelineFragment and dependencies to new architecture

### LINKED ISSUE #3195: Don't pass StatusDisplayOptions when creating view holders
When creating viewholders the current code passes along StatusDisplayOptions to the view holder.

See, for example, https://github.com/tuskyapp/Tusky/blob/develop/app/src/main/java/com/keylesspalace/tusky/adapter/StatusViewHolder.java#L59

I think this causes a number of problems / UI issues that could be solved with a different approach.

The main problem is that if the status display options change (e.g. the user has toggled the "show absolute time" option the only way to currently handle that is to throw away the list of items and refetch it, recreating all the view holders.

This is not too bad if the data is cached locally, but if it's not (any of the network timelines) it's a second fetch of the data, the user may lose their place in the list, and so on.

However, this is made worse in `PreferencesActivity.onSharedPreferenceChanged()`, which will restart `MainActivity` if any of these preferences change (I guess that's to ensure that the list is completely recreated with whatever the new options are). When that happens the user's data and active tab is lost. For example, be on the Notifications tab, open the preferences, change the "show absolute time" setting, press the Back button, and you are dumped back at the Home tab when you exit Preferences.

This doesn't happen for all preference changes, so it's behaviour that the user can't anticipate, and looks a bit like Tusky has crashed and has reset to the default screen.

My initial thoughts for a better UX / implementation would be to:

1. Determine StatusDisplayOptions when an adapter is created, and pass it to the adapter
2. Expose it as a property / provide a method to update it over time
3. Pass the statusdisplayoptions as an additional parameter to each viewholder when the adapter's `onBindViewHolder` method is called.

Then, when any of the status display options actually change none of the underlying data in the adapter has to change, it's purely a data-presentation change. This is faster, and can be triggered with `adapter.notifyItemRangeChanged(...)`. View models can listen for changes to those preferences and expose that to their fragments/activities as appropriate, which can then update the adapter and use the `notify*` methods to update the display of the visible items.

* * * *
- Tusky Version: 20

nikclayton:
Follow the same architecture as Notifications:

- Loading statuses for cached and non-cached timelines uses the Paging3 library.
- Fix the issue where chunks could be missing from a timeline
- Failures during loading are shown in the status list, and the user can retry
- The "Load more" option between pages of statuses is gone, pages of statuses are automatically loaded as the user scrolls
  - The user must still trigger an explicit refresh once they get to the "top" of a timeline
- The architecture is strictly Fragment -> View Model -> Repository, with the Repository managing the PagingSource and RemoteMediator
- The Fragment communicates with the View Models by sending actions
  - The View Models communicate with the Fragment with flows for status data, overall UI state, and success/failure of the different actions
  - The View Models communicates with the network/disk via a Repository
    - There are separate repositories for Cached and Network timelines, and Filters
  - The Fragment and View Models *do not* access the API or database directly
- If an action fails the error is surfaced to the user, with an option to retry
- Delete the "Reading order" preference, it's unnecessary
- Add unit tests for the functionality, updated existing tests as necessary

Other code changes to make this easier:

- Move TimelineKind to a separate type outside the view model, to prevent circular dependencies (e.g., view model depending on a repository and the repository depending on the view model). Implement as a sealed class, and carry kind-specific information in the class instead of passing as additional intent extras.
- Update the "Developer tools" menu to include options for experimenting with missing statuses

Fixes #3195, #3632, https://github.com/tuskyapp/Tusky/issues/3433, https://github.com/tuskyapp/Tusky/issues/3320, https://github.com/tuskyapp/Tusky/issues/3317, https://github.com/tuskyapp/Tusky/issues/3269, https://github.com/tuskyapp/Tusky/issues/2831, https://github.com/tuskyapp/Tusky/issues/2702, https://github.com/tuskyapp/Tusky/issues/3505, https://github.com/tuskyapp/Tusky/issues/3461, #3818


nikclayton:
Before reviewing this one it includes code that's also in the smaller PRs:

- https://github.com/tuskyapp/Tusky/pull/3649
- https://github.com/tuskyapp/Tusky/pull/3651
- https://github.com/tuskyapp/Tusky/pull/3652
- https://github.com/tuskyapp/Tusky/pull/3653

I recommend reviewing those first. Once those are reviewed and merged in to `develop` I can merge those back in to this to shrink the size of this PR a bit.

nikclayton:
This is a large PR, but many of the files have mechanical, type name changes, which can be quickly reviewed and filtered so you can focus on the main body of the PR.

Of the ~ 145 files in the PR, ~ 95 are mechanical changes that can be quickly skimmed.

Also, many of the new lines are new unit tests, which are very similar to the existing ones for Notifications, so should be straightforward to review.

## Recommended order

## Skim `StatusViewData.Concrete` -> `StatusViewData` change

The change that affects probably the greatest lines of code (because placeholders are gone) is the deletion of `StatusViewData.Placeholder` and moving `StatusViewData.Concrete` to just
`StatusViewData`. There are many files where the change is just `StatusViewData.Concrete` -> `StatusViewData`.

So I recommend doing a first pass through the PR just looking at those, and ticking the *Viewed* checkbox to collapse them and get them out of the way.

## Skim `res/values*/strings.xml`

Because there are no placeholders the `load_more_placeholder_text` string has been removed from all the translations.

## New developer tools

- `MainActivity.kt`
- `usecase/DeveloperToolsUseCase.kt`

## TimelineKind changes

Before, a timeline's Kind was a `TimelineViewModel.Kind` type. That introduced a circular dependency (the view model would depend on a paging source, the paging source would depend on the view model for the type).

The dependency was broken by extracting the type to an upper class.

It was also converted from an enum to a sealed class, to allow for subtypes that could hold additional parameters. For example, all the `TimelineKind.User` subclasses have a parameter for the user ID relevant to the timeline.

This also meant that some of the parameters passed to fragments in intents could be removed, as the relevant data is now carried in the TimelineKind type. See, for example, the changes to `AccountPagerAdapter`.

Files that contain just `TimelineKind` changes should be straightforward to review.

## The rest

The rest is cutting over to the new architecture and unit tests. Enjoy...

app/src/main/java/com/keylesspalace/tusky/components/timeline/TimelineFragment.kt
```diff
@@ -293,23 +170,277 @@ class TimelineFragment :
...
+                    if (isSwipeToRefreshEnabled) {
+                        adapter.loadStateFlow
+                            .collect {
+                                if (old == null) {
+                                    Log.d("loadState", "No previous loadState")
+                                    old = it
+                                    return@collect
+                                }
+
+                                if (old?.refresh != it.refresh) {
```
nikclayton, line 339: This section here can be cleaned up, or moved to a function to clean up. Trying to figure out the specific state transitions to get the UI updated is tricky.

app/src/main/java/com/keylesspalace/tusky/adapter/StatusBaseViewHolder.java
```diff
@@ -750,18 +750,19 @@ private void showConfirmFavourite(StatusActionListener listener,
...
-    public void setupWithStatus(@NonNull StatusViewData.Concrete status,
+    public void setupWithStatus(@NonNull StatusViewData status,
                                 @NonNull final StatusActionListener listener,
                                 @NonNull StatusDisplayOptions statusDisplayOptions,
                                 @Nullable Object payloads) {
         if (payloads == null) {
             Status actionable = status.getActionable();
-            setDisplayName(actionable.getAccount().getName(), actionable.getAccount().getEmojis(), statusDisplayOptions);
+            //setDisplayName(actionable.getAccount().getName(), actionable.getAccount().getEmojis(), statusDisplayOptions);
+            setDisplayName(status.getId(), actionable.getAccount().getEmojis(), statusDisplayOptions);
```
nikclayton, line 765:
```suggestion
            setDisplayName(status.getId(), actionable.getAccount().getEmojis(), statusDisplayOptions);
```

Tak:
The newest-first flow is terrible - if I'm hours behind on the timeline, I can't go to the newest posts, I can only load page by page.
(I tried pull-to-refresh and "refresh" from the dots menu, but they both only load the page above where I am.)

Tak:
Voting on polls from the (home) timeline appears to be broken

nikclayton:
> Voting on polls from the (home) timeline appears to be broken

Broken how? I just voted on a test poll I created (voting carried out from this branch @ [d0ed7f3](https://github.com/tuskyapp/Tusky/pull/3436/commits/d0ed7f3ab697fda32644154432c84fb56d08c49d) which is the newest commit at the time I'm writing this) and it worked OK.

Tak:
> > Voting on polls from the (home) timeline appears to be broken
> 
> Broken how? I just voted on a test poll I created (voting carried out from this branch @ [d0ed7f3](https://github.com/tuskyapp/Tusky/pull/3436/commits/d0ed7f3ab697fda32644154432c84fb56d08c49d) which is the newest commit at the time I'm writing this) and it worked OK.

I was on the same commit. Tapping the `Vote` button has no effect in the timeline, I have to go to the detailed view.

nikclayton:
> > > Voting on polls from the (home) timeline appears to be broken
> > 
> > Broken how? I just voted on a test poll I created (voting carried out from this branch @ [d0ed7f3](https://github.com/tuskyapp/Tusky/pull/3436/commits/d0ed7f3ab697fda32644154432c84fb56d08c49d) which is the newest commit at the time I'm writing this) and it worked OK.
> 
> I was on the same commit. Tapping the `Vote` button has no effect in the timeline, I have to go to the detailed view.

I still can't reproduce. I'm on a slightly newer version now, but the voting functionality hasn't been changed.

Here's a video showing it working.

https://youtu.be/ufkHmNfMxlQ

What do the logs show when you try and vote?

Tak:
I don't experience it any longer in `92ac56f` 🤷

Tak:
Argh, I just got it again on the same changeset
The post is https://polyglot.city/@Stoori/110752314874555921 , and it was boosted into my home timeline.
Tapping the vote button from the timeline (after selecting two answers) has no effect, browsing to the detailed view allows me to vote

nikclayton:
> Argh, I just got it again on the same changeset The post is https://polyglot.city/@Stoori/110752314874555921 , and it was boosted into my home timeline. Tapping the vote button from the timeline (after selecting two answers) has no effect, browsing to the detailed view allows me to vote

OK, so it's not the underlying voting functionality, it's UI related.

Here's the existing code:

```java
// In StatusBaseViewHolder.setupPoll()

            // ...
            pollButton.setOnClickListener(v -> {
                int position = getBindingAdapterPosition();
                if (position != RecyclerView.NO_POSITION) {
                    List<Integer> pollResult = pollAdapter.getSelected();
                    if (!pollResult.isEmpty()) {
                        listener.onVoteInPoll(position, pollResult);
                    }
                }
            });
```

`onVoteInPoll()` is defined in `TimelineFragment.kt`, and looks like this:

```kotlin
    override fun onVoteInPoll(position: Int, choices: List<Int>) {
        val statusViewData = adapter.peek(position) ?: return
        val poll = statusViewData.status.poll ?: return
        viewModel.accept(StatusAction.VoteInPoll(poll, choices, statusViewData))
    }
```

My best guess is that the `if (position != RecyclerView.NO_POSITION)` check was returning `false`.

This could happen if, off screen (probably "above" you in the list) more data is being loaded and inserted. This would cause `getBindingAdapterPosition` to return `NO_POSITION` for some of that time, because the `notify*` methods are being called.

However, there are three other possible things that could cause that code to not perform the vote. So I've just pushed https://github.com/tuskyapp/Tusky/pull/3436/commits/75f9f7e6fb6977c173991e60ae72a7d7c51dc94f to show snackbars if any of them occurs.

@Tak please can you try a build with that, and let me know what, if any, errors are shown. Thanks.

Tak:
It looks like following and muting in the hashtag timelines is broken (the buttons / menu items aren't shown)

nikclayton:
> It looks like following and muting in the hashtag timelines is broken (the buttons / menu items aren't shown)

That's weird -- thanks, I'll look at that now.

This is exactly the sort of problem that https://github.com/tuskyapp/Tusky/issues/3869 would catch.

nikclayton:
> It looks like following and muting in the hashtag timelines is broken (the buttons / menu items aren't shown)

Fixed with https://github.com/tuskyapp/Tusky/pull/3436/commits/d6f9c760926b916cd8e19ae7302a7372b98ec5c1, thanks.

mcclure:
Read over the code. I was only partially familiar with this code before the refactor but from a code perspective it looks good to me. So this is a positive code review.

Can you clarify some things?

* Used to there was a distinction between StatusViewData and StatusViewData.Concrete. This patch collapses Concrete into StatusViewData. For what reason is this distinction no longer needed?

* <i>Delete the "Reading order" preference, it's unnecessary</i> Have we confirmed that? I personally have no interest in this feature, but do we have any information about how many users use this preference?

mcclure:
Installed this version of Tusky and will try to test with it tonight and tomorrow. So far it seems fine, it appears that when new statuses load in at the top it looks a little more chaotic than before but I don't know if that is a fluke until I see it happen more times. I *do* run my phone with animations disabled at the OS level.

(BTW from a feature perspective I am personally unsold on the idea of removing "Show More" but not enough to oppose the patch.)

nikclayton:
> Can you clarify some things?
> 
>     * Used to there was a distinction between StatusViewData and StatusViewData.Concrete. This patch collapses Concrete into StatusViewData. For what reason is this distinction no longer needed?

`StatusViewData` was abstract, with `.Concrete` and `.Placeholder` subclasses. The `.Placeholder` subclass was used if the view data was being used to show a "Load more" placeholder.

Now that "Load more" is gone there is no need for the `.Placeholder` subclass, so there is no need to distinguish between an abstract `StatusViewData` and `.Concrete` subclass, they can be the same thing.

>     * Delete the "Reading order" preference, it's unnecessary Have we confirmed that? I personally have no interest in this feature, but do we have any information about how many users use this preference?

The only thing the "Reading order" preference affected was how your reading position changed when you pressed "Load more". With no "Load more" there is no need for this preference.

mcclure:
@nikclayton This answers my questions, thanks

app/src/main/java/com/keylesspalace/tusky/components/timeline/viewmodel/TimelineViewModel.kt
```diff
@@ -34,142 +35,471 @@ import com.keylesspalace.tusky.appstore.MuteEvent
...
-        val votedPoll = poll.votedCopy(choices)
-        updatePoll(votedPoll, status)
+        viewModelScope.launch {
+            eventHub.events
+                .filterIsInstance<PreferenceChangedEvent>()
+                .filter { FILTER_PREF_KEYS.contains(it.preferenceKey) }
+                .distinctUntilChanged()
+                .map { getFilters() }
+                .onStart { getFilters() }
+        }
```
nikclayton, line 474:
```suggestion
```

Remove duplicate code.

mcclure:
I'm testing with [63c3da1](https://github.com/tuskyapp/Tusky/pull/3436/commits/63c3da1c633b418ba42b035d7906b25cb597dc75) , would it be more useful to test with [f893dc2](https://github.com/tuskyapp/Tusky/pull/3436/commits/f893dc2236a1405c991f7cf89bf65e124fe0d1f1) ?

nikclayton:
> I'm testing with [63c3da1](https://github.com/tuskyapp/Tusky/pull/3436/commits/63c3da1c633b418ba42b035d7906b25cb597dc75) , would it be more useful to test with [f893dc2](https://github.com/tuskyapp/Tusky/pull/3436/commits/f893dc2236a1405c991f7cf89bf65e124fe0d1f1) ?

If you get the chance, please.

The only difference should be the behaviour when initially opening a timeline.

With 63c... you might see the "Nothing here" message while data is being loaded the first time.

With f89... you should see the spinning progress bar in the middle of the screen until either (a) the load completes and there is data, (b) the load completes and there is no data ("Nothing here" is shown), or (c) there's an error, and the errorphant is shown.

mcclure:
So, I did install [f893dc2](https://github.com/tuskyapp/Tusky/pull/3436/commits/f893dc2236a1405c991f7cf89bf65e124fe0d1f1) and tried to use it for regular use tonight and had what I would consider a very bad user experience with it.

Here is what my use looked like:
I opened Tusky and flipped between tabs a few times. Everything seems fine. I put down my phone for a couple hours. I pick it up again.
I turn on my phone and I see old posts (as expected). So I tap the top of the tab to go to the top. It shows slightly newer old tweets. I pull down to load more. It… loads more, but I am still not at the top. It just loads a few posts from right after I posted. I tap the top of the tab again. It bounces to the top, and I pull down again, and… it loads more. I am still not at the top. When I looked at my phone and saw old tweets, what I wanted to do was get to the newest tweets. But there was no way to do that. I think I had to tap the top five or six times. When I went to my notifications tab, I don't think I ever even reached the top. I imagine if I had waited two or three days instead of an hour or so. I probably also would not have been able to reach the top.

I did some testing. I realized a few things:
- In this PR, tapping the tab top brings you to the top, whereas double-tapping it brings you to the top and loads more posts (but only one "page" more). I didn't need to pull down, only double tap. But I don't know either why I would double tap, or how I or any other user would *figure out* we are meant to double tap, and at any point double tapping does not do the thing I want (go to the top of the column). It loads a *few* more posts and leaves me still needing to scroll up.
- My "expected behavior" is that tapping the top of the column would take me all the way to the top of the column, not that tapping it once will bring me a page or so up and that tapping it twice is equivalent to manually scrolling a page or so up and pulling down.
- At about this point I realized I didn't actually know what the original Tusky did. I tried testing with Tusky Vanilla and what I realized is that if you tap the top of the column, it takes you to the top of all loaded posts, and if you pull down after that it emits a "load more" followed by the most recent page of posts. So I had this sense that tapping the top of the column would take me to the top, but that was never true, going to the actual top required tapping three times. The old behavior was not my expected behavior, but it was close enough by just sort of fiddling I could get it to do what I wanted without being consciously aware of what I was doing. (By comparison, the new behavior required quite a bit of thought and even after thinking about what it was the interface wanted me to do it potentially *couldn't* do what I wanted.)
- A nice thing about "show more" as an interface affordance is that in the old behavior, what I think I was *actually* doing whenever I checked for new posts was: tap the top and pull down; there are posts; I start reading them from the bottom up. So what I was doing each time was reading "about a page of posts", all of them recent, and reading them from oldest to newest (which is potentially actually nicer than reading in reverse). This was close enough to an ideal experience I never had to really think about what it was doing; if I wanted more, I could scroll down until I saw "show more" and tap that.
- I think my "expected behavior" of this patch would be
    - If I open the app and it is showing old posts, and I scroll up, it will gradually load new posts infinitely, one "page" at a time, and assuming I scroll relatively slowly *without me needing to do a pull to load more gesture*, until it hits the present. This is what all corpo social media apps seem to do\*. 
    - If I tap the tab up top on the other hand what I expect it would do, and what the old Twitter app seems to do? is *move me all the way to the most recent post* (and it does this instantly, so possibly it preloaded it in the background so it would be ready to display them anytime if I tapped to scroll up). If I scroll down from there, it infinitely loads posts beginning from the most recent (and does so without me, unless I scroll very quickly).
- If the "expected behavior" I describe there was the case in Tusky, I think I might still occasionally miss the "show more" behavior because with my "expected behavior" if I am looking at a post and then I zip up to the very top I would probably not be able to find my old place again (especially because mastodon's search capabilities are so poor). However, I would not *mind* this behavior. By contrast, the current PR behavior where it's not possible to scroll all the way to the present except by repeatedly tapping the tab over and over and waiting for pages and pages of posts you're not reading each time seems actually unusable if one's timeline is moving at any speed at all.

To summarize: I find "old Tusky behavior" where you can load more from the top, and load more from the middle, and there are explicit verbs for both of these, more or less usable. I also find the Twitter behavior, where there is an illusion of infinite scroll up and down from the current position as well as a verb for "jump to present", more or less usable. However the current PR seems like an in-between state between the two where the "load more" verb is manual, and the "jump to present" verb doesn't exist, and this is very frustrating to use. 

Am I missing something?

\* If I thought about it I'm not sure I could tell you what mastodon web does. I have a vague sense of it just doing "something reasonable".

Tak:
I agree. I've been using this branch during my vacation, and while the "Load Newest Posts" menu item *works*, the "reading in reverse chronological order" use case (i.e. the default behavior in current Tusky) is still worse than in current Tusky.

I would suggest one of:
- Adopt @mcclure's suggestion
- Reintroduce the reading order preference, and for "newest first", all "manual" loads (refresh menu item, pull-to-refresh, tap home) load the *newest* posts instead of the next newer page


(I've just updated to the head of the branch - I'll give more feedback about the more recent changes after I've gotten to use them for a bit)

nikclayton:
@mcclure 

Thanks for the detailed feedback.

There are some misunderstandings in there I'd like to clear up, so we're working from the same page when it comes to figuring out what specific changes to make.

tl;dr:

- Every time you want to tap on a tab, try the "Load newest posts" menu item. Does that fix the problem?

Much (much) longer info below, including alternative options.

## Mastodon pages

The Mastodon API generally provides information in pages, and each page contains links that can be followed to get the next and previous pages.

> Throughout this, "next" means "the page of older statuses, and would appear at the bottom of the list", "previous" or "prev" means "the page of newer statuses, and would appear at the top of the list"
> 
> Yes, "previous" meaning "newer" is exactly backwards to how you'd expect, but that's the API we've got.

These links are included **even if the page it's linking to doesn't exist yet**. So you make a request for your home timeline, and get a page containing the absolute newest posts, there is still a link to the "previous" page that will contain even newer posts at some point in the future.

Fetching those posts returns an empty list (because there are no newer posts), but it still exists.

So the only way to know you've reached the top of the list is to try and fetch the previous page, and see if it's empty.

## How many pages to fetch?

This poses a question: if the app requests a page of data, and there is a "prev" link (there's always a "prev" link, per the last section), should the app follow the link to fetch the previous page?

It's tempting to think the answer is "yes". Just always fetch the previous page if it exists, and iteratively fetch all the previous pages until you hit a page with no contents. Then stop.

If you do that you risk never finishing.

Consider a rapidly changing timeline, like the Federated timeline, where new posts are appearing all the time.

Suppose 4 new posts appear on this timeline every second (i.e., on average one every 250ms) and the combination of your internet connection speed and phone processing speed means it takes 300ms to process a page of statuses.

Therefore, by the time the app has finished processing a page and checking to see if the previous page has contents, the previous page **does** have contents. So those are processed, 300ms elapses, and you check to see if the new previous page does have contents.

It does too, so you process that, ... ad infinitum. You never run out of pages to process, because they're being produced faster than the app can process them.

The Federated timeline is the easiest one to see this on because it changes the fastest, but it can happen on any timeline.

> The problem is the same for fetching older pages if you're reading down.

To fix this, fetching previous or next pages is capped to about a maximum of 3 in each direction (I say "about 3" because the precise number isn't important). This ensures the user can read, and start scrolling up (or down) as soon as data is available. As they get close to the end of the pages that have already been loaded new pages are fetched in the background, giving the "infinite scroll" illusion.

## How do refresh, tab tapping, and reload interact?

Consider the following state when the user is viewing a timeline.

- The user is viewing the statuses in page 2
- Pages 3, 4, and 5 have been loaded (these are the "previous" pages, containing newer posts)
- Pages 12 through 6 also exist on the server
	- The app does not know that yet
	- The page numbers 1..12 here are to make the diagrams easier to follow. In reality page numbers are much longer, and may have gaps in them -- there is no requirement for a page with ID X to be followed by a page with ID X+1, it can be X+anything (they may not even be numbers)
	- Generating a new page number from an existing page number is not allowed; we can only use the page numbers the server tells us about.

In diagram form:

```mermaid
flowchart TD

  12["12
  (newest)"]:::notloaded 

  12 -.- 5 --- 4 --- 3 --- 2:::current --- 1["1
  (oldest)"]

  classDef notloaded stroke:#000,stroke-dasharray: 5 5
  classDef current stroke:#f00,stroke-width:2px,font-weight:bold
```

- Loaded pages have a solid outline
- Not-loaded pages have a dashed outline
- The current page has a thick red border, and the page number is in bold

Vertical layout makes it clear when the user is scrolling up they are going from page 2 to page 3, to page 4, etc. But it takes a lot of space. I'll use left/right layout from now on, where moving left == scrolling up, moving right == scrolling down.

The same diagram left/right:

```mermaid
flowchart LR

  12["12
  (newest)"]:::notloaded

  12 -.- 5 --- 4 --- 3 --- 2:::current --- 1["1
  (oldest)"]

  classDef notloaded stroke:#000,stroke-dasharray: 5 5
  classDef current stroke:#f00,stroke-width:2px,font-weight:bold
```

### The user scrolls up

As the user scrolls up they reach a point where there are not enough pages ahead of them, and the next newest page is automatically loaded in the background. The precise number of pages considered to be "not enough" is not important. Let's assume it's two.

The user scrolls up from page 2 to page 3. Now there are only two pages loaded before their current position (pages 4 and 5). This is "not enough", so the page previous to page 5 is loaded (page 6).

```mermaid
flowchart LR

  12["12
  (newest)"]:::notloaded

  12 -.- 6 --- 5 --- 4 --- 3:::current --- 2 --- 1["1
  (oldest)"]

  classDef notloaded stroke:#000,stroke-dasharray: 5 5
  classDef current stroke:#f00,stroke-width:2px,font-weight:bold
```

This happens continuously in the background as the user scrolls up (or down).

### The user taps the tab once

Assume the user is at the previous position -- somewhere on page 3, with 3 pages loaded ahead (4, 5, and 6).

Tapping the tab calls `TimelineFragment.onReselect()`.

This jumps them to the top of page 6. There are "not enough" pages loaded ahead of the user, so they are immediately loaded in (pages 7, 8, and 9).

```mermaid
flowchart LR

  12["12
  (newest)"]:::notloaded -.- 9

  9 --- 8 --- 7 --- 6:::current --- 5 -.- 1["1
  (oldest)"]

  classDef notloaded stroke:#000,stroke-dasharray: 5 5
  classDef current stroke:#f00,stroke-width:2px,font-weight:bold
```

### The user taps the tab twice

This is **not** treated differently by the code.

The first tap jumps the user to the top of page 6. Pages 7, 8, and 9 are loaded.

The second tap then jumps the user to the top of page 9. Pages 10, 11, and 12 are loaded.

If there is a third tap the user is jumped to the top of page 12. If there are any pages after page 12 those are loaded as well.

### The user uses the "Load newest posts" menu

The "Load newest posts" option discards the user's reading position, discards the loaded pages, and immediately tries to load the newest pages.

Suppose the user was on page 3, and page 12 is the newest available page. They choose the "Load newest posts" option.

**Was**

```mermaid
flowchart LR

  12["12
  (newest)"]:::notloaded

  12 -.- 6 --- 5 --- 4 --- 3:::current --- 2 --- 1["1
  (oldest)"]

  classDef notloaded stroke:#000,stroke-dasharray: 5 5
  classDef current stroke:#f00,stroke-width:2px,font-weight:bold
```

**Now**

```mermaid
flowchart LR

  12:::current --- 11 --- 10 --- 9 -.- 1:::notloaded

  classDef notloaded stroke:#000,stroke-dasharray: 5 5
  classDef current stroke:#f00,stroke-width:2px,font-weight:bold
```

There are no pages newer than 12, and enough next (i.e., older) pages have been loaded so the user can start scrolling through older content if they want without triggering another load.

#### The user refreshes

A refresh operation is triggered either by swiping down, or from the "Refresh" menu item.

"Refresh" means "Check again to see if there are any new posts, but don't lose my reading position".

Suppose the user completed the previous action, is now at the top of page 12, and there are no new posts, so the system looks like this.

```mermaid
flowchart LR

  12:::current --- 11 --- 10 -.- 1:::notloaded

  classDef notloaded stroke:#000,stroke-dasharray: 5 5
  classDef current stroke:#f00,stroke-width:2px,font-weight:bold
```

Some time later the user refreshes. The newest page is now 20. The system loads the 3 pages previous to page 12 (because two is "not enough"), so pages 13, 14, and 15 are also fetched.

The system looks like this:

```mermaid
flowchart LR

  20:::notloaded -.- 15
  
  15 --- 14 --- 13 --- 12:::current --- 11 --- 10 -.- 1:::notloaded

  classDef notloaded stroke:#000,stroke-dasharray: 5 5
  classDef current stroke:#f00,stroke-width:2px,font-weight:bold
```

Again, the app does not know pages 16 through 20 exist. It only knows up to page 15.

As the user scrolls up pages 16, 17, etc will be transparently loaded in the background, as described earlier.

## Your feedback

Looking at your feedback with all that in mind I've tried to pull out the relevant points needing a UX decision.

> In this PR, tapping the tab top brings you to the top, whereas double-tapping it brings you to the top and loads more posts

As noted, double-tap is not treated specially, but as two single taps. This does lead to the behaviour you observe though.

I can think of a number of possible options here.

1. Leave the current behaviour as is. It provides a way for the user to quickly "page up" through a timeline without having to repeatedly swipe / fling the list. If they want to go all the way to the top there is the "Load newest posts" menu action for that.
2. Explicitly check for a double-tap, and perform the "Load newest posts" action.
3. Change the single-tap behaviour to perform the "Load newest posts" action

I don't have a strong preference either way. I never use the "tap the tab title" functionality day-to-day, so I am not an informed user.

With my UX hat on I have a weak preference for keeping the existing behaviour. My rationale is that the "Load newest posts" functionality (as you noted elsewhere) is destructive, in as much as it loses the user's reading position.

Tapping or double-tapping a tab feels like an operation that is too easy to perform accidentally, and so keeping it as a menu operation (where it requires two taps, but on different parts of the screen to minimise mistakes) seems safest.

I did a quick check of two other Android apps:

- Megoladon
	- Not helpful; there's no tab to tap on, but there is an explicit "Load newest posts" equivalent button that pops up over the timeline. It does a reload, not a refresh.
	- Swipe-to-refresh performs the "Load newest posts" function
- Trunks
	- A single tap on a tab (technically, the title at the top of the screen) jumps you to the top of the loaded content (i.e., the same as this PR)
	- Swipe-to-refresh performs the "Load newest posts" function
	- There is no "refresh" equivalent (i.e., load new posts but keep my reading position)

So neither of those have to make the distinction between refresh or reload; they're "reload only" and you lose your reading position.

I don't have the Twitter app or other social media apps to compare.

> If I open the app and it is showing old posts, and I scroll up, it will gradually load new posts infinitely, one "page" at a time, and assuming I scroll relatively slowly _without me needing to do a pull to load more gesture_, until it hits the present. This is what all corpo social media apps seem to do

If I've understood you correctly this is the behaviour with this PR.

Once you reach the top you do have to explicitly refresh (swipe, or menu) to check again.

When Tusky starts it opportunistically checks for you. So if you've been away from the app for long enough that Android has killed it, it will check again and you can start scrolling without needing to refresh.

> If I tap the tab up top on the other hand what I expect it would do, and what the old Twitter app seems to do? is _move me all the way to the most recent post_ (and it does this instantly, so possibly it preloaded it in the background so it would be ready to display them anytime if I tapped to scroll up). If I scroll down from there, it infinitely loads posts beginning from the most recent (and does so without me, unless I scroll very quickly).

That would be option 3 above ("Change the single-tap behaviour to perform the Load newest posts action").

If you scroll down then older posts are loaded in the background.

Tak:
When loading the first page of posts on a user profile, there's a long\* pause where there's nothing there and no indication that anything's happening

Tak:
Hashtag timelines are now empty until manually refreshed

Tak:
When I try to vote on a boosted poll, I now get:

> statusViewData had null poll

gingerbearding:
Chiming in as a user on the above UX discussion about double taps vs a menu option. I think I'd vastly prefer the menu option. I'm a user who prefers reading in chronological order, not skipping any posts, and the number of times I've accidentally done something in various clients to take me to newest posts is frustratingly high.  Recovery from that would likely be worse if we're auto-loading posts on scroll, as getting back to where I was would likely have to load all pages in between. 

That said, not everyone wants that, so maybe a preference? Double tap to load newest, or long press or something? 

I'll say I do like the sound of the "paging" functionality by tapping repeatedly. That would be especially nice on the federated timeline, where I don't care so much about seeing every post. 

Cheers guys, looking forward to these changes!

nikclayton:
> When loading the first page of posts on a user profile, there's a long* pause where there's nothing there and no indication that anything's happening

Which version? If it's not with https://github.com/tuskyapp/Tusky/pull/3436/commits/f893dc2236a1405c991f7cf89bf65e124fe0d1f1 please try with that.

If it is with that, is it loading, but the progress spinner is off the bottom of the screen? I've seen that in current and earlier versions of Tusky when looking at user profiles that have a lot of bio. information. It pushes the fragment that contains the account's timeline sufficiently far down that the spinner (or errorphant) is not immediately visible, and you have to scroll.

mcclure:
Okay. A few thoughts.

* On using the menu item with Tusky Test, I found I *was* able to jump up to the top. However, I see three problems here.
    * I absolutely would have never found that on my own. If I were a new user and not a community member, I probably would have switched to another app before realizing what I was supposed to do. Until you asked if I'd tried the menu option, I didn't even realize there *was* a menu. (This is made worse because the menu is not visible 50% of the time, if you scroll up even a few pixels the menu disappears.)
    * In my view, this is too awkward of a UI for what is otherwise a "common" action. Instead of tapping the top of the tab I have to scroll down slightly to reveal the menu, tap the upper right, then tap a precise menu item. I jump to top *very quickly*. Whereas I could have performed the old "scroll to top" without thinking, I think this would require thinking.
    * Consider the two verbs we want to offer here: *load more posts from the top of the current scroll* and *load more posts from the present*. I now understand both verbs are available in the PR. However, consider what the PR has really done. Before, *users who prefer to read current posts first* could perform the natural action of just pulling down, whereas *users who prefer not to miss posts* had to specially tap a button. Now, *users who prefer not to miss posts* can just naturally pull down, whereas *users who prefer to read current posts first* have to specially tap a button (one hidden in a menu, but that's not the problem I want to highlight here). This means we have changed the UX, but have not really resolved the UX problem. We have made one thing more natural at the cost of making another thing less natural.
* I **do not** think we should intentionally map double-tap to a significantly different verb than single tab. That is not very discoverable (feels more like an easter egg than a UI feature), and personally with touchscreens I frequently double-tap when I meant to single-tap.

Overall, it is possible this patch is trying to do "too much". The intent, as Nik described in Matrix, was to refactor *code* to use the more modern paging system. However along the way we introduced a fairly significant UI change (removing load more and changing pull-down semantics). Perhaps the solution would be to reintroduce Load More to this PR with the old behavior, merge the PR so we lock in the code improvements, *then* (assuming this is what we want to do) remove Load More in a followup PR, so that when redesigning the UI to accommodate no-Load-More we can take more time to come up with a UI natural for all users…

I am writing this without fully reading Nik's long post, so I will do that now.

mcclure:
Continuing:

> So the only way to know you've reached the top of the list is to try and fetch the previous page, and see if it's empty.

Hm, I'm not sure I understand this. If I pull to read the top of the list, and I get posts which are potentially a few hundred milliseconds out of date because while the request was being serviced more things loaded in, that is fine. The behavior I was complaining about before is "I am looking at posts from 3 hours ago, I pull down, I get posts from 2 hours and 50 minutes ago". (As already discussed, I could have fixed this by using the menu item.)

> This poses a question: if the app requests a page of data, and there is a "prev" link (there's always a "prev" link, per the last section), should the app follow the link to fetch the previous page?

*Assuming we are in an infinite-scroll / no-load-more regime*, I would say the answer is no, not *yet*. Rather: Imagine a single page contains 10 posts. What we should do, in this regime, is load the single page of 10 posts and no more. Then once the user has scrolled up by 5 posts, *then* we should load one more single page. (The numbers 10 and 5 here are examples; we could pick other numbers, but the goal should be to *load more content slightly faster than a typical user would be able to read the content*, so that as they scroll up they are always presented with just a little more.)

This would produce behavior consistent with competing software I have used. It would also result in my goal of "two verbs, both natural": The "load more from current scroll point" verb would be "scroll up"; the "load more from present time" verb would be "tap tab header"

(One question left unanswered in the proposal I make here is "when should new content be loaded if the user is at the 'newest posts' point already?". I don't have a specific answer for this but "when the user executes the pulldown" would be an okay answer.)

> To fix this, fetching previous or next pages is capped to about a maximum of 3 in each direction (I say "about 3" because the precise number isn't important). This ensures the user can read, and start scrolling up (or down) as soon as data is available. As they get close to the end of the pages that have already been loaded new pages are fetched in the background, giving the "infinite scroll" illusion.

I don't think I'm experiencing this with the current PR code. I can test more, but currently, I don't think I'm experiencing this. It seems like I only load more when I do the pulldown gesture.

> The user taps the tab once… Tapping the tab calls TimelineFragment.onReselect(). This jumps them to the top of page 6. There are "not enough" pages loaded ahead of the user, so they are immediately loaded in (pages 7, 8, and 9).

As mentioned I will test more but I don't think I'm experiencing this. However possibly this *is* what is happening and Mastodon.social simply responds to requests so slowly that it *appears* to me nothing is happening until I perform more actions such as a double tap.

> The user taps the tab twice… This is not treated differently by the code.

Well, good. I don't know why I perceived I was seeing different behavior on both "old" and "PR" Tusky. Again perhaps I fooled myself due to Mastodon.social loading content slowly.

**AFTER WRITING THE ABOVE, PERFORMING TESTS WITH OLD AND NEW TUSKY:**

I am finding it very hard to tell whether I am seeing intended behavior or not (testing is slightly awkward because in order to test home timeline I must wait for new posts to be made), and I sincerely believe I am seeing different behavior in timeline ("Home") vs in "Notifications".

Here is what I *sincerely believe* I am seeing: 

IN THE HOME TAB: I cannot under any circumstances get tapping the tab icon, in either Tusky 23 or Tusky PR, to do anything other than scroll to the top. The mere act of being scrolled to the top does not load more posts. To load more posts, I must pull down (or use the menu, I guess). I can wait as long as I like at the top and nothing is loaded, I *must* pull down.

IN THE NOTIFICATIONS TAB: In Tusky PR, tapping once scrolls me to the top, and if I patiently wait about a second and then pull down, it loads more posts. If I *double* tap, it scrolls me a little above the top and shows an unmoving "reloading" icon. Then after a second more posts are loaded.

Things that may be confusing me:
* It is possible the behavior is different when you are "at the present" (which my Home timeline was for these tests) than when you are scrolled back in time (which my Notifications timeline was for these tests). 
* I am very, very confused by this additional column containing the unmoving reload icon. It sometimes appears and sometimes doesn't and I really don't know what the rule is. Possibly what is happening is that I tap a tab column once, it creates a row with the reloading icon above the top but does not scroll past it, and if I tap twice the first tap creates the reloading row and the second tap scrolls up to the reloading row.  However sometimes double tapping does not result in the reload row appearing, so I'm confused (is this inconsistency about home vs notifications tab, or "present vs past"? I do not know).

nikclayton:
> When I try to vote on a boosted poll, I now get:
> 
> > statusViewData had null poll

Perfect, that's what I needed to see the error. Thanks.

Fixed in https://github.com/tuskyapp/Tusky/pull/3436/commits/c25e9c33b83ebf3c5c099de6f93506d741400d7f (regular timelines) and https://github.com/tuskyapp/Tusky/pull/3436/commits/e5c5c73d19c411708996d9786bb34fe9c0c74900 (notification timeline).

nikclayton:
@mcclure Thanks for digging in to this. I hope these explanations are helpful.

I'm going to reply to some of the points out of order to get the stuff we definitely agree on out of the way, and hopefully consolidate on a few things.

Before I do that, one point to make that I think is important.

The Notifications tab has been using this code, or a version of it, since v22.0, with no user complaints that I'm aware of (to either the Tusky account or the \#tusky tag). I introduced these changes to the Notifications tab to cover for exactly this -- get widespread usage of the new UX and see if anyone complained.

> I **do not** think we should intentionally map double-tap to a significantly different verb than single tab.

Agreed.

> Perhaps the solution would be to reintroduce Load More to this PR [...]

The Paging3 libraries don't have a "load more" equivalent, so that won't work[1].

> On using the menu item with Tusky Test, I found I _was_ able to jump up to the top. However, I see three problems here.
>
> - I absolutely would have never found that on my own.

The same could be said of tapping on a tab; it's undiscoverable UI.

We're unlikely to get a single behaviour that pleases all the people all the time (e.g., see @gingerbearding's feedback a couple of posts up).

I am open to the idea of a N-state "What happens when I tap a tab?" preference, where the options are:

1. Do nothing
2. Jump to the newest loaded posts (*this is the current behaviour*)
3. Jump to the newest loaded posts, and refresh to see if there's anything new (this would be new behaviour)
4. Load the absolute newest posts, and jump to them (i.e., same as the "Load newest" menu option)

(precise UI text to be worked on).

However, given the complete lack of complaints from users about this behaviour on the Notifications tab, I think there's an argument for holding off on that until, at minimum, this PR has been merged (post v24.0) and Tusky Nighly users have a chance to provide feedback.

> Before, _users who prefer to read current posts first_ could perform the natural action of just pulling down, whereas _users who prefer not to miss posts_ had to specially tap a button. Now, _users who prefer not to miss posts_ can just naturally pull down, whereas _users who prefer to read current posts first_ have to specially tap a button (one hidden in a menu, but that's not the problem I want to highlight here). This means we have changed the UX

I don't think your description of the current Tusky behaviour here is correct.

The easiest way to see this is on the Federated timeline, because it changes so quickly.

1. Open the Federated timeline
2. Wait a few seconds for new posts to appear on the server
3. Swipe-to-refresh

It doesn't matter if your "Reading order" preference is "Newest first" or "Oldest first", if you do this your reading position is retained, new posts load above your current position in both v23.0 and this PR.

> Rather: Imagine a single page contains 10 posts. What we should do, in this regime, is load the single page of 10 posts and no more. Then once the user has scrolled up by 5 posts, _then_ we should load one more single page. (The numbers 10 and 5 here are examples; we could pick other numbers, but the goal should be to _load more content slightly faster than a typical user would be able to read the content_, so that as they scroll up they are always presented with just a little more.)

That's exactly what happens now. A page size is normally 30 statuses, but can be larger or smaller than that. The combination of `RecyclerView`, `PagingAdapter`, and `PagingSource` communicate so new pages are loaded as the user gets too close to the edge of the currently-loaded pages.

> It seems like I only load more when I do the pulldown gesture.
> 
> [...]
> 
> IN THE HOME TAB: I cannot under any circumstances get tapping the tab icon, in either Tusky 23 or Tusky PR, to do anything other than scroll to the top. The mere act of being scrolled to the top does not load more posts. To load more posts, I must pull down (or use the menu, I guess). I can wait as long as I like at the top and nothing is loaded, I _must_ pull down.

That's correct, and intended behaviour in both cases.

At some point prior Tusky loaded the timeline, all the way to the most recent post you had received at that time.

When that happens Tusky notes that there are no more pages to load, and will not silently attempt to load any more unless the user takes an action to indicate that more content should be loaded, or the current content is stale[2].

Consider starting with this situation. The user is at the top of a timeline, looking at the top of page 5.

```mermaid
flowchart LR

  5:::current --- 4 --- 3 --- 2 --- 1

  classDef notloaded stroke:#000,stroke-dasharray: 5 5
  classDef current stroke:#f00,stroke-width:2px,font-weight:bold
```

Tusky "knows" that the last request to fetch a hypothetical page 6 returned an empty page, so it knows that page 5 is the most recent page.

Time passes. Over on the server new posts arrive, and over time two new pages of posts stack up. The world now looks like this:

```mermaid
flowchart LR

  7:::notloaded --- 6:::notloaded --- 5:::current --- 4 --- 3 --- 2 --- 1

  classDef notloaded stroke:#000,stroke-dasharray: 5 5
  classDef current stroke:#f00,stroke-width:2px,font-weight:bold
```

Tusky still does not know that pages 6 or 7 are fetchable. So when you tap the tab (in either Tusky release, or this PR), Tusky jumps you to the top of page 5 (if you were not already there) and does not try and fetch page 6.

Then you swipe-refresh (or use the refresh menu, or the "Load newest" menu). Now you have positively told the app "Go look for new posts". It does, and either:

You did a refresh, in which case things look like this:

```mermaid
flowchart LR

  7 --- 6 --- 5:::current --- 4 --- 3 --- 2 --- 1

  classDef notloaded stroke:#000,stroke-dasharray: 5 5
  classDef current stroke:#f00,stroke-width:2px,font-weight:bold
```

(pages 6 and 7 are loaded and you can scroll to them, but your position on page 5 has not changed)

Or you did a "Load newest", and it looks like this:

```mermaid
flowchart LR

  7:::current --- 6 --- 5 --- 4 --- 3 --- 2 --- 1

  classDef notloaded stroke:#000,stroke-dasharray: 5 5
  classDef current stroke:#f00,stroke-width:2px,font-weight:bold
```

> It is possible the behavior is different when you are "at the present" (which my Home timeline was for these tests) than when you are scrolled back in time (which my Notifications timeline was for these tests).

That is possible. Your home timeline matched the scenario above. Your notifications timeline was probably more like the following (I'll use the same page numbers as in the previous diagrams, but keep in mind that this is a different timeline).

You start with your Notifications timeline looking like this:

```mermaid
flowchart LR

  7:::notloaded --- 6:::notloaded --- 5 --- 4 --- 3 --- 2:::current --- 1

  classDef notloaded stroke:#000,stroke-dasharray: 5 5
  classDef current stroke:#f00,stroke-width:2px,font-weight:bold
```

You are reading notifications somewhere on page 2. Pages 3, 4, and 5 are also loaded. Pages 6 and 7 exist on the server. To avoid that "infinitely loading pages" problem that I mentioned earlier, Tusky **did not** read past page 5. So at this time it doesn't know if page 6 exists or not.

In the previous scenario Tusky had attempted to read past page 5, got an empty page, and marked page 5 as the newest, and will not try to load past it without the user explicitly refreshing.

In this scenario Tusky has **not** attempted to read past page 5, so does not know if page 5 is the last page or not.

**That's the crucial difference between them.**

You scroll, reading upwards, and in to page 3. The situation is now (very briefly) like this:

```mermaid
flowchart LR

  7:::notloaded --- 6:::notloaded --- 5 --- 4 --- 3:::current --- 2 --- 1

  classDef notloaded stroke:#000,stroke-dasharray: 5 5
  classDef current stroke:#f00,stroke-width:2px,font-weight:bold
```

Then Tusky realises that you're running out of pages ahead of your current position. Since it doesn't know that page 5 is the last page yet it tries to speculatively fetch page 6. That succeeds, so now the situation looks like this:

```mermaid
flowchart LR

  7:::notloaded --- 6 --- 5 --- 4 --- 3:::current --- 2 --- 1

  classDef notloaded stroke:#000,stroke-dasharray: 5 5
  classDef current stroke:#f00,stroke-width:2px,font-weight:bold
```

Page 6 is loaded, and just like with page 5, because "enough" pages are loaded ahead of your current position Tusky doesn't try and see if page 6 is the last page or not. So if you keep scrolling up in to page 4 then Tusky will try and load page 7.

(again -- these examples are illustrative, the exact number of pages that the app tries to read ahead, or the number of statuses on each page, isn't relevant to understanding the underlying mechanism)

Hopefully this explains why, if you're on a timeline that is gaining pages of new posts faster than you are reading them, you can continue to scroll upwards to read them without ever encountering the top.

> I am very, very confused by this additional column containing the unmoving reload icon

Did you mean "row" instead of "column" here?

"unmoving" I guess, because you have animations turned off?

This is the progress indicator showing Tusky trying to load the page immediately above your current position. You only see this if you've scrolled right to the top and Tusky is still loading from the server.

It's present during the load, disappears when the load succeeds. If the load fails it's replaced with an error message and a button allowing you to retry the load.

This is so the user can tell the difference between:

- I did a refresh / reload, and no new data appeared, I guess there's nothing new (and/or "I guess Tusky has a bug")
- I did a refresh / reload, and I got an error message with a retry option, I guess my server is having problems

(`item_load_state_footer.xml` if you want to look deeper -- despite the name the same layout is used for both the header and the footer row)

It's not a very pretty UI. I have considered making it more unobtrusive, like the loading bars you get when you're viewing a thread, and only expanding to an error message if there is an error to show, but I haven't got to that yet.

---

[1] They do have the concept of "placeholder" pages, for when the data hasn't been loaded yet. However, we can't use those, because they require the server to communicate how many items are missing so the placeholders are the correct size. The Mastodon API does not provide this information.

[2] We've been talking about swipe/refresh and reloading as being the things that can cause new data to be loaded. For completeness there are other things as well. For example, if the user changes their filters then the timeline is refreshed to ensure that filters are applied.

mcclure:
> The Notifications tab has been using this code, or a version of it, since v22.0, with no user complaints that I'm aware of (to either the Tusky account or the #tusky tag). I introduced these changes to the Notifications tab to cover for exactly this -- get widespread usage of the new UX and see if anyone complained.

Okay, but then why do I see different behavior in this PR than I see in v23.0? With v23.0 I can get to the most recent notification very easily, with this PR I have to keep pulling and pulling and pulling indefinitely before I reach the top (if I ever do)

> The same could be said of tapping on a tab; it's undiscoverable UI.

I think tab-tapping is a very common idiom in software like this. It's in Twitter at least. It's reasonable to assume users will learn it from using apps on their phone; it's also easy to find "by accident" if you are simply changing tabs (though I'm not sure that's a good thing…). There's also the advantage that tapping on the tab can be done quicker than exposing and navigating a menu, which is an advantage for a frequent action.

(However, if tapping on a tab is too complex/undiscoverable, then this sounds to me like an argument in favor of trying to keep "show more"; even a user who's never found tab-tap can understand "scroll to the top and pull down").

>> Before, users who prefer to read current posts first could perform the natural action of just pulling down
> I don't think your description of the current Tusky behaviour here is correct… It doesn't matter if your "Reading order" preference is "Newest first" or "Oldest first", if you do this your reading position is retained, new posts load above your current position in both v23.0 and this PR.

That's true, but in this case the new posts that load above the "skip to the top" point are post-Load More posts. In other words, you're right that top-tapping+ pulldown won't take you all the way to the top in v23, but it will take you *very close* to the top. You are looking at current posts, if not the very newest.

(In addition in v23 you *can* get to the true top quickly through tab-tap, pulldown, tab-tap, which can be done in a split second. A user can figure that gesture on their own; it's maybe a more complex gesture sequence than it could be, but at least *the app offers a way to get to the top at all* without going through a menu. I feel from other discussions with users on the other hand that they don't notice our menus [and sometimes don't even notice top-bar icons…] and as described above I didn't think to look in the menu until told at least.)

> The Paging3 libraries don't have a "load more" equivalent, so that won't work[1].

I don't understand. If each "page" from the server has both a "back" and a "forward" link, it seems like we can always get "more up" from a previously loaded point, and we can always jump to the top and get "more down" from the timeline head.  Or do you mean that the *Android libraries* we are using are incapable of supporting loading in more items at an arbitrary point selected by us? That seems like a very large limitation; surely random insertion is a must-have feature for list display. I am shocked Google would ship like that.

>> once the user has scrolled up by 5 posts, then we should load one more single page
> That's exactly what happens now
>> At some point prior Tusky loaded the timeline, all the way to the most recent post you had received at that time. … When that happens Tusky notes that there are no more pages to load, and will not silently attempt to load any more unless the user takes an action to indicate that more content should be loaded

I don't understand how these two things are compatible. Either when you scroll up by 5 posts it loads the post ahead to present an infinite scroll, or else there is a hard limit it won't load past without user intervention and the scroll is not infinite.

Meanwhile, this idea of "will not silently attempt to load any more unless the user takes an action"… I don't understand why this is a hard rule. I don't understand why the "this is my load point, I will not load anymore" point is considered important, especially if we have implemented infinite scrolling in the other direction. If the user previously "created" that top point by an action, like pulling down while still in the same column, it *might* make sense to hold fast to that to avoid confusing the user. But the "this is my top point, I will load no more" line is maintained even if the user takes actions— switching from tab to tab, quitting and reopening the application— that would generally cause the user to mentally context switch and not specifically remember they performed a recent action that set that line. If you dismiss Tusky and do not reopen it for two hours, then the next time you come back the "line Tusky won't go above" isn't meaningful anymore— it's just an arbitrary line, potentially hours or days in the past, that Tusky behaves differently around than other points in a usually-infinite scroll.

Is there a reason for leaving the tab and returning *not* to invalidate that last-loaded point?

----

Overall: I appreciate that you're putting a lot of effort into these explanatory diagrams, but here's the point: *I have no intuitive understanding of how to expect the app to work now.* As a user, I am unable to develop a mental model of how it works now. As a *programmer* I can sort of understand what the app is doing, although with this latest set of diagrams I am starting to be a bit bewildered :( But I have special advantages here because I've got the person who wrote the new code explaining it to me personally. Most users won't.

My "expected behavior" is that the user mental model is simple and learnable. I feel that:

* Nothing new loads unless you pull down;  when you do pull down, it loads the newest posts, and if there's a gap between the newest posts and the last-shown posts, there's a skip and an unambiguously-labeled button to expand more is placed i nthe gap.

is both simple and learnable. I also feel that:

* As you scroll up and down, posts continuously scroll in; when you reach the "present" they stop, and pulling down more from that point can eagerly load more

Is also simple and learnable, although I want there to be an easy and "discoverable" way to jump all the way to the present.

I don't feel like what we've got in this PR now is simple and I don't feel it self-explains from use.

I will make an attempt to re-read this latest batch of diagrams tomorrow or something >_>

mcclure:
> This is the progress indicator showing Tusky trying to load the page immediately above your current position. You only see this if you've scrolled right to the top and Tusky is still loading from the server.
> 
> It's present during the load, disappears when the load succeeds. If the load fails it's replaced with an error message and a button allowing you to retry the load…
>
> It's not a very pretty UI. I have considered making it more unobtrusive, like the loading bars you get when you're viewing a thread, and only expanding to an error message if there is an error to show, but I haven't got to that yet.

I think it looks fine— I don't even mind that it's non-animating. If that's because I turned animations off that's especially ok. The problem I'm having is it seems inconsistent when it does or doesn't appear. If I pull down it can take a full second for more posts to appear, But the reload icon doesn't always appear in that time. (I don't *think* this is something that my animation settings would have broken.) If anything maybe it should be more obtrusive? Although it's only really a problem currently because "what is it loading and when" feels confusing to me in the PR interface.

Tak:
> My "expected behavior" is that the user mental model is simple and learnable. I feel that:
> 
>     * Nothing new loads unless you pull down;  when you do pull down, it loads the newest posts, and if there's a gap between the newest posts and the last-shown posts, there's a skip and an unambiguously-labeled button to expand more is placed i nthe gap.
> 
> 
> is both simple and learnable. I also feel that:
> 
>     * As you scroll up and down, posts continuously scroll in; when you reach the "present" they stop, and pulling down more from that point can eagerly load more
> 
> 
> Is also simple and learnable, although I want there to be an easy and "discoverable" way to jump all the way to the present.

Big +1
I feel like tapping the tab button is the *ideal* time to load newest posts, because you're already warping the user's reading position.
However, if we want to make loading newer posts only occur on explicit refresh, I feel like there's no way around having a preference for "what happens when I pull to refresh? load one newer page / load the *newest* page (and ideally warp to top)" if you don't want to degrade the Tusky 23 experience for *somebody*.

nikclayton:
> Big +1 I feel like tapping the tab button is the _ideal_ time to load newest posts, because you're already warping the user's reading position. However, if we want to make loading newer posts only occur on explicit refresh, I feel like there's no way around having a preference for "what happens when I pull to refresh? load one newer page / load the _newest_ page (and ideally warp to top)" if you don't want to degrade the Tusky 23 experience for _somebody_.

So, option 4, of 

1.    Do nothing
2.    Jump to the newest loaded posts (this is the current behaviour)
3.    Jump to the newest loaded posts, and refresh to see if there's anything new (this would be new behaviour)
4.    Load the absolute newest posts, and jump to them (i.e., same as the "Load newest" menu option)

?

Tak:
Option 2 would be the "least surprise" behavior coming from Tusky 23.

But if you also want to make it possible to read oldest-to-newest, then *some* behavior has to change.

*If* we're changing behavior, the most straightforward way is to have a preference for reading order, keep the Tusky23 behavior for tapping a tab, and make pull to refresh either (load newest and jump to top (newest-to-oldest)) or (load one newer page and don't move (oldest-to-newest)).

*but*:
- if you're reading oldest-to-newest, you will naturally scroll to the top as you read and will never\* tap the tab button for the current tab
- if you're reading newest-to-oldest, you will always scroll down and will never\* pull to refresh without tapping the tab button first

So it feels like we could just make pull-to-refresh load one newer page and make tapping the current tab button Option 4 and it would Just Work intuitively for everyone without having a preference.

mcclure:
> I'd need to see a video showing the difference, or a clear list of reproduction steps that demonstrate the problem.

I will take a video. Perhaps the reason I'm having so much trouble with this is what I'm seeing is not how the patch is supposed to work.

mcclure:
… okay, this video came out *very confusing* , one second

mcclure:
Here is a demo video. It shows two things:

1. I am seeing something *very weird* in my day-to-day-use copy of v23, which could indicate a bug in the v23 Paging3 usage as well as possibly the PR Paging3 usage.
2. A demonstration of the rotaty-reload icon behavior I described above as confusing me

In this video, I show three separate copies of Tusky:

1. Vanilla v23.0 (blue)
2. "Andi-hacks" v23.0 (black), which is the shipping v23.0 plus a small group of commits. See: https://github.com/mcclure/Tusky/releases/tag/v23.0-andi-hacks-r0
    * These commits shouldn't have an impact on the behavior of paging3. They are: A change to the display of alt tags; a change to tab-swipe speed; a new button in notifications tab (notification related but should[?] have no effect unless the new button is pressed); a change to bio display in notifications (notification related but should[?] be XML-only); several build-related nikclayton commits (Upgrade AGP to 8.0.2] etc)
3. This PR (green)

I made this video because @nikclayton said "The Notifications tab has been using this code, or a version of it, since v22.0, with no user complaints" [including from me], which is confusing because I noticed no change in the notifications tab when going from v21 to v23+andihacks whereas I see a very large difference going from v23+andihacks to this PR. What I found in the video was that the notification tab behavior of this PR is *the same as* vanilla v23.0, but that v23+andihacks has a notification tab behavior different from *either*, which should not be possible.

Here is the video: https://www.youtube.com/watch?v=YvWRoj_LuiM

0:04: Boot Tusky v23.0.
0:09-0:47: Enter Tusky v23.0 notifications tab. The content is 185 days old. I try to get to the top by a combination of scrolling up and tapping the tab. Progress is very slow and over 30 seconds I am unable to reach the top or even escape day 185. (This is very, very bad. If v23 had been behaving like this for me on my day-to-day use install, I would have stayed on v21.)
0:49: Boot Tusky v23.0+andihacks.
0:50-1:07: I attempt to scroll to the top of the notifications tab. It takes longer than I would have expected (notice an interesting moment where I freeze on the reload-icon box, then switch tabs) but I do eventually get to the top. How is this possible when I could not do it in v23.0? I investigate:
1:15: I scroll down, then up. At 1:15 what I find is **there is a gap in my notifications between 6 minutes ago and 20 hours ago**. Nothing appears in between. There is no indicator of missing content.
1:32: Re-enter Tusky v23.0. I try to replicate what I just saw in v23.0+andihacks. No success. (I do try tab swiping while the reload button is up but maybe my timing was wrong.)
2:06: Scroll down to confirm Tusky v23.0 is a Google Play Tusky v23.0.
2:13: Boot Tusky [this PR].
2:14-End: Without success, I attempt to scroll to the top of the notifications tab. This is where I got a good view of the reload icon behavior:
2:21: The notification tab first displays after a long delay. The reload icon is visible up top, indicating that more posts are being loaded (good)
2:40: I double tap. It scrolls to the top, showing a reload icon.
3:01: I tap once. I got to the top. I am 17 hours in the past. Although more posts are being loaded, no reload icon shows.
3:05: I scroll down. It scrolls smoothly indicating nothing new is being drawn in. There has been a reload icon up there all along, Tusky just wasn't showing it.
3:16: New content loads in. I tap once. I get a reload icon; we've scrolled above it. (I can't explain this? Possibly it's a timing thing or possibly I double tapped and the double tap didn't show well in the video.)

Let's focus on 1:15. There are two possible explanations here. One is that this is about the difference between v23 and v23+andihacks— this is possible, but I don't think this could be it, due to the minor nature of the patches. The other explanation, which I *think* is what is happening, is that there is a bug (which I coincidentally managed to trigger in my v23-andihacks test but not my v23-vanilla test) which can cause the notifications tab to "skip" a large amount of content. In other words, it is as if the "Show More" behavior was retained, but with a less consistent trigger condition, and without any "Show More" button. Although this bug is *bad* (losing user content without indication is very, very wrong!), this bug was what made it possible for me to use v23 at all, because apparently whenever I got too far behind in the feed I would just fiddle with it until I accidentally triggered the bug. I really want to stress this: **If the behavior I saw in my vanilla v23 test above had been the behavior I had seen on first installing v23, I would not have switched from v21 to v23 at all**, I would have stayed on v21. If we have not seen user complaints about the v23 behavior, it could be that users who get "too far behind" in the notifications tab are accidentally triggering this bug.

(A third possibility is that changing filters, which I do pretty frequently, resets you to the top of the feed automatically. I haven't specifically tested this.)

nikclayton:
The bug you note on your andihacks branch at 1:15 -- unless you can reproduce that on v23 or this PR I'm not going to investigate. Given that your branch changes how filters are handled it's possible that gap is due to misconfigured filters. I've never seen it in v23 or on this PR, and no one has reported anything like it elsewhere.

To the contrary, what this PR does do is fix an existing bug caused by the "Load more" behaviour where posts are missing from the timeline (Home, Local, Federated, etc), and there *are* external user complaints about that behaviour.

The rest of the behaviour in the video on v23 this PR looks as intended.

I do note two changes that might make some things more obvious to users.

1. The rows that show if the previous/next page is loading (in your video they show at the top of the list with non-moving progress spinned) should not rely on just the animation to inform the user that something is happening, because the animation may be disabled. They should also have explicit "Loading..." text so the user is aware that something is happening.

2. If there's an active (prepend) page load happening then ignore refresh operations (swipe-refresh and Menu > refresh), since the prepend operation will do the same thing as the refresh.

---

> I think that because this patch changes the user model so significantly, [...]

- Automatically refreshing when the user resumes[1]? That should be straightforward. I'm not 100% sold on the idea, but it's a reasonable experiment. The idea that "The user has just switched to this tab, they want to be able to see new content with the minimum of fuss" is defendable.

- Refreshing on a timer, however, is not. Consider, for a moment, the impact that would have on the load experienced by the users' servers.

> I assume there's some fourth operation that allows us to do in-situ post edits and post deletion on the timeline

No. The data is supposed to be immutable. Not all the existing Tusky code honours this (e.g., by sneaking in mutable data structures where the libraries expect immutable ones).

Fixing this is also on my todo list, I suspect it's the source of subtle bugs.

> Can you elaborate on what the "discard" operation entails? How disruptive of it? Does it entail a screen flash? Does it require flushing our caches of data (I assume no, we are maintaining the cache of content and Paging3 is maintaining the cache of constructed UI objects and it is the cache of UI objects that clears on a "discard"/invalidate?)

A refresh or reload operation[2] discards the existing data (from memory, and from the backing database if there is one), and new data is fetched and cached.

Then the new data is applied to the list via the list adapter. The adapter contains code to diff the old list and the new list, figure out what's changed, and perform the minimum number of screen updates to show the new data.

There is no screen flash. Those changes either appear directly, or smoothly animate in (e.g., see what happens when you view a thread).

For much more on this see https://developer.android.com/codelabs/android-paging#0.

[1] Resume occurs both when Tusky returns to the foreground, or if the user switches tabs.

[2] "refresh" is just "reload, but remember and restore my reading position". Think of them as "My local cache of the data is out of date, it *must* be replaced by up to date content from the remote server".

mcclure:
> The bug you note on your andihacks branch at 1:15 -- unless you can reproduce that on v23 or this PR I'm not going to investigate. Given that your branch changes how filters are handled it's possible that gap is due to misconfigured filters.

Possible D:

mcclure:
Emphasis mine:
> A refresh or reload operation[2] discards the existing data (from memory, and from **the backing database** if there is one), and new data is fetched and cached.
>
>Then the new data is applied to the list via the list adapter. 

Fetched from where? Like… all the way from the server? Could it be fetched from a secondary data store Tusky hosts (so that if we wanted to do "correct" post edits or "load more" page inserts, we could simply change the Tusky device-side store, invalidate, and let paging3 redraw with amended content)? Or is the idea that having switched to paging3, "the data store" is now hosted entirely by paging3 and we no longer have a secondary data store it could be drawn from?

nikclayton:
> Fetched from where? Like… all the way from the server?

Yes. The on-device copy is treated as a local cache that can be discarded when necessary.

> Could it be fetched from a secondary data store Tusky hosts (so that if we wanted to do "correct" post edits or "load more" page inserts, we could simply change the Tusky device-side store, invalidate, and let paging3 redraw with amended content)?

The word "simply" in that sentence is doing an awful lot of work.

In practice, the answer is no.

https://developer.android.com/topic/libraries/architecture/paging/v3-overview has more information on how the library is supposed to be use.

> Or is the idea that having switched to paging3, "the data store" is now hosted entirely by paging3 and we no longer have a post store it could be drawn from?

No.

The data store (e.g., the local database) is still managed by the app. But on receipt of a "refresh" operation the local copy of the data is cleared, and a fresh copy fetched from the server.

This does not mean **all** the local data is cleared, only the data that the server hosts.

So information like "The user has clicked 'show more' on this status" and it should be rendered like that the next time it's displayed, or "This status has a content warning and the user has clicked to show the content of the post" is local view state that the app should maintain and **not** delete when the user refreshes.

Tusky is currently inconsistent about doing that, as I think I mentioned earlier. I've got ideas about how to fix that, but they're orthogonal to this PR.

mcclure:
>> I assume there's some fourth operation that allows us to do in-situ post edits and post deletion on the timeline
>>
>> No. The data is supposed to be immutable. Not all the existing Tusky code honours this (e.g., by sneaking in mutable data structures where the libraries expect immutable ones).
> 
> Fixing this is also on my todo list, I suspect it's the source of subtle bugs.

Okay. Based on a few of your comments but particularly this one, my personal opinion at this point is that paging3 is not appropriate for our use case and we should not be using it, possibly not even for notifications.

nikclayton:
> Based on a few of your comments but particularly this one, my personal opinion at this point is that paging3 is not appropriate or our use case and we should not be using it, possibly not even for notifications.

@mcclure That ship sailed in Jan 2021, with https://github.com/tuskyapp/Tusky/pull/2238

Reading that PR, I'm somewhat amused to note that @charlag identified the mutable data problem and suggested a cleaner way to implement it in https://github.com/tuskyapp/Tusky/pull/2238#issuecomment-1002561050. The approach that was used (https://github.com/tuskyapp/Tusky/pull/2238#issuecomment-1002765152) is some of the technical debt that needs to be cleaned up.

nikclayton:
@mcclure 

> Although more posts are being loaded, no reload icon shows

This is because the progress spinner that indicates if a previous or next page is being loaded is a `ContentLoadingProgressBar` (https://developer.android.com/reference/androidx/core/widget/ContentLoadingProgressBar).

When being shown, this will wait 500ms before appearing. If there's a "hide" in the intervening 500ms then it won't show at all, preventing brief UI adjustments.

Similarly, once it's showing it will stay shown for at least 500ms even if it's told to "hide" before then. Again, to avoid very brief UI adjustments.

So if the load and processing of the previous (or next) page takes < ~500ms you'll not see a progress spinner. If it takes longer you will, and the spinner will be visible for at least 500ms.

This is similar behaviour to what happens when viewing a thread. If you open a thread there's a network request to fetch the status you're viewing, and a second request to fetch the statuses above/below it. If those requests complete in < ~500ms you'll never see the loading indicator, they'll smoothly animate in. If those requests take longer you'll see a loading indicator, so you know there's something to wait for.

nikclayton:
@mcclure 

> Automatically refreshing when the user resumes[1]? That should be straightforward. I'm not 100% sold on the idea, but it's a reasonable experiment. The idea that "The user has just switched to this tab, they want to be able to see new content with the minimum of fuss" is defendable.

I've been trying a build with that for the last 24 hours or so.

I think it's too disruptive, as things that you would not expect to cause a refresh do.

For example, tapping a message to view its thread and then pressing back will trigger a refresh. So will tapping media in a status to view it full screen (image, video, or audio).

This means:

1. The "swipe-refresh" progress spinner appears, which is good, because it tells the user that something has happened, but bad because it's not obvious why it's happened.
2. Any changes to the user's timeline that affect the presentation of that timeline will cause it to adjust, potentially shifting content up or down, or removing it entirely. This happens (approx.) whenever a post in the timeline is edited, or a filter now matches it, or a link preview that wasn't there before is there now.
3. If there is new content then the "scroll by 30dp to show there's new content" logic triggers, so the user's scroll position in the list is changed.

I think the user can expect this if they explicitly triggered a refresh -- reopening the app, swiping, choosing "refresh" from the menu.

But I don't think that activities like "viewing a thread" or "opening an image" (or "swiping to this tab") have this expectation.

I'll look and see if it's possible to do this *only* when the user is resuming the fragment because they've clicked on the tab to switch to it.

mcclure:
> Automatically refreshing when the user resumes[1]? That should be straightforward. I'm not 100% sold on the idea, but it's a reasonable experiment.
>
> I think it's too disruptive, as things that you would not expect to cause a refresh do.

I'm sorry, I think I was unclear.  I was not trying to suggest we should refresh on returning to the program, which probably would be disruptive, but rather that the "water line" above which we do not load further data without user input (assuming I understand correctly and this "water line" exists) should be invalidated, and if the user is already scrolled close enough to the top of the newest page we would (in the absence of the "water line") begin loading in newer data (but not scrolling toward it, which would indeed be disruptive).

nikclayton:
> rather that the "water line" above which we do not load further data without user input (assuming I understand correctly and this "water line" exists) should be invalidated

The "waterline" your describing is **only** created when Tusky has requested the page and the server has returned an empty page (which is how the server says "This page you have requested does not exist").

Then, and only then, is a "waterline" created.

To get Tusky to ask the server again you have to "refresh".

That's **exactly** the same behaviour as the current code -- when you refresh in Tusky at the moment the newest page is explicitly loaded. You **can not scroll up past this page** without performing a refresh.

Let's say you left Tusky reading page 90 of your home timeline, and have returned. In the intervening time pages 91-100 have been created on the server.

**In Tusky 23.0**

- You open the timeline
- It refreshes
- Page 100 is loaded
- Your reading position is page 90
- A "Load more" placeholder is shown between pages 100 and 90.
- **You cannot scroll up past page 100 without explicitly refreshing**
- To load pages 91-99 you must repeatedly tap the "Load more" placeholder. When you do that (and depending on your reading order preference) pages are loaded in order, either 91, 92, 93, ... or 99, 98, 97, ...

**In this PR**

- You open the timeline
- It refreshes
- Pages 91-94 are loaded
- Your reading position is page 90
- You scroll up to read. As you read, pages 95, 96, 97, 98, 99, and 100 are loaded for you.
- Assuming there is nothing past page 100 you stop
- Now, **you cannot scroll up page page 100 without explicitly refresehing**

In both cases what happens when the user reaches the top of page 100 is the same -- the user can't scroll without a refresh.

This PR improves things in the case where new pages become available while you're scrolling and haven't reached the top.

Consider this slight change in the same scenario, where the server is adding new pages to the top of the timeline.

**In Tusky 23.0**

- You open the timeline
- It refreshes
- Page 100 is loaded
- Your reading position is page 90
- A "Load more" placeholder is shown between pages 100 and 90.
- **You cannot scroll up past page 100 without explicitly refreshing**
- To load pages 91-99 you must repeatedly tap the "Load more" placeholder. When you do that (and depending on your reading order preference) pages are loaded in order, either 91, 92, 93, ... or 99, 98, 97, ...
- While you are doing the server adds pages 101, 102, and 103.
- You scroll up and reach the top of page 100. You *must* refresh to see the new pages.

**In this PR**

- You open the timeline
- It refreshes
- Pages 91-94 are loaded
- Your reading position is page 90
- You scroll up to read. As you read, pages 95, 96, 97, 98, 99, and 100 are loaded as you get closer to them.
- While you are doing the server adds pages 101, 102, and 103.
- If these pages are added on the server *before* Tusky has loaded page 100 then page 100 *won't be set as the "waterline"*. So you can continue scrolling past page 100 without needing to "refresh"

So typically, with this PR you will:

1. See more content before you need to explicitly refresh
2. Will have to tap less, because you never need to tap a "Load more" placeholder

PR #3440, opened 2023-03-12
Fix UI error, return the actual error body as a string

nikclayton:
The previous code returned the text representation of the error body type, which resulted in errors appearing in the UI as:

```
okhttp3.ResponseBody$Companion$asResponseBody$1@...
```

This code actually converts the *body* of the error response to a string, so the error is displayed correctly.


nikclayton:
Here's an example of what it looks like (for HTTP 4/5xx errors, errors at the network stack, like "host not found" are displayed correctly) before this fix:

charlag (approved):

connyduck (approved):

Issue #2992, opened 2022-12-05
Remove rxjava


nikclayton:
Tracking issue for the work to remove rxjava. Quoth @connyduck:

> Basically everything should be Kotlin + Coroutines, everything Paginated  AndroidX Paging, if you see something java or Rx feel free to refactor and send a PR
> [...]
> Can't wait for the day when I finally kick Rx out

# Mastodon API

In `MastodonApi.kt`, `Single` is `io.reactivex.rxjava3.core.Single`, removing that from API responses, marking the API functions as `suspend`, and adjusting the call sites should clean up a lot of RX use.

- [x] getFilters, https://github.com/tuskyapp/Tusky/pull/2990
- [x] notifications
- [ ] markersWithAuth
- [ ] notificationsWithAuth
- [x] clearNotifications
- [x] status, https://github.com/tuskyapp/Tusky/pull/2999
- [x] statusRebloggedBy, https://github.com/tuskyapp/Tusky/pull/3005
- [x] statusFavouritedBy, https://github.com/tuskyapp/Tusky/pull/3005
- [x] deleteStatus, https://github.com/tuskyapp/Tusky/pull/3041
- [x] reblogStatus, https://github.com/tuskyapp/Tusky/pull/3042
- [x] unreblogStatus, https://github.com/tuskyapp/Tusky/pull/3042
- [x] favouriteStatus, https://github.com/tuskyapp/Tusky/pull/3042
- [x] bookmarkStatus, https://github.com/tuskyapp/Tusky/pull/3042
- [x] unbookmarkStatus, https://github.com/tuskyapp/Tusky/pull/3042
- [x] pinStatus, https://github.com/tuskyapp/Tusky/pull/3048
- [x] unpinStatus, https://github.com/tuskyapp/Tusky/pull/3048
- [x] muteConversation, https://github.com/tuskyapp/Tusky/pull/3048
- [x] unmuteConversation, https://github.com/tuskyapp/Tusky/pull/3048
- [ ] scheduledStatuses
- [ ] account
- [x] accountFollowers, https://github.com/tuskyapp/Tusky/pull/3005
- [x] accountFollowing, https://github.com/tuskyapp/Tusky/pull/3005
- [x] followAccount, https://github.com/tuskyapp/Tusky/pull/3008
- [x] unfollowAccount, https://github.com/tuskyapp/Tusky/pull/3008
- [x] blockAccount, https://github.com/tuskyapp/Tusky/pull/3008
[...]

Issue #2741, opened 2022-11-01
Hitting follow then back quickly can result in follow silently failing, esp under high server load


mcclure:
On 2022-Oct-31 mastodon.social was under very heavy load due to the Twitter buyout, exposing some Tusky misbehaviors when the server is responding slowly.

**Repro steps:**
* Tap a user avatar to open their user page
* Tap "Follow". Observe that "Follow" has changed to "Unfollow".
* Immediately tap the Android system "Back" button to return to the feed.
* Wait a few seconds then open the user page again to confirm the follow worked.

**Expected behavior:**  
The user is followed.

**Observed behavior** (if the server is under heavy load):  
The user is not followed, even after waiting. According to @connyduck on Matrix, the "Follow" button changes to "Unfollow" optimistically, but the follow request is abandoned when the user page is dismissed.

**Analysis:**

I think "Follow" changing to "Unfollow" is good behavior, but mastodon should do a better job of catching an unusual case where the follow fails or takes so long the user navigates away (hits "back", or even quits Tusky)

Connie's suggested fix: "turn the button into a spinner while the request is in flight", the user would then one assumes wait for it to complete spinning before navigating away  
My suggested fix: Eagerly switch to "unfollow", but if the user starts navigating up block them with a modal dialog until the follow request completes (Connie said this sounds harder than the spinner, and it might not catch the case where Tusky itself is dismissed).

* * * *
- Tusky Version: 19.0
- Android Version: 12
- Android Device: Sony XPeria 5 III
- Mastodon instance (if applicable): mastodon.social

connyduck:
I bet Pete hit that bug https://mastodon.social/@PeteWM1@mas.to/109406278480222481

nikclayton:
The user shouldn't be blocked by a modal dialog. I've fixed some of this for notifications in https://github.com/tuskyapp/Tusky/pull/3159, as a model for how to rewrite the other fragments to handle it.

> If an error occurs during a network operation the user is shown the error, with an opportunity to retry it. If the UI opportunistically showed the operation as succeeding (e.g., try and bookmark a status in airplane mode with the previous code, it will show the bookmark as succeeding) the UI is reset to show the operation did not succeed.

It's only a partial solution, because the state is lost when the user navigates away.

The current problem is because the UI can only show the user one of two states.

- This is the current state of the world
- This is your desired state of the world

Right now, in most places, the default is "This is your desired state of the world". Opportunistically showing "Follow" as succeeding, for example. 

[...]

PR #3446, opened 2023-03-13
Remove Rx from EventHub and TimelineCases

connyduck:
They are quite coupled so it made sense to do them both at the same time.

#2992

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/appstore/EventsHub.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/appstore/EventsHub.kt
@@ -1,20 +1,21 @@
 package com.keylesspalace.tusky.appstore
 
-import io.reactivex.rxjava3.core.Observable
-import io.reactivex.rxjava3.subjects.PublishSubject
+import android.util.Log
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.MutableSharedFlow
 import javax.inject.Inject
 import javax.inject.Singleton
 
 interface Event
-interface Dispatchable : Event
 
 @Singleton
 class EventHub @Inject constructor() {
 
-    private val eventsSubject = PublishSubject.create<Event>()
-    val events: Observable<Event> = eventsSubject
+    private val sharedEventFlow: MutableSharedFlow<Event> = MutableSharedFlow()
+    val events: Flow<Event> = sharedEventFlow
 
-    fun dispatch(event: Dispatchable) {
-        eventsSubject.onNext(event)
+    suspend fun dispatch(event: Event) {
+        Log.d("EventHub", "dispatching $event")
+        sharedEventFlow.emit(event)
     }
 }
--- a/app/src/main/java/com/keylesspalace/tusky/appstore/EventsHub.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/appstore/EventsHub.kt
@@ -1,6 +1,5 @@
 package com.keylesspalace.tusky.appstore
 
-import android.util.Log
 import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.MutableSharedFlow
 import javax.inject.Inject
@@ -15,7 +14,6 @@ class EventHub @Inject constructor() {
     val events: Flow<Event> = sharedEventFlow
 
     suspend fun dispatch(event: Event) {
-        Log.d("EventHub", "dispatching $event")
         sharedEventFlow.emit(event)
     }
 }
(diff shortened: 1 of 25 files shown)


app/src/main/java/com/keylesspalace/tusky/components/preference/AccountPreferencesFragment.kt
```diff
@@ -198,7 +200,6 @@ class AccountPreferencesFragment : PreferenceFragmentCompat(), Injectable {
                     setOnPreferenceChangeListener { _, newValue ->
                         setIcon(getIconForVisibility(Status.Visibility.byString(newValue as String)))
                         syncWithServer(visibility = newValue)
-                        eventHub.dispatch(PreferenceChangedEvent(key))
```
connyduck, line 201: `PreferencesFragment` send the event everytime a pref changes, so these were redundant

app/src/main/java/com/keylesspalace/tusky/appstore/EventsHub.kt
```diff
@@ -1,20 +1,21 @@
...
-import io.reactivex.rxjava3.core.Observable
-import io.reactivex.rxjava3.subjects.PublishSubject
+import android.util.Log
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.MutableSharedFlow
 import javax.inject.Inject
 import javax.inject.Singleton
 
 interface Event
-interface Dispatchable : Event
```
connyduck, line 9: Not sure why there were two interfaces? Seems unnecessay? @charlag?

Tak (approved):
LGTM pending the one thing I found and the test failures

nikclayton:
Nice. Would you mind updating https://github.com/tuskyapp/Tusky/issues/2992 with the progress?

app/src/main/java/com/keylesspalace/tusky/appstore/EventsHub.kt
charlag: I think the idea was to mark only certain events as the ones that can be sent but it makes little sense anyway

app/src/main/java/com/keylesspalace/tusky/components/preference/AccountPreferencesFragment.kt
charlag: I am a bit confused by your comment, where does it do that?

app/src/main/java/com/keylesspalace/tusky/components/timeline/viewmodel/TimelineViewModel.kt
```diff
@@ -120,7 +118,7 @@ abstract class TimelineViewModel(
 
     fun favorite(favorite: Boolean, status: StatusViewData.Concrete): Job = viewModelScope.launch {
         try {
-            timelineCases.favourite(status.actionableId, favorite).await()
+            timelineCases.favourite(status.actionableId, favorite).getOrThrow()
```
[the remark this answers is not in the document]
charlag: or even replace the whole `fold`, idk, it's quite clunky rn

app/src/main/java/com/keylesspalace/tusky/usecase/TimelineCases.kt
```diff
@@ -48,52 +48,42 @@ class TimelineCases @Inject constructor(
...
 
-    /**
-     * Unused yet but can be use for cancellation later. It's always a good idea to save
-     * Disposables.
-     */
-    private val cancelDisposable = CompositeDisposable()
-
-    fun reblog(statusId: String, reblog: Boolean): Single<Status> {
-        val call = if (reblog) {
+    suspend fun reblog(statusId: String, reblog: Boolean): NetworkResult<Status> {
```
charlag, line 51: what's not great about all of these being `suspend` is that they will be cancelled together with the caller's scope. So we might send a network request but not apply `dispatch` because the screen is already gone

charlag (approved):

app/src/main/java/com/keylesspalace/tusky/usecase/TimelineCases.kt
nikclayton: Hah, that's been on my list of things to fix too.

There's an application level scope to use for stuff like this, [MainScope](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/-main-scope.html).

Use it like this:

```
MainScope().launch {
    // do stuff that needs to be run on the UI thread or calls suspend functions
    // as MainScope uses Dispatchers.Main
}
```

(or store it in an `Application` level `val`, or similar. Then use this for network operations that should outlast the current screen.

That's still got the potential for silent failures though. I think the correct fix is https://github.com/tuskyapp/Tusky/issues/2741#issuecomment-1407430453, but that's a lot more work.

connyduck: > what's not great about all of these being suspend is that they will be cancelled together with the caller's scope. So we might send a network request but not apply dispatch because the screen is already gone

yes, but that is also the behavior before these changes

PR #3448, opened 2023-03-13
Show better errors when loading notifications fails

### LINKED ISSUE #3445: Error message handling for notifications still needs work
https://github.com/tuskyapp/Tusky/pull/3440 displayed error text when notifications fail to load because of the server, but the presentation of that can still be improved. I caught an error from mastodon.social a moment ago, and it currently looks like this:

That's obviously a JSON representation of the error. That's lead me to https://docs.joinmastodon.org/entities/Error/, so the right thing to do here is probably to decode this and present the error to user. Probably as:

```
<http error code>: <error>

<error description>
```

I want to avoid hiding this info behind a generic "An error occurred" message because it makes it impossible to troubleshoot user reports of problems.

nikclayton:
The errors are returned as a JSON object, parse it, and show the error message it contains.

Handle the cases where there might be no error message, or the JSON may be malformed.

Add tests.

Fixes #3445

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/components/notifications/NotificationsRepository.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/components/notifications/NotificationsRepository.kt
@@ -23,6 +23,7 @@ import androidx.paging.Pager
 import androidx.paging.PagingConfig
 import androidx.paging.PagingData
 import androidx.paging.PagingSource
+import com.google.gson.Gson
 import com.keylesspalace.tusky.entity.Notification
 import com.keylesspalace.tusky.network.MastodonApi
 import kotlinx.coroutines.flow.Flow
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/keylesspalace/tusky/components/notifications/NotificationsRepository.kt
```diff
@@ -31,7 +32,8 @@ import retrofit2.Response
 import javax.inject.Inject
 
 class NotificationsRepository @Inject constructor(
-    private val mastodonApi: MastodonApi
+    private val mastodonApi: MastodonApi,
+    private val gson: Gson
```
nikclayton, line 36: I'm not completely sure about this. 

This is injected here only so it can be passed down to `NotificationsPagingSource` as a parameter, so it can be provided in the tests as well.

I imagine there's a cleaner way to do this. Thoughts?

connyduck: I think it is fine

connyduck (approved):

Issue #1371, opened 2019-07-07
[Feature request] Group notifications for a toot 


Aldarone:
I'm just gonna copy/paste this toot content since it explains it perfectly

https://fosstodon.org/@rnickson/102400075628490654

> Mastodon has an annoying problem. I mean, it's not a bug. it's working as intended. but in annoying way.
>
> so when people boost/favourite your toot, you get a notification for it. a separate notification for every single action affecting your post. one of my posts has 158 boosts and 114 favourites. this means 272 notifications. sure, fine to know my post is interesting. but can't the notifications be grouped into a one notification, like on the most of social services?
>
> Twitter is a good example.

albjeremias:
https://github.com/mastodon/mastodon/issues/22578#issuecomment-1365571396

![image](https://user-images.githubusercontent.com/26681446/209607164-728d1622-f7df-48ed-8126-287f3e3f18de.png)
![image](https://user-images.githubusercontent.com/26681446/209607186-5b51e87d-83ef-486c-870f-8baf26231410.png)


[design penpot](https://design.penpot.app/#/view/9c5ceddf-c68c-8036-8001-cf95c95013ce?page-id=9c5ceddf-c68c-8036-8001-cf95c95013cf&section=interactions&index=0&share-id=35dd4c70-bf14-80a7-8001-cfa502432c90)

designed using https://penpot.app and open source amazing tool for crafting with :cupid: 

i think this is related:
https://github.com/tuskyapp/Tusky/issues/818

Issue #818, opened 2018-08-28
In-app notifications: Group similar notifications together


arunsathiya:
It would be a good idea to group in-app notifications together, on the `Notifications` tab of the app.

I have arrived at this idea from this issue #817. The idea is that notifications about new followers, new likes on toots can be grouped together periodically to show in this format:

`A, B and C favorited your toot.` and `A, B and C followed you.`

This feels better over seeing an endless chain of new follower and new like notifications on the `Notifications` tab of the app.

* * * *
- Tusky Version: 3 beta 2
- Android Version: 8.1.0
- Android Device: OnePlus 5
- Mastodon instance (if applicable): mastodon.social

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

charlag:
Hi! Thanks for idea. It was mentioned before but I do not see such an issue.
This indeed would be nice but it requires some logical changes for infinite loading so it's not super simple.

kitkatloyalist:
How is this different from "notification channels"?

https://developer.android.com/guide/topics/ui/notifiers/notifications

arunsathiya:
@kitkatloyalist Android's "notification channels" is about the push notifications sent to devices, while the proposal on the original post is around the notifications section inside the app itself.

PR #3452, opened 2023-03-14
818 group notifications

Lakoja:
Fixes [#818](https://github.com/tuskyapp/Tusky/issues/818) and [#1371](https://github.com/tuskyapp/Tusky/issues/1371) at least partly.

The idea here is that only the newest notification about a status is shown. All others are filtered (either on input from the api) or by removing "older" list elements after list update.

This should work as the array "seenFavorites" (to demonstrate) contains all relevant data as the notification list is always created from scratch.

If notifications were to be cached the data storage and calculation about "seen stuff" would be more challenging.


app/src/main/java/com/keylesspalace/tusky/components/notifications/StatusNotificationViewHolder.kt
```diff
@@ -249,10 +251,16 @@ internal class StatusNotificationViewHolder(
             Notification.Type.FAVOURITE -> {
                 icon = getIconWithColor(context, R.drawable.ic_star_24dp, R.color.tusky_orange)
                 format = context.getString(R.string.notification_favourite_format)
+                if (favoritesCount > 1) {
+                    userPart += " +" + (favoritesCount - 1)
```
Lakoja, line 255: How this should be looking exactly is the second question (after the first if this whole thing actually works above).

I'd prefer a text of "and 7 others". However this introduces singular plural problems.

A more fancy display (list of avatars for example) is probably out of scope here.

app/src/main/java/com/keylesspalace/tusky/components/notifications/NotificationsFragment.kt
```diff
@@ -205,10 +205,13 @@ class NotificationsFragment :
         (binding.recyclerView.itemAnimator as SimpleItemAnimator?)!!.supportsChangeAnimations =
             false
 
-        // Signal the user that a refresh has loaded new items above their current position
-        // by scrolling up slightly to disclose the new content
+
         adapter.registerAdapterDataObserver(object : RecyclerView.AdapterDataObserver() {
             override fun onItemRangeInserted(positionStart: Int, itemCount: Int) {
+                removeDuplicateOlderEntries(positionStart)
```
Lakoja, line 211: I guess this is a good location? "Consolidate complete list after inserts..."

I tried it below in "pagingData.collectLatest" first but I have no real idea what that does (or my idea does not match reality really).

app/src/main/java/com/keylesspalace/tusky/components/notifications/NotificationsFragment.kt
```diff
@@ -434,6 +438,27 @@ class NotificationsFragment :
...
+        for (pos in dataList.size - 1 downTo  positionStart) {
+            val notificationViewData = dataList[pos]
+
+            val status = notificationViewData.statusViewData?.status
+                ?: continue
+
+            if (!viewModel.hasNewestNotificationId(notificationViewData.type, status.id, notificationViewData.id)) {
+                Log.w(TAG, "Removing old notification at "+pos+" for "+status.id)
+
+                adapter.notifyItemRemoved(pos)
```
Lakoja, line 455: Simply "notify removed" seems odd as I don't have removed something but I want it to be removed in the list.

However I guess it might be correct as the backing list is api calls and what is presented in the list is or should not be exactly the api data.

app/src/main/java/com/keylesspalace/tusky/components/notifications/NotificationsFragment.kt
```diff
@@ -241,6 +244,7 @@ class NotificationsFragment :
                     viewModel.pagingData.collectLatest { pagingData ->
                         Log.d(TAG, "Submitting data to adapter")
                         adapter.submitData(pagingData)
+                        // TODO why is this called always _before_ anything from NotificationsPagingSource is loaded (and with what data)?
```
nikclayton, line 247: Short:

This code will run any time `viewModel.pagingData` is modified. So the position of this block of code (the inner `launch` block) doesn't especially matter within the larger `repeatOnLifecycle` block.

Long:

One of the first things the viewmodel does is start fetching data, using `NotificationsRepository`. When it has new data it is emitted in to `pagingData`.

The block of code here runs in a coroutine (i.e., it's asynchronous) and waits for new data to be emitted in `pagingData`. When it does the block of code inside `collectLatest` is run.

When `collectLatest` has run the code inside the lambda it goes back to waiting for more data to be emitted. When more data is emitted `collectLatest` runs the lambda again. And so on -- every time there's new data the lambda inside `collectLatest` is re-run.

There's two possible scenarios at start up:

1. This code is run **after** the viewmodel has fetched data and emitted it. In which case the lambda inside `collectLatest` will run immediately.
2. This code is run **before** the viewmodel has fetched data and emitted it. In which case the `collectLatest` call will **block** until data is available. But that's fine, because this is all running in a coroutine (because of the `launch`), so `collectLatest` blocking doesn't block anything else. When the data is emitted the `collectLatest` call executes the lambda, and then blocks again waiting for new data.

app/src/main/java/com/keylesspalace/tusky/components/notifications/NotificationsFragment.kt
nikclayton: This is OK for prototyping, but in the final version this probably belongs in the viewmodel.

Rule of thumb.
- Fragment displays data, can make decisions about what the UI of that data looks like (e.g., is a status shown expanded or collapsed)
- ViewModel determines what the actual data is

So if data is going to be removed from the list it should probably happen in the viewmodel.

This is not a hard and fast rule, but it does make testing easier.

app/src/main/java/com/keylesspalace/tusky/components/notifications/NotificationsFragment.kt
nikclayton: This is a good argument for doing this in the viewmodel, so that the data removal happens before the data is inserted in to the adapter (per my previous comment).

app/src/main/java/com/keylesspalace/tusky/components/notifications/NotificationsFragment.kt
```diff
@@ -434,6 +438,27 @@ class NotificationsFragment :
         }
     }
 
+    private fun removeDuplicateOlderEntries(positionStart: Int) {
+        val dataList = adapter.snapshot().items
+
+        Log.w(TAG, "found elements in adapter "+dataList.size)
```
nikclayton, line 444: If you plan to keep these in the final code use `Log.d` please (or possibly `Log.v` if there are a lot of them per update).

app/src/main/java/com/keylesspalace/tusky/components/notifications/NotificationsFragment.kt
```diff
@@ -434,6 +438,27 @@ class NotificationsFragment :
...
+
+        for (pos in dataList.size - 1 downTo  positionStart) {
```
nikclayton, line 446: Instead of `...size - 1` you can use `dataList.lastIndex`.

I think that's better because it helps to clearly communicate that this is an indexing operation.

(I need to do that in a few other places in the code, including in code I've written recently, I only noticed this the other day).

You could also do this:

```
for ((pos, notificationViewData) in dataList.reversed().withIndex()) {
    // ...
}
```

Arguably that's even better (clearly communicates it's using the index and the value, and it's going in reverse).

app/src/main/java/com/keylesspalace/tusky/components/notifications/StatusNotificationViewHolder.kt
nikclayton: Different strings for pluralisation is fine, see https://developer.android.com/guide/topics/resources/string-resource#Plurals (the app does this elsewhere).

Thinking about the UX, if this showed something like:

```
Lakoja and 7 others favourited your post
```

I would expect to be able to tap on that and see a dialog that listed everyone. Tapping on anyone in that dialog should take me to their profile page.

app/src/main/java/com/keylesspalace/tusky/components/notifications/NotificationsFragment.kt
Lakoja: There are two types of filtering here:

One happens in the view model (`pagingData.filter`) but the one here is to remove old entries that are made only obsolete by other new entries.

app/src/main/java/com/keylesspalace/tusky/components/notifications/NotificationsFragment.kt
Lakoja: I always see this in the log:

(me pulling refresh)
* submitting data (here from collectLatest)
* then 3 fetching data (from the NotificationsPagingSource)

I would have expected the fetch to happen before the submit.

(Or conversely: How do the fetched iteams after the submit end up in the list?)

nikclayton: Yep. So lets walk through what happens.

Here's a sample log I got just now from doing swipe-refresh on my notifications:

```
Submitting data to adapter
load() with Refresh for key: 131541099
--> GET https://mastodon.social/api/v1/notifications/131541099
--> GET https://mastodon.social/api/v1/notifications?max_id=131541099&limit=90
<-- 200 https://mastodon.social/api/v1/notifications/131541099 (182ms, unknown-length body)
<-- 200 https://mastodon.social/api/v1/notifications?max_id=131541099&limit=90 (1377ms, unknown-length body)
load() with Prepend for key: 131541099
--> GET https://mastodon.social/api/v1/notifications?min_id=131541099&limit=30
<-- 200 https://mastodon.social/api/v1/notifications?min_id=131541099&limit=30 (197ms, unknown-length body)
```

The swipe is caught, and the swipe listener will be called.

The listener was set earlier in the code with `binding.swipeRefreshLayout.setOnRefreshListener(this)`, so the swipe results in a call to `NotificationsFragment.onRefresh()`.

That does:

```
override fun onRefresh() {
    binding.progressBar.isVisible = false
    adapter.refresh()
}
```

Per the documentation for the adapter's `refresh()` method:

> `refresh` triggers the creation of a new `PagingData` with a new instance of `PagingSource` to represent an updated snapshot of the backing dataset.

So this creates a new `PagingData`, which means there's a new value in `viewModel.pagingData`, which means anything that's collecting values from that flow is going to be run, which means this code in the fragment runs:

```
launch {
    viewModel.pagingData.collectLatest { pagingData ->
        Log.d(TAG, "Submitting data to adapter")
        adapter.submitData(pagingData)
    }
}
```

That explains the first log message. The `PagingData` is new, so it's collected, and submitted to the adapter.

The new (empty) `PagingData` has now been submitted to the adapter.

The adapter will now try and collect pages of data from this `PagingData`. It's new, so there aren't any pages, so the adapter eventually calls the `load` method on the `PagingSource` that is wrapped by the `PagingData`.

That triggers the `load() with...` log messages.

It's only then that the API calls happen, and the `--> GET` and similar log entries come from okhttp.

Finally, the adapter receives the new pages of data, diffs them against the old pages, inserting or deleting items as appropriate.

If you haven't already used it I strongly recommend Android Studio's `Navigate` menu, especially the `Declaration or usages` feature (`Ctrl+B`). You can use to, for example, navigate from the `adapter.refresh()` call in `fun onRefresh()` and drop down through the different layers of code in the framework libraries where all this happens.

Lakoja: Ah, the important thing is that the PagingData is empty initially.
(That isn't obvious I guess. I would expect elements to be transmitted/collected and not placeholder to trigger something else. :-) )
(While debugging you cannot really see that PagingData is empty. At least not I.)

app/src/main/java/com/keylesspalace/tusky/components/notifications/NotificationsFragment.kt
Lakoja: Ok, done.

app/src/main/java/com/keylesspalace/tusky/components/notifications/NotificationsFragment.kt
Lakoja: (See below - this is only the second filtering.)

app/src/main/java/com/keylesspalace/tusky/components/notifications/NotificationsViewModel.kt
```diff
@@ -500,6 +516,41 @@ class NotificationsViewModel @Inject constructor(
...
+        } ?: return true
+
+        val highestNotificationId = trackerArray[statusId]
+
+        return if (highestNotificationId == null || isEqualOrNewer(notificationId, highestNotificationId)) {
+            trackerArray[statusId] = notificationId
+
+            true
+        } else {
+            // TODO edge case: a newer favorite has been removed: the old notification will not be added again
```
Lakoja, line 533: I wonder if this edge case is worth a fix.

(It will be correct for an app restart.)

Lakoja: Have been trying to do the heuristic now. But I didn't really find a suitable place in all that paging and flowing.

(Added a few Todo comments, though.)

Lakoja:
This (now) leads to a bogus entry being displayed in the list.
(The text "List is empty" is not present in Tusky. The first element ist still present.)

grafik

app/src/main/java/com/keylesspalace/tusky/components/notifications/NotificationsViewModel.kt
```diff
@@ -499,6 +522,41 @@ class NotificationsViewModel @Inject constructor(
...
+            false
+        }
+    }
+
+    /**
+     * NOTE this currently assumes that the ids are integers. Can that change? If it does all notifications are unequal.
+     */
+    fun isEqualOrNewer(thisId: String, idToCompare: String): Boolean {
+        try {
+            return thisId.toInt() >= idToCompare.toInt()
```
Lakoja, line 552: There are non-int ids out there...

(From the chat about problems with Hometown notifications:

`01GZ6K8XSQ6RMPXTFJPZRCXSXJ
01GZ6JS7V2K93VB54S312BC2WN
01GZ6HWAQSQMTCVTT2RXJC0ENE
01GZ6HVX4N5VE2MVATZHGXGRFJ
01GZ6HJQT3DSE358C2BQ0WN6KK`

This would still work with the frequent "length and lexicographical" comparissons I've seen in the code...

nikclayton:
FYI: Not ignoring this, but focusing on the 22 release at the moment. I'll get back to this when 22 is out in the wild...

nikclayton:
Note to self -- also need to think about how this affects grouping of Mastodon notifications when they're displayed as Android notifications.

app/src/main/java/com/keylesspalace/tusky/components/notifications/NotificationsViewModel.kt
nikclayton: `StringUtils.kt` already has `String.isLessThan`, which would the right thing to use here.

Lakoja: Ok, done.

Lakoja:
> This (now) leads to a bogus entry being displayed in the list. (The text "List is empty" is not present in Tusky. The first element is the next one and still present.)

That was due to a crash in `load()` for empty return sets.

Lakoja:
After rebase (and two fixes) this is again working and is discussable.

Issue #3435, opened 2023-03-11
Thead view does not show loading animation


Lakoja:
Currently my instance is slow in loading conversations.
So if a I tap a status in the home timeline it instantly shows that status but nothing more. After some seconds then the remaining status below are shown.

If you (initially) scroll the screen up you will see that this is possible and then at the bottom a loading animation line is shown.
The screen is by a measure of the title bar higher than it needs to be.

(For most status that is. If it has a high image or much text it is probably not too "high" because it of course extends below the fold.)

* * * *

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

nikclayton:
The intended effect is:

1. Delay the loading animation for 500ms, try and fetch the missing data. If the data is not loaded within 500ms then show the progress bar.
2. In parallel, fetch the statuses above/below the thread. This is a single API call. Again, there is a loading progress bar for this (it's https://github.com/tuskyapp/Tusky/blob/develop/app/src/main/res/layout/fragment_view_thread.xml#L26), but it's delayed for 500ms.

Are you sure it's taking "some seconds" to show the statuses without any progress indication?

Lakoja:
> The intended effect is:

Ah, now I get it.
That code (`getProgressBarJob`) is rather new.

You can see parts of the problem when opening a single status (not too long) with not replies: The resulting view can be moved some pixels upwards (to move the thread title out of view).
Only in the now visible space was there the indeterminate progress bar line.

I'm not really sure what exactly was slow as I have not yet found out how to simulate it.
If I add a delay in the async "contextCall" in `ViewThreadViewModel#loadThread` the gui hangs for that period (or is very slugish). Which surprises me because I would have guessed that the api call in there is sometimes slow.

nikclayton:
What did you try in the `contextCall` code?

I just modified it to this to add a 10 second delay, and tested:

```
val contextCall = async { delay(10000); api.statusContext(id) }
```

That doesn't delay the UI for me, but it does demonstrate that the loading animation isn't shown. Hmm

nikclayton:
As a quick fix, try this as `fragment_view_thread.xml`:

```
<?xml version="1.0" encoding="utf-8"?>
<androidx.coordinatorlayout.widget.CoordinatorLayout xmlns:android="http://schemas.android.com/apk/res/android"
    xmlns:app="http://schemas.android.com/apk/res-auto"
    android:layout_width="match_parent"
    android:layout_height="match_parent">

[...]

PR #3453, opened 2023-03-14
3435: Have loading progress visible

Lakoja:
Fixes https://github.com/tuskyapp/Tusky/issues/3435

This especially moves the progress bar to the top (where I would have expected it anyway).

I only wonder if this is the only location where we have a progress display like this?
The circular ones are all only for empty lists?

### DIFF
--- a/app/src/main/res/layout/fragment_view_thread.xml
+++ b/app/src/main/res/layout/fragment_view_thread.xml
@@ -4,34 +4,34 @@
     android:layout_width="match_parent"
     android:layout_height="match_parent">
 
-    <androidx.swiperefreshlayout.widget.SwipeRefreshLayout
-        android:id="@+id/swipeRefreshLayout"
+    <LinearLayout
         android:layout_width="match_parent"
         android:layout_height="match_parent"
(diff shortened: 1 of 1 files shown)


nikclayton:
> This especially moves the progress bar to the top (where I would have expected it anyway).

There should probably be two, to indicate that stuff is being loaded above and below the focused status.

> I only wonder if this is the only location where we have a progress display like this? The circular ones are all only for empty lists?

Pretty much. I haven't got around the migrating the others yet (where appropriate)

connyduck:
Yes, can we have two bars?

Lakoja:
> Yes, can we have two bars?

If that is the intended meaning of a line progress bar (something is loaded exactly _here_).

I would have understood it simply as a "something is loaded". Two seems a bit "blinky" to me.
In most other locations it would be a single one because only one location can be updated. Having two here would also look a bit inconsistent.

(I can do it however if this is waived. :-) )

Lakoja:
> Yes, can we have two bars?

As we discovered in the material docu (https://m3.material.io/components/progress-indicators/guidelines) there shouldn't be several (more than one) progress bars.

PR #3456, opened 2023-03-16
3159: Correctly refresh timestamps

Lakoja:
Improves https://github.com/tuskyapp/Tusky/pull/3159

Status refresh is done on all list elements.

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/components/notifications/NotificationsFragment.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/components/notifications/NotificationsFragment.kt
@@ -19,6 +19,7 @@ package com.keylesspalace.tusky.components.notifications
 
 import android.app.Dialog
 import android.content.DialogInterface
+import android.content.SharedPreferences
 import android.os.Bundle
 import android.util.Log
 import android.view.LayoutInflater
@@ -217,22 +218,14 @@ class NotificationsFragment :
             }
         })
 
-        /**
-         * Collect this flow to notify the adapter that the timestamps of the visible items have
-         * changed
-         */
+        // update post timestamps
         val updateTimestampFlow = flow {
-            while (true) { delay(60000); emit(Unit) }
-        }.onEach {
-            layoutManager.findFirstVisibleItemPosition().let { first ->
-                first == RecyclerView.NO_POSITION && return@let
-                val count = layoutManager.findLastVisibleItemPosition() - first
-                adapter.notifyItemRangeChanged(
-                    first,
-                    count,
-                    listOf(StatusBaseViewHolder.Key.KEY_CREATED)
-                )
+            while (true) {
+                delay(60000);
+                emit(Unit)
             }
+        }.onEach {
+            adapter.notifyItemRangeChanged(0, adapter.itemCount, listOf(StatusBaseViewHolder.Key.KEY_CREATED))
         }
 
         viewLifecycleOwner.lifecycleScope.launch {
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/keylesspalace/tusky/components/notifications/NotificationsFragment.kt
```diff
@@ -366,24 +359,19 @@ class NotificationsFragment :
...
-                                    count,
-                                    null
-                                )
-                            }
+                            adapter.notifyItemRangeChanged(0, adapter.itemCount, null)
 
                             if (!it.useAbsoluteTime) {
                                 updateTimestampFlow.collect()
                             }
+                            // TODO there is a way to stop this?
```
Lakoja, line 374: If I understand this code correctly (collectLatest) it should always do something on changes to the display options.

So also (only) "useAbsoluteTime" option could change and if it is set to true the flow is unnecessary.

However currently this does not seem to be a problem as there is a sort of full restart on display preferences change (you will always return to the home timeline which is reloaded).

app/src/main/java/com/keylesspalace/tusky/components/notifications/NotificationsFragment.kt
```diff
@@ -366,24 +359,19 @@ class NotificationsFragment :
                 }
 
                 // Update status display from statusDisplayOptions. If the new options request
-                // relative time display collect the flow to periodically re-bind the UI.
+                // relative time display collect the flow to periodically update the timestamp in the list gui elements.
                 launch {
                     viewModel.statusDisplayOptions
                         .collectLatest {
+                            // NOTE this this also triggered (emitted?) on resume.
```
Lakoja, line 366: But how?

app/src/main/java/com/keylesspalace/tusky/components/notifications/NotificationsFragment.kt
```diff
@@ -217,22 +218,14 @@ class NotificationsFragment :
...
-         * Collect this flow to notify the adapter that the timestamps of the visible items have
-         * changed
-         */
+        // update post timestamps
         val updateTimestampFlow = flow {
-            while (true) { delay(60000); emit(Unit) }
-        }.onEach {
-            layoutManager.findFirstVisibleItemPosition().let { first ->
-                first == RecyclerView.NO_POSITION && return@let
-                val count = layoutManager.findLastVisibleItemPosition() - first
```
Lakoja, line 229: Refresh all elements otherwise with scrolling you will easily see list elements with a wrong timestamp.

app/src/main/java/com/keylesspalace/tusky/components/notifications/NotificationsFragment.kt
```diff
@@ -366,24 +358,19 @@ class NotificationsFragment :
...
-                                    count,
-                                    null
-                                )
-                            }
+                            adapter.notifyItemRangeChanged(0, adapter.itemCount, null)
 
                             if (!it.useAbsoluteTime) {
                                 updateTimestampFlow.collect()
                             }
+                            // TODO there is a way to stop this?
```
nikclayton, line 373: It stops automatically. 

https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/collect-latest.html

"The crucial difference from collect is that when the original flow emits a new value then the action block for the previous value is cancelled."

So when there's a new value for statusdisplayoptions the previous update job (if it exists) is cancelled automatically.

app/src/main/java/com/keylesspalace/tusky/components/notifications/NotificationsFragment.kt
nikclayton: The repeatOnLifecycle call.

connyduck (approved):

Issue #3434, opened 2023-03-11
Alt text editor too small


Lakoja:
Adding an image description is a very tedious task (also) because the text field is so small and navigating in it (to read it or correct it) is finicky at best.

In this dialog text should be as important or more important (place-wise) than the image.

Would could gain some additional space with using an activity here. That would also make the back button work (which does not do anything here at the moment).

![grafik](https://user-images.githubusercontent.com/1618905/224484014-2b27a123-e93c-4a37-ae26-2b859e28aa93.png)

* * * *

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

PR #3458, opened 2023-03-17
3434: Make description dialog (text field) more usable

Lakoja:
Fixes https://github.com/tuskyapp/Tusky/issues/3434

This should be in style with the focus dialog.
Few paddings or margins to maximize usable space.
I want to highlight the text box to show where something can/needs to be done (accessible).

However I'm not quite content: less contrast for the text field.
The focus dialog has this reversed: dialog background is brighter but the actual todo zone, the image, has darker background for the unused parts.

grafik


connyduck:
I think there should just be a proper `TextInputLayout` used. It also has support for max length. Check `EditProfileActivity `to see an example.

Lakoja:
> I think there should just be a proper `TextInputLayout` used. It also has support for max length. Check `EditProfileActivity` to see an example.

It looks like this now:

grafik

app/src/main/res/layout/dialog_image_description.xml
```diff
@@ -0,0 +1,39 @@
...
+    <com.google.android.material.textfield.TextInputLayout
+        style="@style/TuskyTextInput"
+        android:id="@+id/imageDescriptionTextLayout"
+        android:layout_width="match_parent"
+        android:layout_height="0dp"
+        android:layout_weight="1"
+        android:layout_marginStart="?dialogPreferredPadding"
+        android:layout_marginTop="?dialogPreferredPadding"
+        android:layout_marginEnd="?dialogPreferredPadding"
+        app:hintEnabled="false"
```
Lakoja, line 25: This uses the hint on the actual text field.

The one on the InputLayout would be slicker but our text is much too long for that.
(It would be cut when floating - only one line shown - and it would still be cut when moved to the border when the input field is focused.)

app/src/main/res/layout/dialog_image_description.xml
```diff
@@ -0,0 +1,39 @@
...
+        app:counterEnabled="true"
+        app:counterTextColor="?android:textColorTertiary">
```
connyduck, line 27: Looks like the counter is hidden because the TextInputLayout stretches too high.

Also, it probably makes sense to set the `counterMaxLength`.

Lakoja: I ignored that for the moment:

- 1.500 is a rather high limit (to reach)
- Showing the limit (below) would take some vertical space
- The text field _will_ adhere to the limit and stop accepting input (which is some sort of queue in the right direction of what's happening)

connyduck: ok, but then please remove the counter if it isn't shown anyway

Lakoja: Done.

connyduck (approved):

Issue #3457, opened 2023-03-16
long poll options get truncated (particularly a small screen issue)


Eli-the-Bearded:
Poll web link:
https://broken.graphics/@alinanorakari/110031900991775229

Screenshot in Tusky:
![Screenshot_20230316-104945](https://user-images.githubusercontent.com/3791569/225710158-a9136a6e-3d79-4247-88ca-02668895baa6.png)

* * * *
- Tusky Version: 21.0
- Android Version: 11
- Android Device: Jelly2 (3" diagonal screen)
- Mastodon instance (if applicable):

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

PR #3459, opened 2023-03-17
3457: Allow multi-line poll options

Lakoja:
Fixes https://github.com/tuskyapp/Tusky/issues/3457

I'm not sure however that radio button or checkbox are top/start aligned with the text. I don't see an alignment at all for them. (and/or how does the text make room for them?)


app/src/main/res/layout/item_poll.xml
```diff
@@ -11,8 +11,8 @@
         android:layout_height="wrap_content"
         android:layout_marginTop="6dp"
         android:background="@drawable/poll_option_background"
+        android:maxLines="3"
```
Lakoja, line 14: There is also a lenght-limit on the poll option text (so we could get rid of this option). But maybe only for Mastodon?

connyduck:
Can you show me a screenshot please?

Lakoja:
> Can you show me a screenshot please?

grafik

connyduck (approved):
thx

app/src/main/res/layout/item_poll.xml
connyduck: I wouldn't overthink this

PR #3460, opened 2023-03-17
Fix doubled line from PR 3188 (support mastodon filter api)

Lakoja:
Remove accidentally doubled migration line.


Lakoja:
Introduced with #3188

connyduck (approved):

PR #3462, opened 2023-03-18
add lint to ci checks

connyduck:
follow up to https://github.com/tuskyapp/Tusky/pull/3363


charlag (approved):

nikclayton:
Can this be done as a separate parallel step that also blocks merges?

Otherwise this delays getting results from CI.

connyduck:
> Can this be done as a separate parallel step that also blocks merges?

Separate parallel step: No, because we only have 1 concurrency on bitrise
Blocking merge: yes (just changed it)

Tak (approved):

app/lint-baseline.xml
```diff
@@ -1764,6 +1756,39 @@
             column="25"/>
     </issue>
 
+    <issue
+        id="LogConditional"
+        message="The log call Log.v(...) should be conditional: surround with `if (Log.isLoggable(...))` or `if (BuildConfig.DEBUG) { ... }`"
```
Tak, line 1761: Oh god, what is even the point of different log levels if we have to do this anyway 😭

Issue #3469, opened 2023-03-19
Thread view: Refresh leads to a jump down


Lakoja:
This happens quite often to me and is at least a bit annoying.

When refreshing a thread view (on the top) it might jump back down to the initial opening status.

* Tap on a reply (from a longer conversation) in a timeline
* The thread view with that status in the middle is opened, the remaining status is added above and below
* I scroll up enthusiastically to start from the beginning
* With this I trigger the pull-to-refresh
* The view jumps back down

I'm not 100% certain what the proper thing to do is. But jumping down is wrong.
Probably we just want to stay at the position that was visible before the refresh.

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

Lakoja:
Seems at least related to #3158 (there was a `isInitialLoad` bool introduced).

Lakoja:
And in turn this seems to be related to #3118 where and additional progress bar (and "LoadingThread") was introduced.

PR #3471, opened 2023-03-20
3469: Do not jump in the list on the (second) refresh

Lakoja:
Fixes https://github.com/tuskyapp/Tusky/issues/3469

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/components/viewthread/ViewThreadFragment.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/components/viewthread/ViewThreadFragment.kt
@@ -178,7 +178,11 @@ class ViewThreadFragment :
                         threadProgressBar = getProgressBarJob(binding.threadProgressBar, 500)
                         threadProgressBar.start()
 
-                        adapter.submitList(listOf(uiState.statusViewDatum))
+                        if (viewModel.isInitialLoad) {
+                            adapter.submitList(listOf(uiState.statusViewDatum))
+
+                            // else this "submit one and then all on success below" will always center on the one
+                        }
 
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/keylesspalace/tusky/components/viewthread/ViewThreadFragment.kt
```diff
@@ -178,7 +178,11 @@ class ViewThreadFragment :
                         threadProgressBar = getProgressBarJob(binding.threadProgressBar, 500)
                         threadProgressBar.start()
 
-                        adapter.submitList(listOf(uiState.statusViewDatum))
+                        if (viewModel.isInitialLoad) {
+                            adapter.submitList(listOf(uiState.statusViewDatum))
+
+                            // else this "submit one and then all on success below" will always center on the one
+                        }
```
Lakoja, line 185: NOTE / question the `scrollToPositionWithOffset` below in success is normally not necessary. The "submit 1" here results in exactly that.

However I guess that is a safe-guard. This location here ("ThreadLoad") is maybe not called always or not always before?

connyduck (commented):
Wouldn't it be better to not even send the `LoadingThread` state when refreshing? 🤔

Lakoja:
> Wouldn't it be better to not even send the `LoadingThread` state when refreshing? 🤔

Hm, the second progress "bar" here, the `threadProgressBar` was introduced with #3118.

The idea is probably that the initial progress bar (round) is not suitable anymore once the starting post is displayed.

But I guess in this case (and our discussion about consolidating progresses https://github.com/tuskyapp/Tusky/pull/3474) only the recyclerview's progress bar would be sufficient until the whole conversation is loaded?

PR #3474, opened 2023-03-21
3430: Make list refresh/retry consistent

### LINKED ISSUE #3430: Wrong progress indicators with swipe/refresh in Bookmarks
Swiping to refresh in Bookmarks if they initially failed to load (e.g., poor network, or the server did not respond in time) shows both the "swipe to refresh" circular progress spinner, and the "loading content" progress spinner at the same time.

To reproduce:

1. Open Tusky
2. Enable airplane mode (to simulate poor network connection)
3. Open bookmarks. Observe the sad elephant
4. Disable airplane mode
5. Swipe down to refresh
6. Observe both progress spinners appear

@connyduck This is related to our discussion over in the Notifications PR. Which progress spinner should appear in this case? The swipe one, because the user has swiped, or the non-swipe one, because no content has been loaded yet?

* * * *
- Tusky Version: 21

Lakoja:
Touches / potentially fixes #3430

I have made the different lists more consistent:
* Everything should have a SwipeRefreshLayout
* Everything with an error screen ("message view") should have a retry button

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/ListsActivity.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/ListsActivity.kt
@@ -101,6 +101,9 @@ class ListsActivity : BaseActivity(), Injectable, HasAndroidInjector {
             DividerItemDecoration(this, DividerItemDecoration.VERTICAL)
         )
 
+        binding.swipeRefreshLayout.setOnRefreshListener { viewModel.retryLoading() }
+        binding.swipeRefreshLayout.setColorSchemeResources(R.color.tusky_blue)
+
         lifecycleScope.launch {
             viewModel.state.collect(this@ListsActivity::update)
         }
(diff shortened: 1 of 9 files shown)


connyduck:
There is also `FiltersActivity` which doesn't show an error at all currently. Other places seem fine now.

Small nitpick: I don't think full progressbar and swipe-to-refresh-spinner should be shown at the same time, but it is already better than before.

Lakoja:
> There is also `FiltersActivity` which doesn't show an error at all currently. Other places seem fine now.

I also adapted that now. That was a larger change (also introduction of "loading states").

> Small nitpick: I don't think full progressbar and swipe-to-refresh-spinner should be shown at the same time, but it is already better than before.

Yes, I didn't touch the original "two progressbars" of the issue.
We should decide which single progressbar is shown in all list load and reload cases, I think.

PR #3485, opened 2023-03-23
RFC: Try to (also) provide notifications for the list from a simple list

Lakoja:
Well more like a "request for pointers to the right concepts".

I tried to integrate notifications coming in via the fetcher to the normal notifications timeline (if open).

However I'm struggling again mightily with paging sources (and /or the data flow surrounding them).

Where would I put in "other notifications" which are not coming from the current NotificationsPagingSource?
Currently I would guess that I need a "combining source" (is that a Mediator?).


Lakoja:
From Nik (from the chat):

The current code only has a "read-only snapshot" of the data in memory.
If I wanted to also write that we would need to add our own caching. Either heavily in the database or at least in our own in-memory cache.
(https://developer.android.com/topic/libraries/architecture/paging/v3-network-db)

Issue #3492, opened 2023-03-27
Long display name doesn't display right


fin-w:
When I use a long display name on my own account, when it appears in the notifications tab (someone faved / boosted my toot) it overlaps the timestamp on the right hand side of the notification it's in. The app seems to account for the display name plus username together being too long (adding ellipsis correctly) but the threshold at which point ellipsis is added to a display name when it's too long on its own, seems to be the configured to the wrong length. If the display name is long enough it will display ellipsis, but this is when the display name is already overlapping the timestamp, and the ellipsis are also over the timestamp
![image](https://user-images.githubusercontent.com/41450706/227992868-e854c3eb-5da7-4324-91a0-bf577445b9a8.png)

ALSO, related, and therefore not added as a separate issue: The long display name is cut off at the nearest full word instead of cut off with ellipsis, within the menu that you open from the left. The display name is shown at the top of the menu next to the profile pic, and doesn't seem to ever use ellipsis and so it's not clear that it's been cut short. When the display name is only a single long word, there are still no ellipsis but the first (and only) word in the display name is shown until the app cuts it short (at 25 chars, for me, if that's useful).
![image](https://user-images.githubusercontent.com/41450706/227993296-f9fd605d-4de8-4c88-8734-81d97e9e7432.png)

* * * *
- Tusky Version: 21.0
- Android Version: 13
- Android Device: Moto G7 Power

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

PR #3495, opened 2023-03-29
3492: Correctly shorten name in drawer and notifications

Lakoja:
Fixes https://github.com/tuskyapp/Tusky/issues/3492

### DIFF
--- a/app/src/main/res/layout/item_status_notification.xml
+++ b/app/src/main/res/layout/item_status_notification.xml
@@ -24,52 +24,59 @@
         android:textSize="?attr/status_text_medium"
         tools:text="Someone favourited your status" />
 
-    <RelativeLayout
+    <LinearLayout
         android:id="@+id/status_name_bar"
+        android:orientation="horizontal"
         android:layout_width="match_parent"
         android:layout_height="wrap_content"
         android:layout_below="@+id/notification_top_text"
         android:layout_toEndOf="@+id/notification_status_avatar"
         android:paddingBottom="4dp">
 
-        <TextView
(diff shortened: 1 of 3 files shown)


app/src/main/res/layout/item_status_notification.xml
```diff
@@ -24,52 +24,59 @@
...
+                android:maxLines="1"
+                android:textColor="?android:textColorTertiary"
+                android:textSize="?attr/status_text_medium"
+                tools:text="\@Entenhausen" />
+
+        </LinearLayout>
 
         <TextView
             android:id="@+id/status_meta_info"
             android:layout_width="wrap_content"
```
Lakoja, line 71: Have a width and stay visible.

However this does only work as second element in the LinearLayout thus the wrapping additional LinearLayout above is necessary.

Lakoja:
Linter says: Nested weights are bad for performance

However that is not an error.
At least not until somebody finds another solution that works (this one here does).

connyduck:
Here is how it works with ConstraintLayout: https://github.com/tuskyapp/Tusky/commit/b67806ee0e8ea7692da31b8385e07b7d247cddba

Lakoja:
> Here is how it works with ConstraintLayout: [b67806e](https://github.com/tuskyapp/Tusky/commit/b67806ee0e8ea7692da31b8385e07b7d247cddba)

Ok, using that now, thank's.

connyduck (approved):

fin-w:
Many thanks for your effort! It's appreciated :)

Issue #3503, opened 2023-03-31
Consistent message view when swipe-to-refresh


Lakoja:
Partly a followup of #3408:

Every list with a Swipe-To-Refresh should consistently have the proper layering with a message view (for errors and empty).

Otherwise (for example) the Retry button for errors is not usable.

PR #3504, opened 2023-03-31
3503: SwipeRefreshLayout must be higher level

Lakoja:
Fixes: #3503

### DIFF
--- a/app/src/main/res/layout/fragment_timeline.xml
+++ b/app/src/main/res/layout/fragment_timeline.xml
@@ -1,23 +1,33 @@
 <?xml version="1.0" encoding="utf-8"?>
 <androidx.constraintlayout.widget.ConstraintLayout xmlns:android="http://schemas.android.com/apk/res/android"
     xmlns:app="http://schemas.android.com/apk/res-auto"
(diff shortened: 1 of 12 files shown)


app/src/main/res/layout/fragment_timeline.xml
```diff
@@ -1,23 +1,33 @@
...
-        android:visibility="gone"
-        app:layout_constrainedHeight="true"
-        app:layout_constraintBottom_toBottomOf="parent"
-        app:layout_constraintLeft_toLeftOf="parent"
-        app:layout_constraintRight_toRightOf="parent"
-        app:layout_constraintTop_toTopOf="parent"
-        tools:visibility="visible" />
+        android:layout_height="match_parent">
+
+        <FrameLayout
```
Lakoja, line 13: SwipeToRefreshLayout must more or less be top-level and wrapping (to not be before other elements or behind other elements).

However it may only have one child.

connyduck:
Makes sense, but should be changed in the tablet layout as well

connyduck (approved):

PR #3507, opened 2023-04-03
3488 improve profile list

### LINKED ISSUE #3488: Profile page: Work with lists
Followup of #3416 

When working with lists of a profile page (Add/remove from lists in the context menu) the respective dialog lets some to be desired:

- There should be a loading progress
- There should be a button for managing (adding) lists

Lakoja:
Fixes #3488 

Working with lists from a profile page and in the normal "lists view" from the drawer now use the same fragment view code.

(also) RFC regarding joining different list lists

grafik


app/src/main/java/com/keylesspalace/tusky/components/account/list/ListsForAccountFragment.kt
```diff
@@ -117,7 +123,7 @@ class ListsForAccountFragment : DialogFragment(), Injectable {
             }
         }
 
-        viewLifecycleOwner.lifecycleScope.launch {
+        lifecycleScope.launch {
```
Lakoja, line 126: Not quite sure what I'm doing with lifecycles here.

With the usage of onCreateDialog() however there seems to never be a point during creation (like "onCreated") where the view is present, so viewLifecycleOwner is always null.

app/src/main/java/com/keylesspalace/tusky/TabPreferenceActivity.kt
```diff
@@ -317,6 +317,7 @@ class TabPreferenceActivity : BaseActivity(), Injectable, ItemInteractionListene
                     showProgressBarJob.cancel()
                     adapter.addAll(lists)
                     if (lists.isEmpty()) {
+                        // TODO do like in ListsForAccountFragment (full binding in dialog with message view)
```
Lakoja, line 320: Mostly the draft is about that I have the feeling that all list lists should be joined more or less.

For example the list selection in the tab preferences is basically the same as here (dialog which loads a list list and you can tap an entry). Also visually all list lists should look the same.

So a general lists fragment (?) could be used in at least 3 locations (third being the ListActivity itself).

app/src/main/java/com/keylesspalace/tusky/components/account/list/ListsForAccountFragment.kt
```diff
@@ -60,46 +67,45 @@ class ListsForAccountFragment : DialogFragment(), Injectable {
         viewModel.setup(requireArguments().getString(ARG_ACCOUNT_ID)!!)
     }
 
-    override fun onStart() {
-        super.onStart()
-        dialog?.apply {
-            window?.setLayout(
-                LinearLayout.LayoutParams.MATCH_PARENT,
```
Lakoja, line 67: Do we need this dialog to be full screen?
(In that case simply using a RelativeLayout is enough and also needed for the alert dialog buttons to be at the bottom.)

Having it small has the advantage that you can also tap anywhere outside to cancel.

(What happens anyway if the list is too long for full screen? There is no scrolling yet here?)

app/src/main/java/com/keylesspalace/tusky/components/account/list/ListsForAccountFragment.kt
```diff
@@ -49,7 +53,11 @@ class ListsForAccountFragment : DialogFragment(), Injectable {
     lateinit var viewModelFactory: ViewModelFactory
 
     private val viewModel: ListsForAccountViewModel by viewModels { viewModelFactory }
-    private val binding by viewBinding(FragmentListsForAccountBinding::bind)
+
+    private var _binding: FragmentListsForAccountBinding? = null
+
+    // This property is only valid between onCreateDialog and onDestroyView
+    private val binding get() = _binding!!
```
connyduck, line 60: Move the binding completely into `onCreateDialog`. It is not needed anywhere else.

app/src/main/java/com/keylesspalace/tusky/components/account/list/ListsForAccountFragment.kt
connyduck: If it was fullscreen it doesn't need to be a dialog, could just be a regular frament. I think it is better if it looks like a dialog, and yes of course it should scroll.

app/src/main/java/com/keylesspalace/tusky/components/account/list/ListsForAccountFragment.kt
connyduck: If that is the case, I think it is ok to just use `lifecycleScope.launch`

app/src/main/res/layout/fragment_lists_for_account.xml
```diff
@@ -27,30 +26,18 @@
...
-    <Button
-        android:id="@+id/doneButton"
-        style="@style/TuskyButton.TextButton"
-        android:layout_width="wrap_content"
-        android:layout_height="wrap_content"
-        android:layout_gravity="end"
-        android:text="@string/button_done" />
-
-</LinearLayout>
+</LinearLayout >
```
connyduck, line 43: This looks like it can be done with a single `FrameLayout`? Nested `LinearLayout`/`ConstraintLayout` is definitely too much.

app/src/main/java/com/keylesspalace/tusky/TabPreferenceActivity.kt
connyduck: Yes consolidating all those similar things would be really cool

app/src/main/java/com/keylesspalace/tusky/components/account/list/ListsForAccountFragment.kt
```diff
@@ -1,4 +1,4 @@
-/* Copyright 2022 kyori19
+/* Copyright Tusky Contributors
```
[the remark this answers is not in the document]
connyduck: whatever

app/src/main/java/com/keylesspalace/tusky/components/account/list/ListsForAccountFragment.kt
Lakoja: load() does still use it.

(That ties in with my observation / todo above for using a LoadState above.)

app/src/main/res/layout/fragment_lists_for_account.xml
Lakoja: Done

app/src/main/java/com/keylesspalace/tusky/components/account/list/ListSelectionFragment.kt
```diff
@@ -16,90 +16,116 @@
...
+        binding.listsView.adapter = adapter
+
+        val dialogBuilder = AlertDialog.Builder(context)
+            .setView(binding.root)
+            .setTitle(R.string.select_list_title)
+            .setNeutralButton(R.string.select_list_manage) { _, _ ->
+                val listIntent = Intent(context, ListsActivity::class.java)
+                startActivity(listIntent)
+            }
+            .setNegativeButton(if (accountId != null) R.string.button_done else android.R.string.cancel, null)
```
Lakoja, line 98: Generally this fragment now serves two leaders (simple list selection or list member management). This is decided with account id being null or not.

However in most locations this is a bit cumbersome or hard to follow.

Also: I wasn't able to use this fragment in the ListsActivity (which would make three functions).
Mostly because I don't see how a simple list list could be extended to _also_ support membership info sometimes.
(Would need two wrapped flows for the load state??)

So I would be still open to suggestion how this could be solved more architectually clean and easier to understand.

app/src/main/java/com/keylesspalace/tusky/components/account/list/ListSelectionFragment.kt
```diff
@@ -16,90 +16,116 @@
...
+    private val binding get() = _binding!!
 
     private val adapter = Adapter()
 
+    private var selectListener: ListSelectionListener? =  null
+    private var accountId: String? = null
+
+    override fun onAttach(context: Context) {
+        super.onAttach(context)
+        selectListener = context as? ListSelectionListener
```
Lakoja, line 76: I like this implicit configurability: If the caller implements the interface it will be informed of a selection - otherwise nothing happens.

app/src/main/java/com/keylesspalace/tusky/TabPreferenceActivity.kt
Lakoja: (but also see "architectually clean" question below)

Lakoja:
Now linter complains: R.string.select_list_empty appears to be unused

Which is technically correct but not new in this PR.

(And how do you remove translation keys?)

nikclayton:
Reviewing this sent me down a rabbit hole of looking at all the problems with the current UX.

I wrote up some findings and a proposed solution at https://github.com/tuskyapp/Tusky/issues/3488#issuecomment-1666927377. WDYT?

Lakoja:
> Now linter complains: R.string.select_list_empty appears to be unused

Removed two ununused strings now.

Lakoja:
> Reviewing this sent me down a rabbit hole of looking at all the problems with the current UX.
> 
> I wrote up some findings and a proposed solution at [#3488 (comment)](https://github.com/tuskyapp/Tusky/issues/3488#issuecomment-1666927377). WDYT?

Sound like a couple of followup issues. :-)

So maybe this is an answer to my "architecture" questions here: My current proposal is good enough for a medium issue and is an improvement. All else may come in future issues/PRs.

Tak (approved):
Looks ok to me - @connyduck do your unresolved comments still apply?

app/src/main/java/com/keylesspalace/tusky/TabPreferenceActivity.kt
```diff
@@ -273,69 +268,24 @@ class TabPreferenceActivity : BaseActivity(), Injectable, ItemInteractionListene
...
+        listSelectDialog = null
+
+        val newTab = createTabDataFromId(LIST, listOf(list.id, list.title))
+        currentTabs.add(newTab)
+        currentTabsAdapter.notifyItemInserted(currentTabs.size - 1)
+        updateAvailableTabs()
+        saveTabs()
     }
 
     private fun getProgressBarJob(progressView: View, delayMs: Long) = this.lifecycleScope.launch(
```
connyduck, line 291: function is now unused

connyduck (approved):
The unused method should be removed, but overall its good

app/src/main/java/com/keylesspalace/tusky/TabPreferenceActivity.kt
Lakoja: removed

connyduck (approved):

PR #3513, opened 2023-04-05
Do not crash on empty relationship response

Lakoja:
Found while looking at Google crash reports.

I checked a few other "[0]" in the code. Most are guarded by a "isNotEmpty" but some are not.
For example in `searchAutocompleteSuggestions`.

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/components/account/AccountViewModel.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/components/account/AccountViewModel.kt
@@ -85,7 +85,7 @@ class AccountViewModel @Inject constructor(
                 mastodonApi.relationships(listOf(accountId))
                     .fold(
                         { relationships ->
-                            relationshipData.postValue(Success(relationships[0]))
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/keylesspalace/tusky/components/account/AccountViewModel.kt
```diff
@@ -85,7 +85,7 @@ class AccountViewModel @Inject constructor(
                 mastodonApi.relationships(listOf(accountId))
                     .fold(
                         { relationships ->
-                            relationshipData.postValue(Success(relationships[0]))
+                            relationshipData.postValue(if (relationships.isNotEmpty()) Success(relationships[0]) else Error())
```
Lakoja, line 88: It was similar to this in 2020 but the size check was then removed by a sort-of refactoring.

connyduck:
Do we know when this will be null? Opening the profile of a deleted account or something like that?

connyduck (approved):

PR #3517, opened 2023-04-07
3516 provied occurrence log

### LINKED ISSUE #3516: Track system events in debug version
There should be - at least in test versions - the possibility to see a list of recent important application events.
Especially api calls and crashes.

Lakoja:
Fixes #3516 

Draft: Needs more work and discussion

### DIFF
--- a/app/src/main/res/layout/item_occurrence.xml
+++ b/app/src/main/res/layout/item_occurrence.xml
@@ -1,0 +1,39 @@
+<?xml version="1.0" encoding="utf-8"?>
+<LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:tools="http://schemas.android.com/tools"
+    android:layout_width="match_parent"
+    android:layout_height="wrap_content"
+    android:gravity="center_vertical"
+    android:orientation="horizontal">
+
+    <TextView
+        android:id="@+id/code"
+        android:layout_width="60dp"
+        android:layout_height="match_parent"
+        android:lines="1"
+        android:textSize="?attr/status_text_large"
+        tools:text="500" />
+
+    <LinearLayout
+        android:layout_width="0dp"
+        android:layout_weight="1"
+        android:layout_height="wrap_content"
+        android:gravity="center_vertical"
+        android:orientation="vertical">
+
+        <TextView
+            android:id="@+id/what"
+            android:layout_width="match_parent"
+            android:layout_height="wrap_content"
+            android:textSize="?attr/status_text_medium"
+            android:textStyle="bold"
+            tools:text="Example crash" />
+
+        <TextView
+            android:id="@+id/whenDate"
+            android:layout_width="match_parent"
+            android:layout_height="wrap_content"
+            android:textSize="?attr/status_text_medium"
+            tools:text="2048:04:01" />
+    </LinearLayout>
+</LinearLayout>
--- a/app/src/main/res/layout/item_occurrence.xml
+++ b/app/src/main/res/layout/item_occurrence.xml
@@ -3,37 +3,59 @@
     xmlns:tools="http://schemas.android.com/tools"
     android:layout_width="match_parent"
     android:layout_height="wrap_content"
-    android:gravity="center_vertical"
-    android:orientation="horizontal">
+    android:orientation="vertical"
+    android:layout_margin="6dp">
 
     <TextView
-        android:id="@+id/code"
-        android:layout_width="60dp"
-        android:layout_height="match_parent"
-        android:lines="1"
-        android:textSize="?attr/status_text_large"
-        tools:text="500" />
+        android:id="@+id/what"
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content"
+        android:textColor="?android:textColorPrimary"
+        android:textSize="?attr/status_text_medium"
+        android:textStyle="bold"
+        android:layout_marginBottom="4dp"
+        tools:text="Example crash" />
 
     <LinearLayout
-        android:layout_width="0dp"
-        android:layout_weight="1"
+        android:layout_width="match_parent"
         android:layout_height="wrap_content"
-        android:gravity="center_vertical"
-        android:orientation="vertical">
+        android:orientation="horizontal">
 
         <TextView
-            android:id="@+id/what"
-            android:layout_width="match_parent"
-            android:layout_height="wrap_content"
-            android:textSize="?attr/status_text_medium"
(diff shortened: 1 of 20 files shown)


app/src/main/java/com/keylesspalace/tusky/components/occurrence/OccurrenceActivity.kt
```diff
@@ -0,0 +1,231 @@
...
+import javax.inject.Inject
+
+class OccurrenceActivity : BaseActivity(), Injectable, HasAndroidInjector {
+
+    @Inject
+    lateinit var viewModelFactory: ViewModelFactory
+
+    // TODO what's this?
+    @Inject
+    lateinit var dispatchingAndroidInjector: DispatchingAndroidInjector<Any>
```
connyduck, line 57: That is needed for activities that contain fragments for injection to work

app/src/main/java/com/keylesspalace/tusky/components/occurrence/OccurrenceActivity.kt
```diff
@@ -0,0 +1,231 @@
...
+import com.keylesspalace.tusky.util.getRelativeTimeSpanString
+import com.keylesspalace.tusky.util.viewBinding
+import com.keylesspalace.tusky.util.visible
+import dagger.android.DispatchingAndroidInjector
+import dagger.android.HasAndroidInjector
+import kotlinx.coroutines.runBlocking
+import java.text.DateFormat
+import javax.inject.Inject
+
+class OccurrenceActivity : BaseActivity(), Injectable, HasAndroidInjector {
```
connyduck, line 50: I'd call it DebugActivity/DebugLogActivity or something like that

app/src/main/java/com/keylesspalace/tusky/components/occurrence/OccurrenceActivity.kt
```diff
@@ -0,0 +1,231 @@
...
+            if (occurrence.finishedAt != null) {
+                durationMs = occurrence.finishedAt.time - occurrence.startedAt.time
+                duration = getDurationStringAllowMillis(currentLocale, durationMs)
+            }
+            holder.binding.duration.text = duration
+            holder.binding.duration.setTextColor(
+                if (durationMs >= 1000) {
+                    Color.MAGENTA
+                } else if (durationMs >= 400) {
+                    // TODO these colors are a bit of a problem: e. g. on light mode yellow is hardly visible and green is difficult
```
connyduck, line 200: make them resources and use different ones for light and dark mode

app/src/main/res/layout/item_occurrence.xml
```diff
@@ -0,0 +1,71 @@
...
+    <TextView
+        android:id="@+id/trace"
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content"
+        android:textColor="?android:textColorPrimary"
+        android:textSize="?attr/status_text_medium"
+        android:layout_marginTop="4dp"
+        android:visibility="gone"
+        tools:text="Trace line 1" />
+</LinearLayout>
```
connyduck, line 71: I'd always use ConstraintLayout for these kind of layout.
It goes something like this:
- Just need to center or frame something -> FrameLayout
- Just need some items below/next to each other -> LinearLayout
- Everything else -> ContaintLayout because it avoids nesting of other layouts and should be faster

app/src/main/java/com/keylesspalace/tusky/components/occurrence/OccurrenceActivity.kt
Lakoja: Hm, it's not used for debugging.

It should/could have the same name as in the ui (and there "debug" would be somewhat misleading).

It shows (hopefully) significant events in the app life.

app/src/main/res/layout/item_occurrence.xml
Lakoja: Ok, I will try.

(My reasoning would go something like this: a "line" could be considered a group and thus can be grouped. If for example the top level is a vertical linear layout I would know that its contents are "line-ish" things.)

app/src/main/java/com/keylesspalace/tusky/components/occurrence/OccurrenceEntity.kt
```diff
@@ -0,0 +1,73 @@
...
+    val what: String,
+    val startedAt: Date, // TODO or use LocalDateTime (or Long)?
+    val finishedAt: Date? = null,
+    val code: Int? = null,
+    val callTrace: String,
+) {
+    companion object {
+        fun reduceTrace(stackTrace: Array<StackTraceElement>): String {
+            // TODO conditions/transforms here are a bit arbitrary...
+            // TODO we could maybe record more (full stack?) in the db and only reduce it for display?
```
Lakoja, line 39: These two are the most interesting / to be discussed todos.

(What exactly to record and what to show.)

app/src/main/java/com/keylesspalace/tusky/components/occurrence/OccurrenceRepository.kt
```diff
@@ -0,0 +1,108 @@
...
+
+        val occurrence = OccurrenceEntity(
+            accountId = accountManager.activeAccount?.id,
+            type = OccurrenceEntity.Type.APICALL,
+            what = what,
+            startedAt = Calendar.getInstance().time,
+            callTrace = ""
+//            callTrace = OccurrenceEntity.reduceTrace(Throwable().stackTrace),
+        )
+        // TODO all stack traces here have no hint where they might have originated (always ThreadPool)
```
Lakoja, line 38: This _should_ have a solution.
Otherwise we only have "something happened but you don't know where/why".

app/src/main/res/layout/item_occurrence.xml
Lakoja: Using ConstraintLayout now.

Cursed two more hours and still have no idea how ConstraintLayout works (with respect to "layout_weight"). By chance I found that width="0dp" does it...

app/src/main/java/com/keylesspalace/tusky/components/occurrence/OccurrenceActivity.kt
Lakoja: Done.

Lakoja:
It would probably also be a good idea to record error log(cat) entries.

However I found no possibility to transparently intercept calls to Log.e().

So if we want that it would need a rewrite of logging.

PR #3521, opened 2023-04-10
Disable wildcard imports for Java and Kotlin files

nikclayton:
Out-of-the-box Android Studio will perform wildcard imports if the import list is >= 3. This is a lint failure.

Adjust the default to 999 so that, in practice, wildcard imports never happen.


connyduck (approved):
nice

PR #3526, opened 2023-04-12
3525: Properly hide list; update list in any case (empty)

### LINKED ISSUE #3525: Delete last filter: list is not empty
Unter Account settings > Filters
If you delete the last filter that filter is still shown in the list.

* * * *
- Tusky Version: Current develop (16e1a)

Lakoja:
Fixes #3525

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/components/filters/FiltersActivity.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/components/filters/FiltersActivity.kt
@@ -87,7 +87,7 @@ class FiltersActivity : BaseActivity(), FiltersListener {
                             binding.filtersList.show()
                         }
 
-                        if (binding.filtersList.adapter == null || (binding.filtersList.adapter as FiltersAdapter).filters != state.filters) {
+                        if ((binding.filtersList.adapter as? FiltersAdapter)?.filters != state.filters) {
                             binding.filtersList.adapter = FiltersAdapter(this@FiltersActivity, state.filters)
                         }
                     }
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/keylesspalace/tusky/components/filters/FiltersActivity.kt
```diff
@@ -81,9 +84,12 @@ class FiltersActivity : BaseActivity(), FiltersListener {
                             binding.messageView.show()
                         } else {
                             binding.messageView.hide()
-                            binding.filtersList.adapter = FiltersAdapter(this@FiltersActivity, state.filters)
```
Lakoja, line 84: Must be done in any case: Also an empty list is valid.

app/src/main/java/com/keylesspalace/tusky/components/filters/FiltersActivity.kt
```diff
@@ -81,9 +84,12 @@ class FiltersActivity : BaseActivity(), FiltersListener {
...
                             binding.filtersList.show()
                         }
+
+                        if (binding.filtersList.adapter == null || (binding.filtersList.adapter as FiltersAdapter).filters != state.filters) {
```
Tak, line 90: Why not just `?.`?

Tak (approved):

app/src/main/java/com/keylesspalace/tusky/components/filters/FiltersActivity.kt
Lakoja: How exactly would that look?

Tak: Sorry, missed this reply in my notifications.
I'm thinking like `(binding.filtersList.adapter as FiltersAdapter)?.filters != state.filters` - but it's just a minor nitpick

PR #3540, opened 2023-04-16
Search: Do not lose changes to the edit field when switching tabs

### LINKED ISSUE #3527: Search: search text is lost when switching tabs (before search)
The search view has 3 tabs (Posts, Accounts, Hashtags) for which you can search.
The search field above is emptied when you switch between these.
Or more precisely: it always shows the last search in all fields (also empty before the first search).

Why do we have three types (the web ui doesn't have it?)?
It's unusual to burden the user with this choice - at least before the search.

If there have to be three types it is not helpful to have three different search fields (or the one being emptied?).

 At least I normally decide on the type after I entered my search text...
(Especially as these tabs are below the field.)

Lakoja:
Fixes #3527 

What is then more prominent after this: There is no info on what has been searched.

There should probably be another text field below the search field with "Search results for 'search term':"


jsoref:
Please change the description from `loose` to `lose`

nikclayton (approved):

PR #3541, opened 2023-04-16
Migrate to AGP 8.0.2 / Android Studio Flamingo / Java 17

nikclayton:
- Update AGP in version catalog to 8.0.2
- Set Java version to 17
- Enable non-final resource IDs


connyduck:
needs changes to bitrise.yml as well https://devcenter.bitrise.io/en/infrastructure/build-stacks/managing-java-versions.html

nikclayton:
Marking back as draft at the moment. I've had a number of instances where moving between different Tusky branches with this enabled has got Gradle or some other part of the build system in to a weird state where everything needs to be stopped and all caches invalidated. That, plus occasional errors like this when the affected code hasn't changed:

```
WARNING: An illegal reflective access operation has occurred
WARNING: Illegal reflective access by org.jetbrains.kotlin.kapt3.util.ModuleManipulationUtilsKt (file:/C:/Users/nik/.gradle/caches/modules-2/files-2.1/org.jetbrains.kotlin/kotlin-annotation-processing-gradle/1.8.20/52361dad4f2c22f4e39c173001b1c4c2b4e61cb6/kotlin-annotation-processing-gradle-1.8.20.jar) to constructor com.sun.tools.javac.util.Context()
WARNING: Please consider reporting this to the maintainers of org.jetbrains.kotlin.kapt3.util.ModuleManipulationUtilsKt
WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
WARNING: All illegal access operations will be denied in a future release

> Task :app:kaptBlueDebugKotlin
java.io.FileNotFoundException: C:\Windows\sqlite-3.36.0-27da3296-d7b4-47bd-a213-c3d4befb54c1-sqlitejdbc.dll.lck (Access is denied)
```

(note the file path in the last error -- something's gone wrong if it's trying to write there instead of the temporary directories)

So I don't think this is ready for prime time yet.

nikclayton:
Those `java.io.FileNotFoundException` errors I mentioned earlier were resolved in the PR that migrated Room to use ksp instead of kapt, so I'm trying this again. So far so good.

connyduck (approved):

gradle/libs.versions.toml
```diff
@@ -1,5 +1,5 @@
 [versions]
-agp = "7.4.2"
+agp = "8.0.2"
```
Goooler, line 2:
```suggestion
agp = "8.1.0"
```

PR #3552, opened 2023-04-21
Refresh timelines when filters are added/edited/removed

### LINKED ISSUE #3546: Create v2 filter doesn't invalidate relevant timeline
Editing a v1 filter (i.e., client side) trigger a refresh of the relevant timeline, because the filters are stored as preferences and the preference change is noted.

Saving a v2 filter (server side) does not, because there's no event that the filter has been saved. The user has to pull-to-refresh before the timeline updates.

* * * *
- Tusky Version: 21

Tak:
Fixes #3546

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/components/timeline/viewmodel/TimelineViewModel.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/components/timeline/viewmodel/TimelineViewModel.kt
@@ -27,6 +27,7 @@ import com.keylesspalace.tusky.appstore.BlockEvent
 import com.keylesspalace.tusky.appstore.DomainMuteEvent
 import com.keylesspalace.tusky.appstore.Event
 import com.keylesspalace.tusky.appstore.EventHub
+import com.keylesspalace.tusky.appstore.FilterUpdatedEvent
 import com.keylesspalace.tusky.appstore.MuteConversationEvent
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/keylesspalace/tusky/components/timeline/viewmodel/TimelineViewModel.kt
```diff
@@ -271,6 +272,11 @@ abstract class TimelineViewModel(
             is PreferenceChangedEvent -> {
                 onPreferenceChanged(event.preferenceKey)
             }
+            is FilterUpdatedEvent -> {
+                if (filterContextMatchesKind(kind, event.filterContext)) {
+                    fullReload()
```
Tak, line 277: `invalidate()` isn't enough here because the application of the filters occurs on the server

Tak:
I'm pretty sure that filters are not applied to notifications?

nikclayton:
> I'm pretty sure that filters are not applied to notifications?

Filters can be applied to notifications, per https://docs.joinmastodon.org/methods/filters/#create says (emphasis mine):

> ...
> context[]
>     required, Array of String. Where the filter should be applied. Specify at least one of home, **notifications**, public, thread, account.
> ...

Lakoja (approved):

Issue #3545, opened 2023-04-18
Can't delete final filter on screen (if list one screen or more)


nikclayton:
If you have one or more screens worth of filters configured for an account you can't delete the last one, because the FAB obscures the 'X' delete icon at the right of the list.

* * * *
- Tusky Version: 21

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

PR #3553, opened 2023-04-21
Add delete button to edit filter activity

Tak:
Adds workaround for #3545


app/src/main/java/com/keylesspalace/tusky/components/filters/EditFilterActivity.kt
```diff
@@ -256,6 +262,32 @@ class EditFilterActivity : BaseActivity() {
         }
     }
 
+    private fun deleteFilter() {
+        originalFilter?.let { filter ->
```
Lakoja, line 266: Can this be flattened a bit?

5 or 6 levels seem a bit much.

(A more linear call graph maybe? No fold for example? Could also reduce the error case to one.)

nikclayton (approved):

PR #3557, opened 2023-04-23
Disable lint checks for unused resource IDs

nikclayton:
The check doesn't catch some instances where resources are used through viewbinding, and has too many false positives to be useful.


nikclayton:
@Tak -- the diff to lint-baseline.xml is mechanically generated, so doesn't need a review.

Tak (approved):

Issue #3550, opened 2023-04-20
Missing padding in short notifications


connyduck:
On short notifications, there is space missing between the divider and the avatar.

![Screenshot_20230420-201353](https://user-images.githubusercontent.com/10157047/233453126-e083437e-4e23-4673-a4dc-a4b51e7a2bbe.png)

* * * *
- Tusky Version: Tusky Test 21.0-test-16e1ab20
- Android Version: 12
- Android Device: Fairphone 4
- Mastodon instance (if applicable): seen on chaos.social and mastodon.social

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

PR #3559, opened 2023-04-23
Add padding below status content

nikclayton:
Fixes https://github.com/tuskyapp/Tusky/issues/3550


connyduck (changes requested):
Now it is too much padding on longer notifications 

Screenshot_20230424_092205

nikclayton:
Should be good now.

PR #3571, opened 2023-04-25
#3570: Also provide a "direct" back with the hardware button

### LINKED ISSUE #3570: Using search: cannot directly go back
If you actually search something in the search view then press (phone button) "back", nothing happens.
Only the second back press goes back.

Lakoja:
Fixes #3570

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -130,7 +130,7 @@
             android:theme="@style/Base.Theme.AppCompat" />
         <activity
             android:name=".components.search.SearchActivity"
-            android:launchMode="singleTop"
+            android:launchMode="singleTask"
             android:exported="false">
             <intent-filter>
                 <action android:name="android.intent.action.SEARCH" />


app/src/main/AndroidManifest.xml
```diff
@@ -130,7 +130,7 @@
             android:theme="@style/Base.Theme.AppCompat" />
         <activity
             android:name=".components.search.SearchActivity"
-            android:launchMode="singleTop"
+            android:launchMode="singleTask"
```
Lakoja, line 133: I'm not 100% sure this will not introduce problems. The documentation says "Meanwhile all of the other activities on top of it are destroyed."

However I _think_ that singleTop would have had the same problem.

Also I didn't see a case where you can get to the search Activity without it being on top of the stack.

Lakoja: I suspect it was for the same reason I try to fix here: Once a search is executed the intent filter here below starts the SearchActivity again - only this time with a search term.

This seems to be there from the beginning.

Tak (approved):

Issue #3562, opened 2023-04-24
Image zoom does not work consistently


Lakoja:
This happens for single images:

If I do this: tap image in status, try pinch zoom,
this mostly only works for a few pixels - then zooming is stopped. It works better on the second attempt.
If I then use a single finger to move the viewport around this works at most for one time left to right. Then it again stops.
You often see the blu-ish page changing indicator to the left or right light up.

For several images it often and mostly undesired (when trying to zoom) switches to the next image.

Partly related #2297 (but much more of a problem).

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

Lakoja:
Technical:

Actually if I disable user input for the view pager the above problems mostly go away.
So it seems to be the ViewPager2 "grabbing" the input when it feels it should page change and then the touch is cancelled (and PhotoViewAttacher stops).

Lakoja:
Out in the wild as well:

https://universeodon.com/@pixelpusher220/110232778207441656

Lakoja:
The project (https://github.com/Baseflow/PhotoView) also seems to be unmaintained (unchanged for two years; most issues are unanswered).

PR #3572, opened 2023-04-25
#3562: Only allow the view pager if it makes sense

Lakoja:
Concerns #3562

Disable the view pager (gesture input) when an image is zoomed.

This is only a draft (or RFC): The enabling and disabling of the view pager is quite hack-ish.
It works better with this but not perfect on my device.
(What happens if for example the behavior/contract of "isUserInputEnabled changes or behaves slightly different on different devices?)

Especially there is and always will be a conflict between 2 to 3 parties here:
- The view pager wants to switch pages (and also highlight what it will do starting with very small finger movements)
- The photo view wants to pinch zoom and move the zoomed image around; also a negative pinch zoom might dismiss it
- We want to add the dismiss feature for swiping down

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/ViewMediaActivity.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/ViewMediaActivity.kt
@@ -82,6 +82,7 @@ class ViewMediaActivity : BaseActivity(), ViewImageFragment.PhotoActionsListener
     private var attachments: ArrayList<AttachmentViewData>? = null
     private val toolbarVisibilityListeners = mutableListOf<ToolbarVisibilityListener>()
     private var imageUrl: String? = null
+    private var hasMultiple: Boolean = false
 
     fun addToolbarVisibilityListener(listener: ToolbarVisibilityListener): Function0<Boolean> {
         this.toolbarVisibilityListeners.add(listener)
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/keylesspalace/tusky/ViewMediaActivity.kt
```diff
@@ -113,13 +114,16 @@ class ViewMediaActivity : BaseActivity(), ViewImageFragment.PhotoActionsListener
...
+        hasMultiple = attachments != null && attachments!!.size > 1
+
         binding.viewPager.adapter = adapter
         binding.viewPager.setCurrentItem(initialPosition, false)
         binding.viewPager.registerOnPageChangeCallback(object : ViewPager2.OnPageChangeCallback() {
             override fun onPageSelected(position: Int) {
                 binding.toolbar.title = getPageTitle(position)
             }
         })
+        binding.viewPager.isUserInputEnabled = hasMultiple
```
Lakoja, line 126: At least for a single image this is a 100% improvement: have no (useless) view paging there

Lakoja:
There is some code in com.github.chrisbanes.photoview.OnGestureListener#onDrag (in PhotoViewAttacher) that seems to try to deal with this - according to the comment "... parent ... allow touch event ... edges ...".

Also there is an old PR that wants to improve some of it:
https://github.com/Baseflow/PhotoView/pull/742

Lakoja:
> Are you still looking in to this?

Not exactly. It was merely a conversation starter/research result.

However the change to only enable the pager if there are at least two images (the first three change lines here) would be an easy and problem-free 30% improvement.

nikclayton:
@Lakoja -- I think this can be closed now that https://github.com/tuskyapp/Tusky/pull/3894 is merged. Is that right?

Issue #3578, opened 2023-04-26
22 beta: Loading notifications slow in Friendica?


kr428:
Compared to the latest stable version, 22 beta and nightly are considerably slower when fetching timelines on Friendica, especially notifications. Known problem? Any way to debug this?

connyduck:
There have been changes to the notification view, so this is definitely possible 🤔 
Since you specified Friendica, does that mean other servers are fine?

nikclayton:
It's probably because the number of notifications loaded to populate the initial list has increased. I'll tune this down before the next beta release.

kr428:
Seems quite a bit faster on a Mastodon instance. Not sure how that makes a difference though.

nikclayton:
> Seems quite a bit faster on a Mastodon instance. Not sure how that makes a difference though.

Tusky makes the same API calls in both cases, so one or both of:

1. The Mastodon instance is running on faster hardware or has a faster network connection, so can respond faster.
2. The Mastodon code for computing the API response and returning it is faster than the Friendica equivalent

I'm not very familiar with either of those codebases, so I don't know which one of those is correct.

kr428:
Hmmm that's odd then, because on earlier versions (tusky 21), difference between the two systems (same instances for both Mastodon and Friendica) weren't that noticeable. Friendica's always a tad slower for my account it seems, but so far it never bothered. But another thing I noticed: For some things (like, having new followers), I see the 22 beta displaying more details on for the user account / profile. Could this cause issues too? Does that make additional requests?

nikclayton:
> Hmmm that's odd then, because on earlier versions (tusky 21), difference between the two systems (same instances for both Mastodon and Friendica) weren't that noticeable. Friendica's always a tad slower for my account it seems, but so far it never bothered. But another thing I noticed: For some things (like, having new followers), I see the 22 beta displaying more details on for the user account / profile. Could this cause issues too? Does that make additional requests?

That feature doesn't make any new requests. Tusky always received the information in the past, the difference is that now it displays it as well.

I've just compared the previous notifications code and the new code again.

In both cases the old and the new code requested 30 or more notifications. 30 is the limit imposed by the Mastodon API (see the description of the `limit` parameter at https://docs.joinmastodon.org/methods/notifications/#query-parameters), so the server shouldn't be having to do any more work per-request.

The new code does also fetch a page of notifications immediately before your reading position and immediately after your reading position -- so it's making three notifications requests where the previous code made one.

[...]

PR #3579, opened 2023-04-26
Set Notifications page size to 15

nikclayton:
This should balance loading enough notifications to allow the user to immediately start scrolling through them, without loading so many notifications that the server is slow to respond.

Fixes https://github.com/tuskyapp/Tusky/issues/3578

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/components/notifications/NotificationsRepository.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/components/notifications/NotificationsRepository.kt
@@ -71,6 +71,6 @@ class NotificationsRepository @Inject constructor(
 
     companion object {
         private const val TAG = "NotificationsRepository"
-        private const val PAGE_SIZE = 30
+        private const val PAGE_SIZE = 15
     }
 }


Lakoja:
Tried this on my instance:

- There 6 api requests when switching from the home timeline to notifications (3 for notifications, 2 for the home timeline* and 1 for filters)
- For a refresh there are 3 (only for notifications)
- One of these always takes rather long (1-2s for ?max_id=409295&limit=90&exclude_types%5B%5D=follow_request)
- This does not change noticeably with page size 15 instead of 30

* ViewPager (?) always does strange things when switching

Lakoja:
> > ```
> > * This does not change noticeably with page size 15 instead of 30
> > ```
> 
> Are you sure? That `limit=90` should change to `limit=45` (i.e., 3x15 instead of 3x30).
> 
> Note that in either case this should be clamped to a maximum of 30 by the Mastodon API (see the description of the `limit` parameter at https://docs.joinmastodon.org/methods/notifications/#query-parameters).

Yes, that was only one example.

But the clamping would explain why the page size change doesn't change the duration of the request.

connyduck (approved):

nikclayton:
Closing this one, https://github.com/tuskyapp/Tusky/pull/3598 is the better fix.

Issue #3580, opened 2023-04-27
22 beta: Notification error in pixelfed


kr428:
When connecting current 22 beta release (or nightly) to pixelfed.social, notifications timeline will display notifications for a moment and then switch to an error screen. Reloading will again make notifications disappear for a moment and then trigger the same behaviour. Related to #3578 ?

kr428:
Any way to provide debug info on what's happening there...? Logs?

nikclayton:
The builds on Play / F-Droid don't include logs unfortunately.

Are you able to build and test from the source code locally? That would provide logs.

nikclayton:
OK. It's handy you can build from source, since you can reproduce the problem.

## Initial setting

The value of `max_id` is initially set in `NotificationsViewModel`, around line 460. Check the logs for a line like

```
Restoring at ...
```

The value at the end of that line should be the ID of the notification you were at the last time you used the app. The default value is 0, which gets converted to a null.

So if you see a null or a long string of digits on that log line things are OK. What would not be OK is if this is the empty string.

## Passing to NotificationsRepository

Still in `NotificationsViewModel`, `lastNotificationId` is passed to `getNotifications()` as the `initialKey` parameter.

From there it is passed to `repository.getNotificationsStream()`, which sets it as the `initialKey` for the `Pager`.

## Use in `NotificationsPagingSource`.

The `Pager` makes calls to `NotificationsPagingSource.load()` when it needs data. The first time it does this it passes in the `initialKey` value in the `LoadParams` type (`key` property).

Which is then used in `getInitialPage()`.

## Best guess

My best guess is that the `val lastNotificationId = when ...` code in `NotificationsViewModel` is getting a key that is unexpectedly empty (i.e., the empty string, **not** null).

Can you check that? If it is then the fix is straightforward, although I also need to figure out how that can happen.

kr428:
Hmmm. I'm unsure. Added a breakpoint to see whether / how the lastNotificationId is set - and this seemed to happen just once, and in this case it seemed to have a meaningful value. Switching to the notification tab after that - displays notifications. See the "Prepend for key: 141...". This seems to work, and this is apparently what's being used in the request url for min_id:

```
D/NotificationsPagingSource: load() with Prepend for key: 14186984
I/okhttp.OkHttpClient: --> GET https://pixelfed.social/api/v1/notifications?min_id=14186984&limit=30&exclude_types%5B%5D=follow_request
[...]

PR #3581, opened 2023-04-27
Fix IndexOutOfBoundsException in onPause

nikclayton:
Use `getOrNull` instead of `get`, which was occasionally throwing IndexOutOfBoundsException.

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/components/notifications/NotificationsFragment.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/components/notifications/NotificationsFragment.kt
@@ -453,7 +453,7 @@ class NotificationsFragment :
         // Save the ID of the first notification visible in the list
         val position = layoutManager.findFirstVisibleItemPosition()
         if (position >= 0) {
-            adapter.snapshot()[position]?.id?.let { id ->
+            adapter.snapshot().getOrNull(position)?.id?.let { id ->
                 viewModel.accept(InfallibleUiAction.SaveVisibleId(visibleId = id))
             }
         }


nikclayton:
Fixes issue in 22.0 beta 1.

I'm not clear how this can happen, unless there's a race condition between `findFirstVisibleItemPosition` and calling `adapter.snapshot()`.

app/src/main/java/com/keylesspalace/tusky/components/notifications/NotificationsFragment.kt
```diff
@@ -453,7 +453,7 @@ class NotificationsFragment :
         // Save the ID of the first notification visible in the list
         val position = layoutManager.findFirstVisibleItemPosition()
         if (position >= 0) {
-            adapter.snapshot()[position]?.id?.let { id ->
+            adapter.snapshot().getOrNull(position)?.id?.let { id ->
```
Lakoja, line 456: We have a few "crashes but I don't know what leads to it" now (I fixed two).

Maybe this would be a case for an error log - being shown in the new application event log.

nikclayton:
> Is that a crash from the Google console?
> 
> Or this one #3580?

It shows up in the Play Console, but it could also be the root of #3580. The Play Console is refusing to show the stack trace at the moment ("An unexpected error has occurred. Please try again. (50A27A07)")

app/src/main/java/com/keylesspalace/tusky/components/notifications/NotificationsFragment.kt
nikclayton: This is the sort of thing that could be covered, yes.

Tak (approved):

PR #3583, opened 2023-04-27
Add missing languages to locales_config.xml

nikclayton:
- Update the language names, where necessary, with the language names from the icu4j v73.1 data
- Sort the list of language names using the ICU collator

This makes some baseline lint errors obsolete, remove them.


nikclayton:
Why have the cutoff at all? E.g., with "ber", if I've read the docs correctly and the user's device has that set as the language then Tusky will use those translated strings, even if it's just 10%, falling back to English for the untranslated ones.

The XML file controls https://developer.android.com/guide/topics/resources/app-languages#use-localeconfig, so users can set a different language from the device language if they want.

There's also a possible argument that user seeing an incomplete translation is also incentivised to join the translation project to improve things.

connyduck:
Well I don't like Apps that present me half baked translations, but I guess thats me.

If you want to go forward with all, the lists in donottranslate.xml need to be updated as well.

nikclayton:
Yeah, unfortunately Weblate doesn't let you set a threshold of minimum number or percentage of translated strings before pushing (https://github.com/WeblateOrg/weblate/issues/3436).

nikclayton:
Flipped to draft, I'm unsure how to represent the name of the Berber languages, I've contacted the person who's done the translation for advice.

nikclayton:
Also, I'm not sure yet of the best practices for representing a sorted list of languages when some of those languages are written in different scripts.

nikclayton:
> How about using their script name as first part and English as a second part in brackets? And sort by English. (Except the default one and frequently used at the top.)

I did some more research, and decided that using the ICU collation code makes the most sense.

Ideally this would happen on the device, but some of the necessary features aren't present until the Q release, so a fixed order is the best we can do at the moment.

The ordering here is using the ICU collation order, after case folding each entry. This appears to match the order that they appear in the Android System language chooser as well.

I didn't do this by hand, I wrote a tool to do that. I'll be preparing a separate PR with that, so that in the future this can be generted automatically at build time

PR #3574, opened 2023-04-25
Translations update from Weblate

nailyk-weblate:
Translations update from [Weblate](https://weblate.tusky.app) for [Tusky/Tusky](https://weblate.tusky.app/projects/tusky/tusky/).

Current translation status:

Weblate translation status

PR #3584, opened 2023-04-27
Add entry for "or" to locales_config.xml and donottranslate.xml

nikclayton:


nikclayton:
This is necessary to unblock https://github.com/tuskyapp/Tusky/pull/3574

connyduck:
`language_entries` and `language_values` in `donottranslate.xml` need to be updated as well.

Sidenote: Starting from AGP 8.1 / Android Studio Giraffe it will be possible to auto-generate the localeConfig

PR #3585, opened 2023-04-27
Remove useClasspathSnapshot flag

Goooler:
Some of them are the default values for now.


gradle.properties
```diff
@@ -3,18 +3,10 @@ org.gradle.caching=true
...
-
-# https://blog.jetbrains.com/kotlin/2022/07/a-new-approach-to-incremental-compilation-in-kotlin/
-kotlin.incremental.useClasspathSnapshot=true
+# https://docs.gradle.org/8.1/userguide/configuration_cache.html
+org.gradle.configuration-cache=true
 
 # Disable buildFeatures flags by default
-android.defaults.buildfeatures.aidl=false
-android.defaults.buildfeatures.buildconfig=false
-android.defaults.buildfeatures.renderscript=false
```
nikclayton, line 15: Don't these defaults change with AGP 8? We're still using 7.4.2. Or am I misreading the docs?

Goooler: Oooops, I forget that we are still on AGP 7.

Goooler:
The only flag change will be handled by #3835.

PR #3593, opened 2023-04-29
update minSdk to 24, cleanup code

### LINKED ISSUE #2607: Stopping support for Android 6.x
Tusky will stop supporting Android 6.x at the beginnig of 2023
Current usage of Android 6.x according to Google Play is 0,4% (250 devices).

connyduck:
closes #2607


Lakoja (approved):

nikclayton:
Will merge after v22 is released.

app/src/main/java/com/keylesspalace/tusky/util/StatusParsingHelper.kt
```diff
@@ -36,31 +35,3 @@ fun String.parseAsMastodonHtml(tagHandler: TagHandler? = null): Spanned {
...
-            builder!!.append(ASCII_HYPHEN)
-        } else if (replacing) {
-            builder!!.append(character)
-        }
-    }
-    return if (replacing) builder else content
-}
-
-private const val SOFT_HYPHEN = '\u00ad'
-private const val ASCII_HYPHEN = '-'
```
Tak, line 66: 🎉

Tak (approved):

charlag:
Did we check the usage? do we strictly need to do this?

connyduck:
> Did we check the usage? do we strictly need to do this?

Usage is in the issue. I think the simpler code and app size improvements are easily worth it.

charlag:
Sorry that I've missed the usage, I am a bit surprised tbh but I agree, it is totally worth it if that's the case. Too bad we don't have F-Droid statistics but I don't think it would make a difference.

PR #3595, opened 2023-04-29
cleanup trending code

connyduck:
- reduced boilerplate
- removed dependency on BottomSheetActivity and LinkListener
- calculation of maxTrendingValue only running once instead once per tag
- items now get diffed correctly
- don't show progress bar (only refresh spinner) during pull-to-refresh to avoid the whole screen blinking
- log error


app/lint.xml
```diff
@@ -24,6 +24,9 @@
          a noun and a verb -->
     <issue id="DuplicateStrings" severity="ignore" />
 
+    <!-- Unconditional log calls are no problem, they are thrown out of bytecode by the code shrinker -->
+    <issue id="LogConditional" severity="ignore" />
```
Tak, line 28: ❤️

Tak (approved):

app/src/main/java/com/keylesspalace/tusky/util/ViewDataUtils.kt
```diff
@@ -71,9 +69,20 @@ fun Notification.toViewData(
...
 }
 
-@JvmName("tagToViewData")
-fun TrendingTag.toViewData(): TrendingViewData.Tag {
-    return TrendingViewData.Tag(
-        tag = this
-    )
+fun List<TrendingTag>.toViewData(): List<TrendingViewData.Tag> {
+    val maxTrendingValue = flatMap { tag -> tag.history }
+        .mapNotNull { it.uses.toLongOrNull() }
```
charlag, line 74: it sucks that it's gonna box them but I guess no nice way around it

charlag (approved):

PR #3610, opened 2023-05-02
Add a document for API debugging

nikclayton:
Explains how users can perform API requests and send results to make debugging easier.


connyduck:
Why not put it into faqs?

nikclayton:
It's not really frequently asked. And the info is probably equally useful for developers.

PR #3614, opened 2023-05-03
Save the user's reading position in the home timeline

nikclayton:
- Add a field to AccountEntity to hold the reading position
- Provide a method to save in the viewmodel to save the position
- Save the position when TimelineFragment pauses

Does not restore the position yet.


nikclayton:
This is the "save" half of https://github.com/tuskyapp/Tusky/pull/3601. As I noted there, #3436 needs this to be in a release *before* that PR is merged, so that it can restore the user's reading position after the local timeline cache is wiped.

Lakoja:
> This is the "save" half of #3601. As I noted there, #3436 needs this to be in a release _before_ that PR is merged, so that it can restore the user's reading position after the local timeline cache is wiped.

Does it need to be wiped?
One could just remove the placeholders (from db). The state then would not be different to the current situation where there might be any number of status missing at any position. And in any case all that is "old stuff" which will soon be outdated (deleted) anyway.
If the "Load more code" is not removed not even the placeholders need to be removed.

I'm just asking if the alternative (keep a special release order and depenencies) is too complicated for what is achieved.

nikclayton:
> > This is the "save" half of #3601. As I noted there, #3436 needs this to be in a release _before_ that PR is merged, so that it can restore the user's reading position after the local timeline cache is wiped.
> 
> Does it need to be wiped? One could just remove the placeholders (from db).

That's more complicated than just wiping and reloading, *and* it drops valid statuses from the timeline.

app/src/main/java/com/keylesspalace/tusky/components/timeline/viewmodel/CachedTimelineViewModel.kt
```diff
@@ -289,6 +289,14 @@ class CachedTimelineViewModel @Inject constructor(
         }
     }
 
+    override fun saveReadingPosition(statusId: String) {
+        accountManager.activeAccount?.let { account ->
+            Log.d(TAG, "Saving position at: $statusId")
+            account.lastVisibleHomeTimelineStatusId = statusId
```
Lakoja, line 295: TimelineFragment is used for more timelines than the home timeline (?)

nikclayton: It is.

Which kind of timelines is CachedTimelineViewModel used for though...?

connyduck (approved):

app/src/main/java/com/keylesspalace/tusky/components/timeline/viewmodel/CachedTimelineViewModel.kt
nikclayton: Hint: https://github.com/tuskyapp/Tusky/blob/develop/app/src/main/java/com/keylesspalace/tusky/components/timeline/TimelineFragment.kt#L102-L108

PR #3615, opened 2023-05-03
Ignore LogConditional issues

nikclayton:
Logs are stripped in release builds, this is an unnecessary lint


connyduck (approved):
got that in https://github.com/tuskyapp/Tusky/pull/3595 already though

PR #3621, opened 2023-05-04
mklanguages: Manage language lists in donottranslate.xml

nikclayton:
mklanguages processes values-*/strings.xml entries and uses that to determine which languages can be used.

It uses the ICU database to generate the display name for each language, and to list them in sorted order.

Add `runtools` scripts to make it easy to run this and other tools.

Update donottranslate.xml with comments that show the language code and English name for each language.


nikclayton:
Oh, something else this code could do in the future -- if you want to have some threshold on the "quality" of a translation before it's listed in the languages arrays, this could would be the place to do that decision making.

Some future criteria might be:

- # of strings translated
- Are some "core" strings translated (e.g., all top level menu and timeline names)?
- Is everything associated with a particular timeline translated

Don't want to define them now, that's just some ideas for the future.

connyduck:
So much added complexity just for some languages that need to be updated like 4 times a year? I wouldn't do that.

nikclayton:
Here's the alternative:

Every time we add a new language someone has to:

- Figure out the correct way to write the name of that language
- Figure out how the name of the language should be sorted relative to the other languages
- Modify the `language_entries` and `language_values` arrays and keep them in sync

Because it's only done a few times a year it requires whoever does that to effectively do it from scratch every time -- anything they learned the last time they did it (and there's no guarantee they were the last person to do it) they've probably forgotten.

So it's incredibly easy to get wrong. This PR is proof -- the previous version of the `donottranslate.xml` file had `Esperanto` and `Español` in the wrong order because someone (I don't know who, and I'm not looking it up because this isn't about assigning blame) got it wrong. There were similar mistakes for the different variants of Chinese.

https://github.com/tuskyapp/Tusky/pull/3574/files is another example.  That adds "Odia" as a translation.

What's the correct name to use for that language (it's not "Odia"), and where should it go in the sorted list?

With this PR I get the answer in about a second, and a change I can commit and submit as a PR, confident that it's correct. Without it you have to do a bunch of time consuming research. Then whoever reviews the PR has to confirm that the research is correct.

nikclayton:
> Ok, can you add tests for this?

Done.

Issue #3539, opened 2023-04-16
Check for notifications at startup / when notifications are enabled


nikclayton:
If I'm reading the code correctly, enabling pull notifications schedules periodic checks for notifications using WorkManager (`NotificationManager.enablePullNotifications()`).

But these checks are scheduled for some time in the future.

It's probably a good idea to perform an explicit check for new notifications in at least two places.

1. When pull notifications are enabled (i.e., in `enablePullNotifications`).
2. When `MainActivity` starts (in `onCreate` at least, possibly even in `onResume`).

That should give users a better experience, with notifications appearing more frequently.

Code to do this (if added to `NotificationHelper.java`) would look like:

```
/** WorkManager tag for periodic notification checks */
private static final String NOTIFICATION_PULL_TAG = "pullNotifications";

/** WorkManager tag for oneshot notification checks */
private static final String NOTIFICATION_ONESHOT = "oneshotNotifications";

// ...

public static void checkNotifications(Context context) {
    WorkManager workManager = WorkManager.getInstance(context);
    OneTimeWorkRequest oneShot = new OneTimeWorkRequest.Builder(NotificationWorker.class)
        .setConstraints(
            new Constraints.Builder()
                .setRequiredNetworkType(NetworkType.CONNECTED)
                .build()
        )
        .setExpedited(OutOfQuotaPolicy.RUN_AS_NON_EXPEDITED_WORK_REQUEST)
        .build();
    workManager.enqueueUniqueWork(NOTIFICATION_ONESHOT, ExistingWorkPolicy.KEEP, oneShot);
}
```

* * * *
- Tusky Version: 21

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

PR #3626, opened 2023-05-05
Only fetch and display a given notification once

### LINKED ISSUE #3625: Do not show notifications if already seen
Notifications that were already seen in the notifications tab should not be shown as system notifications again.

* * * *
- Tusky Version: this _should_ be older than v22

Maybe related / same to #3616

nikclayton:
When fetching:
- Maintain a marker with the position of the newest fetched notification
- Use the marker to determine which notifications to fetch
- Fetch notifications with min_id to ensure that none are lost
- Update the marker as necessary
- Perform a one-time immediate fetch of notifications on startup

When creating notifications:
- Identify each notification with `tag=${MastodonNotificationId}`, `id=${account.id}`
- Remove `activeNotifications` field, it's no longer necessary
- Use the tag/id tuple to reliably identify existing notifications and avoid creating duplicates
- Cancelling notifications for an account must iterate over all the notifications, and individually remove the notifications that exist for that account.
- Limit notifications to a maximum of 40 (excluding summary notifications)
- Remove notifications (oldest first) to get under this limit
- Rate limit notification creation to 1 per second, so the OS won't drop them

Adjust the summary notification:
- Ensure the summary notification and the child notifications have the same group key
- Dismiss the summary notification if there is only one child notification

NotificationClearBroadcastReceiver is no longer needed, so remove it, and the need for deletePendingIntent.

Fixes #3625, #3539


app/src/main/java/com/keylesspalace/tusky/components/notifications/NotificationFetcher.kt
```diff
@@ -31,34 +30,47 @@ class NotificationFetcher @Inject constructor(
...
-            }
-        }
-        return result
+        if (notifications.isEmpty()) return mutableListOf()
+
+        // Notifications are returned in order, most recent first. Save the newest notification ID
+        // in the marker.
+        val newMarkerId = notifications.first().id
+        Log.d(TAG, "updating notification marker to: $newMarkerId")
+        mastodonApi.updateMarkersWithAuth(authHeader, notificationsLastReadId = newMarkerId)
```
Lakoja, line 71: This must also be done in the notification tab?

nikclayton: No, because the notification tab (specifically, NotificationsViewModel) updates `account.lastNotificationId` when the user navigates away, and `account.lastNotificationId` is used in the algorithm earlier in this code for figuring out which notification to show.

The notifications tab should show an unbroken timeline of the user's notifications, unaffected by the Android Notifications that are generated. When the user returns to this tab their reading position (i.e., the notification they return to) should be the same one as when they left. That's why the notification tab has a separate "marker" for the user's position.

However Android Notifications do not need to show every single Mastodon Notification. If the user has already seen it in the notifications tab then they don't need to see an Android Notification about it as well. So the "marker" for the Android Notifications is set to whichever is larger of either the previous value, or the marker that the notification tab uses.

Lakoja: Then we are talking of the same. (Wanted to avoid exactly that: get Android notifications for a previously seen Mastodon notification)

Lakoja: > No, because the notification tab (specifically, NotificationsViewModel) updates `account.lastNotificationId` when the user navigates away

"navigates away" means I still get an Android notification about a notification I am looking at at the moment in the timeline?
Shouldn't it be "once loaded"?

nikclayton: There's probably more to it than that. I just experimented with the GMail app, which has the following behaviours that I think we should imitate.

## Active GMail Android Notifications are cleared on open

- Have one or more GMail Android Notifications
- Open GMail
- The Android Notifications are cleared

Proposed Tusky behaviour -- clear all Tusky Android Notifications for the active account when the user navigates to the Notifications tab.

## No Android Notifications while GMail is open

- Open GMail
- From another e-mail app on a different device, send a message to the GMail account you have open on the phone
- GMail updates to show the new e-mail, but does not trigger an Android Notification

Proposed Tusky behaviour -- do not send an Android Notification while the Notifications tab is open. However, do update the marker if new notifications are fetched by the NotificationFetcher code.

Lakoja: > Proposed Tusky behaviour -- clear all Tusky Android Notifications for the active account when the user navigates to the Notifications tab.

That should be the case: #2528

> Proposed Tusky behaviour -- do not send an Android Notification while the Notifications tab is open. However, do update the marker if new notifications are fetched by the NotificationFetcher code.

I see that falling in the category "update things without user interaction / in the background".
One aspect of that would be some information badges with "here is new content".
In the case of the notifications timeline it could update the list silently (and still show the "new stuff" badge on the tab icon).

nikclayton: > > Proposed Tusky behaviour -- clear all Tusky Android Notifications for the active account when the user navigates to the Notifications tab.
> 
> That should be the case: #2528

That's not happening -- there's a bug somewhere.

I've got an emulator running with the code from this PR, and the active account has 5 Android Notifications (i.e., if I pull down the notification shade I see five notifications).

When I switch to the notification tab `NotificationHelper.clearNotificationsForActiveAccount()` *is* called.

But inside that function `account.getActiveNotifications()` returns `"[]"`, so the notifications are never cancelled.

I put breakpoints on every place in the code where `AccountEntity.activeNotifications` is modified, and launched a debug build. The breakpoints were never hit, so the list of active notifications is not being updated / fetched.

I'll keep digging.

Lakoja: > > That should be the case: #2528
> 

> When I switch to the notification tab `NotificationHelper.clearNotificationsForActiveAccount()` _is_ called.
> 
> But inside that function `account.getActiveNotifications()` returns `"[]"`, so the notifications are never cancelled.


That sounds "basic" as if notification removal does not/never work for you?
It works fine for me (either pull or now push notifications).

However a quick look in the code I found the `NotificationClearBroadcastReceiver` which does exactly that: Only set the field to empty array but does not actually clear any notifications (?).
I don't remember however ever stumbling across this code or problem therein.

nikclayton: From experimenting with the code it would clear the record of all the Android notifications if the user removed (e.g., swiped left) on a single notification.

That's obviously bad, so I've also fixed that in this PR -- instead of trying to keep track of the active notifications it's easier to get the active notifications when they're needed and iterate over them.

This also simplifies the code -- there's no need for the database column, marshalling data to/from JSON, etc.

nikclayton: This ensures that notifications are cleared when the notifications tab is active.

It doesn't disable showing new notifications if they come in while the tab is open. That can get pushed to v23, I'm focused on trying to fix the known issues with v22 so we can get beta 3 out of the door and then hopefully a final release.

app/src/main/java/com/keylesspalace/tusky/components/notifications/NotificationHelper.java
```diff
@@ -516,17 +520,14 @@ public static void disablePullNotifications(Context context) {
 
     public static void clearNotificationsForActiveAccount(@NonNull Context context, @NonNull AccountManager accountManager) {
         AccountEntity account = accountManager.getActiveAccount();
-        if (account != null && !account.getActiveNotifications().equals("[]")) {
-            Single.fromCallable(() -> {
-                account.setActiveNotifications("[]");
-                accountManager.saveAccount(account);
-
-                NotificationManager notificationManager = (NotificationManager) context.getSystemService(Context.NOTIFICATION_SERVICE);
-                notificationManager.cancel((int) account.getId());
```
Lakoja, line 525: Very odd indeed: This (old code) did work for me all the time.
It also did just now in the emulator: I had 19 notifications (lines) there when starting, switching to the notification tab removed them all.
Probably it always was a summary notification (so only one) and removing that one does ok.

nikclayton: Possibly. I tend to bounce between a bunch of different branches, launching in the same emulator, so it wouldn't surprise me if the database and real life got out of sync.

nikclayton:
Worth noting -- using the marker API means that if the user has multiple devices notifications will be distributed amongst them somewhat randomly (from the perspective of the user)

E.g., if the user has device A and device B, A is checking for notifications every 15 minutes on the hour, and B is checking for notifications every 15 minutes starting at 5 minutes past the hour and the following notifications come in:

N1 @ 00:00
N2 @ 00:01
N3 @ 00:07
N4 @ 00:16
N5 @ 00:21

then device A will receive notifications N1, N2, N4, and device B will receive notifications N3 and N5.

Note that device A and device B don't have to be two different devices running Tusky. One of them could be Tusky, the other could be the user signed in to the web app (e.g., I just checked on mastodon.social, and opening the web app makes GET and POST requests to the markers API for notifications and the home timeline).

We might want to put this in a preference in v23

app/src/main/java/com/keylesspalace/tusky/components/notifications/NotificationHelper.java
```diff
@@ -290,19 +289,14 @@ private static NotificationCompat.Builder newNotification(Context context, Notif
...
-
         NotificationCompat.Builder builder = new NotificationCompat.Builder(context, getChannelId(account, body))
                 .setSmallIcon(R.drawable.ic_notify)
                 .setContentIntent(summary ? summaryResultPendingIntent : eventResultPendingIntent)
-                .setDeleteIntent(deletePendingIntent)
                 .setColor(context.getColor(R.color.notification_color))
                 .setGroup(account.getAccountId())
                 .setAutoCancel(true)
                 .setShortcutId(Long.toString(account.getId()))
+                .setGroup(GROUP_KEY_GENERAL)
```
connyduck, line 299: setting group twice?

nikclayton: Good point. I thik the account ID (String) is the better choice, so that notifications from different accounts don't get grouped together. I've made that change.

connyduck (approved):

nikclayton:
Looking at this code some more I think there's another problem.

If I'm reading it properly only one summary notification can be generated. That summary notification will be assigned to whichever notification channel (follows, boosts, mentions, etc...) was assigned to the last notification seen.

I think that's a problem, because:

1. The user may have muted that notification channel. I assume that will mute the summary notification as well. But the summary notification might be summarising different types of notification (follow, boost, etc), so now the entire summary is muted when only the summary for a specific channel should be muted.
2. The Android documentation is silent on what happens if notifications in different channels are part of the same summary group, but I suspect that's an edge case that's not handled well.

Also, with one summary the user can choose to dismiss all notifications with one swipe, which might not be appropriate.

So I think this needs to be modified to create N summary notifications, one per notification channel that contains 2+ notifications.

- The summary notification will be muted if the channel is muted
- The user can easily dismiss all notifications of a particular type (e.g., boosts) without affecting the other notifications

I'll work on that now.

nikclayton:
OK, I think that's fixed. The new code:

1. Creates a summary notification for each notification group (follow, reblog, mention, etc). This ensures that if one channel is muted the summary notifications for other channels are not affected
2. Updates those summary notifications every time new notifications come in

It also rate limits the calls to notify(), if you do too many too quickly Android will silently drop the notification. This may be the cause of some "missing notification" reports in the past.

Android also has a limit on the number of active notifications, and will not create new ones if the app is at that limit. To work around that, this code will also delete older Android notifications if it needs space to show newer ones.

Note that in this case you can't, e.g., update the summary notification and say something like "40 notifications shown, 20 hidden", because on API 24 and above Android manages the text of the summary notification for you, overriding whatever text the app uses for it.

Possible bug, but I'm not sure I'm worried about it for v22:

- If the user swipes a notification away the summary notification for that group will not be modified.

app/src/main/java/com/keylesspalace/tusky/components/notifications/NotificationFetcher.kt
```diff
@@ -1,64 +1,138 @@
...
     private val accountManager: AccountManager,
     private val context: Context
 ) {
     fun fetchAndShow() {
+        // This should never be run on the main thread (see
+        // https://developer.android.com/guide/background/persistent/threading/worker). Belt-and-braces
+        // check to make sure.
+        if (Looper.getMainLooper().isCurrentThread) {
+            throw IllegalStateException("fetching notifications on the main thread")
+        }
```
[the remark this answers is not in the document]
nikclayton: That's not what I see.

I just checked this by doing the following:

In `MainActivity.onCreate()` I put the following code immediately after the `super.onCreate(savedInstanceState)` call:

```kotlin
NotificationFetcher(mastodonApi, accountManager, this).fetchAndShow()
```

and commented out the `throw IllegalStateException` on here.

Rebuilt and restarted the app and the notifications were fetched.

Then uncommented the `throw IllegalStateException`, repeated the test, and the exception was thrown as expected.

Looking in to this some more the check that I've got here can be replaced with an `@WorkerThread` annotation on the class. That won't be a runtime error, but it will cause the lint checks to fail, and since that will blocks PRs I think that's OK.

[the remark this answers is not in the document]
connyduck: yes an annotation is a better idea

app/src/main/java/com/keylesspalace/tusky/components/notifications/NotificationHelper.java
```diff
@@ -233,51 +234,135 @@ public static void make(final Context context, Notification body, AccountEntity
...
+            setSoundVibrationLight(account, summaryBuilder);
+
+            // TODO: Use the batch notification API available in NotificationManagerCompat
+            // 1.11 and up (https://developer.android.com/jetpack/androidx/releases/core#1.11.0-alpha01)
+            // when it is released.
+            notificationManager.notify(summaryTag, accountId, summaryBuilder.build());
+
+            // Android will rate limit / drop notifications if they're posted too
+            // quickly. There is no indication to the user that this happened
+            try { Thread.sleep(1000); } catch (InterruptedException ignored) { }
```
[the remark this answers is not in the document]
nikclayton: Yeah, I think that documentation is wrong.

To prove this, do the following:

Modify `NotificationFetcher.fetchNewNotifications()`, and replace the block of code that calculates `minId` with this:

```kotlin
//        val minId = when (val marker = fetchMarker(authHeader, account)) {
//            null -> account.lastNotificationId.takeIf { it != "0" }
//            else -> if (marker.lastReadId > account.lastNotificationId) marker.lastReadId else account.lastNotificationId
//        }
        val minId = account.lastNotificationId
```

This will fetch new notifications based on your reading position in the Notifications tab.

Then comment out the `Thread.sleep` calls here and in `NotificationFetcher`.

Build and launch the app. Switch to the notifications tab, and scroll down a few pages, so that your reading position is before the newest notifications.

Switch back to the home tab. Then relaunch the app.

Tusky will re-fetch the Mastodon notifications that were above your reading position in the Notifications tab and display them as Android notifications.

In my emulator (running API 31) I see the following errors logged by the system when it does that (without the sleep):

```
NotificationService               E  Muting recently noisy 0|com.keylesspalace.tusky|2|147077155|10151
NotificationService               E  Muting recently noisy 0|com.keylesspalace.tusky|2|com.keylesspalace.tusky.notification.group_summary.CHANNEL_FAVOURITEmastodon.social:109241526707263377|10151
NotificationService               E  Package enqueue rate is 67.90989. Shedding 0|com.keylesspalace.tusky|2|148269585|10151. package=com.keylesspalace.tusky
```

If I'm reading that correctly it means that Android is both muting some notification channels because of the rate of new notifications (so users will not be informed about them) and dropping (i.e., "Shedding") some notifications entirely.

I suspect this is the source of some of the user complaints about missing notifications. If you get a lot in one go (say, a post has recently been boosted or favourited a lot) generating a lot of notifications then Android will mute and/or drop them.

That link says:

> Beginning with Android 8.1 (API level 27), apps can't make a notification sound more than once per second. If your app posts multiple notifications in one second, they all appear as expected, but only the first notification per second makes a sound.
> However, Android also applies a rate limit when updating a notification. If you post updates to a single notification too frequently, such as many in less than one second, the system might drop updates.

I don't think the first paragraph is relevant here.

The second paragraph should be relevant, and that's what I think is wrong. The code in this PR calls `.notify()` on each Android notification exactly once, so it's not "post(ing) updates to a single notification too frequently".

https://cs.android.com/android/platform/superproject/+/master:frameworks/base/services/core/java/com/android/server/notification/NotificationManagerService.java;l=6905-6937?q=%22Package%20enqueue%20rate%22 shows that this is checking `appEnqueueRate`.

And if you go poking around [frameworks/base/services/core/java/com/android/server/notification/NotificationUsageStats.java](https://cs.android.com/android/platform/superproject/+/refs/heads/master:frameworks/base/services/core/java/com/android/server/notification/NotificationUsageStats.java) it looks like this is the enqueue rate for all notifications, *not* a per-notification rate.

And if you look at https://cs.android.com/android/platform/superproject/+/refs/heads/master:frameworks/base/services/core/java/com/android/server/notification/NotificationManagerService.java;l=8129-8133;drc=7346c436e5a11ce08f6a80dcfeb8ef941ca30176 (where `isAlertRateLimited(pkg)` is called you) can see that the rate limits are definitely not calculated per notification, but across all the notifications posted by a single package (`pkg = record.getSbn().getPackageName()` -- `Sbn` is the status bar notification, `getPackageName()` is the name of the package that posted the notification).

https://cs.android.com/android/platform/superproject/+/refs/heads/master:frameworks/base/services/core/java/com/android/server/notification/AlertRateLimiter.java;l=23-35;drc=7346c436e5a11ce08f6a80dcfeb8ef941ca30176 is the rate limiter, which seems to set an allowed interval of 1 per 1000ms.

So I think Android applies undocumented rate limits to notifications, grouped by the package that posted them. Exceeding these limits will see notifications muted or dropped.

nikclayton:
While I'm here, there's at least one more bug which this doesn't fix yet.

`fetchNotifications` (`fetchNewNotifications` in this PR) makes one call to get new notifications (`mastodonApi.notificationsWithAuth`).

That will return N new notifications, starting with the oldest new one, up to `limit`. `limit` is not set in this code, and is documented as being default=15 max=30 at https://docs.joinmastodon.org/methods/notifications/#get.

If we can trust these documented defaults this means that in any given notification fetch (i.e., approx every 15 minutes) Tusky will get the batch of next-newest 15 notifications and display them. If the user has more notifications than that they will have to wait another ~ 15 minutes until they are fetched.

That's probably also a source of user complaints about missing notifications.

Fix is to either:

1. Use the NotificationsRepository and fetch the notifications that way, taking advantage of its existing code for paging up (prepending) the list of notifications. Not sure how well Pager3 is suited to that though.
2. Write some additional code here that repeatedly fetches new notifications until there are no new ones.

connyduck (approved):

PR #3633, opened 2023-05-08
replace deprecated getParcelable* methods with compat versions

connyduck:
I'm still missing a compat method for `getSerializableExtra` with we also use in a few places though


app/src/main/java/com/keylesspalace/tusky/components/compose/dialog/FocusDialog.kt
```diff
@@ -38,7 +38,7 @@ fun <T> T.makeFocusDialog(
     existingFocus: Focus?,
     previewUri: Uri,
     onUpdateFocus: suspend (Focus) -> Unit
-) where T : Activity, T : LifecycleOwner {
+) where T : AppCompatActivity, T : LifecycleOwner {
```
connyduck, line 41: Otherwise Glide complains: Deprecated.
Use androidx Activitys instead (ie FragmentActivity, or AppCompatActivity)

Tak (approved):

charlag (approved):
why didn't they make a `reified` version of it, uh

PR #3634, opened 2023-05-08
FilterResult.status_matches is a List<String>?

nikclayton:
This was a Mastodon documentation bug, and is being fixed on their side in https://github.com/mastodon/mastodon/issues/24900.


app/src/main/java/com/keylesspalace/tusky/entity/FilterResult.kt
```diff
@@ -5,5 +5,14 @@ import com.google.gson.annotations.SerializedName
 data class FilterResult(
     val filter: Filter,
     @SerializedName("keyword_matches") val keywordMatches: List<String>?,
```
nikclayton, line 7:
```suggestion
    @SerializedName("keyword_matches") val keywordMatches: List<String>?
```

Tak (approved):

nikclayton:
They're updating the docs and confirmed it's `List<String>?`, so I've updated this to match.

PR #3635, opened 2023-05-08
don't crash on unexpected json responses

connyduck:


app/src/main/java/com/keylesspalace/tusky/components/accountlist/AccountListFragment.kt
```diff
@@ -313,7 +313,7 @@ class AccountListFragment : Fragment(R.layout.fragment_account_list), AccountAct
 
                 val linkHeader = response.headers()["Link"]
                 onFetchAccountsSuccess(accountList, linkHeader)
-            } catch (exception: IOException) {
+            } catch (exception: Exception) {
                 onFetchAccountsFailure(exception)
```
charlag, line 317: seems a bit too broad but okay

charlag (approved):

nikclayton:
I'm not sure this is the right thing to do.

Catching IO exceptions is fine -- IO failing is an expected problem.

JSON parsing failing isn't an expected problem. It means there's a bug, either in the server or in Tusky. If that happens we need to know about it so we can fix it.

https://github.com/tuskyapp/Tusky/pull/3436 does a better job of showing errors to the user and allowing retry, but that's for errors that can reasonably be retried -- fetching bad JSON a second time isn't going to fix the problem.

So I'm worried that if this goes in as-is we risk getting error reports along the lines of "Tusky isn't working", but with no information to debug, and no crashes appearing in the Play Console to track down the problem.

connyduck:
I would agree if we were in control of the server software. Since we aren't and we are not only dealing with one, but multiple server implementations, we should treat server responses almost if they were user input. Mostly because crash = "oh no that app sucks" vs error screen = "oh no whats up with my instance again".

Yes, an easy way to get logs from devices is needed in one form or another, but I see this discussion is already happening elsewhere.

Note that these two places I'm adressing here are the only ones where we don't handle JSON exceptions, everywhere else they are handled.

nikclayton:
> I would agree if we were in control of the server software. Since we aren't and we are not only dealing with one, but multiple server implementations, we should treat server responses almost if they were user input.

I agree.

> Mostly because crash = "oh no that app sucks" vs error screen = "oh no whats up with my instance again".

I agree with that too. But I don't think the error handling / display code in Tusky is up to dealing with this properly yet.

1. That'll show an error if it occured as part of a refresh. But not (I think) if it's a prepend or append operation.
2. The documentation for MediatorResult.Error notes that it's supposed to be used for recoverable errors that can be retried. If the server is handing back bogus JSON there's no point in retrying, so JSON exceptions should be special-cased
3. This would probably have made the debugging experience worse when I was looking at the JSON issue in 22 beta 2

If this PR also included code changes so that everywhere we have something like:

```kotlin
if ((loadState.refresh as LoadState.Error).error is IOException) {
    binding.statusView.setup(R.drawable.elephant_offline, R.string.error_network, null)
} else {
    binding.statusView.setup(R.drawable.elephant_error, R.string.error_generic, null)
}
```

it had an explicit path to handle JSON errors that allowed the user to report the problem, something like (untested):

```kotlin
when (loadState.refresh as LoadState.Error).error {
    IOException -> binding.statusView.setup(R.drawable.elephant_offline, R.string.error_network, null),
    JsonParseException -> /* show a view or launch an activity to get detailed reports of JSON errors */,
    else -> binding.statusView.setup(R.drawable.elephant_error, R.string.error_generic, null)
}
```

then I think this would be very helpful. But as is I worry that it will make things harder to diagnose.

nikclayton:
Although that example in my previous comment still doesn't handle JSON errors in JSON loaded from prepend or append operations properly.

Tak (approved):

PR #3642, opened 2023-05-12
3591 detect push changes

### LINKED ISSUE #3591: UnifiedPush not working on tusky 22
Running lineageOs with Up-Example app I'm able to receive the test notification, on tusky I don't receive any notification 

By the way Im running 0 google services

Any clue or it's tusky broken ? 

Thanks !

* * * *
- Tusky Version: 22 beta, 21 
- Android Version: 13
- Android Device: samsung

- [ x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

Lakoja:
Probably fixes #3591, but is a huge step forward regardless.

Draft because the TODO count is still too high: Architecture decisions, feature decisions or at least "this is fine for now" decisions.

This improves the following for push notifications (ordered by importance):

- Possibility to change the distributor (detects that the previously used one is missing, unregisters and registers another one on the server)
- Only update (push) notification settings on server if needed
- Show currently used distributor as (unchangeable) setting
- Allow to only fetch notifications for one account (the one for which a push message was received)

Maybe main problem with also this code: All of this is (still) very edge-casey. For example there are a number of different paths and intentions in which the various methods of the new PushNotificationManager can be called.
It is rather involved testing this and it is still not possible for the user to "change anything" if something is wrong.
And: if there are few people using this the effort in programming here is also a problem.


app/src/main/java/com/keylesspalace/tusky/components/notifications/PushNotificationManager.kt
```diff
@@ -0,0 +1,311 @@
...
+        return distributors.isNotEmpty()
+    }
+
+    // TODO! there should be an actual decision (possibility) to say "I don't want to use push notifications for Tusky".
+
+    fun canEnablePushNotifications(): Boolean =
+        isUnifiedPushAvailable() && !anyAccountNeedsMigration()
+
+    fun hasPushNotificationsEnabled(account: AccountEntity): Boolean =
+        isUnifiedPushAvailable() && !account.unifiedDistributorName.isNullOrEmpty()
```
nikclayton, line 59:
```suggestion
        isUnifiedPushAvailable() && account.unifiedDistributorName.isNotBlank()
```

app/src/main/java/com/keylesspalace/tusky/components/notifications/PushNotificationManager.kt
```diff
@@ -0,0 +1,311 @@
...
+        accountManager.accounts.forEach {
+            val notificationGroupEnabled = Build.VERSION.SDK_INT < 28 ||
+                nm.getNotificationChannelGroup(it.identifier)?.isBlocked == false
+            val shouldEnable = it.notificationsEnabled && notificationGroupEnabled
+
+            if (shouldEnable) {
+                enableUnifiedPushNotificationsForAccount(it)
+            } else {
+                disableUnifiedPushNotificationsForAccount(it)
+            }
```
nikclayton, line 77: Why is a method called `enablePushNotifications` possibly disabling push notifications?

nikclayton:
> Draft because the TODO count is still too high: Architecture decisions, feature decisions or at least "this is fine for now" decisions.

Can you describe those separately to the code? E.g., a high level architecture sketch that describes what you think this *should* do?

That's helpful for several reasons:

1. I don't need to try and reverse engineer the architecture from the code
2. It's easier to see if the architecture you intended matches the architecture in the code
3. It's possible to have a conversation about the larger architecture without needing to get in to the specifics of the implementation

app/src/main/java/com/keylesspalace/tusky/components/notifications/PushNotificationManager.kt
Lakoja: unifiedDistributorName is nullable

app/src/main/java/com/keylesspalace/tusky/components/notifications/PushNotificationManager.kt
Lakoja: (it's moved old code)

Could be called `doYourStuff` or `manage` instead - as it is in `PushNotificationManager`.

Maybe `enableOrDisablePerAccount`?

app/src/main/java/com/keylesspalace/tusky/components/notifications/NotificationsFragment.kt
```diff
@@ -554,6 +554,9 @@ class NotificationsFragment :
     }
 
     private fun showFilterDialog() {
+        // TODO why do we have two different "notification filter" settings; the ones here (only for ui) and the
+        //   ones in the account settings (NotificationPreferencesFragment)?
```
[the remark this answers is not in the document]
Lakoja: ok, removed the comment

Lakoja:
> > Draft because the TODO count is still too high: Architecture decisions, feature decisions or at least "this is fine for now" decisions.
> 
> Can you describe those separately to the code? E.g., a high level architecture sketch that describes what you think this _should_ do?

There is no larger architecture here, yet. Maybe not needed (and there was none).

It's basically "do push notification registration if needed on start".

Most questions make sense at their specific location. Extracting them would loose valuable information.

Thus they can and should be discussed in-line here.

Most of them are in `PushNotificationManager`.

The biggest there on the top: Should we join pull and push notification code more?

Lakoja:
Maybe one general question still:

The linked issue has be closed (by author). So it probably makes sense to add another one "improve push notifications" and change the references here (?)

Lakoja:
Updating / rebasing this another time.

As people's main problem with Tusky is missing notifications.

Lakoja:
Newer version: Based on current code and recent refactorings: https://github.com/tuskyapp/Tusky/pull/4896

PR #3646, opened 2023-05-13
make links in bios of follow request and follow notifications work

connyduck:
Also fixes a bug where FollowRequestViewHolder would not show the bio if it has been previously bound to an account without bio.

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/adapter/FollowRequestViewHolder.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/adapter/FollowRequestViewHolder.kt
@@ -25,11 +25,14 @@ import com.keylesspalace.tusky.components.notifications.NotificationsPagingAdapt
 import com.keylesspalace.tusky.databinding.ItemFollowRequestBinding
 import com.keylesspalace.tusky.entity.TimelineAccount
 import com.keylesspalace.tusky.interfaces.AccountActionListener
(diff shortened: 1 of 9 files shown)


app/src/main/java/com/keylesspalace/tusky/adapter/FollowRequestViewHolder.kt
```diff
@@ -92,9 +96,11 @@ class FollowRequestViewHolder(
...
             binding.accountNote.hide()
         } else {
-            binding.accountNote.text =
-                account.note.parseAsMastodonHtml()
-                    .emojify(account.emojis, binding.accountNote, animateEmojis)
+            binding.accountNote.show()
+
+            val emojifiedNote = account.note.parseAsMastodonHtml()
+                .emojify(account.emojis, binding.accountNote, animateEmojis)
+            setClickableText(binding.accountNote, emojifiedNote, emptyList(), null, linkListener)
```
nikclayton, line 103: Good catch.

Please can you also change the `TextView` that contains the note in the layout file to `ClickableSpanTextView`, so that the touch targets are expanded appropriately.

If you can't do that today I'll make a change to the PR directly, so that this can go in to beta 3.

nikclayton (changes requested):

nikclayton (approved):

PR #3649, opened 2023-05-13
Move cache pruning to a WorkManager worker

nikclayton:
- Extend what was `NotificationWorkerFactory` to `WorkerFactory`. This
  can construct arbitrary Workers as long as they provide their own
  Factory for construction.

  The per-Worker factory contains any injected components just for that
  worker type, keeping `WorkerFactory` clean.

- Move `NotificationWorkerFactory` to the new model.

- Implement `PruneCacheWorker`, and remove the code from
 `CachedTimelineViewModel`.

- Create the periodic worker in `TuskyApplication`, ensuring that the
  database is only pruned when the device is idle.

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/worker/PruneCacheWorker.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/worker/PruneCacheWorker.kt
@@ -34,9 +34,9 @@ class PruneCacheWorker(
     private val accountManager: AccountManager
 ) : CoroutineWorker(appContext, workerParams) {
     override suspend fun doWork(): Result {
-        accountManager.activeAccount?.id?.let { accountId ->
-            Log.d(TAG, "Pruning database using account ID: $accountId")
-            appDatabase.timelineDao().cleanup(accountId, MAX_STATUSES_IN_CACHE)
+        for (account in accountManager.accounts) {
+            Log.d(TAG, "Pruning database using account ID: ${account.id}")
+            appDatabase.timelineDao().cleanup(account.id, MAX_STATUSES_IN_CACHE)
         }
         return Result.success()
     }
--- a/app/src/main/java/com/keylesspalace/tusky/worker/PruneCacheWorker.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/worker/PruneCacheWorker.kt
@@ -15,14 +15,13 @@
  * see <http://www.gnu.org/licenses>.
  */
 
-package com.keylesspalace.tusky.components.timeline
+package com.keylesspalace.tusky.worker
 
 import android.content.Context
 import android.util.Log
 import androidx.work.CoroutineWorker
 import androidx.work.ListenableWorker
 import androidx.work.WorkerParameters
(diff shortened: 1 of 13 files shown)


nikclayton:
This was originally in https://github.com/tuskyapp/Tusky/pull/3436.

In that PR it's a necessary step to remove the database as a dependency in CachedTimelineViewModel (from the view model any access to the database should be via the repository).

I've split it out in to its own PR because it's an independent change, so will reduce the amount of work that has to be done to review 3436 (which is getting close to being ready for review).

gradle/libs.versions.toml
```diff
@@ -91,6 +91,7 @@ androidx-swiperefreshlayout = { module = "androidx.swiperefreshlayout:swiperefre
 androidx-test-junit = { module = "androidx.test.ext:junit", version.ref = "androidx-junit" }
 androidx-viewpager2 = { module = "androidx.viewpager2:viewpager2", version.ref = "androidx-viewpager2" }
 androidx-work-runtime = { module = "androidx.work:work-runtime", version.ref = "androidx-work" }
+androidx-work-runtime-ktx = { module = "androidx.work:work-runtime-ktx", version.ref = "androidx-work" }
```
connyduck, line 94: I'd remove `androidx-work-runtime` in return, `androidx-work-runtime-ktx` includes it anyway

app/src/main/java/com/keylesspalace/tusky/worker/PruneCacheWorker.kt
```diff
@@ -0,0 +1,58 @@
...
+
+/** Prune the database cache of old statuses. */
+class PruneCacheWorker(
+    appContext: Context,
+    workerParams: WorkerParameters,
+    private val appDatabase: AppDatabase,
+    private val accountManager: AccountManager
+) : CoroutineWorker(appContext, workerParams) {
+    override suspend fun doWork(): Result {
+        accountManager.activeAccount?.id?.let { accountId ->
```
connyduck, line 37: why not for all accounts?

connyduck (commented):
very nice

gradle/libs.versions.toml
nikclayton: Done.

app/src/main/java/com/keylesspalace/tusky/worker/PruneCacheWorker.kt
nikclayton: Consistency with the previous code. But now you've pointed it out I agree, that does make sense to do all of them. Done.

connyduck (approved):

Lakoja:
With this (?) I get a regression for the notification worker (pull requests):

java.lang.IllegalStateException: Expedited WorkRequests require a Worker to provide an implementation for `getForegroundInfo()`

This is only visible in Logcat but prohibits notifications.

At the moment I cannot easily fix it (either by disabling the OneTimeWorkRequest or trying to move the NotificationWorker to a CoroutineWorker).

PR #3651, opened 2023-05-14
Throttle UI actions instead of debouncing

nikclayton:
Introduce Flow<T>.throttleFirst(). In a flow this emits the first value, and each value afterwards that is > some timeout after the previous value.

This prevents accidental double-taps on UI elements from generating multiple-actions.

The previous code used debounce(). That has a similar effect, but with debounce() the code has to wait until after the timeout period has elapsed before it can process the action, leading to an unnecessary UI delay.

With throttleFirst a value is emitted immediately, there's no need to wait. It's subsequent values that are potentially throttled.

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/util/FlowExtensions.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/util/FlowExtensions.kt
@@ -19,25 +19,29 @@ package com.keylesspalace.tusky.util
 
 import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.flow
+import kotlin.time.Duration
+import kotlin.time.ExperimentalTime
+import kotlin.time.TimeMark
+import kotlin.time.TimeSource
 
 /**
  * Returns a flow that mirrors the original flow, but filters out values that occur within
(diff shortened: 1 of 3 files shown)


nikclayton:
This was originally in https://github.com/tuskyapp/Tusky/pull/3436. I've split it out so that reviewing that PR will be a bit easier.

app/src/main/java/com/keylesspalace/tusky/util/FlowExtensions.kt
```diff
@@ -0,0 +1,61 @@
...
+ *
+ * produces the following emissions.
+ *
+ * ```text
+ * 1, 4, 5
+ * ```
+ *
+ * @see kotlinx.coroutines.flow.debounce
+ */
+fun <T> Flow<T>.throttleFirst(timeoutMillis: Long): Flow<T> = flow {
```
connyduck, line 51: Tests for this would be awesome

nikclayton: Done.

connyduck (approved):

PR #3653, opened 2023-05-14
Make AccountPreferenceDataStore injectable

nikclayton:
This will make tests that need it easier.

- Rename from AccountPreferenceHandler
- Inject its dependencies
- Create an injectable CoroutineScope it can use for launching coroutines
- Use it in AccountPreferences


nikclayton:
This was originally in https://github.com/tuskyapp/Tusky/pull/3436. I've split it out so that reviewing that PR will be a bit easier.

This PR doesn't introduce tests that need this functionality, but 3436 does.

app/src/main/java/com/keylesspalace/tusky/settings/AccountPreferenceDataStore.kt
```diff
@@ -1,15 +1,21 @@
...
+import kotlinx.coroutines.launch
+import javax.inject.Inject
 
-class AccountPreferenceHandler(
-    private val account: AccountEntity,
+class AccountPreferenceDataStore @Inject constructor(
     private val accountManager: AccountManager,
-    private val dispatchEvent: (PreferenceChangedEvent) -> Unit
+    private val eventHub: EventHub,
+    @ApplicationScope val externalScope: CoroutineScope
```
charlag, line 16: why is it called `external`?

charlag (approved):

app/src/main/java/com/keylesspalace/tusky/settings/AccountPreferenceDataStore.kt
nikclayton: That's from https://developer.android.com/topic/architecture/data-layer#make_an_operation_live_longer_than_the_screen, where the examples there (and in the accompanying blog post at https://medium.com/androiddevelopers/coroutines-patterns-for-work-that-shouldnt-be-cancelled-e26c40f142ad) use the name `externalScope`.

The naming choice is not properly explained, but I think the intent is to communicate that although this is a (coroutine) scope, the code that uses it should not make assumptions about the lifetime of the scope.

So it can't be assumed to be a particular `lifecycleScope`, or `viewModelScope`, `GlobalScope`, etc. Just one that the code that created this object knows will last long enough for the operations.

If it's injected then it will be an `@ApplicationScope`, but if the object is constructed "by hand" then it could be any scope.

app/src/main/java/com/keylesspalace/tusky/di/CoroutineScopeModule.kt
```diff
@@ -0,0 +1,44 @@
...
+ * Scope for potentially long-running tasks that should outlive the viewmodel that
+ * started them. For example, if the API call to bookmark a status is taking a long
+ * time, that call should not be cancelled because the user has navigated away from
+ * the viewmodel that made the call.
+ *
+ * @see https://developer.android.com/topic/architecture/data-layer#make_an_operation_live_longer_than_the_screen
+ */
+@Retention(AnnotationRetention.BINARY)
+@Qualifier
+annotation class ApplicationScope
```
connyduck, line 37: I don't think I understand what this annotation does? Without it everything works just the same, no?

nikclayton: It doesn't.

`AccountPreferenceDataStore.kt:16` has this line:

```
@ApplicationScope val externalScope: CoroutineScope
```

Dagger needs to know how to provide that `CoroutineScope`. But there may be many different CoroutineScopes with different dispatchers etc.

So this annotation here means that when a `@ApplicationScope ...: CoroutineScope` is injected, use this provider.

If you remove the `@ApplicationScope` annotation you get this error:

```
tusky\di\AppComponent.java:9: error: [Dagger/MissingBinding] kotlinx.coroutines.CoroutineScope cannot be provided without an @Provides-annotated method.
```

because Dagger doesn't know how to provide the correct `CoroutineScope` instance.

connyduck: No, I mean delete completely. Basically it is just here in case another injectable Scope is added in the future. Superflous if you ask me.

app/src/main/java/com/keylesspalace/tusky/settings/AccountPreferenceDataStore.kt
connyduck, line 16: externalScope could be private

nikclayton: Done.

app/src/main/java/com/keylesspalace/tusky/di/CoroutineScopeModule.kt
nikclayton: From a pure code perspective you could remove the annotation and the module could just be:

```kotlin
@Module
class CoroutineScopeModule {
    @Provides
    fun providesCoroutineScope() = CoroutineScope(SupervisorJob() + Dispatchers.Default)
}
```

But the annotation provides additional context to the next person who's looking at the code, as well as a handy place to put additional explanatory Javadoc.

connyduck (approved):

Issue #3494, opened 2023-03-29
Font diff between notification action and content


nikclayton:
![Screenshot_20230329-134749.png](https://user-images.githubusercontent.com/773100/228527350-c9dc6171-f03d-45f1-9238-538e95c97357.png)

Per this screenshot there's a font diff between the first line, the notification action, and subsequent lines, the content.

So it looks weird that the username, repeated on both lines, is displayed differently.

nikclayton:
Pixel 4a

nikclayton:
I see this on a physical device, but not an emulator.

Some checking shows it's the `@android:style/TextAppearance.DeviceDefault.Small` in `item_status_notification.xml`. That doesn't appear to be used in any other layout, so can be safely deleted.

PR #3663, opened 2023-05-16
Remove android:textAppearance="@android:style/TextAppearance.DeviceDefault.Small"

nikclayton:
It caused text size differences between the text in this view and all the other textviews in this layout.

It's not used in other layouts.

Fixes https://github.com/tuskyapp/Tusky/issues/3494


connyduck (approved):
nice find

Issue #3295, opened 2023-02-10
Additional consistency in status layouts


nikclayton:
# Background

Continuing work on https://github.com/tuskyapp/Tusky/issues/3037 I want to display the list of scheduled statuses (and also eventually the list of draft statuses) in the same fashion as statuses are displayed on other timelines.

I checked, and there's quite a lot of inconsistency in how that's done. So this is a cleanup opportunity.

At the moment, statuses are displayed using:

- `item_status`
- `item_status_detailed`
- `item_status_edit`
- `item_status_notification`
- `item_follow_request` (special case, just the user info)

In general, a status display consists of four main sections, displayed vertically one after another.

- "Action" information ("X favourited this status", optional)
- "Metadata" (avatar, display name, user name, timestamp, etc)
- Status body (content warning, text, media, poll, preview card)
- "Action buttons" (reply, favourite, bookmark, etc)

How those sections are displayed differs.

Sometimes that's probably reasonable, like the differences between `item_status` and `item_status_detailed` around text size, layout choices, and so on.

The other times I think it's unintentional.

# Differences

Differences I've noted so far (I'm ignoring `item_status_detailed`, as that's deliberately a different presentation).

## Display name and username split

Those are on one line in `item_status` and `item_status_notification`

Two lines on `item_status_followrequest`.

`item_status_edit` doesn't display the username, and unlike the other views, doesn't use a heavier font weight for the poster's name.

## Location of the content warning

It's to the right of the avatar on `item_status` and `item_status_notification`.

It's below the avatar on `item_status_edit`.

## Location of the "Action" information

The exact location and alignment of this is inconsistent between:

- `item_status`
- `item_status_edit`
- `item_follow_request`
- `item_status_notification`

## Content alignment

The text that appears to the right of the avatar (display name, etc) is aligned inconsistently.

Sometimes it's aligned with the top of the avatar (`item_follow_request`). Other times it's aligned below the top of the avatar (`item_status`). Sometimes it's aligned below it (`item_status_notification`).

## Heavier text

To stand out, some text needs to be "bold". Some views set `android:textStyle="normal|bold"`. Others change the font weight for the same view to `medium` instead. `medium` is not the same as `bold`, and looks subtly different.

## Layout grid

Android recommendation is to align items to an 8dp grid, dropping to a 4dp grid for smaller pieces of text.

Tusky's all over the place.

- Some things use a 16dp grid, other things are 14dp, or 12dp, or 17dp (!)
[...]

PR #3673, opened 2023-05-18
Initial work to regularise item layouts

nikclayton:
The current layout "grid" is very inconsistent, 8dp, 4dp, some layouts use 16dp grid, 14dp, or 12dp, or 17dp (!). There's no guidance for contributors on what to use, and it's easy for mistakes to be made.

Start to fix this:

- Add a document for contributors with an initial layout guide
- Create dimension resources for an 8dp grid and use them in `item_status.xml` and `item_status_detailed.xml`.
- Apply margin, where necessary, to only view top and start
- Add a lint check to block future use of raw dimensions
- Update the lint baseline to grandfather in existing usage

Relates to: https://github.com/tuskyapp/Tusky/issues/3295


nikclayton:
The lint check's pretty handy -- it means that cleanup work on existing layouts can be done bit by bit (instead of trying to land a PR that changes all the layouts at once) while preventing new problems from creeping in.

The lint checks show up automatically in Android Studio as well -- if you're editing a layout that uses a raw dimension value you get a warning.

I'm going to mark this as draft, create some screenshots showing before/after, and then mark as ready for review.

nikclayton:
Here's the before / after of the regular status view

nikclayton:
Here's an annotated version of that:

tusky-compare

nikclayton:
And there's the same for the detailed view.

nikclayton:
Annotated:

tusky-compare-detailed

PR #3684, opened 2023-05-22
Fix ghost tap on media tab

maxiin:
In an account media tab, if there is media that was hidden due to being tagged as sensitive and the account had the "alwaysShowSensitiveMedia" set to true, the media would show revealed, but after tapping on the image nothing would happen, because the media would still try to "be revealed"
This fixes the problem caching the alwaysShowSensitiveMedia variable from the active account, is it a good solution?

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/components/account/media/AccountMediaRemoteMediator.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/components/account/media/AccountMediaRemoteMediator.kt
@@ -29,7 +29,7 @@ import retrofit2.HttpException
 class AccountMediaRemoteMediator(
     private val api: MastodonApi,
     private val accountManager: AccountManager,
-    private val viewModel: AccountMediaViewModel,
+    private val viewModel: AccountMediaViewModel
 ) : RemoteMediator<String, AttachmentViewData>() {
 
     override suspend fun load(
--- a/app/src/main/java/com/keylesspalace/tusky/components/account/media/AccountMediaRemoteMediator.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/components/account/media/AccountMediaRemoteMediator.kt
@@ -60,7 +60,7 @@ class AccountMediaRemoteMediator(
             }
 
             val attachments = statuses.flatMap { status ->
-                AttachmentViewData.list(status, accountManager.activeAccount)
+                AttachmentViewData.list(status, accountManager.activeAccount?.alwaysShowSensitiveMedia ?: false)
             }
 
             if (loadType == LoadType.REFRESH) {
--- a/app/src/main/java/com/keylesspalace/tusky/components/account/media/AccountMediaRemoteMediator.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/components/account/media/AccountMediaRemoteMediator.kt
@@ -20,6 +20,7 @@ import androidx.paging.LoadType
 import androidx.paging.PagingState
 import androidx.paging.RemoteMediator
 import com.keylesspalace.tusky.components.timeline.util.ifExpected
+import com.keylesspalace.tusky.db.AccountManager
 import com.keylesspalace.tusky.network.MastodonApi
 import com.keylesspalace.tusky.viewdata.AttachmentViewData
 import retrofit2.HttpException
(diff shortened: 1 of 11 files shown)


connyduck (changes requested):
Nice find!

I think for fixing this you should investigate why `AttachmentViewData.isRevealed` is false when `alwaysShowSensitiveMedia` is true.
Looks like when mapping attachments to `AttachmentViewData` the `alwaysShowSensitiveMedia` needs to be taken into account and `isRevealed` set to the correct value - right now it is just `!sensitive`. In return the check in `AccountMediaGridAdapter` might no longer be necessary.

maxiin:
Good solution, will try to change to that!

maxiin:
I got a bit confused with the injectable system since I've never used it before, but I think this is the best solution I could come up with.

connyduck:
Yes thats better!

1) Instead of having the whole `Account` as parameter to `list`, can you give it the `alwaysShowSensitiveMedia` Boolean directly? And since `list` is called from a couple of places where `alwaysShowSensitiveMedia` does not matter, you could give it a default `false` and only pass the correct value where needed (which is in `AccountMediaRemoteMediator`).
2) There is a `alwaysShowSensitiveMedia` check in `AccountMediaGridAdapter` that is no longer needed now.
3) You need to fix the code formatting so our CI checks suceeds

connyduck:
You can set a default parameter like this `fun list(status: Status, alwaysShowSensitiveMedia: Boolean = false)` then you don't need to pass it all the time `list(status)`

maxiin:
Sorry! been busy with work, forgot to check, will try to fix today

maxiin:
Fixed, was following bitrise and it failed on unit tests, but I could not replicate in my machine, is it a bitrise error?

connyduck (approved):

connyduck:
Maybe something to add to the v23 release @nikclayton?

app/src/main/java/com/keylesspalace/tusky/components/account/media/AccountMediaRemoteMediator.kt
```diff
@@ -58,7 +60,7 @@ class AccountMediaRemoteMediator(
             }
 
             val attachments = statuses.flatMap { status ->
-                AttachmentViewData.list(status)
+                AttachmentViewData.list(status, accountManager.activeAccount?.alwaysShowSensitiveMedia ?: false)
```
nikclayton, line 63: The active account can change over the lifecycle of this object (there's a race condition between the active account changing and the object being destroyed).

The safe thing to do is add an `activeAccount` property, like

```
val activeAccount = accountManager.activeAccount!!
```

and then use it here.

app/src/main/java/com/keylesspalace/tusky/components/notifications/NotificationsFragment.kt
```diff
@@ -499,7 +499,7 @@ class NotificationsFragment :
 
     override fun onViewMedia(position: Int, attachmentIndex: Int, view: View?) {
         val status = adapter.peek(position)?.statusViewData?.status ?: return
-        super.viewMedia(attachmentIndex, list(status), view)
+        super.viewMedia(attachmentIndex, list(status, false), view)
```
nikclayton, line 502: What's this `false` for?

app/src/main/java/com/keylesspalace/tusky/viewdata/AttachmentViewData.kt
```diff
@@ -34,16 +34,17 @@ data class AttachmentViewData(
     val id = attachment.id
 
     companion object {
+
```
nikclayton, line 37:
```suggestion
```

nikclayton (changes requested):

nikclayton:
@maxiin Ping -- do you have time to bring this over the finish line? If not, I'm happy to make the changes and get it merged in ahead of the next release.

nikclayton:
> @nikclayton Hi, sorry! I can't work on this right now, you can take over, yes

Thanks. https://github.com/tuskyapp/Tusky/pull/3930 is your PR, with a couple of modifications. I'll get that in shortly.

nikclayton:
@maxiin Thanks for doing the vast majority of the work on this. #3930 is merged, so I'm closing this one.

PR #3685, opened 2023-05-23
Save and restore the user's reading position more frequently

nikclayton:
The previous code gets the user's reading position *once*, when the viewmodel is created, and uses that whenever it needs to be restored.

This is a problem. Suppose the user is a few days behind on their notifications, and opens the app.

The reading position is restored, as expected. They scroll up to start reading newer notifications.

Then they change their notification filters. This causes the notifications list to change, and when it does their reading position is set back to what it was when they first switched to the Notifications tab.

Fix this by:

NotificationsFragment:
- Save the reading position whenever the user stops scrolling.

NotificationsViewModel:
- Use the saved reading position whenever the list of notifications can change, not just when the view model is created.

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/components/notifications/NotificationsFragment.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/components/notifications/NotificationsFragment.kt
@@ -41,6 +41,7 @@ import androidx.recyclerview.widget.DiffUtil
 import androidx.recyclerview.widget.DividerItemDecoration
 import androidx.recyclerview.widget.LinearLayoutManager
 import androidx.recyclerview.widget.RecyclerView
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/keylesspalace/tusky/components/notifications/NotificationsFragment.kt
```diff
@@ -193,6 +194,18 @@ class NotificationsFragment :
                     }
                 }
             }
+
+            override fun onScrollStateChanged(recyclerView: RecyclerView, newState: Int) {
```
nikclayton, line 198:
```suggestion
            @Suppress("SyntheticAccessor")
            override fun onScrollStateChanged(recyclerView: RecyclerView, newState: Int) {
```

connyduck (approved):

Issue #2663, opened 2022-08-11
Old toots stuck on top of home timeline


LittleFox94:
Hi,

After going on my profile and scrolling to an old toot (knew I tooted it 2022-07-25, so just scrolled there), my home timeline now has toots from roughly that period stuck on top - fresh toots starting ~8 screen sizes below them.

The stuck ones seem to be non-contiguous, I think I should have had more toots in home as it shows. The fresh ones are refreshed, but stay at their place below the stuck ones.

I tried purging the app cache and switching tabs and accounts, without success.

This is pretty obscure, not sure what other info could be useful here or how to reproduce it.. some slightly rusted android dev skills are available, so feel free to ask for any adb thing (or something else) that might help diagnose the problem :)

* * * *
- Tusky Version: 19.0
- Android Version: 12 (SQ3A.220705.003)
- Android Device: Pixel 6 Pro
- Mastodon instance (if applicable): mastodon.social

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate (browsed issues for some pages).

connyduck:
Those old toots stuck on top, could they be from a different account that is logged into Tusky or are they definitely from the right one, just in the wrong order?

(logging out and in again or clearing the app data completely will solve the problem temporarily)

LittleFox94:
Now that you ask, they actually are from the Home timeline of another account.

And I noticed something even more strange: going to their authors profile via their avatar on the toot brings me to a completely different profile - for every toot by the same author, I go to the same other profile, but multiple authors could go to the same profile (clicking "toby" on stuck toot always brings up "mmcc", clicking any of two different bots gives me Eugen, clicking me gives another account than the one for "toby")

LittleFox94:
Those stuck toots could belong to a logged-in account on an instance that is currently offline - maybe some weird form of "old cache + cannot refresh -> not cleared from view"?

connyduck:
Ok so this seems to be another incarnation of the "ghost timeline" bug that was supposed to be fixed by https://github.com/tuskyapp/Tusky/commit/f419e83c16438456559163463785ec68ef98fdd3

Basically what happens is that the timeline of one account gets written to the database with the id of another.

It could be reproduced by logging out of one account while a refresh was still ongoing, but since that is fixed there must be another way that it can happen as well 🤔

PR #3688, opened 2023-05-25
Don't overwrite the active account when composing from a notification

nikclayton:
Fix a bug where the active account can be overwritten.

1. Have two accounts logged in to Tusky, A and B
2. Open Tusky as account A
3. Send a DM to account B (doesn't have to be a DM, just anything that creates a notification for account B)
4. Restart Tusky so the Android notification for the DM is displayed immediately. You are still acting as account A.
5. Drag down the Android notification, you should see two options, "Quick reply" and "Compose"
6. Tap "Compose"
7. ComposeActivity will start. You are now acting as account B. Compose and send a reply
8. You'll be returned to the "Home" tab.

The UI will show you are still account A (i.e., it's account A's avatar at the top of the screen, if you have the "Show username in toolbars" option turned on it will be account A's username in the toolbar).

But you are now seeing the home timeline for account B.

Fix this.

ComposeViewModel
- Do not rely on the active account in sendStatus(), receive the account ID as a parameter

ComposeActivity
- Use either the account ID from the intent, or the current active account. **Do not** change the active account
- Pass the account ID to use in the sendStatus() call

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/components/compose/ComposeActivity.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/components/compose/ComposeActivity.kt
@@ -142,6 +142,9 @@ class ComposeActivity :
     private lateinit var emojiBehavior: BottomSheetBehavior<*>
     private lateinit var scheduleBehavior: BottomSheetBehavior<*>
 
+    /** The account that is being used to compose the status */
+    private lateinit var activeAccount: AccountEntity
+
     private var photoUploadUri: Uri? = null
 
     private val preferences by unsafeLazy { PreferenceManager.getDefaultSharedPreferences(this) }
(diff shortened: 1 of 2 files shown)


nikclayton:
Possibly implicated in https://github.com/tuskyapp/Tusky/issues/2663 ...?

connyduck:
Very nice find!

I believe the correct fix for this is to not open `ComposeActivity` directly, but `MainActivity` and then `ComposeActivity`. This way `MainActivity` can also switch accounts, and it is consistent with other places where we open `ComposeActivity` from the outside (e.g. when sharing to Tusky or using a shortcut).

nikclayton:
Sure, but can we fix the bug while preserving existing behaviour first, and then move on to changing the behaviour?

connyduck (approved):
I think the behavior is wrong and it is the bug, but sure

nikclayton:
I agree. I've filed https://github.com/tuskyapp/Tusky/issues/3695 to track fixing this so it's consistent with other parts of the code.

Issue #3689, opened 2023-05-25
very old (1y+) notifications are shown


nehemiagurl:
Starting today, very old notifications in two of my instances are shown.  Clicking the bell icon to auto-scroll all the way to the top doesn't help: it just loads newer, but still 1y+ old notifications. Same with the three dot refresh button.
Closing the app and force stopping did not assist. Switching accounts back and forth didn't either.
![Screenshot_20230525-181258_Tusky](https://github.com/tuskyapp/Tusky/assets/109424320/3c1ed0cc-b37c-44c5-bdf2-25614ab48691)
![Screenshot_20230525-181340_Tusky](https://github.com/tuskyapp/Tusky/assets/109424320/4f576203-9d44-4c61-be3b-40f1c616cc61)

~Weirdly enough this doesn't happen in all of my accounts, just two of them for unclear reasons. It's also not the admin privileges that give the sign-up and report notifications that are to blame, because I am also sign in with the main admin account for both instances and in them this doesn't happen.~
EDIT: okay now that I switched away from them and back it does happen in them as well. so it seems the admin-type notifications (signups and reports) are to blame.

* * * *
- Tusky Version: 22.0 beta 5
- Android Version: 12
- Android Device: Galaxy A12
- Mastodon instance (if applicable): tooot.im, toki.social

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

nikclayton:
Thanks for the report, I've been trying to track this down. And sorry for the hassle.

You mentioned you're signed in to multiple accounts.

Do you remember if the following has happened:

1. You were using one of your accounts.
2. You received an **Android** notification about a message in one of the other accounts
3. From the **Android** notification there was a "Compose" option, so you could reply directly from the notification.
4. Did you use that "Compose" feature

?

If you did I think this is a bug I've identified and am fixing in https://github.com/tuskyapp/Tusky/pull/3688.

If you didn't then something else is going on.

nikclayton:
# tl;dr

The only thing I can think of is that there's another bug like https://github.com/tuskyapp/Tusky/pull/3688 where the active account is being changed. Something like the following.

- User is on two servers, A and B. A is extremely busy, thousands of users, lots of activity. B is quiet.
- **If** notification IDs on a server are implemented as monotonically increasing ints, and are per-server, not per user, this would mean that the notification IDs on server are A very high, and the notification IDs on server B are very low(*)
[...]

PR #3692, opened 2023-05-27
Notification tab cleanups

nikclayton:
- Use NO_POSITION instead of hardcoding 0.
- Don't set a state restoration policy, PagingDataAdapter already does that
- Return the closest item, not just the closest page, in getRefreshKey


nikclayton:
Noticed while I was going over the code trying to figure out https://github.com/tuskyapp/Tusky/issues/3689

connyduck (approved):

PR #3693, opened 2023-05-27
Blackbox notification data collection and optional reporting

nikclayton:
Collect data about notification fetching that the user can choose to report

- Add BlackBox, which records logs to a circular buffer
- Add extra logging in to the black box for notification related actions
- Add an activity to show the user the data, and allow them to report it


nikclayton:
This is kinda janky, but I want to pull it all out before the release happens, so I don't care that it's not using string resources (no need to bother translators with this) or that the UI isn't super polished.

I just need to get enough information to fix the problem.

I've drafted a doc / FAQ for users to reference from the beta release notes and post a link to from the Tusky account at https://gist.github.com/nikclayton/ecc2d4bceb653bd671a05b305b10d876. We can discuss that too.

app/src/main/java/com/keylesspalace/tusky/BlackBox.kt
```diff
@@ -0,0 +1,78 @@
...
+    /** Get an item from the buffer */
+    @Suppress("UNCHECKED_CAST")
+    operator fun get(index: Int): T = when {
+        capacity == 0 || index > capacity || index < 0 -> throw IndexOutOfBoundsException("$index")
+        capacity == data.size -> data[(head + index) % data.size]
+        else -> data[index]
+    } as T
+
+    /** Buffer as a list. */
+    @Suppress("UNCHECKED_CAST")
```
connyduck, line 65: Redundant suppression

connyduck (approved):
There are a few small things that could be improved about the code, but for a quick temporary solution this is defintitely good enough

PR #3697, opened 2023-05-29
Write notification account information to the correct account

nikclayton:
Don't use `accountManager.activeAccount` throughout the code.

Instead, get the active account once, and use that over the life of the viewmodel.

As shown in https://github.com/tuskyapp/Tusky/issues/3689#issuecomment-1567219338 the active account can change before the view model is destroyed, and if that happens account information for the account will be written to the wrong account.

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/BaseActivity.java
+++ b/app/src/main/java/com/keylesspalace/tusky/BaseActivity.java
@@ -212,6 +212,13 @@ public abstract class BaseActivity extends AppCompatActivity implements Injectab
         }
     }
 
+    // TODO: This changes the accountManager's activeAccount property, but does not do any
+    // of the work that AccountManager.setActiveAccount() does. In particular:
+    //
+    // - The current active account is not saved
+    // - The account passed as parameter here goes not have its `isActive` property set
+    //
+    // Is that deliberate? Or is this a bug?
     public void openAsAccount(@NonNull String url, @NonNull AccountEntity account) {
         accountManager.setActiveAccount(account);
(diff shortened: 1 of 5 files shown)


nikclayton:
Logging (and inspecting the database) shows that this is now writing the data to the correct account.

For notifications only -- this doesn't fix the broader data corruption issue, which would be a much more involved PR, and not one I want to tackle during the beta period.

There's one comment in `BaseActivity.java.openAsAccount()` that I'd like feedback on. If that's a real problem then I'll create an issue for that as well so it doesn't get lost.

app/src/main/java/com/keylesspalace/tusky/BaseActivity.java
```diff
@@ -212,6 +212,13 @@ public void showAccountChooserDialog(CharSequence dialogTitle, boolean showActiv
         }
     }
 
+    // TODO: This changes the accountManager's activeAccount property, but does not do any
+    // of the work that AccountManager.setActiveAccount() does. In particular:
+    //
+    // - The current active account is not saved
+    // - The account passed as parameter here goes not have its `isActive` property set
+    //
+    // Is that deliberate? Or is this a bug?
```
connyduck, line 221: I think it is a bug. Does not really make sense this way.

nikclayton: OK. I've filed https://github.com/tuskyapp/Tusky/issues/3702 to track looking in to that in more detail after v22 is released.

connyduck (approved):

Issue #3576, opened 2023-04-26
Don't prompt to save empty draft


bjorn3:
Nor prompt to save a draft containing only mentions without any other text. If you accidentally clicked on the button to draft a toot, the question to save the draft when clicking away doesn't have any value and is somewhat annoying.

* * * *
- Tusky Version: 21.0
- Android Version: Not relevant
- Android Device: Not relevant
- Mastodon instance (if applicable): Not relevant

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

connyduck:
Can you please check if that happens on 22 beta 1 as well? I can't reproduce it there, it might be fixed already.

bjorn3:
Looks like it still happens if you go to someones profile, click to draft a new toot, delete the mention that is automatically added and finally click to close the drafting view.

Lakoja:
I can partly confirm this. It is a bit edge-case-y:

* On a profile tap "create"
* Delete the mention
* Exit
There will be a needless "Save?" dialog.

However there is no dialog if you exit directly as second step above.

Tusky has a notion of "initial state" for status composing and compares the state on exit with the initial one. In the example above the initial state includes the mention.

Expanding from that the case "empty on exit" would need at least two new checks:

* "is new and empty": no dialog
* "is edit draft and empty": new dialog with "Do you want to delete the draft?"

connyduck:
This is fixed

PR #3699, opened 2023-05-30
Improve prompt when draft is empty

pt2121:
address #3576 

I'm sort of following https://github.com/tuskyapp/Tusky/issues/3576#issuecomment-1527504752.

When the user is closing the compose view,
- if it's new and empty, don't show a prompt.
- if it's an existing draft and now empty, ask if the user wants to delete it or continue editing. I don't think there is much value in saving an empty draft.

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/components/compose/ComposeActivity.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/components/compose/ComposeActivity.kt
@@ -76,6 +76,7 @@ import com.keylesspalace.tusky.R
 import com.keylesspalace.tusky.adapter.EmojiAdapter
 import com.keylesspalace.tusky.adapter.LocaleAdapter
 import com.keylesspalace.tusky.adapter.OnEmojiSelectedListener
+import com.keylesspalace.tusky.components.compose.ComposeViewModel.ConfirmationPromptType
 import com.keylesspalace.tusky.components.compose.dialog.CaptionDialog
 import com.keylesspalace.tusky.components.compose.dialog.makeFocusDialog
 import com.keylesspalace.tusky.components.compose.dialog.showAddPollDialog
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/keylesspalace/tusky/components/compose/ComposeActivity.kt
```diff
@@ -1124,7 +1124,12 @@ class ComposeActivity :
     private fun handleCloseButton() {
         val contentText = binding.composeEditField.text.toString()
         val contentWarning = binding.composeContentWarningField.text.toString()
-        if (viewModel.didChange(contentText, contentWarning)) {
+        if (viewModel.composeKind == ComposeKind.NEW && viewModel.isEmpty(contentText, contentWarning)) {
```
nikclayton, line 1127: This sort of business logic probably belongs in the view model, probably in the existing `didChange` function (perhaps renamed).

The function could then return an enum that corresponds to the type of confirmation dialog the activity should show (if any).

Something like:

```kotlin
when (viewModel.didChange(contextText, contentWarning)) {
    NO_CHANGE -> { /* Stop uploads and finish */ }
    SAVE_OR_DISCARD -> getSaveAsDraftOrDiscardDialog(...)
    UPDATE_OR_DISCARD -> getUpdateDraftOrDiscardDialog(...)
    CONTINUE_OR_DISCARD -> getContinueEditingOrDiscardDialog(...)
}
```

app/src/main/res/values/strings.xml
```diff
@@ -481,6 +481,7 @@
     <string name="action_remove">Remove</string>
     <string name="lock_account_label">Lock account</string>
     <string name="lock_account_label_description">Requires you to manually approve followers</string>
+    <string name="compose_delete_draft">Delete this draft?</string>
```
nikclayton, line 484:
```suggestion
    <string name="compose_delete_draft">Delete draft?</string>
```

For consistency with "Save draft?"

nikclayton (changes requested):

app/src/main/java/com/keylesspalace/tusky/components/compose/ComposeActivity.kt
pt2121: Refactored and moved the logic to the view model.

app/src/main/java/com/keylesspalace/tusky/components/compose/ComposeViewModel.kt
```diff
@@ -486,6 +511,14 @@ class ComposeViewModel @Inject constructor(
     private companion object {
         const val TAG = "ComposeViewModel"
     }
+
+    enum class ConfirmationPromptType {
```
nikclayton, line 515: Nit: For consistency with other code, `ConfirmationKind` would probably be better (see how `...Kind` is used as a suffix fora  lot of enums or sealed classes like this elsewhere in the code).

app/src/main/java/com/keylesspalace/tusky/components/compose/ComposeViewModel.kt
```diff
@@ -213,7 +213,28 @@ class ComposeViewModel @Inject constructor(
         this.markMediaAsSensitive.value = this.markMediaAsSensitive.value != true
     }
 
-    fun didChange(content: String?, contentWarning: String?): Boolean {
+    fun handleCloseButton(contentText: String?, contentWarning: String?): ConfirmationPromptType {
+        return if (didChange(contentText, contentWarning)) {
+            when (composeKind) {
+                ComposeActivity.ComposeKind.NEW -> if (isEmpty(contentText, contentWarning)) {
```
nikclayton, line 219: Nit: Please import `ComposeActivity.ComposeKind`, so the leading `ComposeActivity` can be dropped from these branches.

app/src/main/java/com/keylesspalace/tusky/components/compose/ComposeViewModel.kt
```diff
@@ -223,6 +244,10 @@ class ComposeViewModel @Inject constructor(
         return modifiedInitialState || textChanged || contentWarningChanged || mediaChanged || pollChanged || didScheduledTimeChange
     }
 
+    private fun isEmpty(content: String?, contentWarning: String?): Boolean {
```
nikclayton, line 247: I'm a little concerned about this function.

It takes two parameters, but it operates on five values in total (`modifiedInitialState`, `media`, and `poll` being the others). So a casual reading of the code could lead to confusion.

I don't see a particularly clean way to fix that without a considerable rewrite of the viewmodel, so I'm mentioning this in case you can think of something better. If not, there's a rewrite coming, so it's not a big issue if this goes in as is.

nikclayton (changes requested):

app/src/main/java/com/keylesspalace/tusky/components/compose/ComposeViewModel.kt
pt2121: I basically follow `didChange` but I agree that it is a bit confusing.
That said I don't see a good solution for this either. If we already have a plan for a rewrite, I'd leave this as is for now.

Issue #3706, opened 2023-06-02
[22b6] Edit Profile Fields Don't Support Text Editing Features After Activity Swap


gudenau:
Going into the "edit profile" pane, switching to a different activity and switching back disables text editing features (selecting, copying, pasting, etc.) in the "Profile metadata" fields. Fields added after an activity switch have the text editing features until another activity switch.

[YouTube video with the bug.](https://youtu.be/q6ut51sscwI)

* * * *
- Tusky Version: 22.0b6
- Android Version: 13 (TP1A.220624.014)
- Android Device: Pixel 4 XL
- Mastodon instance (if applicable): fosstodon.org

- [X] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

PR #3707, opened 2023-06-02
Ensure textview fields can be copy/pasted

nikclayton:
The Android libraries have a bug where a TextView can forget that it contains selectable text, can be pasted in to, etc.

See https://issuetracker.google.com/issues/37095917

Fix this with an extension method that toggles the selectable state to re-enable it, and use this on the profile fields when editing an account.

Fixes https://github.com/tuskyapp/Tusky/issues/3706


connyduck (approved):
Wtf, why don't they fix this in the support library

PR #3708, opened 2023-06-02
Add a menu option to load the newest notifications

nikclayton:
- Create a flow with new items (arbitrary ints) when a reload from the top should happen
- Combine this flow with notificationFilter, so changes to either of them trigger a reload
- Provide a menu item in NotificationsFragment to initiate the reload
- Handle the action in the view model


nikclayton:
Not sold on this being necessary yet, except for debugging. I've pushed it so we can have a discussion if people think it would be valuable.

connyduck:
I would like to have this option, yes

PR #3720, opened 2023-06-09
make sure the active account is always correctly saved to database

### LINKED ISSUE #3702: BaseActivity.java::openAsAccount may be buggy
`openAsAccount` changes the accountManager's activeAccount property, but does not do any of the work that AccountManager.setActiveAccount() does. In particular:
  
- The current active account is not saved
- The account passed as parameter here goes not have its `isActive` property set

This is probably a bug, per feedback on https://github.com/tuskyapp/Tusky/pull/3697/files#r1212176622, but needs proper investigation (it's called from a number of places which might depend on this behaviour).

* * * *
- Tusky Version: 21

connyduck:
closes #3702

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/BaseActivity.java
+++ b/app/src/main/java/com/keylesspalace/tusky/BaseActivity.java
@@ -212,15 +212,8 @@ public abstract class BaseActivity extends AppCompatActivity implements Injectab
         }
     }
 
-    // TODO: This changes the accountManager's activeAccount property, but does not do any
-    // of the work that AccountManager.setActiveAccount() does. In particular:
-    //
-    // - The current active account is not saved
-    // - The account passed as parameter here goes not have its `isActive` property set
-    //
-    // Is that deliberate? Or is this a bug?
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/keylesspalace/tusky/BaseActivity.java
```diff
@@ -212,15 +212,8 @@ public void showAccountChooserDialog(CharSequence dialogTitle, boolean showActiv
         }
     }
 
-    // TODO: This changes the accountManager's activeAccount property, but does not do any
-    // of the work that AccountManager.setActiveAccount() does. In particular:
-    //
-    // - The current active account is not saved
-    // - The account passed as parameter here goes not have its `isActive` property set
-    //
-    // Is that deliberate? Or is this a bug?
```
connyduck, line 221: It is definitely a bug, the setter should have never been public, and was probably only called because it looks the same as the method from Java. I tested it, works fine with this change.

nikclayton (approved):

PR #3728, opened 2023-06-09
Experiment with linting and testing using Github actions

nikclayton:
Create an action that lints, tests, and builds the green debug variant.


nikclayton:
Want to run with this for a bit and get a sense of how much faster it is (if any) than the Bitrise checks. If it's faster it could take the place of the current `primary` Bitrise workflow.

connyduck (approved):

PR #3732, opened 2023-06-10
improve account switching intents

### LINKED ISSUE #3695: Route ComposeActivity intents via MainActivity
[Copying out of https://github.com/tuskyapp/Tusky/pull/3688]

"Compose" intents from Android notifications (e.g., a notification about a message you can reply to) are currently set directly to `ComposeActivity`, where it makes sure the message is sent with the correct ID.

To be consistent with the way other intents are handled it would be better to send these through `MainActivity` so the account switch can happen there (https://github.com/tuskyapp/Tusky/pull/3688#issuecomment-1563422030).

* * * *
- Tusky Version: 21

connyduck:
Before: Intent creation all over the place, account switching logic in both `ComposeActivity` and `MainActivity`
After: Intents only created in `MainActivity` Companion, account switching only in `MainActivity`

closes #3695


Tak (approved):

app/lint-baseline.xml
```diff
@@ -3705,12 +3705,12 @@
 
     <issue
         id="SyntheticAccessor"
-        message="Access to `private` method `forwardShare` of class `MainActivity` requires synthetic accessor"
-        errorLine1="                                    forwardShare(intent)"
-        errorLine2="                                    ~~~~~~~~~~~~">
+        message="Access to `private` method `forwardToComposeActivity` of class `MainActivity` requires synthetic accessor"
+        errorLine1="                                    forwardToComposeActivity(intent)"
+        errorLine2="                                    ~~~~~~~~~~~~~~~~~~~~~~~~">
```
connyduck, line 3710: Not sure if this rule makes any sense. Worrying about method limits is so 2017.

app/src/main/java/com/keylesspalace/tusky/MainActivity.kt
```diff
@@ -1049,8 +1065,74 @@ class MainActivity : BottomSheetActivity(), ActionButtonActivity, HasAndroidInje
...
+        private const val REDIRECT_URL = "redirectUrl"
+        private const val OPEN_DRAFTS = "draft"
+        private const val ACCOUNT_ID = "accountId"
+        private const val COMPOSE_OPTIONS = "composeOptions"
+        private const val NOTIFICATION_TYPE = "notificationType"
+        private const val NOTIFICATION_TAG = "notificationTag"
+        private const val NOTIFICATION_ID = "notificationId"
+
+        /**
+         * switches the active account to the provided accountId and then stays on MainActivity
```
nikclayton, line 1077:
```suggestion
         * Switches the active account to the provided accountId and then stays on MainActivity
```

app/src/main/java/com/keylesspalace/tusky/MainActivity.kt
```diff
@@ -1049,8 +1065,74 @@ class MainActivity : BottomSheetActivity(), ActionButtonActivity, HasAndroidInje
...
+         */
+        @JvmStatic
+        fun accountSwitchIntent(context: Context, accountId: Long): Intent {
+            return Intent(context, MainActivity::class.java).apply {
+                putExtra(ACCOUNT_ID, accountId)
+            }
+        }
+
+        /**
+         * switches the active account to the accountId and takes the user to correct place according to the notification they clicked
```
nikclayton, line 1087:
```suggestion
         * Switches the active account to the accountId and takes the user to the correct place according to the notification they clicked
```

app/src/main/java/com/keylesspalace/tusky/MainActivity.kt
```diff
@@ -1049,8 +1065,74 @@ class MainActivity : BottomSheetActivity(), ActionButtonActivity, HasAndroidInje
...
+         */
+        @JvmStatic
+        fun openNotificationIntent(context: Context, type: Notification.Type, accountId: Long): Intent {
+            return accountSwitchIntent(context, accountId).apply {
+                putExtra(NOTIFICATION_TYPE, type)
+            }
+        }
+
+        /**
+         * switches the active account to the accountId and then opens the ComposeActivity with the provided options
```
nikclayton, line 1097:
```suggestion
         * Switches the active account to the accountId and then opens ComposeActivity with the provided options
```

app/src/main/java/com/keylesspalace/tusky/MainActivity.kt
```diff
@@ -1049,8 +1065,74 @@ class MainActivity : BottomSheetActivity(), ActionButtonActivity, HasAndroidInje
...
+            options: ComposeActivity.ComposeOptions,
+            accountId: Long = -1,
+            notificationTag: String? = null,
+            notificationId: Int = -1
+        ): Intent {
+            return accountSwitchIntent(context, accountId).apply {
+                action = Intent.ACTION_SEND // so it can be opened via shortcuts
+                putExtra(COMPOSE_OPTIONS, options)
+                putExtra(NOTIFICATION_TAG, notificationTag)
+                putExtra(NOTIFICATION_ID, notificationId)
```
nikclayton, line 1113: Should these be conditional on `notificationTag` being non-null and `notificationId` not being `-1`?

app/src/main/java/com/keylesspalace/tusky/MainActivity.kt
```diff
@@ -1049,8 +1065,74 @@ class MainActivity : BottomSheetActivity(), ActionButtonActivity, HasAndroidInje
...
+        /**
+         * switches the active account to the accountId and then opens the ComposeActivity with the provided options
+         * @param notificationId optional id of the notification that should be cancelled when this intent is opened
+         * @param notificationTag optional tag of the notification that should be cancelled when this intent is opened
+         */
+        @JvmStatic
+        fun composeIntent(
+            context: Context,
+            options: ComposeActivity.ComposeOptions,
+            accountId: Long = -1,
```
nikclayton, line 1105: Should this have a default? Would it be clearer at the call site if the expected account was always specified?

app/src/main/java/com/keylesspalace/tusky/MainActivity.kt
```diff
@@ -1049,8 +1065,74 @@ class MainActivity : BottomSheetActivity(), ActionButtonActivity, HasAndroidInje
...
+            return Intent(context, MainActivity::class.java).apply {
+                putExtra(ACCOUNT_ID, accountId)
+            }
+        }
+
+        /**
+         * switches the active account to the accountId and takes the user to correct place according to the notification they clicked
+         */
+        @JvmStatic
+        fun openNotificationIntent(context: Context, type: Notification.Type, accountId: Long): Intent {
```
nikclayton, line 1090: Please can all of these `...Intent` methods have a consistent param order; `context` first, then `accountId` (possibly called `tuskyAccountId`, per the earlier comment), and then the params that are specific to that intent.

app/src/main/java/com/keylesspalace/tusky/components/compose/ComposeActivity.kt
```diff
@@ -206,22 +205,7 @@ class ComposeActivity :
...
-        // If started from an intent then compose as the account ID from the intent.
-        // Otherwise use the active account. If null then the user is not logged in,
-        // and return from the activity.
-        val intentAccountId = intent.getLongExtra(ACCOUNT_ID_EXTRA, -1)
-        activeAccount = if (intentAccountId != -1L) {
-            accountManager.getAccountById(intentAccountId)
-        } else {
-            accountManager.activeAccount
-        } ?: return
+        activeAccount = accountManager.activeAccount ?: return
```
nikclayton, line 208: +1

I'll do a sweep through the code after this PR has gone in to catch other places where this should be done.

app/src/main/java/com/keylesspalace/tusky/MainActivity.kt
```diff
@@ -1049,8 +1065,74 @@ class MainActivity : BottomSheetActivity(), ActionButtonActivity, HasAndroidInje
...
+
+        /**
+         * switches the active account to the accountId and then tries to resolve and show the provided url
+         */
+        @JvmStatic
+        fun redirectIntent(context: Context, url: String, accountId: Long): Intent {
+            return Intent(context, MainActivity::class.java).apply {
+                putExtra(ACCOUNT_ID, accountId)
+                putExtra(REDIRECT_URL, url)
+            }
```
nikclayton, line 1125:
```suggestion
            return accountSwitchIntent(context, accountId).apply {
                putExtra(REDIRECT_URL, url)
            }
```

?

app/src/main/java/com/keylesspalace/tusky/util/ShareShortcutHelper.kt
```diff
@@ -72,7 +71,7 @@ fun updateShortcut(context: Context, account: AccountEntity) {
         val intent = Intent(context, MainActivity::class.java).apply {
             action = Intent.ACTION_SEND
             type = "text/plain"
-            putExtra(NotificationHelper.ACCOUNT_ID, account.id)
+            putExtra(ShortcutManagerCompat.EXTRA_SHORTCUT_ID, account.id.toString())
         }
```
nikclayton, line 75: Should this be moved in to a function in `MainActivity`, for consistency with all the others?

nikclayton (changes requested):

app/src/main/java/com/keylesspalace/tusky/MainActivity.kt
```diff
@@ -1049,8 +1065,74 @@ class MainActivity : BottomSheetActivity(), ActionButtonActivity, HasAndroidInje
...
+        }
+
+        /**
+         * switches the active account to the accountId and then opens the ComposeActivity with the provided options
+         * @param notificationId optional id of the notification that should be cancelled when this intent is opened
+         * @param notificationTag optional tag of the notification that should be cancelled when this intent is opened
+         */
+        @JvmStatic
+        fun composeIntent(
```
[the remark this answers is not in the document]
connyduck: `openNotificationIntent` is coming from notifications, `openComposeIntent` is going to compose, they are not the same

app/src/main/java/com/keylesspalace/tusky/MainActivity.kt
connyduck: I think default is fine. Added an explanation to the Kdoc.

app/src/main/java/com/keylesspalace/tusky/MainActivity.kt
```diff
@@ -232,10 +241,10 @@ class MainActivity : BottomSheetActivity(), ActionButtonActivity, HasAndroidInje
             } else if (openDrafts) {
                 val intent = DraftsActivity.newIntent(this)
                 startActivity(intent)
-            } else if (accountRequested && savedInstanceState == null) {
+            } else if (accountRequested && savedInstanceState == null && intent.hasExtra(NOTIFICATION_TYPE)) {
```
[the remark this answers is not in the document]
connyduck: That is a safeguard to prevent the same event from happening over and over again when the activity is recreated and it should actually be everywhere

app/src/main/java/com/keylesspalace/tusky/util/ShareShortcutHelper.kt
connyduck: Was wondering that as well, but it is very specific to the Shortcutmanager and should not be used anywhere else. Can still move if you think its better.

app/src/main/java/com/keylesspalace/tusky/BaseActivity.java
```diff
@@ -214,9 +214,9 @@ public void showAccountChooserDialog(CharSequence dialogTitle, boolean showActiv
 
     public void openAsAccount(@NonNull String url, @NonNull AccountEntity account) {
         accountManager.setActiveAccount(account.getId());
-        Intent intent = new Intent(this, MainActivity.class);
+        Intent intent = MainActivity.redirectIntent(this, url, account.getId());
         intent.setFlags(Intent.FLAG_ACTIVITY_NEW_TASK | Intent.FLAG_ACTIVITY_CLEAR_TASK);
```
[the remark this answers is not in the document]
connyduck: yes, lets move them

app/src/main/java/com/keylesspalace/tusky/MainActivity.kt
```diff
@@ -182,30 +184,39 @@ class MainActivity : BottomSheetActivity(), ActionButtonActivity, HasAndroidInje
...
+            val notificationId = intent.getIntExtra(NOTIFICATION_ID, -1)
+            if (notificationId != -1) {
+                // opened from a notification action, cancel the notification
+                val notificationManager = getSystemService(NOTIFICATION_SERVICE) as NotificationManager
+                notificationManager.cancel(intent.getStringExtra(NOTIFICATION_TAG), notificationId)
+            }
+
             /** there are two possibilities the accountId can be passed to MainActivity:
-             * - from our code as long 'account_id'
+             * - from our code as Long Intent Extra ACCOUNT_ID
```
nikclayton, line 198:
```suggestion
             * - from our code as Long Intent Extra TUSKY_ACCOUNT_ID
```

nikclayton (approved):

PR #3744, opened 2023-06-12
Catch missing classes in WorkerFactory

### LINKED ISSUE #3740: [Crash] nightly build crashes to homescreen instantly
Hi, since this morning Tusky nightly crashes right back to my homescreen when I want to open the app. I see a short blue-gray splashscreen for a few hundred ms, and then the app crashes.

My phone then shows me a warning "An error occurred with Tusky" that prompts me to clear the cache (which doesn't help), and then to wait for an update that has a fix.

If there's any specific logs that could help you track this down, please lmk what exactly you need and I'll supply those!

* * * *
- Tusky Version: 22.0-ac49222d
- Android Version: 12
- Android Device: Galaxy s10e
- Mastodon instance (if applicable): corteximplant.com, not sure it's related.

nikclayton:
When NotificationWorker was moved from `...components.notifications` to `...worker`, installing Tusky with this change doesn't remove any future periodic jobs queued under the old class name.

So when `Class.forName()` is called the old class name is not found, and the exception is thrown.

Handle this the same way `androidx.work.WorkerFactory` does -- catch the exception, log it, and return null.

Fixes #3740

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/worker/WorkerFactory.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/worker/WorkerFactory.kt
@@ -18,11 +18,13 @@
 package com.keylesspalace.tusky.worker
 
 import android.content.Context
+import android.util.Log
 import androidx.work.ListenableWorker
 import androidx.work.WorkerFactory
 import androidx.work.WorkerParameters
 import javax.inject.Inject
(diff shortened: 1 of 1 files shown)


nikclayton:
Fixes an active fairly serious crash on Nightly.

connyduck (approved):

Issue #3742, opened 2023-06-12
Some Trending Hashtags are cut-off


foss-:
Follow-up to https://github.com/tuskyapp/Tusky/pull/3149

Some trending hashtags are cut-off. Interestingly, when checking different accounts, a hashtag that shows fine in one account e.g. `#Berlin` may show cut-off in another account.

* * * *
- Tusky Version: 22
- Android Version: 13.0
- Android Device: Xiaomi 10T Lite
- Friendica instance: squeet.me, pirati.ca

No pattern I see, why some hashtags are cut-off or where they are cut-off.

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

<img src="https://github.com/tuskyapp/Tusky/assets/5708172/10fe5f60-e696-459f-ba6f-88f0e600f880" width="300">

PR #3745, opened 2023-06-12
fix trending tags being cut off

connyduck:
closes https://github.com/tuskyapp/Tusky/issues/3742


Tak (approved):

nikclayton (changes requested):
There's a landscape layout file that needs the same changes too.

nikclayton (approved):

PR #3747, opened 2023-06-13
Handle status edit histories with < 2 entries

### LINKED ISSUE #3743: App crashes upon trying to view edit history
Trying to view the edit history of [this post](https://spore.social/@tamarasiuda/110522895217050705) results in a crash.
Edit history is available when viewing on OP instance, but not when viewing remotely (see screenshot) which I assume is the source of the crash.
Screenshot_20230612-190544_Firefox

* * * *
- Tusky Version: 22.0
- Android Version: 12
- Android Device: Samsung Galaxy A12 
- Mastodon instance (if applicable): spore.social (posted), multiple (viewing)

nikclayton:
This can happen if the edit history has not been propogated to the user's server.

Try and fetch the edit history from the original poster's server. This may succeed (if the status was public). If it doesn't then show a specific error message.

Fixes #3743

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/components/viewthread/edits/ViewEditsFragment.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/components/viewthread/edits/ViewEditsFragment.kt
@@ -72,9 +72,6 @@ class ViewEditsFragment :
 
     private lateinit var statusId: String
 
-    /** The status's URL on the original server */
-    private var url: String? = null
-
     override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
         requireActivity().addMenuProvider(this, viewLifecycleOwner, Lifecycle.State.RESUMED)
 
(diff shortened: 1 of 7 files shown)


connyduck:
I'm very much against doing calls to instances other than the one the user logged in with.

- It breaks privacy. As a user I would definitely not approve my client to leak my IP address to foreign instances. Also we would have to change our [privacy policy](https://tusky.app/privacy-policy/).
- Malicious instances could probably exploit this somehow e.g. by making Tusky load 1000s of images.
- It only works with public posts on Mastodon Api compatible servers, but there are others
- Opens the door for other calls to foreign instances, and that is complexity I would very much like to keep out of our code

app/src/main/java/com/keylesspalace/tusky/components/viewthread/edits/ViewEditsFragment.kt
```diff
@@ -86,6 +86,7 @@ class ViewEditsFragment :
         (binding.recyclerView.itemAnimator as SimpleItemAnimator).supportsChangeAnimations = false
 
         statusId = requireArguments().getString(STATUS_ID_EXTRA)!!
+
```
nikclayton, line 89:
```suggestion
```

nikclayton:
OK. Updated to remove that feature and be more specific in the error message.

connyduck (approved):

PR #3751, opened 2023-06-14
Default to hiding filtered statuses

### LINKED ISSUE #3746: About the new handling of filters in v22
I'm really disappointed to see that Tusky has started aping Mastodon's approach to filtered toots, where a disruptive message  in your timeline asks if maybe you want to see the message anyway. I assume most people don't since they actively filtered a term to begin with.

Until the latest version, filtered toots just... didn't show. Perfection. Now, to avoid the annoying "SHOW ANYWAY" flags, I am expected to edit each of my filters and select "hide completely" to switch back to the former default.

Now, I have a lot of filters, between 20 and 70 on each of my four accounts. Did it occur to the developers that a) changing the default display settings for terms *that users already opted out of* was a bad idea, and b) maybe it shouldn't be up to users to revert your brain fart on a granular, per filter level?

**Tl;dr**
1. Change the default display setting for filtered toots back to "hide completely".
2. Allow users to change display settings for all filters rather than for individual terms.

* * * *
- Tusky Version: 22

nikclayton:
Tusky v21 and before didn't support the warn/hide distinction in filters, so everything was hidden.

Tusky v22.0 does, but since the default is "warn" for filters this means that they are now seeing warnings for content they've filtered, and the warnings include the very text they're trying to avoid.

Fix this by adding a new account-level preference that overrides the filter action for all their filters. Values are either NONE (use the filter's action), WARN (ignore the filter's action, show a warning), and HIDE (ignore the filter's action, hide the status).

The default value for the preference is HIDE, to match the v21.0 behaviour.

Fixes #3746

### DIFF
--- a/app/schemas/com.keylesspalace.tusky.db.AppDatabase/52.json
+++ b/app/schemas/com.keylesspalace.tusky.db.AppDatabase/52.json
@@ -1,0 +1,1009 @@
+{
+  "formatVersion": 1,
+  "database": {
+    "version": 52,
+    "identityHash": "ab31230e96647bd0476290b31bd8c956",
+    "entities": [
+      {
+        "tableName": "DraftEntity",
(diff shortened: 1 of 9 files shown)


nikclayton:
Note that this is a merge in to **main**.

Given the disruptive nature of the issue -- specifically, users are seeing content that they have very unambiguously said they didn't want to see, and that previous versions of Tusky did not show them -- my plan is, once this is approved, to:

1. Merge to `main`
2. Prep and release 22.1 with just this fix in it
3. Merge this back in to `develop`

Play Console suggests that (a) ~ 44% of users haven't updated to 22.0 yet, so this would benefit them, and (b) there appears to be no way to pause the v22.0 rollout.

nikclayton:
Here's what the new preference looks like:



After users have had a couple of releases to update their filter settings this can be removed (in say, v24).

connyduck:
see #3746

PR #3760, opened 2023-06-18
Show notifications from workers

### LINKED ISSUE #3754: Possible "pull" notification regression
With this (?) I get a regression for the notification worker (pull requests):

```
java.lang.IllegalStateException: Expedited WorkRequests require a Worker to provide an implementation for `getForegroundInfo()`
```

This is only visible in Logcat but prohibits notifications.

At the moment I cannot easily fix it (either by disabling the OneTimeWorkRequest or trying to move the NotificationWorker to a CoroutineWorker).

_Originally posted by @Lakoja in https://github.com/tuskyapp/Tusky/issues/3649#issuecomment-1594285314_

nikclayton:
Fix a crash where workers, in some conditions, should show a notification, but didn't.

These new notifications are sent to a dedicated channel with no importance.

Convert `NotificationWorker` to a `CoroutineWorker` and remove its use of `runBlocking`.

Fixes #3754

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/components/notifications/NotificationHelper.java
+++ b/app/src/main/java/com/keylesspalace/tusky/components/notifications/NotificationHelper.java
@@ -37,6 +37,7 @@ import android.util.Log;
 
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
+import androidx.annotation.StringRes;
 import androidx.core.app.NotificationCompat;
 import androidx.core.app.RemoteInput;
 import androidx.core.app.TaskStackBuilder;
@@ -77,7 +78,12 @@ import java.util.concurrent.TimeUnit;
 
 public class NotificationHelper {
 
-    private static int notificationId = 0;
+    /** ID of notification shown when fetching notifications */
+    public static final int NOTIFICATION_ID_FETCH_NOTIFICATION = 0;
+    /** ID of notification shown when pruning the cache */
+    public static final int NOTIFICATION_ID_PRUNE_CACHE = 1;
+    /** Dynamic notification IDs start here */
+    private static int notificationId = NOTIFICATION_ID_PRUNE_CACHE + 1;
 
     /**
      * constants used in Intents
@@ -121,6 +127,7 @@ public class NotificationHelper {
     public static final String CHANNEL_SIGN_UP = "CHANNEL_SIGN_UP";
     public static final String CHANNEL_UPDATES = "CHANNEL_UPDATES";
     public static final String CHANNEL_REPORT = "CHANNEL_REPORT";
+    public static final String CHANNEL_LISTENABLE_WORKER = "CHANNEL_LISTENABLE_WORKER";
 
     /**
      * WorkManager Tag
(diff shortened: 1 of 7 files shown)


nikclayton:
Might need a bit more work -- experimenting on an API 30 emulator, sometimes the "Fetching notifications..." notification is cancelled, sometimes it's not. Not sure why yet...

Lakoja:
> Might need a bit more work -- experimenting on an API 30 emulator, sometimes the "Fetching notifications..." notification is cancelled, sometimes it's not. Not sure why yet...

Works ok in my tests for the moment.

However I was wondering (also earlier) about the `Thread.sleep` in the code. Shouldn't that be a `delay`?

nikclayton:
> However I was wondering (also earlier) about the `Thread.sleep` in the code. Shouldn't that be a `delay`?

Yep, good catch. Fixed.

app/src/main/java/com/keylesspalace/tusky/worker/NotificationWorker.kt
```diff
@@ -17,27 +17,36 @@
...
-    override fun doWork(): Result {
+) : CoroutineWorker(appContext, params) {
+    val notification: Notification = NotificationHelper.createWorkerNotification(applicationContext, R.string.notification_notification_worker)
+
+    override suspend fun doWork(): Result {
         notificationsFetcher.fetchAndShow()
         return Result.success()
     }
 
+    override suspend fun getForegroundInfo() = ForegroundInfo(NOTIFICATION_ID_FETCH_NOTIFICATION, notification)
```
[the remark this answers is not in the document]
nikclayton: It works on Android 12 because `getForegroundInfo()` isn't called on Android 12, there `getForegroundInfoAsync()` is called instead, https://developer.android.com/reference/kotlin/androidx/work/CoroutineWorker#getForegroundInfo()

app/src/main/java/com/keylesspalace/tusky/components/notifications/NotificationHelper.java
```diff
@@ -121,6 +127,7 @@ public class NotificationHelper {
     public static final String CHANNEL_SIGN_UP = "CHANNEL_SIGN_UP";
     public static final String CHANNEL_UPDATES = "CHANNEL_UPDATES";
     public static final String CHANNEL_REPORT = "CHANNEL_REPORT";
+    public static final String CHANNEL_LISTENABLE_WORKER = "CHANNEL_LISTENABLE_WORKER";
```
Lakoja, line 130: CHANNEL_BACKGROUND_TASKS ? (listenable sound quite technical/specific)

Lakoja (approved):

app/src/main/java/com/keylesspalace/tusky/worker/PruneCacheWorker.kt
```diff
@@ -41,6 +48,8 @@ class PruneCacheWorker(
         return Result.success()
     }
 
+    override suspend fun getForegroundInfo() = ForegroundInfo(NOTIFICATION_ID_PRUNE_CACHE, notification)
```
connyduck, line 51: Why here as well? This one is not expedited?

app/src/main/java/com/keylesspalace/tusky/components/notifications/NotificationHelper.java
connyduck: agree

app/src/main/java/com/keylesspalace/tusky/worker/PruneCacheWorker.kt
nikclayton: Belt and braces. If future code launches this as an expedited worker it won't work without this.

But the failure will be effectively silent (there's no user notification that it happened), and unless you're (a) checking the logs when it happens, or (b) notice that the database is growing without bounds, it won't be noticed.

connyduck: ok

app/src/main/java/com/keylesspalace/tusky/components/notifications/NotificationHelper.java
nikclayton: Done.

connyduck (approved):

PR #3762, opened 2023-06-19
Simplify repeated code that shows errors

nikclayton:
Instead of repeating the same if/else check on the error type when setting up the background message, move this in to BackgroundMessageView.

Provide different `setup()` variants, including one that just takes a throwable and a handler, and figures out the correct drawables and error message.

Update and simplify call sites.


nikclayton:
Pulled this out of https://github.com/tuskyapp/Tusky/pull/3436 since it can stand on its own and makes reviewing that PR a bit easier.

connyduck (approved):
nice

PR #3777, opened 2023-06-29
Migrate "room" from "kapt" to "ksp"

nikclayton:
- Add ksp plugin
- Switch room to use ksp instead of kapt
- `ArrayList` -> `List` in a few places to fix errors about unbound generics


nikclayton:
Upgraded from Android Studio EE to FF and kapt started reporting errors. Migrating to ksp was easier than troubleshooting kapt.

connyduck (approved):

Issue #1469, opened 2019-09-05
Can't tell when post is part of a thread


realityfabric:
[Issue text goes here].

* * * *
- Tusky Version: 8.1
- Android Version: 9
- Android Device: Motorola G7

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

On Mastodon (in the web version) threads will indicate that they are part of a thread, if they are not the first toot in the thread. On Tusky, there is no indication that it's a self-reply unless you open the toot.

It would be nice to know if a post is a part of a thread without needing to open the post.

charlag:
Hi
Do I misunderstand or you're talking about reply/arrow icon which changes it's look when it's a reply? We have it as well.

Tak:
The web ui has a big "SHOW THREAD" thing now (which imo we should _not_ adopt)

![Screenshot_20190914-105348~2](https://user-images.githubusercontent.com/142250/64905883-5b0a5b80-d6de-11e9-91c9-0dc56dd26c43.png)

realityfabric:
Yes, the show thread thing. Why shouldn't we adopt it? It's very confusing to see posts and not realize there's context to them. It only applies to replies to self.

Tak:
Well, my suggestion is because I don't like the way it looks, and we already have the reply arrow thing that @charlag mentioned as a visual indicator.
However, I'm just some person (i.e. not a Tusky maintainer), and it's just my opinion. 😃

realityfabric:
Ah, I see that now. I literally never would have realized that the double arrow vs single arrow even existed, let alone that the double arrow means you'd be replying to a reply.

The double arrow is sleek and unobtrusive, and also at the bottom of the post (seen *after* reading it). Sleekness is nice and all, but even knowing it exists it's only a marginal improvement. I'm not on Tusky to see the most beautiful, sleekest app, but to have a good experience seeing posts on the fediverse.

charlag:
The thing is, Mastodon has more screen real estate usually so they can afford that.
Mastodon did what we do initially. Maybe we can find some better indicator?

KWierso:
Excuse the terrible mockup, but maybe the thread indicator could be the unicode 25B2/25BC symbols up hear the relative timestamp for a toot? 
![Screenshot_20191214-013637-01](https://user-images.githubusercontent.com/172215/70846890-ff518a80-1e12-11ea-9f22-9a4ee6821dd9.jpeg)

charlag:
[...]

PR #3778, opened 2023-06-29
3771 mark post as reply

### LINKED ISSUE #3771: Show that a post is part of a thread
Hi! Thanks for creating Tusky!

I'd love it if toots in the timeline showed that they're part of a thread. Currently they are presented in the feed in the same way as standalone posts, and you don't know that you're missing the context.

This is how it looks in the Megalodon app:

timeline in the Megalodon app

* * * *
- Tusky Version: 22.0
- Android Version: 11
- Android Device: N/A

Lakoja:
Fixes #3771 / partly addresses #1469

grafik

This shows in a timeline the status info "In reply to < username>" just like the existing "< username> boosted" above a status.

This is a more or less low-hanging fruit as the information is available and the web gui does this (but see review comments here for details).

This is NOT the requested "Show thread" functionality - which would still need to be discussed; i.e. every reply makes a thread?

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/adapter/StatusViewHolder.java
+++ b/app/src/main/java/com/keylesspalace/tusky/adapter/StatusViewHolder.java
@@ -30,6 +30,7 @@ import com.keylesspalace.tusky.R;
 import com.keylesspalace.tusky.entity.Emoji;
 import com.keylesspalace.tusky.entity.Filter;
 import com.keylesspalace.tusky.entity.Status;
+import com.keylesspalace.tusky.entity.TimelineAccount;
 import com.keylesspalace.tusky.interfaces.StatusActionListener;
 import com.keylesspalace.tusky.util.CustomEmojiHelper;
 import com.keylesspalace.tusky.util.NumberUtils;
@@ -38,6 +39,7 @@ import com.keylesspalace.tusky.util.StatusDisplayOptions;
 import com.keylesspalace.tusky.util.StringUtils;
 import com.keylesspalace.tusky.viewdata.StatusViewData;
 
+import java.util.Collections;
 import java.util.List;
 
 import at.connyduck.sparkbutton.helpers.Utils;
@@ -63,21 +65,37 @@ public class StatusViewHolder extends StatusBaseViewHolder {
     public void setupWithStatus(@NonNull StatusViewData.Concrete status,
                                 @NonNull final StatusActionListener listener,
                                 @NonNull StatusDisplayOptions statusDisplayOptions,
-                                @Nullable Object payloads) {
+                                @Nullable Object payloads,
+                                boolean showStatusInfo) {
         if (payloads == null) {
-
             boolean sensitive = !TextUtils.isEmpty(status.getActionable().getSpoilerText());
             boolean expanded = status.isExpanded();
 
             setupCollapsedState(sensitive, expanded, status, listener);
 
             Status reblogging = status.getRebloggingStatus();
-            if (reblogging == null || status.getFilterAction() == Filter.Action.WARN) {
+            boolean isReply = status.getStatus().getInReplyToId() != null;
+            boolean isReplyOnly = isReply && reblogging == null;
+
+            boolean hasStatusContext = reblogging != null || isReply;
+
+            if (!hasStatusContext || !showStatusInfo || status.getFilterAction() == Filter.Action.WARN) {
                 hideStatusInfo();
             } else {
-                String rebloggedByDisplayName = reblogging.getAccount().getName();
-                setRebloggedByDisplayName(rebloggedByDisplayName,
-                        reblogging.getAccount().getEmojis(), statusDisplayOptions);
+                String accountName = "";
+                List<Emoji> emojis = Collections.emptyList();
+                if (reblogging != null) {
+                    accountName = reblogging.getAccount().getName();
+                    emojis = reblogging.getAccount().getEmojis();
+                } else if (isReply) {
+                    TimelineAccount repliedTo = status.getInReplyToAccount();
+                    if (repliedTo != null) {
+                        accountName = repliedTo.getName();
+                        emojis = repliedTo.getEmojis();
+                    }
+                }
+
+                setStatusInfoText(isReplyOnly, accountName, emojis, statusDisplayOptions);
                 statusInfo.setOnClickListener(v -> listener.onOpenReblog(getBindingAdapterPosition()));
             }
 
@@ -88,19 +106,27 @@ public class StatusViewHolder extends StatusBaseViewHolder {
         setFavouritedCount(status.getActionable().getFavouritesCount());
         setReblogsCount(status.getActionable().getReblogsCount());
 
-        super.setupWithStatus(status, listener, statusDisplayOptions, payloads);
+        super.setupWithStatus(status, listener, statusDisplayOptions, payloads, showStatusInfo);
     }
 
-    private void setRebloggedByDisplayName(final CharSequence name,
-                                           final List<Emoji> accountEmoji,
-                                           final StatusDisplayOptions statusDisplayOptions) {
+    private void setStatusInfoText(final boolean isReply,
+                                   final CharSequence name,
+                                   final List<Emoji> accountEmoji,
+                                   final StatusDisplayOptions statusDisplayOptions) {
+
         Context context = statusInfo.getContext();
-        CharSequence wrappedName = StringUtils.unicodeWrap(name);
-        CharSequence boostedText = context.getString(R.string.post_boosted_format, wrappedName);
-        CharSequence emojifiedText = CustomEmojiHelper.emojify(
-                boostedText, accountEmoji, statusInfo, statusDisplayOptions.animateEmojis()
-        );
-        statusInfo.setText(emojifiedText);
+        if (name.length() > 0) {
+            CharSequence wrappedName = StringUtils.unicodeWrap(name);
+            CharSequence statusContextText = context.getString(isReply ? R.string.post_replied_format : R.string.post_boosted_format, wrappedName);
+            CharSequence emojifiedText = CustomEmojiHelper.emojify(
+                statusContextText, accountEmoji, statusInfo, statusDisplayOptions.animateEmojis()
+            );
+            statusInfo.setText(emojifiedText);
+        } else {
+            statusInfo.setText(context.getString(R.string.post_replied));
+        }
+        statusInfo.setCompoundDrawablesWithIntrinsicBounds(isReply ? R.drawable.ic_reply_all_18dp : R.drawable.ic_reblog_18dp, 0, 0, 0);
+
         statusInfo.setVisibility(View.VISIBLE);
     }
 
--- a/app/src/main/java/com/keylesspalace/tusky/components/timeline/TimelineTypeMappers.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/components/timeline/TimelineTypeMappers.kt
@@ -206,8 +206,8 @@ fun TimelineStatusWithAccount.toViewData(moshi: Moshi, isDetailed: Boolean = fal
             // no url for reblogs
             url = null,
             account = this.reblogAccount!!.toAccount(moshi),
-            inReplyToId = null,
-            inReplyToAccountId = null,
+            inReplyToId = status.inReplyToId,
+            inReplyToAccountId = status.inReplyToAccountId,
             reblog = reblog,
             content = "",
             // lie but whatever?
@@ -269,6 +269,7 @@ fun TimelineStatusWithAccount.toViewData(moshi: Moshi, isDetailed: Boolean = fal
     }
     return StatusViewData.Concrete(
         status = status,
+        inReplyToAccount = this.inReplyToAccount?.toAccount(moshi),
         isExpanded = this.status.expanded,
         isShowingContent = this.status.contentShowing,
         isCollapsed = this.status.contentCollapsed,
--- a/app/src/main/java/com/keylesspalace/tusky/components/timeline/viewmodel/NetworkTimelineRemoteMediator.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/components/timeline/viewmodel/NetworkTimelineRemoteMediator.kt
@@ -20,18 +20,26 @@ import androidx.paging.ExperimentalPagingApi
 import androidx.paging.LoadType
 import androidx.paging.PagingState
 import androidx.paging.RemoteMediator
+import com.keylesspalace.tusky.components.timeline.toAccount
 import com.keylesspalace.tusky.components.timeline.util.ifExpected
 import com.keylesspalace.tusky.db.AccountManager
+import com.keylesspalace.tusky.db.AppDatabase
+import com.keylesspalace.tusky.entity.TimelineAccount
 import com.keylesspalace.tusky.util.HttpHeaderLink
 import com.keylesspalace.tusky.util.toViewData
 import com.keylesspalace.tusky.viewdata.StatusViewData
+import com.squareup.moshi.Moshi
 import retrofit2.HttpException
 
 @OptIn(ExperimentalPagingApi::class)
 class NetworkTimelineRemoteMediator(
     private val accountManager: AccountManager,
-    private val viewModel: NetworkTimelineViewModel
+    private val viewModel: NetworkTimelineViewModel,
+    db: AppDatabase,
+    private val moshi: Moshi,
 ) : RemoteMediator<String, StatusViewData>() {
+
+    private val accountDao = db.timelineAccountDao()
 
     private val statusIds = mutableSetOf<String>()
 
@@ -80,7 +88,13 @@ class NetworkTimelineRemoteMediator(
                 val expanded = oldStatus?.isExpanded ?: activeAccount.alwaysOpenSpoiler
                 val contentCollapsed = oldStatus?.isCollapsed ?: true
 
+                var inReplyToAccount: TimelineAccount? = null
+                if (status.inReplyToAccountId != null) {
+                    inReplyToAccount = accountDao.get(status.inReplyToAccountId)?.toAccount(moshi)
+                }
+
                 status.toViewData(
+                    inReplyToAccount = inReplyToAccount,
                     isShowingContent = contentShowing,
                     isExpanded = expanded,
                     isCollapsed = contentCollapsed
--- a/app/src/main/java/com/keylesspalace/tusky/db/TimelineStatusEntity.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/db/TimelineStatusEntity.kt
@@ -111,10 +111,16 @@ data class TimelineAccountEntity(
 data class TimelineStatusWithAccount(
     @Embedded
     val status: TimelineStatusEntity,
+
     // null when placeholder
-    @Embedded(prefix = "a_")
+    @Embedded(prefix = "author_")
     val account: TimelineAccountEntity? = null,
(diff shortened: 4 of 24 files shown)


app/src/main/java/com/keylesspalace/tusky/db/TimelineStatusEntity.kt
```diff
@@ -111,6 +111,8 @@ data class TimelineStatusWithAccount(
     val status: TimelineStatusEntity,
     @Embedded(prefix = "a_")
     val account: TimelineAccountEntity? = null, // null when placeholder
-    @Embedded(prefix = "rb_")
+    @Embedded(prefix = "rp_")
```
Lakoja, line 114: Note: The source of this account data is a bit unexpected.

Most other information comes directly from the Api. This here is mixed-in only locally. This could be surprising?

app/src/main/res/drawable/ic_reply_24dp.xml
```diff
@@ -5,6 +5,6 @@
     android:viewportHeight="24.0"
     android:viewportWidth="24.0">
     <path
-        android:fillColor="#fff"
+        android:fillColor="?android:attr/textColorTertiary"
```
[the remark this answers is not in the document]
nikclayton: I think the correct thing to do is to leave `fillColor` as is (or, better, use `@android:color/white`), and set `android:tint` on the `vector`. See `ic_edit_24dp.xml` for an example.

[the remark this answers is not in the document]
Lakoja: Ok.
(Changed it back to fff, which is the most consistent with other icons - not all do this, though.)

app/src/main/java/com/keylesspalace/tusky/components/timeline/viewmodel/NetworkTimelineRemoteMediator.kt
```diff
@@ -29,9 +33,13 @@ import retrofit2.HttpException
 @OptIn(ExperimentalPagingApi::class)
 class NetworkTimelineRemoteMediator(
     private val accountManager: AccountManager,
-    private val viewModel: NetworkTimelineViewModel
+    private val viewModel: NetworkTimelineViewModel,
+    db: AppDatabase,
+    private val gson: Gson
```
nikclayton, line 38: I think all of these can be injected here.

Then you can inject the remote mediator in to the view model. That way the view model doesn't need to receive `db` or `gson` as parameters.

app/src/main/java/com/keylesspalace/tusky/db/TimelineStatusEntity.kt
nikclayton: "rp_" is "reply"?

What's "bo_"? "boost"?

That's probably sufficiently confusing that it should just be "reply_" and "reblog_".

I prefer "reblog" to "boost" because we're pretty consistent about using the API terms for these things in the code, not the UI terms (e.g., it's always "status" not "post", etc).

app/src/main/res/values/strings.xml
```diff
@@ -80,6 +80,7 @@
 
     <string name="post_username_format">\@%s</string>
     <string name="post_boosted_format">%s boosted</string>
+    <string name="post_replied_format">In reply to %s</string>
```
nikclayton, line 83:
```suggestion
    <string name="post_replied_format">%s replied</string>
```

?

Or is the %s here the name of the original author, and not the person who replied?

nikclayton (changes requested):
A couple of low-hanging things I spotted. Later today I'll check this out locally and take a proper look.

app/src/main/res/values/strings.xml
Lakoja: The person who replied is the status owner (and thus listed one line below).

%s is the "replied to account name".

app/src/main/java/com/keylesspalace/tusky/db/TimelineStatusEntity.kt
Lakoja: Ok, changed it to "author", "replied" and "reblogger" (all 3 TimelineStatusEntity).

(Generally these 2 queries and tedious to change and to understand.)

app/src/main/java/com/keylesspalace/tusky/components/timeline/viewmodel/NetworkTimelineRemoteMediator.kt
Lakoja: The remote mediator has the view model as dependency

(and thus a dependency cylce)

nikclayton:
I haven't tested this yet, but I've just observed the following:

1. `item_status.xml` (and related layouts) have a `R.id.status_info` view at the top, which is where the "X boosted" message goes.
2. `StatusBaseViewHolder.java` has a `setIsReply` method that's called whenever the status is a reply

So, in `StatusViewHolder.java`, wouldn't it be simpler to just do this:

```java
protected void setIsReply(boolean isReply) {
    super setIsReply(isReply);
    if (isReply) {
        statusInfo.text = getString(R.string.post_replied_format, ...)
    }
}
```

?

Also, what should happen if the status was a reply, and it was reblogged? https://chaos.social/@jollysea/110668273086228868 is an example post that is a reply that was reblogged by @mcclure.

1. The "X boosted" message is more important
2. The "In reply to Y" message is more important
3. The message should be "X boosted this reply to Y"
4. Show two lines of text

Note that that status above is an interesting edge case for this, because it's mcc reblogging a reply to her own post, so although option 3 might be generally good, it would lead to "mcc boosted this reply to mcc". Which might be more mcc than a single line can handle :-)

Lakoja:
> So, in `StatusViewHolder.java`, wouldn't it be simpler to just do this:
> 
> ```java
> protected void setIsReply(boolean isReply) {
>     super setIsReply(isReply);
>     if (isReply) {
>         statusInfo.text = getString(R.string.post_replied_format, ...)
>     }
> }
> ```
> 
> ?

Instead of what?
At least all the changes before and in (now) `setStatusInfoText()` would be necessary as they also deal with emojis and a maybe different name.
And at least at the moment the `setIsReply()` only deals with two different reply cases. The reblog information is orthogonal.

Lakoja:
> 
> Also, what should happen if the status was a reply, and it was reblogged? https://chaos.social/@jollysea/110668273086228868 is an example post that is a reply that was reblogged by @mcclure.
> 
>     3. The message should be "X boosted this reply to Y"
> 

I kind of like option 3.
However that might be too much information? (and will often be too long, names tend to be long...)
2 lines will also be too much, I think.

At the moment (in my code) reblog is more important than reply. So an "if ... else if" logic.

app/src/main/java/com/keylesspalace/tusky/components/timeline/viewmodel/NetworkTimelineRemoteMediator.kt
```diff
@@ -71,7 +80,13 @@ class NetworkTimelineRemoteMediator(
                 val expanded = oldStatus?.isExpanded ?: activeAccount.alwaysOpenSpoiler
                 val contentCollapsed = oldStatus?.isCollapsed ?: true
 
+                var inReplyToAccount:TimelineAccount? = null
+                if (status.inReplyToAccountId != null) {
+                    inReplyToAccount = accountDao.get(status.inReplyToAccountId)?.toAccount(gson)
```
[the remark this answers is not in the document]
nikclayton: I don't think the account will always exist in the local database.

Consider: You follow user B, they reply to user A, who you do not follow.

This works for a reblog, because the `reblog` field contains the complete status, which includes the full information of the original author (https://docs.joinmastodon.org/entities/Status/#reblog).

A status that's a reply doesn't, it only contains `in_reply_to_account_id` (https://docs.joinmastodon.org/entities/Status/#in_reply_to_account_id).

It *might* also contain a `mentions` object, but AFAIK there's no requirement that a reply also mention the person they're replying to, and even if there is a `mentions` object it doesn't contain the display name of the person that's being replied to.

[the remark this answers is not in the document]
Lakoja: It does work so far (for a week) on my timeline.

Not sure if the api actually provides arbitrary account information but I'm pretty sure we wouldn't want to do an api call for this information.

At the moment if an account is missing the text line would be missing.

(So is there an alternative?)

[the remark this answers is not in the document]
nikclayton: Two strings, "Replied" and "Replied to %s", select the correct one based on whether or not we know the account info.

app/src/main/res/layout/item_status.xml
```diff
@@ -17,7 +17,7 @@
         android:layout_marginStart="14dp"
         android:layout_marginTop="@dimen/status_reblogged_bar_padding_top"
         android:layout_marginEnd="14dp"
-        android:drawableStart="@drawable/ic_reblog_18dp"
+        android:drawableStart="@drawable/ic_reblog_24dp"
```
nikclayton, line 20: Why is this switching to the bigger icon? It causes the effective left margin of the text to change.

nikclayton: Also, you could copy `ic_reply_24dp.xml` to ic_reply_18dp.xml`, and then use that for replies.

app/src/main/java/com/keylesspalace/tusky/components/timeline/viewmodel/NetworkTimelineRemoteMediator.kt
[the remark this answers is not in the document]
Lakoja: ok

nikclayton:
@Lakoja Can you take a look at https://github.com/Lakoja/Tusky/pull/4, and see if you want to merge that in to this PR? I figured that was easier than more back-and-forth.

You'll want to merge `develop` in to this PR first to minimse the diffs.

Some quick screenshots:

## It shows reply and boost, with a bullet between them

image

image

--- 

## If the account isn't known, but mentioned, it uses the @-form of the user name

image

---

## Long lines are truncated / ellipsised

image

---

I'm not 100% convinced that moving the left margin over is a good idea, but as you can see from the screenshots, sometimes the line does get quite long.

WDYT?

app/src/main/res/layout/item_status.xml
Lakoja: The size is determined by the component (height of TextView).

Already before: The text behind the drawable is not alligned with the name or the content below. This does not change here.

So I could leave it as 18 but that would not change it graphically.

Lakoja:
> Some quick screenshots:

> WDYT?

I find the lines with two information quite hard to parse visually.
(thus) I still wouldn't present two informations there.

nikclayton:
> > Some quick screenshots:
> 
> > WDYT?
> 
> I find the lines with two information quite hard to parse visually. (thus) I still wouldn't present two informations there.

I think that might be a problem, because with this PR we're trying to use the statusInfo to show two difference pieces of information.

Currently the statusInfo answers the "Why am I seeing this in my timeline?" question. It tells the user "You are seeing this because someone you follow has boosted it".

This PR extends that to also answer the "Can I tap on this post to read a thread?" question.

I'm not sure that only ever showing one bit of information is a good idea, because it means that users will not be able to rely on the information in the statusInfo -- what's shown will vary, and it will be difficult for users to figure out why it's varying.

By which I mean that some replies -- replies from people that the user follows -- will be marked as replies, and the user will know they can tap the post to see the thread.

But other replies -- replies from people they don't follow that have been boosted by someone they do -- will not be marked as replies.

[I know the "reply" icon at the bottom of the post will change as well, but as we see from the bug reports, this is also something that user's don't always notice]

If we flip that example around, and decide that showing that it's part of a thread is more important than showing that it's a boost, then we get the problem that a user can see "user-i-dont-follow replied to ..." in the statusInfo, and not know why this post is appearing in their timeline.

The first screenshot in my previous comment is an example of that. I don't follow rain, but I do follow Norm.

Both of those seem like a recipe for user confusion.

I think we can collapse some of the information down. For example, if someone boosts their own post there's no need to show their  display name again.

And we could experiment with putting some of the information elsewhere. I'm not sure a two-line statusInfo is a good idea, but maybe the "reply" icon could appear next to the display name of the poster (i.e., the displayname that appears to the right of the avatar)? That would leave the statusInfo for boosts (and later, hashtags).

But... that would mean the left-edge of the display names would jump around a bit as the user scrolls through their timeline.

I'll try some mockups and see what happens.

nikclayton:
Here's the latest variant (I'm still waiting for a self-boost to cross my timeline).

## Replying to themselves

image

nikclayton:
Some of https://github.com/tuskyapp/Tusky/issues/3883 is relevant to this as well. There's one, possibly two additional values that can be included in the status_info, but they're mutually exclusive with the others (I think). They are:

- "Sent you a direct message"
- "Replied to your direct message"

or similar. Possibly with "direct message" in that text in bold, or bold tusky_red, or something like that, to make it even more distinctive.

app/src/main/java/com/keylesspalace/tusky/components/timeline/TimelineTypeMappers.kt
```diff
@@ -207,8 +207,8 @@ fun TimelineStatusWithAccount.toViewData(gson: Gson, isDetailed: Boolean = false
             id = status.serverId,
             url = null, // no url for reblogs
             account = this.reblogAccount!!.toAccount(gson),
-            inReplyToId = null,
-            inReplyToAccountId = null,
+            inReplyToId = status.inReplyToId,
```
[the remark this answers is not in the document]
Tak: But the boost itself is not the reply, right? The *actionable* status is the reply

app/src/main/java/com/keylesspalace/tusky/adapter/StatusViewHolder.java
```diff
@@ -65,19 +67,34 @@ public void setupWithStatus(@NonNull StatusViewData.Concrete status,
...
         if (payloads == null) {
-
             boolean sensitive = !TextUtils.isEmpty(status.getActionable().getSpoilerText());
             boolean expanded = status.isExpanded();
 
             setupCollapsedState(sensitive, expanded, status, listener);
 
             Status reblogging = status.getRebloggingStatus();
-            if (reblogging == null || status.getFilterAction() == Filter.Action.WARN) {
+            boolean isReply = status.getStatus().getInReplyToId() != null;
```
Tak, line 76: Don't you need to use `status.getActionable` here, and on line 90?

Tak (commented):
Doesn't this need a database migration for `TimelineAccountEntity`?

app/src/main/java/com/keylesspalace/tusky/components/timeline/TimelineTypeMappers.kt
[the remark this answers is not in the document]
Lakoja: What is boosted ("status") can be a reply (too).

This info is "unused".

Lakoja:
> Doesn't this need a database migration for `TimelineAccountEntity`?

There is no change there?

All info here is present (in rare cases the "replied to account info" might be missing and then the name is not show).

app/src/main/java/com/keylesspalace/tusky/adapter/StatusViewHolder.java
Lakoja: I was searching for some time for the right location with the right information.
"actionable" was not it if I remember correctly.

It is working and plausible at the moment.

(Would need to look deep(er) into the code if you want me to be 100% certain about this.)

mcclure:
What does this patch do if a post is BOTH a reply AND an RT?

app/src/main/java/com/keylesspalace/tusky/components/timeline/TimelineTypeMappers.kt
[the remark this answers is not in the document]
Tak: Ah, I see now, this is here as a hack to populate the inreplyto timeline account of the concrete statusviewdata when it's a boost of a reply.
…but as you say, we're not displaying that information, so it could be reverted to avoid confusing the issue (although the status/actionable thing is already confusing)

app/src/main/java/com/keylesspalace/tusky/adapter/StatusViewHolder.java
Tak: You're right, the inreplyto timeline account is only available on status, but without the timeline type mapper change, the correct thing to check for the inreplyto id is the actionable status. (And if `status.getActionable().getInReplyToId()` is nonnull, then `status.getInReplyToAccount()` will be as well.)

Tak:
The current PR code displays this when a reply is boosted
Screenshot 2023-10-06 at 11 12 35 AM

Tak:
Oh, it looks super weird in the thread view - is that intentional?

Tak:
I guess this change should be applied to the notification timeline as well? (Maybe in a second pass after the refactor lands?)

Lakoja:
> Oh, it looks super weird in the thread view - is that intentional?

No.
(Super weird: it shows "Replied" above every status.)

I fixed it by not showing the status info in the thread view.

But I guess the introduced flag should/could be part of "StatusDisplayOptions" object?

Lakoja:
> I guess this change should be applied to the notification timeline as well? (Maybe in a second pass after the refactor lands?)

Yes, probably and yes.

gradle/libs.versions.toml
```diff
@@ -120,6 +120,7 @@ glide-compiler = { module = "com.github.bumptech.glide:ksp", version.ref = "glid
 glide-core = { module = "com.github.bumptech.glide:glide", version.ref = "glide" }
 glide-okhttp3-integration = { module = "com.github.bumptech.glide:okhttp3-integration", version.ref = "glide" }
 gson = { module = "com.google.code.gson:gson", version.ref = "gson" }
+kapt = { module = "androidx.room:room-compiler", version.ref = "androidx-room" }
```
Lakoja, line 123: Not sure why this is needed _now_ (starting today)?

If I don't have this there will be a build error that the (automatically generated) AppDatabase_Impl does not implement all needed dao factory methods.

app/src/main/java/com/keylesspalace/tusky/adapter/StatusViewHolder.java
Lakoja: What is the meaning of "actionable" in current code?

I see that it only has a value if "reblog" is true (Status.kt:59), so the thinking is probably "the item the action was taken for"?

(The "reply stuff" I have aligned more with the extsting "inReplyToAccountId".)

Tak: Yeah, my understanding is that the "outer" status is the boost itself, and the actionable status is the status that was boosted

Tak:
imo the changes look ok, but it will have to be rebased again after #4026 lands (after the 25.0 release) 😐 
I'm really sorry that this has been in stasis for almost a year

connyduck:
Should we in turn get rid of the double arrow on the reply button that currently indicates replies? It wouldn't be necessary anymore.

gradle/libs.versions.toml
connyduck: This should not be necessary as we migrated Room to ksp. If you get an error, try a clean build.

Lakoja: Removed.

Lakoja:
Replaced by: https://github.com/tuskyapp/Tusky/pull/4834

PR #3801, opened 2023-07-04
refactor domain blocks to paging

connyduck:
Adds Paging, ViewModel, error snackbars, removes Rx, renames everything to "domain blocks" to match the Masto Api

part of #2992


app/src/main/java/com/keylesspalace/tusky/components/domainblocks/DomainBlocksRemoteMediator.kt
```diff
@@ -0,0 +1,56 @@
...
+import androidx.paging.RemoteMediator
+import com.keylesspalace.tusky.network.MastodonApi
+import com.keylesspalace.tusky.util.HttpHeaderLink
+import retrofit2.HttpException
+import retrofit2.Response
+
+@OptIn(ExperimentalPagingApi::class)
+class DomainBlocksRemoteMediator(
+    private val api: MastodonApi,
+    private val viewModel: DomainBlocksViewModel
```
nikclayton, line 15: This (passing the viewmodel down to the paging source) is a layering violation. I appreciate there's quite a few of them in the code at the moment, but I'm trying to clear them up in https://github.com/tuskyapp/Tusky/pull/3436.

The approach in that PR is to have the viewmodel create a repository.

The repository manages the cache of items (domain blocks, here), and provides the cache to both the paging source and the remote mediator.

The paging source only reads from the cache. The remote mediator writes to the cache and invalidates (see later comment) the paging source when it does so.

app/src/main/java/com/keylesspalace/tusky/components/domainblocks/DomainBlocksFragment.kt
```diff
@@ -0,0 +1,108 @@
...
+            }
+        }
+    }
+
+    private fun showUnmuteError(domain: String) {
+        showSnackbar(
+            getString(R.string.error_unblocking_domain, domain),
+            R.string.action_retry
+        ) { viewModel.unblock(domain) }
+    }
```
nikclayton, line 83: In the Notifications code the error classes have `@StringRes val message: Int` properties so that the code using the error doesn't need to worry about it.

They also contain the action that failed, so the retry code can be generic.

They also contain the throwable that caused the error, so more accurate information can be shown to the user.

Any reason not to do that here too?

app/src/main/java/com/keylesspalace/tusky/components/domainblocks/DomainBlocksFragment.kt
```diff
@@ -0,0 +1,108 @@
...
+import com.keylesspalace.tusky.di.ViewModelFactory
+import com.keylesspalace.tusky.util.hide
+import com.keylesspalace.tusky.util.show
+import com.keylesspalace.tusky.util.viewBinding
+import com.keylesspalace.tusky.util.visible
+import kotlinx.coroutines.flow.collectLatest
+import kotlinx.coroutines.launch
+import javax.inject.Inject
+
+class DomainBlocksFragment : Fragment(R.layout.fragment_domain_blocks), Injectable {
```
nikclayton, line 25: Should the user be able to refresh this, with a swipe, or a menu item?

app/src/main/java/com/keylesspalace/tusky/components/domainblocks/DomainBlocksAdapter.kt
```diff
@@ -0,0 +1,27 @@
...
+        return BindingHolder(binding)
+    }
+
+    override fun onBindViewHolder(holder: BindingHolder<ItemMutedDomainBinding>, position: Int) {
+        getItem(position)?.let { instance ->
+            holder.binding.mutedDomain.text = instance
+            holder.binding.mutedDomainUnmute.setOnClickListener {
+                onUnmute(instance)
+            }
+        }
```
nikclayton, line 25:
```suggestion
        getItem(position)?.let { domain ->
            holder.binding.mutedDomain.text = domain
            holder.binding.mutedDomainUnmute.setOnClickListener {
                onUnmute(domain)
            }
        }
```

Couldn't figure out why this used both "instance" (which apparently is deprecated in favour of "server") and "domain".

app/src/main/java/com/keylesspalace/tusky/components/domainblocks/DomainBlocksFragment.kt
connyduck: yes, I'd add this in a separate PR

Tak (approved):

PR #3789, opened 2023-07-03
new issue templates

connyduck:
inspired by https://github.com/mastodon/mastodon/tree/main/.github/ISSUE_TEMPLATE

PR #3802, opened 2023-07-04
Show additional bug report info in AboutActivity

nikclayton:
Make it easier for people to find information we need for a bug report, and show it on AboutActivity.

New info is:

- Device manufacturer (e.g., "Google") and model (e.g., "Pixel 4a (5G)")
- Android version (e.g., "13")
- SDK version (e.g., "33")
- Active account (e.g., "@Tusky@mastodon.social")
- Server's version (e.g., "4.1.2+nightly-20230627")

All info is copyable to make it easy to include in a bug report. A button to copy the information is also shown, to make that as easy as possible.


nikclayton:
I figure if we're going to get more structured info from people for bug reports (https://github.com/tuskyapp/Tusky/pull/3789) then it's a good idea to make it easy for users to get that info.

Here's a before / after:

 


The "after" is on an emulator, hence the "sdk_gphone64_x86_64" value for the model.

evelynmarie:
Not sure how I feel about the fancy formatting being ditched for the license info and website / issue links.  The before looks better in terms of that, IMO.

app/src/main/java/com/keylesspalace/tusky/AboutActivity.kt
```diff
@@ -32,6 +40,28 @@ class AboutActivity : BottomSheetActivity(), Injectable {
...
+            accountManager.activeAccount?.let { account ->
+                val instanceInfo = instanceInfoRepository.getInstanceInfo()
+                binding.accountInfo.text = getString(
+                    R.string.about_account_info,
+                    account.username,
+                    account.domain,
+                    instanceInfo.version
+                )
+                binding.accountInfoTitle.isVisible = true
+                binding.accountInfo.isVisible = true
```
nikclayton, line 61: Note to self: This should use `.show()`.

connyduck:
I like the idea but the design is indeed not ideal. Maybe more like the Wikipedia App

nikclayton:
> I like the idea but the design is indeed not ideal. Maybe more like the Wikipedia App 

Here's what it looks like now:

evelynmarie:
> > I like the idea but the design is indeed not ideal. Maybe more like the Wikipedia App 
> 
> 
> 
> Here's what it looks like now:
> 
> 
> 
> 
> 
> 

now this i can get behind - this actually looks good. but, perhaps an option could be added to show / hide the detailed information re: device and account? perhaps as like an advanced setting. idk if people wanna have that info be visible all the time.

nikclayton:
> > > I like the idea but the design is indeed not ideal. Maybe more like the Wikipedia App
> > 
> > 
> > Here's what it looks like now:
> > 
> 
> now this i can get behind - this actually looks good. but, perhaps an option could be added to show / hide the detailed information re: device and account? perhaps as like an advanced setting. idk if people wanna have that info be visible all the time.

The only time people are likely going to look at this screen is if they need the information for a bug report, so hiding the information and requiring additional work on the part of the user to get it seems unnecessary.

Lakoja (approved):

PR #3813, opened 2023-07-07
Uses the system theme as default theme

SpaceFox:
A small polish update to set the "System Design" as default theme.

This to avoid unexpected behaviour on first application usage, and to force the user to change settings few seconds only after the first Tusky opening – which is a quite bad first impression.


app/src/main/java/com/keylesspalace/tusky/util/ThemeUtils.kt
```diff
@@ -35,7 +35,7 @@ private const val THEME_DAY = "day"
 private const val THEME_BLACK = "black"
 private const val THEME_AUTO = "auto"
 private const val THEME_SYSTEM = "auto_system"
-const val APP_THEME_DEFAULT = THEME_NIGHT
+const val APP_THEME_DEFAULT = THEME_SYSTEM
```
nikclayton, line 38: Because you're changing the default this will affect everyone who is upgrading, and not just new users. There are many places in the code that do

```
val theme = sharedPreferences.getString("appTheme", APP_THEME_DEFAULT)
```

We have a mechanism to handle this, so that people who are upgrading do not see the UI suddenly change out from under them.

First you will need to update `PrefKeys.SCHEMA_VERSION`, in https://github.com/tuskyapp/Tusky/blob/develop/app/src/main/java/com/keylesspalace/tusky/settings/SettingsConstants.kt#L44.

Then you will need to modify `upgradeSharedPreferences()` (https://github.com/tuskyapp/Tusky/blob/develop/app/src/main/java/com/keylesspalace/tusky/TuskyApplication.kt#L121) so that, on upgrade, if the user's current theme preference is not set, it's set to the `THEME_NIGHT` constant.

nikclayton (changes requested):

app/src/main/java/com/keylesspalace/tusky/util/ThemeUtils.kt
SpaceFox: Thank you for your really detailed comments! I thought it was too easy... I do the requested changes and push them somewhere this week or the next one.

app/src/main/java/com/keylesspalace/tusky/util/ThemeUtils.kt
```diff
@@ -35,7 +35,8 @@ private const val THEME_DAY = "day"
...
+const val APP_THEME_DEFAULT_COMPAT = THEME_NIGHT
```
SpaceFox, line 39: I’m not sure of this one, the other idea I have is to set _one_ constant (`THEME_NIGHT`) from above as public, which is quite strange. Let me know what do you prefer.

app/src/main/java/com/keylesspalace/tusky/util/ThemeUtils.kt
SpaceFox: Done, I hope I have well understand your guidelines :)

app/src/main/java/com/keylesspalace/tusky/TuskyApplication.kt
```diff
@@ -130,6 +131,14 @@ class TuskyApplication : Application(), HasAndroidInjector {
             editor.remove(PrefKeys.MEDIA_PREVIEW_ENABLED)
         }
 
+        if (oldVersion < 2023071701) {
+            // Default value for appTheme is now THEME_SYSTEM, so let’s save the old default value as preference to
+            // avoid upgrading user to see their theme change without action.
+            if (sharedPreferences.contains(PrefKeys.APP_THEME)) {
+                editor.putString(PrefKeys.APP_THEME, APP_THEME_DEFAULT_COMPAT);
```
nikclayton, line 138:
```suggestion
                editor.putString(PrefKeys.APP_THEME, APP_THEME_DEFAULT_COMPAT)
```

app/src/main/java/com/keylesspalace/tusky/TuskyApplication.kt
```diff
@@ -130,6 +131,14 @@ class TuskyApplication : Application(), HasAndroidInjector {
             editor.remove(PrefKeys.MEDIA_PREVIEW_ENABLED)
         }
 
+        if (oldVersion < 2023071701) {
+            // Default value for appTheme is now THEME_SYSTEM, so let’s save the old default value as preference to
+            // avoid upgrading user to see their theme change without action.
+            if (sharedPreferences.contains(PrefKeys.APP_THEME)) {
+                editor.putString(PrefKeys.APP_THEME, APP_THEME_DEFAULT_COMPAT)
+            }
```
nikclayton, line 139: I think this check is backwards -- currently, this code will overwrite the user's previous preference, if they had set it.

What it should do is set the preference to the old default if there's no value, so this should be:

```suggestion
            if (!sharedPreferences.contains(PrefKeys.APP_THEME)) {
                editor.putString(PrefKeys.APP_THEME, APP_THEME_DEFAULT_COMPAT)
            }
```

app/src/main/java/com/keylesspalace/tusky/BaseActivity.java
```diff
@@ -74,7 +76,7 @@ protected void onCreate(@Nullable Bundle savedInstanceState) {
         /* There isn't presently a way to globally change the theme of a whole application at
          * runtime, just individual activities. So, each activity has to set its theme before any
          * views are created. */
-        String theme = preferences.getString("appTheme", ThemeUtils.APP_THEME_DEFAULT);
+        String theme = preferences.getString(APP_THEME, ThemeUtils.APP_THEME_DEFAULT);
```
nikclayton, line 79: Thanks for making these changes too.

app/src/main/java/com/keylesspalace/tusky/util/ThemeUtils.kt
nikclayton: There's probably no harm in making all those constants public, which would then make the code in `TuskyApplication` a bit more explicit about exactly which value is being set.

nikclayton (changes requested):

app/src/main/java/com/keylesspalace/tusky/util/ThemeUtils.kt
SpaceFox: Done.

app/src/main/java/com/keylesspalace/tusky/TuskyApplication.kt
SpaceFox: Shame on me >_< I must stop developing tired.

nikclayton:
Just tried this, and it doesn't work as expected from a fresh install, because the default value, if the preference is missing, is 0, and that's less than the value it's being compared against.

So even with a fresh install you still get the dark theme.

Unfortunately, the two cases where the preference is missing are:

- this is a fresh install of the app
- the user is updating from a version before this schema version preference was introduced

There needs to be a mechanism to distinguish between those two cases.

It might be possible to do that by inspecting the database. I'll investigate tomorrow.

nikclayton:
@SpaceFox Yeah -- simplest approach is probably, in `TuskyApplication.kt`:

1. Inject an `AppDatabase` instance
2. From the database instance, get an `AccountDao` instance, and call `loadAll()` on that

That'll return a list of accounts. If it's empty then this is a fresh install[1]. If it's not empty then this is an upgrade.

So something like (untested) around line 79 of TuskyApplication.kt:

```kotlin
val upgrading = db.accountDao().loadAll().isNotEmpty()
if (upgrading) {
    val oldVersion = sharedPreferences.getInt(PrefKeys.SCHEMA_VERSION, 0)
    if (oldVersion != SCHEMA_VERSION) {
        upgradeSharedPreferences(oldVersion, SCHEMA_VERSION)
    }
}
```

[1] There's a chance the user has also logged out of all their accounts, and is now logging in from scratch. That's probably sufficiently unlikely that we don't need to worry about it...

connyduck:
Please don't overengineer this D:
Why not check the `oldVersion` in the preferences migration? If it is 0 its a new install

nikclayton:
> Please don't overengineer this D: Why not check the `oldVersion` in the preferences migration? If it is 0 its a new install

Ah, yeah, that'll do it :-)

nikclayton:
Hi @SpaceFox , are you able to make the suggested changes?

SpaceFox:
Hi,

I’m back from holidays, I pushed the suggested updates.

nikclayton:
Thank you. Sorry it took longer than I'd like to check this out.

Just in case it's needed for future reference, I checked the following scenarios and it all behaves as expected:

## Upgrade

- Wipe and install a fresh `develop`
- Don't change settings
- Update with PR
- Should still be dark theme

PASS

## Upgrade with theme (black)

- Wipe and install a fresh `develop`
- Change theme to black
- Update with PR
- Should still be black theme

PASS


## Upgrade with theme (dark)

- Wipe and install a fresh `develop`
- Explicitly set the theme to dark (e.g., set to black, exit prefs, then set back to dark)
- Update with PR
- Should still be dark theme

PASS

## Fresh install

- Wipe and install from PR
- Should be light theme

PASS

PR #3817, opened 2023-07-07
Prepare 23.0 beta 2 (versionCode 112)

nikclayton:


connyduck (approved):
Imho this is stable enough to be released as v23.0 already

nikclayton:
> Imho this is stable enough to be released as v23.0 already

I can understand the sentiment. But of ~ 6,000 people enrolled in the beta programme, only ~ 1,500 have downloaded and installed 23.0 beta 1, so there's still not as much stability data as I'd like.

Issue #3756, opened 2023-06-17
To many and doubled Android notifications


Lakoja:
Since #3626 every notification is shown (sent to the notification manager) as a single Android notification.
Only afterwards a summary notification is created which should remove (group) all single notifications.

However this does not work with Android 11 and lower.
Here all these notifications (single and summary are shown).

![grafik](https://github.com/tuskyapp/Tusky/assets/1618905/8c85b5d9-7744-4332-ba6c-d4e8db3ac8cd)

Two smaller issues:
* If a summary notification is about to be shown it does not make much sense for the user to show (potentially many) single notifications beforehand. They are especially shown only every second (rate-limited). It would be better to just show the final summary notification.
* The summary notifications are split by type (favorite, boost, ...) so often multiple are shown. However with Android 11 this only shows the title(s) "5 new interactions" and "2 new interactions" with my example and some names. Even with Android 12 this split is probably confusing but with Android 11 it makes no sense.

* * * *
- Tusky Version: 22
- Android Version: <= 11

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

nikclayton:
Do they get re-grouped when you close the notification shade? I've observed a bug in Android where if you have the shade down as notifications are being created, and then create the summary they are not collapsed in to the summary.

But closing the notification shade causes them to be grouped, and then reopening the notification shade correctly shows them as grouped.

You mentioned Android 11 and lower, but https://developer.android.com/develop/ui/views/notifications/group says:

> Note: If your app sends four or more notifications and doesn't specify a group, the system [automatically groups](https://developer.android.com/develop/ui/views/notifications/group#automatic-grouping) them on Android 7.0 and higher.

> If a summary notification is about to be shown it does not make much sense for the user to show (potentially many) single notifications beforehand. They are especially shown only every second (rate-limited). It would be better to just show the final summary notification.

I agree with you. But my understanding of https://developer.android.com/develop/ui/views/notifications/group#group-summary (and the example code that they show there) is that the individual notifications must be created first, and then the summary notification is created.

> The summary notifications are split by type (favorite, boost, ...) so often multiple are shown. However with Android 11 this only shows the title(s) "5 new interactions" and "2 new interactions" with my example and some names. Even with Android 12 this split is probably confusing but with Android 11 it makes no sense.

According to https://developer.android.com/develop/ui/views/notifications/group#group-summary this is on API 23 and below. 

[...]

PR #3821, opened 2023-07-08
3820: Group android notifications and properly use the "group summary alert"

### LINKED ISSUE #3820: Doubled notification sound for multiple favorite notifications
Breaking out a smaller step from #3756:

> The "firstOfBatch" logic should be fixed to a "firstOfGroup"; for that a grouping of the single notifications by "group id" should be done before the notify() loop

Currently if there are multiple notifications of one type (e.g. favorite) there will be two notification sounds: One for the first and one for the summary notification for that type.

This can be tested / worked with in a dev version when modifying the two values remoteMarkerId and localMarkerId in NotificationFetcher#fetchNewNotifications`; setting it to a fixed value some hundred below their current value.
That will yield a few notifications at once each time.

Lakoja:
Fixes #3820 

This mainly corrects the "first of batch" logic.
(Each group/batch has a list so the question "only one per this group?" can be answered.)

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/components/notifications/NotificationFetcher.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/components/notifications/NotificationFetcher.kt
@@ -17,6 +17,7 @@ import com.keylesspalace.tusky.util.isLessThan
 import kotlinx.coroutines.delay
 import javax.inject.Inject
 import kotlin.math.min
+import kotlin.time.Duration.Companion.milliseconds
 
 /** Models next/prev links from the "Links" header in an API response */
 data class Links(val next: String?, val prev: String?) {
@@ -106,10 +107,12 @@ class NotificationFetcher @Inject constructor(
                                 notificationListEntry.value.size == 1
                             )
                             notificationManager.notify(notification.id, account.id.toInt(), androidNotification)
-                        }
 
-                        // Leave some time between notifications from different types (to be able to distinguish them)
-                        delay(1000)
+                            // Android will rate limit / drop notifications if they're posted too
+                            // quickly. There is no indication to the user that this happened.
+                            // See https://github.com/tuskyapp/Tusky/pull/3626#discussion_r1192963664
+                            delay(1000.milliseconds)
+                        }
                     }
 
                     NotificationHelper.updateSummaryNotifications(
--- a/app/src/main/java/com/keylesspalace/tusky/components/notifications/NotificationFetcher.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/components/notifications/NotificationFetcher.kt
@@ -97,14 +97,14 @@ class NotificationFetcher @Inject constructor(
                     // 1.11 and up (https://developer.android.com/jetpack/androidx/releases/core#1.11.0-alpha01)
                     // when it is released.
 
-                    notificationsByType.forEach { notificationGroup ->
-                        notificationGroup.value.forEach { notification ->
+                    notificationsByType.forEach { notificationsGroup ->
+                        notificationsGroup.value.forEach { notification ->
                             val androidNotification = NotificationHelper.make(
                                 context,
                                 notificationManager,
                                 notification,
                                 account,
-                                notificationGroup.value.size == 1
+                                notificationsGroup.value.size == 1
                             )
                             notificationManager.notify(notification.id, account.id.toInt(), androidNotification)
 
--- a/app/src/main/java/com/keylesspalace/tusky/components/notifications/NotificationFetcher.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/components/notifications/NotificationFetcher.kt
@@ -14,8 +14,9 @@ import com.keylesspalace.tusky.entity.Notification
 import com.keylesspalace.tusky.network.MastodonApi
 import com.keylesspalace.tusky.util.HttpHeaderLink
 import com.keylesspalace.tusky.util.isLessThan
-import kotlinx.coroutines.delay
+import java.util.*
 import javax.inject.Inject
+import kotlin.collections.HashMap
 import kotlin.math.min
 import kotlin.time.Duration.Companion.milliseconds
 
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/keylesspalace/tusky/components/notifications/NotificationFetcher.kt
```diff
@@ -65,23 +66,24 @@ class NotificationFetcher @Inject constructor(
                         }
                     }
 
+                    val notificationsGrouped = groupByGroupId(account, notifications)
```
nikclayton, line 69: Can you use https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/group-by.html here?

app/src/main/java/com/keylesspalace/tusky/components/notifications/NotificationFetcher.kt
```diff
@@ -11,8 +11,9 @@ import com.keylesspalace.tusky.entity.Marker
 import com.keylesspalace.tusky.entity.Notification
 import com.keylesspalace.tusky.network.MastodonApi
 import com.keylesspalace.tusky.util.isLessThan
-import kotlinx.coroutines.delay
+import java.util.*
```
nikclayton, line 14: Lint won't like this (wildcard imports)

app/src/main/java/com/keylesspalace/tusky/components/notifications/NotificationFetcher.kt
```diff
@@ -65,23 +66,24 @@ class NotificationFetcher @Inject constructor(
...
+                    notificationsGrouped.forEach { notificationList ->
+                        notificationList.forEach { notification ->
+                            val androidNotification = NotificationHelper.make(
+                                context,
+                                notificationManager,
+                                notification,
+                                account,
+                                notificationList.size == 1
+                            )
+                            notificationManager.notify(notification.id, account.id.toInt(), androidNotification)
```
nikclayton, line 85: Why remove the delay? This will result in notifications being dropped.

Lakoja: See my findings and the documentation link in https://github.com/tuskyapp/Tusky/issues/3756#issuecomment-1605342221

The only thing that drops (grouped) notifications for me is the maximum total number.

app/src/main/java/com/keylesspalace/tusky/components/notifications/NotificationFetcher.kt
Lakoja: Ok

It's amazing how many helper functions Kotlin has.
Even if it hadn't it would have made sense to write our own extension for this - as this reads much nicer and more concise now...

app/src/main/java/com/keylesspalace/tusky/components/notifications/NotificationFetcher.kt
Lakoja: Changed

app/src/main/java/com/keylesspalace/tusky/components/notifications/NotificationFetcher.kt
nikclayton: Right, that's what confuses me. Because in that comment you note that you saw this message being logged if the delay is removed.

> Package has already posted or enqueued 25 notifications. Not showing more. package=com.keylesspalace.tusky"

So, remove the delay and Android will refuse to show some of the notifications.

Lakoja: My findings were from my device (and Android 10). With the emulator Logcat shows something like "Silencing noisy Tusky" as an error after 15 or so (group) notifications without delay.

If they are actually "dropped" I need to check again, also on my device.

However dropping we will have in any case for lots of notifications: a) in our code (50) and b) on actual devices (25 on mine).

So maybe we can (must) leave the delay in here.

But we should (must) switch to batched notifications as soon as possible, I think.

Lakoja: Added delay again. It might help on some devices.

app/src/main/java/com/keylesspalace/tusky/components/notifications/NotificationHelper.java
```diff
@@ -710,7 +710,7 @@ public static boolean filterNotification(@NonNull NotificationManager notificati
     }
 
     @Nullable
-    private static String getChannelId(AccountEntity account, Notification notification) {
+    public static String getChannelId(AccountEntity account, Notification notification) {
```
nikclayton, line 713: Is this change needed now?

app/src/main/java/com/keylesspalace/tusky/components/notifications/NotificationFetcher.kt
```diff
@@ -65,23 +65,29 @@ class NotificationFetcher @Inject constructor(
                         }
                     }
 
+                    val notificationsGrouped = notifications.groupBy { it.type }
```
nikclayton, line 68:
```suggestion
                    val notificationsByType = notifications.groupBy { it.type }
```

?

app/src/main/java/com/keylesspalace/tusky/components/notifications/NotificationFetcher.kt
```diff
@@ -65,23 +65,29 @@ class NotificationFetcher @Inject constructor(
...
-                            account,
-                            index == 0
-                        )
-                        notificationManager.notify(notification.id, account.id.toInt(), androidNotification)
-                        // Android will rate limit / drop notifications if they're posted too
-                        // quickly. There is no indication to the user that this happened.
-                        // See https://github.com/tuskyapp/Tusky/pull/3626#discussion_r1192963664
-                        delay(1000.milliseconds)
+
+                    notificationsGrouped.forEach { notificationListEntry ->
```
nikclayton, line 75:
```suggestion
                    notificationsGrouped.forEach { notificationGroup ->
```

?

app/src/main/java/com/keylesspalace/tusky/components/notifications/NotificationFetcher.kt
Lakoja: ok

app/src/main/java/com/keylesspalace/tusky/components/notifications/NotificationFetcher.kt
Lakoja: ok (`notificationsGroup`)

Tak (approved):
Looks good to me, but let's hold off on merging anything for a bit

connyduck (approved):

Tak:
@Lakoja If you want to resolve the conflicts, we can merge this for the next release

PR #3822, opened 2023-07-08
3819: Fix the tab list display (in TabPreferenceActivity)

### LINKED ISSUE #3819: Selecting a list for a tab shows unformatted IDs
In Tusky 23 beta 1 and 2, when going to the account preferences and choosing which tabs to show, adding a list tab brings up the lists...but instead of showing the title, it shows each list as something like  `MastoList(id=123, title=Tech Projects)`

They're still functional, they just look awful.

Confirmed with current Mastodon and the latest GoToSocial RC

Lakoja:
Fixes #3819

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/entity/MastoList.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/entity/MastoList.kt
@@ -1,4 +1,4 @@
-/* Copyright 2017 Andrew Dawson
+/* Copyright 2023 Tusky contributors
  *
  * This file is a part of Tusky.
  *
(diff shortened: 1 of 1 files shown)


connyduck:
Not sure if that is the ideal fix. I sometimes use toString for debugging but this breaks it for MastoList 🤔

Lakoja:
That sounds a bit unpredictable (toString for "object representation during debug")?

However, I find it also odd that the api/ArrayAdapter would use toString for something... But as that class exists and esp. can replace the custom class ListSelectionAdapter fully, I guess it is ok.

connyduck:
That is the thing, for data classes it is predictable

nikclayton:
https://github.com/tuskyapp/Tusky/pull/3823 is probably the cleaner fix. @Lakoja, I tried adding you as a reviewer to that PR but Github wouldn't list your account for some reason.

nikclayton:
> However, I find it also odd that the Android api/ArrayAdapter would use toString for something

`ArrayAdapter` assumes that you're handing it an array of strings (or things that can usefully be represented as strings).

PR #3823, opened 2023-07-08
Display the list's title in the array adapter

### LINKED ISSUE #3819: Selecting a list for a tab shows unformatted IDs

nikclayton:
Fixes #3819

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/TabPreferenceActivity.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/TabPreferenceActivity.kt
@@ -21,6 +21,7 @@ import android.os.Bundle
 import android.util.Log
 import android.view.Gravity
 import android.view.View
+import android.view.ViewGroup
 import android.widget.ArrayAdapter
 import android.widget.FrameLayout
 import android.widget.LinearLayout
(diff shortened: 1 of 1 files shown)


connyduck (approved):

app/src/main/java/com/keylesspalace/tusky/TabPreferenceActivity.kt
```diff
@@ -273,7 +274,13 @@ class TabPreferenceActivity : BaseActivity(), Injectable, ItemInteractionListene
     }
 
     private fun showSelectListDialog() {
-        val adapter = ArrayAdapter<MastoList>(this, android.R.layout.simple_list_item_1)
+        val adapter = object : ArrayAdapter<MastoList>(this, android.R.layout.simple_list_item_1) {
+            override fun getView(position: Int, convertView: View?, parent: ViewGroup): View {
+                val view = super.getView(position, convertView, parent)
+                getItem(position)?.let { item -> (view as TextView).text = item.title }
```
Lakoja, line 280: (If I remeber correctly there was a sort of "?as" )

Lakoja (approved):

app/src/main/java/com/keylesspalace/tusky/TabPreferenceActivity.kt
nikclayton: `as?` would silently fail. So if the layout was changed but this wasn't we'd not know unless someone told us.  Using `as` like this means it'll crash, which will show up in the logs.

PR #3835, opened 2023-07-11
Kotlin 1.9.0

Goooler:
https://kotlinlang.org/docs/whatsnew19.html


Goooler:
The ktlint failure related to https://github.com/JLLeitschuh/ktlint-gradle/issues/692.

connyduck:
> The ktlint failure related to [JLLeitschuh/ktlint-gradle#692](https://github.com/JLLeitschuh/ktlint-gradle/issues/692).

Yes than please undo that change, I don't think we should wait for an update to the ktlint plugin.

Goooler:
The lint issues seem should be fixed in a separate PR by creating a baseline file?

nikclayton:
> The lint issues seem should be fixed in a separate PR by creating a baseline file?

I don't think so, these look like issues that would crash on API < 24, e.g., 

```
/bitrise/src/app/src/main/java/com/keylesspalace/tusky/adapter/AccountFieldEditAdapter.kt:36: Error: Call requires API level 24 (current min is 23): java.lang.Iterable#forEach [NewApi]
        fields.forEach { field ->
```

So I don't think we should ignore them.

connyduck:
Seems like a false positive, it is correctly using the Kotlin stdlib extension function and not the Java method. This PR is not crashing on Api 23, I tried it out. No idea why lint suddenly reports it though?

connyduck:
@Gooler you can recreate the baseline file by deleting `app/lint-baseline.xml` and running `./gradlew app:lintGreenDebug`

connyduck (approved):

nikclayton:
> Seems like a false positive, it is correctly using the Kotlin stdlib extension function and not the Java method. This PR is not crashing on Api 23, I tried it out. No idea why lint suddenly reports it though?

Looks like https://issuetracker.google.com/issues/185418482 :sigh:

connyduck:
> Execution failed for task ':app:lintAnalyzeGreenDebug'.
> A failure occurred while executing com.android.build.gradle.internal.lint.AndroidLintWorkAction
   > com/android/tools/lint/client/api/IssueRegistry has been compiled by a more recent version of the Java Runtime (class file version 61.0), this version of the Java Runtime only recognizes class file versions up to 55.0

Not sure what exactly that means? But maybe it will be resolved once https://github.com/tuskyapp/Tusky/pull/3541 is merged?

nikclayton:
@Goooler -- thanks!

PR #3840, opened 2023-07-12
Enable gradle build cache for Bitrise builds

nikclayton:
Read/write to the cache for nightly / release builds, read from the cache for other builds.


nikclayton:
This might speed up builds, per https://devcenter.bitrise.io/en/dependencies-and-caching/remote-build-caching/remote-build-cache-for-gradle.html#configuring-remote-build-caching-for-gradle-in-local-builds

PR #3844, opened 2023-07-12
Set statusView width/height to match_parent

nikclayton:
Fixes an issue where the view's width is only enough to wrap the image, resulting in a very narrow view that obscures the text of the error.


connyduck (approved):
I think we fixed this for every view individually now 😅

PR #3855, opened 2023-07-14
WIP: Check for updates

nikclayton:
This is v22 with some code to check for updates. More details when I've done more testing. In the meantime, I need an APK from CI so it's been signed with the right keys.


connyduck:
I'm against deploying different flavors to F-Droid and Google Play. Unnecessary additional complexity. Also, including a Google Play library will give us a "Upstream non free" antifeature on F-Droid which is a no-go imo.

nikclayton:
> I'm against deploying different flavors to F-Droid and Google Play. Unnecessary additional complexity.

Some more complexity, yes, but we don't know it's unnecessary yet. At the time of writing approx. 30K users (about 18%) are still on a version < 22 (which was released over a month ago

> Also, including a Google Play library will give us a "Upstream non free" antifeature on F-Droid which is a no-go imo.

Doesn't have to. At the moment the release process attaches APKs from Google Play to the GitHub release, but we could change that, and attach an APK built from the fdroid flavor instead.

Then "Upstream non-free" wouldn't apply.

charlag:
I would also not like different flavors again unless we *really really* need jt

Tak:
I thought it was an explicit project goal not to rely on google services and/or non-free dependencies?

connyduck:
Yes @nikclayton please don't waste you time on this. 

(Theoretically we could implement a custom update check, e.g. by hosting a version.json on tusky.app and periodically querying that, but that will add a new step to releasing and its probably not as easy as it sounds...)

nikclayton:
https://github.com/tuskyapp/Tusky/pull/3882 is a much cleaner version of this which works for F-Droid and GitHub releases too, and doesn't introduce "Upstream non free" on F-Droid.

PR #3867, opened 2023-07-20
Work around "Illegal reflective access" warnings in kapt

nikclayton:
See https://youtrack.jetbrains.com/issue/KT-30589/Kapt-An-illegal-reflective-access-operation-has-occurred for details.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -186,3 +186,20 @@ dependencies {
     androidTestImplementation libs.androidx.room.testing
     androidTestImplementation libs.androidx.test.junit
 }
+
+// Work around warnings of:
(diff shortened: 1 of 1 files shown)


nikclayton:
The warning breaks incremental compilation, so fixing this significantly improves the developer experience.

charlag:
the linked SO answer says it's not needed with JDK 16 anymore though?

nikclayton:
> the linked SO answer says it's not needed with JDK 16 anymore though?

We currently use 11.

nikclayton:
Now on Java 17, so this is not necessary.

PR #3870, opened 2023-07-22
Clear listeners, hide icons when appropriate

nikclayton:
Preferences are shown using view holders.

The previous code did not clear the listeners or hide the icons if necessary.

The practical upshot of this was that if you had two or more slider preferences, *and* they were situated more than a screen's height apart, the viewholder from the first one would get reused.

And if the first one enabled icons then the second one would show them. And clicking on the second one would also call the listeners for the first one.


app/src/main/java/com/keylesspalace/tusky/view/SliderPreference.kt
```diff
@@ -130,6 +131,8 @@ class SliderPreference @JvmOverloads constructor(
 
         binding.root.isClickable = false
 
+        binding.slider.clearOnChangeListeners()
+        binding.slicer.clearOnSliderTouchListeners()
```
nikclayton, line 135:
```suggestion
        binding.slider.clearOnSliderTouchListeners()
```

Lakoja (approved):

connyduck (approved):

PR #3882, opened 2023-07-26
Prompt the user if a new version is available

nikclayton:
Analytics show that up to a fifth of users are using an older version of Tusky, sometimes going back six months or more. While this won't help them, it will help other people not get in to that situation.

On resuming MainActivity, and approximately once per day, check and see if a newer version of Tusky is available, and prompt the user to update by going to the relevant install location (Google Play, F-Droid, or GitHub).

The dialog prompt allows them to ignore this specific version, or disable all future update notifications. This is also exposed through the preferences, so the user can adjust it there too.

A different update check method is used for each installation location.

- F-Droid: Use the F-Droid API to query for the newest released version
- GitHub: Use the GitHub API to query for the newest release, and check the APK filename attached to that release
- Google Play: Use the Play in-app-updates library (https://developer.android.com/guide/playcore/in-app-updates) to query for the newest released version

These are kept in different build flavours (source sets), so that e.g., the build for the F-Droid store can only query the F-Droid API, the UI strings are specific to F-Droid, etc. This also ensures that the update service libraries are specific to that build and do not "cross-contaminate".

Note that this *does not* update the app, it takes the user to either the relevant store page (F-Droid, Play) or GitHub release page. The user must still start the update from that page.


bitrise.yml
```diff
@@ -66,15 +66,15 @@ workflows:
         - retry_on_failure: "no"
         - gradlew_path: ./gradlew
         - gradle_options: --no-daemon
-        - gradle_task: ktlintCheck lintGreenDebug
+        - gradle_task: ktlintCheck lintGreen
```
nikclayton, line 69:
```suggestion
        - gradle_task: ktlintCheck lintGreenFdroidDebug
```

nikclayton: Use the correct lintGreenFdroidDebug task

bitrise.yml
```diff
@@ -66,15 +66,15 @@ workflows:
...
     - android-unit-test@1.0:
         inputs:
         - project_location: $PROJECT_LOCATION
         - module: app
-        - variant: greenDebug
+        - variant: green
     - android-build:
         inputs:
-        - variant: greenDebug
+        - variant: green
```
nikclayton, line 77:
```suggestion
        - variant: greenFdroidDebug
```

bitrise.yml
```diff
@@ -66,15 +66,15 @@ workflows:
...
         - gradlew_path: ./gradlew
         - gradle_options: --no-daemon
-        - gradle_task: ktlintCheck lintGreenDebug
+        - gradle_task: ktlintCheck lintGreenFdroidDebug
     - android-unit-test@1.0:
         inputs:
         - project_location: $PROJECT_LOCATION
         - module: app
-        - variant: greenDebug
+        - variant: green
```
nikclayton, line 74:
```suggestion
        - variant: greenFdroidDebug
```

nikclayton:
Note to self: Needs adjusting to disable for the F-Droid nightly repo, as there's no API to query the version.

https://releases.nailyk.fr/repo/index-v2.json appears to be the JSON file that contains the data from @nailyk-fr 's repository, but they've noted that once the user installs Tusky Nightly from there they never get prompted to update. Inspecting that JSON file, the versionCode for the nightly build looks wrong.

The nightly build config for Bitrise has `change-android-versioncode-and-versionname@1.3: {}` as a step, which looks important.

That's https://github.com/bitrise-steplib/steps-change-android-versioncode-and-versionname, 

By default that sets the new versionCode to the Bitrise build number, and leaves the versionName unmodified. Since the Nightly version of test appends the git SHA to the versionName this works out OK.

PR #3890, opened 2023-07-27
3532: Show badge on conversations tab on new conversations

### LINKED ISSUE #3532: Show indicator for new direct messages
Sort of part of #3405 (but simpler; more easily implementable):

When there are new (unread) direct messages there should be an indicator (badge) on the tab icon.

Lakoja:
Fixes #3532

(Old PR, now closed: https://github.com/tuskyapp/Tusky/pull/3533)

Listens on new notifications and if a "direct mention" is detected a badge (red dot) is shown on the conversations tab if present.

I am missing things like this a lot and also big accounts are unhappy with the usability so far: https://mastodon.social/@pallenberg/110129889996182814

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/MainActivity.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/MainActivity.kt
@@ -336,7 +336,7 @@ class MainActivity : BottomSheetActivity(), ActionButtonActivity, HasAndroidInje
                     }
                     is NewNotificationsEvent -> {
                         directMessageTab?.let { tab ->
-                            if (event.accountId == accountManager.activeAccount?.accountId) {
+                            if (event.accountId == activeAccount.accountId) {
                                 val hasDirectMessageNotification =
                                     event.notifications.any { it.type == Notification.Type.MENTION && it.status?.visibility == Status.Visibility.DIRECT }
 
@@ -347,12 +347,12 @@ class MainActivity : BottomSheetActivity(), ActionButtonActivity, HasAndroidInje
                         }
                     }
                     is NotificationsLoadingEvent -> {
-                        if (event.accountId == accountManager.activeAccount?.accountId) {
+                        if (event.accountId == activeAccount.accountId) {
                             showDirectMessageBadge(false)
                         }
                     }
                     is ConversationsLoadingEvent -> {
-                        if (event.accountId == accountManager.activeAccount?.accountId) {
+                        if (event.accountId == activeAccount.accountId) {
                             showDirectMessageBadge(false)
                         }
                     }
--- a/app/src/main/java/com/keylesspalace/tusky/MainActivity.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/MainActivity.kt
@@ -67,8 +67,11 @@ import com.google.android.material.tabs.TabLayout.OnTabSelectedListener
 import com.google.android.material.tabs.TabLayoutMediator
 import com.keylesspalace.tusky.appstore.AnnouncementReadEvent
 import com.keylesspalace.tusky.appstore.CacheUpdater
+import com.keylesspalace.tusky.appstore.ConversationsLoadingEvent
 import com.keylesspalace.tusky.appstore.EventHub
 import com.keylesspalace.tusky.appstore.MainTabsChangedEvent
+import com.keylesspalace.tusky.appstore.NewNotificationsEvent
+import com.keylesspalace.tusky.appstore.NotificationsLoadingEvent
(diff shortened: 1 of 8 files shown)


app/src/main/java/com/keylesspalace/tusky/MainActivity.kt
```diff
@@ -294,11 +300,32 @@ class MainActivity : BottomSheetActivity(), ActionButtonActivity, HasAndroidInje
...
                         setupTabs(false)
                     }
-
                     is AnnouncementReadEvent -> {
                         unreadAnnouncementsCount--
                         updateAnnouncementsBadge()
                     }
+                    is NewNotificationsEvent -> {
+                        directMessageTab?.let { tab ->
+                            if (event.accountId == accountManager.activeAccount?.accountId) {
```
nikclayton, line 309: Don't use `accountManager.activeAccount` here (and elsewhere), as that can change over the life of the activity, and you get the activity thinking it's working with one account and it's actually working with another.

```
private lateinit var activeAccount: AccountEntity

// ...

    // Somewhere in onCreate
    activeAccount = accountManager.activeAccount!!
```

should do it.

nikclayton (changes requested):
Interesting idea, but I'm not sure about the approach.

Conversations in Mastodon have an "unread" flag, and conversations are currently stored in the database.

So what if:

1. This was a count of unread conversations
2. A new UI affordance to mark a conversation as read was introduced (https://docs.joinmastodon.org/methods/conversations/#read). There's no "Mark a conversation as unread" equivalent unfortunately
3. Introduce a `ConversationsRepository` to act as the point-of-truth for conversation information
4. Inject the ConversationsRepository in to MainActivity, and let MainActivity get a flow of updates to the count of unread conversations.
5. Inject the ConversationsRepository in to the incoming notification handler, and if it sees that a notification is about a conversation then it tells the ConversationsRepository to update itself

That should keep everything architecturally clean.

Lakoja:
>     2. A new UI affordance to mark a conversation as read was introduced (https://docs.joinmastodon.org/methods/conversations/#read). There's no "Mark a conversation as unread" equivalent unfortunately
> 
>     3. Introduce a `ConversationsRepository` to act as the point-of-truth for conversation information

I like the idea of a single source of truth - and didn't know we had "conversations" in the db.

The "mark as read" management must be automatic however.

Two questions pop up:

- When Tusky sees a notification about a conversation it tells the repository to update itself which means a (at least one) followup request and an actual update of local db information - "just" for a notification?
- Architectually: It is not the job of the NotificationsFetcher (or Handler) to deal with anything else than notification business? So especially not analysing content and call methods accordingly. That's why I went with an event: everything can have a single responsibility topic.
(I know that question is attackable as the notification event receiver has to deal with notifications AND conversations but not/never more.)

Lakoja:
> > The "mark as read" management must be automatic however.
> 
> Not yet, because the user can't undo it. The user might view a conversation, make a mental note to come back to it later, and not want it marked as read.

My plan for this is that it is a small helper. It should not need the user to do (constant) extra work.

> 
> There is going to be support for this ([mastodon/mastodon#25509](https://github.com/mastodon/mastodon/pull/25509)) in the future though.

Lakoja:
> > My plan for this is that it is a small helper. It should not need the user to do (constant) extra work.
> 
> My point is that if there is no way to undo an action, Tusky should not perform that action without explicit opt in from the user.

Then this is our main and blocking beef here. :-)

This is an indicator. It will show and disappear to convey information. All of this is done automatic. (think: indicator for notifications in Android)

It _will_ get difficult to make this automatic (needs to track what status are visible in a timeline).

As a simple first step it is therefor probably sufficient as is done in this PR: hide the indicator again if the conversations tab is opened.

app/src/main/java/com/keylesspalace/tusky/MainActivity.kt
Tak: This is valid

Tak (approved):
Apart from the active account thing, it looks good to me

app/src/main/java/com/keylesspalace/tusky/MainActivity.kt
Lakoja: Ok, I changed that in MainActivity.

There are two more locations with "activeAccount" though in the NotificationFragment and ConversationFragment on loading.
I left i unchanged there.

When testing this seems to work ok (loading in one account does not delete the badge of the other one).

However I'm not quite sure in both locations when the "activeAccount" is what and if it is ok to directly use accountManager.activeAccount in the fragments.

PR #3892, opened 2023-07-27
Start/end the thread divider in the middle of the avatar

nikclayton:
Previously, the thread indicator would start at the top of the avatar for the status at the start of the thread, and end at the top of the avatar for the status at the end of the thread.

If these avatars were partially transparent the thread indicator could either (a) poke out of the top of the avatar at the start of the thread, (b) not properly connect with the avatar at the end of the thread, or (c) both.

Partially fix this by making the divider start/stop in the middle of the avatar. This assumes that this area will typically have opaque content, even if some of the rest of the avatar is transparent. This is not always true, but it's still better than the current behaviour.


nikclayton:
Note: This is stacked on top of https://github.com/tuskyapp/Tusky/pull/3874, change the base before viewing the PR.

nikclayton:
Here are before / after screenshots. In these screenshots I have artifically reduced the alpha of the avatars to 0.2f so the boundaries of the divider can be seen properly. Obviously, this alpha reduction is not present in the PR.

charlag:
I would rather propose to start it below and and above? I see no reason to go inside on either side?

nikclayton:
> I would rather propose to start it below and and above? I see no reason to go inside on either side?

It's necessary for avatars that have partial or full transparency in places.

There's an example in https://github.com/tuskyapp/Tusky/pull/3874

charlag:
I would say that the line doesn't have to touch the avatar, it would be even more aesthetically pleasing if it didn't

nikclayton:
> I would say that the line doesn't have to touch the avatar, it would be even more aesthetically pleasing if it didn't

Then it's weird if the avatar is has areas of transparency.

For example, these two with transparent areas at the top:

image

image

---

This one with transparent areas top and bottom

image

mcclure:
Reviewed code, LGTM.

Also tested, saw no problems. (I did see #3934 but on testing, this also occurs on vanilla Tusky 23, so it is not a regression and should not hold back this PR.)

mcclure (approved):

PR #3893, opened 2023-07-27
3891 better hashtag filtering

### LINKED ISSUE #3891: Filtering hashtags filters too much, not helping
grafik

There is a mute button on a hashtag timeline. I noticed several problems there:

1. The added filter also filters the word without being a hashtag; so muting "#X" will mute "X"
2. The buttons here give no feedback what they are doing
3. The added filter only hides but does not remove completely
4. The home timeline is not re-rendered after a change

(2) There should be at least a dialog that for example "Hashtag X was followed". But furthermore the next steps are not obvious: You can see and especially tweak the "mute filter" under Account settings > Filters. I guess no one will find that easily.

(3) Can be discussed what the default here is.
But this actually got me started on this problem when reading a complaint that muting a hashtag still shows it; albeit as "hidden; do you want to see it?".

(4) It can be workarounded by refreshing the view yourself. But I fear many will not know this.

* * * *
- Tusky Version: v22+
(Code introduced with #3188)

Lakoja:
Fixes #3891 (the first two observerations; the rest is either ok or has another issue)

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/StatusListActivity.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/StatusListActivity.kt
@@ -174,6 +174,7 @@ class StatusListActivity : BottomSheetActivity(), HasAndroidInjector {
      */
     private fun updateMuteTagMenuItems() {
         val tag = hashtag ?: return
+        val hashedTag = if (tag.startsWith('#')) tag else "#" + tag
 
         muteTagItem?.isVisible = true
         muteTagItem?.isEnabled = false
@@ -184,7 +185,7 @@ class StatusListActivity : BottomSheetActivity(), HasAndroidInjector {
                 { filters ->
                     mutedFilter = filters.firstOrNull { filter ->
                         filter.context.contains(Filter.Kind.HOME.kind) && filter.keywords.any {
-                            it.keyword == tag
+                            it.keyword == hashedTag
                         }
                     }
                     updateTagMuteState(mutedFilter != null)
@@ -194,7 +195,7 @@ class StatusListActivity : BottomSheetActivity(), HasAndroidInjector {
                         mastodonApi.getFiltersV1().fold(
                             { filters ->
                                 mutedFilterV1 = filters.firstOrNull { filter ->
-                                    tag == filter.phrase && filter.context.contains(FilterV1.HOME)
+                                    hashedTag == filter.phrase && filter.context.contains(FilterV1.HOME)
                                 }
                                 updateTagMuteState(mutedFilterV1 != null)
                             },
@@ -224,10 +225,10 @@ class StatusListActivity : BottomSheetActivity(), HasAndroidInjector {
 
     private fun muteTag(): Boolean {
         val tag = hashtag ?: return true
-        val hashedTag = if (tag.startsWith('#')) tag else "#" + tag
 
         lifecycleScope.launch {
             var filterCreateSuccess = false
+            val hashedTag = if (tag.startsWith('#')) tag else "#" + tag
 
             mastodonApi.createFilter(
                 title = "#$tag",
@@ -236,7 +237,7 @@ class StatusListActivity : BottomSheetActivity(), HasAndroidInjector {
                 expiresInSeconds = null
             ).fold(
                 { filter ->
-                    if (mastodonApi.addFilterKeyword(filterId = filter.id, keyword = tag, wholeWord = true).isSuccess) {
+                    if (mastodonApi.addFilterKeyword(filterId = filter.id, keyword = hashedTag, wholeWord = true).isSuccess) {
                         mutedFilter = filter
                         // TODO the preference key here ("home") is not meaningful; should probably be another event if any
                         eventHub.dispatch(PreferenceChangedEvent(filter.context[0]))
@@ -249,7 +250,7 @@ class StatusListActivity : BottomSheetActivity(), HasAndroidInjector {
                 { throwable ->
                     if (throwable is HttpException && throwable.code() == 404) {
                         mastodonApi.createFilterV1(
-                            tag,
+                            hashedTag,
                             listOf(FilterV1.HOME),
                             irreversible = false,
                             wholeWord = true,
--- a/app/src/main/java/com/keylesspalace/tusky/StatusListActivity.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/StatusListActivity.kt
@@ -186,9 +186,7 @@ class StatusListActivity : BottomSheetActivity(), HasAndroidInjector {
                 { filters ->
                     mutedFilter = filters.firstOrNull { filter ->
                         // TODO shouldn't this be an exact match (only one keyword; exactly the hashtag)?
-                        filter.context.contains(Filter.Kind.HOME.kind) && filter.keywords.any {
-                            it.keyword == hashedTag
-                        }
+                        filter.context.contains(Filter.Kind.HOME.kind) && filter.title == hashedTag
                     }
                     updateTagMuteState(mutedFilter != null)
                 },
--- a/app/src/main/java/com/keylesspalace/tusky/StatusListActivity.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/StatusListActivity.kt
@@ -27,6 +27,7 @@ import at.connyduck.calladapter.networkresult.fold
 import com.google.android.material.snackbar.Snackbar
 import com.keylesspalace.tusky.appstore.EventHub
 import com.keylesspalace.tusky.appstore.PreferenceChangedEvent
+import com.keylesspalace.tusky.components.filters.FiltersActivity
 import com.keylesspalace.tusky.components.timeline.TimelineFragment
 import com.keylesspalace.tusky.components.timeline.viewmodel.TimelineViewModel.Kind
 import com.keylesspalace.tusky.databinding.ActivityStatuslistBinding
@@ -132,6 +133,8 @@ class StatusListActivity : BottomSheetActivity(), HasAndroidInjector {
                     {
                         followTagItem?.isVisible = false
                         unfollowTagItem?.isVisible = true
+
+                        Snackbar.make(binding.root, getString(R.string.following_hashtag_success_format, tag), Snackbar.LENGTH_LONG).show()
                     },
                     {
                         Snackbar.make(binding.root, getString(R.string.error_following_hashtag_format, tag), Snackbar.LENGTH_SHORT).show()
@@ -152,6 +155,8 @@ class StatusListActivity : BottomSheetActivity(), HasAndroidInjector {
                     {
                         followTagItem?.isVisible = true
                         unfollowTagItem?.isVisible = false
+
+                        Snackbar.make(binding.root, getString(R.string.unfollowing_hashtag_success_format, tag), Snackbar.LENGTH_SHORT).show()
                     },
                     {
                         Snackbar.make(binding.root, getString(R.string.error_unfollowing_hashtag_format, tag), Snackbar.LENGTH_SHORT).show()
@@ -219,8 +224,11 @@ class StatusListActivity : BottomSheetActivity(), HasAndroidInjector {
 
     private fun muteTag(): Boolean {
         val tag = hashtag ?: return true
+        val hashedTag = if (tag.startsWith('#')) tag else "#" + tag
 
         lifecycleScope.launch {
+            var filterCreateSuccess = false
+
             mastodonApi.createFilter(
                 title = "#$tag",
                 context = listOf(FilterV1.HOME),
@@ -230,8 +238,9 @@ class StatusListActivity : BottomSheetActivity(), HasAndroidInjector {
                 { filter ->
                     if (mastodonApi.addFilterKeyword(filterId = filter.id, keyword = tag, wholeWord = true).isSuccess) {
                         mutedFilter = filter
-                        updateTagMuteState(true)
+                        // TODO the preference key here ("home") is not meaningful; should probably be another event if any
                         eventHub.dispatch(PreferenceChangedEvent(filter.context[0]))
+                        filterCreateSuccess = true
                     } else {
                         Snackbar.make(binding.root, getString(R.string.error_muting_hashtag_format, tag), Snackbar.LENGTH_SHORT).show()
                         Log.e(TAG, "Failed to mute #$tag")
@@ -248,8 +257,8 @@ class StatusListActivity : BottomSheetActivity(), HasAndroidInjector {
                         ).fold(
                             { filter ->
                                 mutedFilterV1 = filter
-                                updateTagMuteState(true)
                                 eventHub.dispatch(PreferenceChangedEvent(filter.context[0]))
+                                filterCreateSuccess = true
                             },
                             { throwable ->
                                 Snackbar.make(binding.root, getString(R.string.error_muting_hashtag_format, tag), Snackbar.LENGTH_SHORT).show()
@@ -262,6 +271,16 @@ class StatusListActivity : BottomSheetActivity(), HasAndroidInjector {
                     }
                 }
             )
+
+            if (filterCreateSuccess) {
+                updateTagMuteState(true)
+                Snackbar.make(binding.root, getString(R.string.muting_hashtag_success_format, tag), Snackbar.LENGTH_LONG).apply {
+                    setAction(R.string.action_view_filters) {
+                        startActivityWithSlideInAnimation(Intent(this@StatusListActivity, FiltersActivity::class.java))
+                    }
+                    show()
+                }
+            }
         }
 
         return true
@@ -281,6 +300,7 @@ class StatusListActivity : BottomSheetActivity(), HasAndroidInjector {
                 } else {
                     mastodonApi.deleteFilter(filter.id)
                 }
+
             } else if (mutedFilterV1 != null) {
                 mutedFilterV1?.let { filter ->
                     if (filter.context.size > 1) {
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/keylesspalace/tusky/StatusListActivity.kt
```diff
@@ -221,17 +229,23 @@ class StatusListActivity : BottomSheetActivity(), HasAndroidInjector {
...
                 filterAction = Filter.Action.WARN.action,
                 expiresInSeconds = null
             ).fold(
                 { filter ->
-                    if (mastodonApi.addFilterKeyword(filterId = filter.id, keyword = tag, wholeWord = true).isSuccess) {
-                        mutedFilter = filter
-                        updateTagMuteState(true)
+                    if (mastodonApi.addFilterKeyword(filterId = filter.id, keyword = hashedTag, wholeWord = true).isSuccess) {
+                        // must be requested again; otherwise does not contain the keyword (but server does)
+                        mutedFilter = mastodonApi.getFilter(filter.id).getOrNull()
```
Lakoja, line 244: This is a bit odd (third api call).

We could open the filters list any time - as is done as fallback below.

app/src/main/java/com/keylesspalace/tusky/StatusListActivity.kt
```diff
@@ -178,8 +185,9 @@ class StatusListActivity : BottomSheetActivity(), HasAndroidInjector {
             mastodonApi.getFilters().fold(
                 { filters ->
                     mutedFilter = filters.firstOrNull { filter ->
+                        // TODO shouldn't this be an exact match (only one keyword; exactly the hashtag)?
```
nikclayton, line 188: That's probably worth checking.  This was originally used with v1 filters, where you could only specify one keyword (or phrase) per filter.

nikclayton: Although you need to be careful here, since always adding the "#" will mean that hashtags that were previously shown as muted will no longer be shown as muted, possibly confusing users. It might be worth checking both.

app/src/main/java/com/keylesspalace/tusky/StatusListActivity.kt
```diff
@@ -169,6 +175,7 @@ class StatusListActivity : BottomSheetActivity(), HasAndroidInjector {
      */
     private fun updateMuteTagMenuItems() {
         val tag = hashtag ?: return
+        val hashedTag = if (tag.startsWith('#')) tag else "#" + tag
```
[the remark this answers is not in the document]
nikclayton: And elsewhere the same idiom is used.

app/src/main/java/com/keylesspalace/tusky/StatusListActivity.kt
nikclayton: Maybe couple it with checking the title of the filter?

app/src/main/java/com/keylesspalace/tusky/StatusListActivity.kt
```diff
@@ -221,17 +229,23 @@ class StatusListActivity : BottomSheetActivity(), HasAndroidInjector {
...
-                        updateTagMuteState(true)
+                    if (mastodonApi.addFilterKeyword(filterId = filter.id, keyword = hashedTag, wholeWord = true).isSuccess) {
+                        // must be requested again; otherwise does not contain the keyword (but server does)
+                        mutedFilter = mastodonApi.getFilter(filter.id).getOrNull()
+
+                        // TODO the preference key here ("home") is not meaningful; should probably be another event if any
                         eventHub.dispatch(PreferenceChangedEvent(filter.context[0]))
+                        filterCreateSuccess = true
                     } else {
                         Snackbar.make(binding.root, getString(R.string.error_muting_hashtag_format, tag), Snackbar.LENGTH_SHORT).show()
```
nikclayton, line 250: I'd remove the '#' from the format string and just pass in `hashedTag` here. It's a little confusing that the code uses `hashedTag` everywhere, then reverts to `tag` here.

Also, per the earlier check, `tag` may already start with a `#`, so including the `#` in the format string means it would appear doubled.

app/src/main/java/com/keylesspalace/tusky/StatusListActivity.kt
nikclayton: Why do you need this?

app/src/main/java/com/keylesspalace/tusky/StatusListActivity.kt
```diff
@@ -262,6 +276,24 @@ class StatusListActivity : BottomSheetActivity(), HasAndroidInjector {
                     }
                 }
             )
+
+            if (filterCreateSuccess) {
+                updateTagMuteState(true)
```
nikclayton, line 281: Can this be moved outside the `if` block, and just be `updateTagMuteState(filterCreateSuccess)` ?

app/src/main/java/com/keylesspalace/tusky/StatusListActivity.kt
```diff
@@ -262,6 +276,24 @@ class StatusListActivity : BottomSheetActivity(), HasAndroidInjector {
...
+                Snackbar.make(binding.root, getString(R.string.muting_hashtag_success_format, tag), Snackbar.LENGTH_LONG).apply {
+                    setAction(R.string.action_view_filter) {
+                        val intent = if (mutedFilter != null) {
+                            Intent(this@StatusListActivity, EditFilterActivity::class.java).apply {
+                                putExtra(EditFilterActivity.FILTER_TO_EDIT, mutedFilter)
+                            }
+                        } else {
+                            Intent(this@StatusListActivity, FiltersActivity::class.java)
+                        }
```
nikclayton, line 290:
```suggestion
                        val intent = mutedFilter?.let {
                            Intent(this@StatusListActivity, EditFilterActivity::class.java).apply {
                                putExtra(EditFilterActivity.FILTER_TO_EDIT, it)
                            }
                        } ?: Intent(this@StatusListActivity, FiltersActivity::class.java)
```

Better would be to add static factory methods to `FiltersActivity` and `EditFilterActivity` that return the intent (like, e.g., `AccountListActivity.newIntent()`)

nikclayton (changes requested):

app/src/main/java/com/keylesspalace/tusky/StatusListActivity.kt
Lakoja: What? This call?
To get the correct current filter with that id.

Otherwise the filter opened will be incomplete (and cannot directly be saved for example).

app/src/main/java/com/keylesspalace/tusky/StatusListActivity.kt
Lakoja: (tried to use `tag` for everything/all outputs and `hashedTag` only for the actual api code and comparisson.)

app/src/main/java/com/keylesspalace/tusky/StatusListActivity.kt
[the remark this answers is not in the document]
Lakoja: ok

app/src/main/java/com/keylesspalace/tusky/StatusListActivity.kt
Lakoja: It was "inside" before.

Moving it outside would set it to a different (equally not-meaningful) state when there is an error.

app/src/main/java/com/keylesspalace/tusky/StatusListActivity.kt
```diff
@@ -281,6 +313,7 @@ class StatusListActivity : BottomSheetActivity(), HasAndroidInjector {
                 } else {
                     mastodonApi.deleteFilter(filter.id)
                 }
+
```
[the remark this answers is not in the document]
Lakoja: ok

app/src/main/java/com/keylesspalace/tusky/StatusListActivity.kt
Lakoja: Will do if we keep this.

See "normally open filter overview or the specific one" discussion above.

app/src/main/java/com/keylesspalace/tusky/StatusListActivity.kt
Lakoja: It would probably be wise to just remove the "if" in the declaration of `hashedTag`. The current code always assumes that `tag` (this.hashtag) is without #.

Otherwise I would also change the error code strings (which would be a mess).

app/src/main/java/com/keylesspalace/tusky/StatusListActivity.kt
Lakoja: Hm, as that filter can be changed freely afterwards (i.e. adding other keywords) this is always a heuristic, I guess.

With tapping "unmute" one could destroy (involuntarily) some config work.

Or conversely with an exact match, the state of the button could be wrong; and you add and remove a doubled filter.

app/src/main/java/com/keylesspalace/tusky/StatusListActivity.kt
```diff
@@ -262,6 +276,24 @@ class StatusListActivity : BottomSheetActivity(), HasAndroidInjector {
...
+            if (filterCreateSuccess) {
+                updateTagMuteState(true)
+                Snackbar.make(binding.root, getString(R.string.muting_hashtag_success_format, tag), Snackbar.LENGTH_LONG).apply {
+                    setAction(R.string.action_view_filter) {
+                        val intent = if (mutedFilter != null) {
+                            Intent(this@StatusListActivity, EditFilterActivity::class.java).apply {
+                                putExtra(EditFilterActivity.FILTER_TO_EDIT, mutedFilter)
+                            }
+                        } else {
+                            Intent(this@StatusListActivity, FiltersActivity::class.java)
```
nikclayton, line 289: So v2 filters get EditFilterActivity, and v1 filters get FiltersActivity, is that right?

app/src/main/java/com/keylesspalace/tusky/StatusListActivity.kt
```diff
@@ -152,6 +156,8 @@ class StatusListActivity : BottomSheetActivity(), HasAndroidInjector {
                     {
                         followTagItem?.isVisible = true
                         unfollowTagItem?.isVisible = false
+
+                        Snackbar.make(binding.root, getString(R.string.unfollowing_hashtag_success_format, tag), Snackbar.LENGTH_SHORT).show()
```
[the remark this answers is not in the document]
Lakoja: That one got the "LONG" (only) for the action button.

app/src/main/java/com/keylesspalace/tusky/StatusListActivity.kt
Lakoja: Sounds reasonable (and a bit involved).

I'd do it in another issue or at least another PR - as the problem is not new here.

app/src/main/java/com/keylesspalace/tusky/StatusListActivity.kt
Lakoja: At the moment, yes.

(As it not checks `mutedFilterV1`.)

app/src/main/java/com/keylesspalace/tusky/StatusListActivity.kt
Lakoja: Another observation (which could be done in this PR): This code in EditFilterViewModel and StatusListActivity is double code - esp. for this rather complex "fallback to v1" case. They could be joined.

app/src/main/java/com/keylesspalace/tusky/StatusListActivity.kt
```diff
@@ -178,8 +185,9 @@ class StatusListActivity : BottomSheetActivity(), HasAndroidInjector {
             mastodonApi.getFilters().fold(
                 { filters ->
                     mutedFilter = filters.firstOrNull { filter ->
+                        // TODO shouldn't this be an exact match (only one keyword; exactly the hashtag)?
                         filter.context.contains(Filter.Kind.HOME.kind) && filter.keywords.any {
-                            it.keyword == tag
+                            it.keyword == hashedTag
```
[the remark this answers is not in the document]
Lakoja: Ok

(Push pending the 3 api calls or 1 api call discussion)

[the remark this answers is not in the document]
Lakoja: However this case can never be removed (or filters need to be migrated).

app/src/main/java/com/keylesspalace/tusky/StatusListActivity.kt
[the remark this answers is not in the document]
nikclayton: I meant the one on line 138, that's LENGTH_LONG with no action associated with the snackbar.

The success snackbars should persist for consistent lengths. Over in https://github.com/tuskyapp/Tusky/blob/develop/app/src/main/java/com/keylesspalace/tusky/components/notifications/NotificationsFragment.kt#L305-L322 I used `LENGTH_SHORT`, 

The error snackbars should persist indefinitely until the user dismisses them. This is so the user has the chance to note down the error (or take a screenshot) in case they need to report the problem (either to us, or to their server admin).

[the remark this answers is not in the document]
Lakoja: Ok, changed that one to SHORT.

The one with the action remains LONG.

app/src/main/java/com/keylesspalace/tusky/StatusListActivity.kt
Lakoja: Ok, did that now (remove if)

app/src/main/java/com/keylesspalace/tusky/StatusListActivity.kt
Lakoja: These are the two and a half open questions here:

* Go for a single `createFilter`: this requires testing and/or research why it was "terrible" in the past; and related the question if this is necessary at all here (it is not if always the filter list is displayed; see "FiltersActivity vs. EditFilterActivity" below)
* Join double code (here)?

Tak: `createFilter`/`updateFilter`: It's like the comment said - it just didn't work. The filter would be created, but the keywords would be missing or not updated.
See the discussion [starting here](https://matrix.to/#/!PWkUGGpeMZWDPmdfRE:matrix.org/$1673017151510578sqjwI:matrix.org?via=m.charlag.space&via=matrix.org&via=tchncs.de)

It's possible that something has been fixed in mastodon since then, I didn't debug the server 🤷

Lakoja: Thanks. Can you try it again?

Adding 
```
 @Field("keywords_attributes[][keyword]") keywords: List<String>? = null,
 @Field("keywords_attributes[][whole_word]") wholeWord: List<Boolean>? = null
```

to `createFilter`

and to its call for example in `muteTag`

```
                listOf("#$tag"),
                listOf(true)
```

(Which server are you testing?)

Tak: At the time, I was testing on my instance (glitch) and also mastodon.gamedev.place (vanilla masto.host mastodon)

app/src/main/java/com/keylesspalace/tusky/StatusListActivity.kt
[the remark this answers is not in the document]
nikclayton: @Lakoja 

Yes. Thinking about it some more I think filters need to be migrated. Having to special case certain filters deep in the filters code forever is a losing strategy; it's technical debt we'll never be rid of. And I say that as the person who, I think, wrote the code with the bug in the first place.

Here's what I think is a better strategy.

1. Create a new per-account preference that stores a boolean. This boolean is false if the filters haven't been migrated, else true.
2. Check this preference in MainActivity. If it's true, do nothing. If it's false, migrate the filters, and then set the preference to true.

The migration strategy is as follows:

- Fetch and iterate through all the user's filters.
- If the filter title does not start `#` then ignore it. Yes, this will ignore filters if the user changed the title after it was created. If the user does that I assume they're sophisticated enough to know what they're doing.
- Extract the non-`#` part of the filter title (i.e. `#foo` becomes `foo`). If a keyword in the filter matches that then prepend the ''#', and save the filter.

We can mention this in the release notes and link to a longer blog post about it.

Now, the mute/unmute code can assume that filtered tags always have a `#` and can be simpler.

---

Optional bonus extra work: In the filter editing code, put a check in if the user tries to edit any filter where the title starts `#`. The dialog can explain that these filters are created to mute hashtags, and that editing them may cause problems.

Or, disallow editing these filters entirely -- separate them visually, and if the user tries to edit it (e.g., taps on it) put up a dialog that says this is a filter that mutes a hashtag, and the only operation that can be done is to delete it.

Let me know if you would prefer me to do that work. This is the last thing that's blocking the 24.0 release, so I'd like to get this resolved soon.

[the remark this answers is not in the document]
nikclayton: That last idea ("disallow editing these filters entirely") might fit in which how followed hashtags are displayed.

The user can go to "Account preferences" and there's a "Followed hashtags" section.

Maybe there should be a "Muted hashtags" section as well, which shows just the filters that affect hashtags. As before, the only options would be to unmute the hashtag (so, delete the filter), or change the filter type from warn to hide, and vice-versa.

I'm not completely sold on this idea -- if the user manages their filters from apps other than Tusky there's the possibility to get things in to a weird state that'll be difficult to recover from within Tusky.

But it might also be easier for users to understand.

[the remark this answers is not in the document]
Lakoja: How about this (just commited):

Only compare the title.

This has the advantages:
- Works with old and new code
- Thus does not require migration (which is _quite_ complex as it involves the api)
- Possibility of some failure (someone editing that filter) is rather limited
- Leaves the possibility to still do some migration/fix in the future (and/or limit editing of those filters)

What is the state of "V1" here? Where is that needed or used?
(Because this fix does not work for that and also cannot be migrated?)

Lakoja:
If the current "matching" (title) is ok,

there would only remain one decision: Directly go to the filter (3 api calls instead of 2) or always go to the filter list from the snackbar action?

And then there would be one or two follow-up issues:
- Reduce 3 api calls for creation to 1 (and test)
- Do a precise matching (hashtag as keyword somewhere); delete only that keywork and the filter only if that is the only content. This can then be discussed and migrated. But seems to me neither is necessary.

Issue #2297, opened 2022-01-19
Panning a zoomed-in image stops under certain conditions


kuba-orlik:
Steps to reproduce:

1. Find a post with an image attached
2. Open the image
3. Zoom in the image so its height is larger than the screen height
4. Pan the image so the upper edge of the image is aligned with the upper edge of the screen
5. Try to pan the image left or right while moving the finger slightly downwards
6. See that the panning is stopped as if there is no more picture in that direction (even though there is)

It seems that the application confuses the upper edge meeting the end of the screen with the side edge meeting the end of the screen. 

Here's a video of that:

https://video.kuba-orlik.name/videos/watch/ad085c55-deaf-4e31-859e-44f678512bdb

* * * *
- Tusky Version: 15.1
- Android Version: 11
- Android Device: Pixel 3a

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

cobratbq:
I would love to see this fixed. Especially crippling for very wide or very high images.

charlag:
Hi, we are aware that image viewer (still) doesn't work as expected, we might need to replace/rewrite it again

9Lukas5:
I'm gonna give this issue a bump, as I really like using Tusky and this is the number one thing bothering me regularly scrolling through the Fediverse :grimacing:

nikclayton:
Fixed in https://github.com/tuskyapp/Tusky/pull/3894

PR #3894, opened 2023-07-27
Fix image zoom / pan / scroll / swipe

nikclayton:
Migrate to touchimageview from photoview, and adjust the touch logic to correctly handle single finger drag, two finger pinch/stretch, flings, taps, and swipes.

As before, the features are:

- Single tap, show/hide controls and media description
- Double tap, zoom in/out
- Single finger drag up/down, scale/translate image, dismiss if scrolled too far
- Single finger drag left/right
  - When not zoomed, swipe to next image if multiple images present
  - When zoomed, scroll to edge of image, then to next image if multiple images present
- Two finger pinch/zoom, zoom in/out on the image

Behaviour differences to previous code

1. Bug fix: The image can't get "stuck" when zoomed, and impossible to scroll
2. Bug fix: Pinching is not mis-interpreted as a fling, closing the image
3. Bug fix: The zoom state of images is not lost or misinterpreted when the user swipes through multiple images
4. Bug fix: Double-tap zooms all the way, instead of stopping
5. Tapping outside the image does not dismiss it, controls and description show/hide

Fixes https://github.com/tuskyapp/Tusky/issues/3562, https://github.com/tuskyapp/Tusky/issues/2297


Lakoja:
Testing with my device (Android 10); observations:

Works very well, esp. the "swiping through images while zoomed".

Except two nitpicks (first one can probably be ignored):

- Closing with zoom out (far) does not work
- Sometimes (?) closing with swipe down does not work anymore once you zoomed in or out once

app/src/main/java/com/keylesspalace/tusky/fragment/ViewImageFragment.kt
```diff
@@ -108,95 +111,129 @@ class ViewImageFragment : ViewMediaFragment() {
...
                 }
-                result
-            }
-        }
 
-        var lastY = 0f
+                // The user is starting or stopping a pinch-zoom. If starting then disable touch
+                // events on the parent, so it does not attempt to scroll horizontally. If stopping
+                // then re-enable, to allow the user to swipe horizontally.
+                if (event.pointerCount >= 2 || view.canScrollHorizontally(1) && view.canScrollHorizontally(-1)) {
```
Lakoja, line 139: The two `canScrollHorizontally` are an "is zoomed and not edge"?

app/src/main/java/com/keylesspalace/tusky/fragment/ViewImageFragment.kt
```diff
@@ -108,95 +111,129 @@ class ViewImageFragment : ViewMediaFragment() {
...
+                    return false
+                }
+            }
+        )
 
-            // Clicking outside the photo closes the viewer.
-            setOnOutsidePhotoTapListener { photoActionsListener.onDismiss() }
-            setOnClickListener { onMediaTap() }
+        binding.photoView.setOnTouchCoordinatesListener(object : OnTouchCoordinatesListener {
+            var lastY: Float? = null
```
Lakoja, line 126: `oneFingerY` ?

Because it serves two functions: track movement and detect change between one and two fingers.

nikclayton:
>     * Closing with zoom out (far) does not work

That was a bug in the old code. If you zoomed out and took your fingers off the screen but didn't take them off at exactly the same time one finger would remain on the screen. If that finger then moved it could be mis-interpreted as a fling.

>     * Sometimes (?) closing with swipe down does not work anymore once you zoomed in or out once

Can't reproduce. The only thing I can think of is that the image was slightly zoomed in, and fling up/down is disabled if the image is zoomed in.

(zooming out shouldn't matter, because if you zoom out past 1.0 (i.e., the image is smaller) then the zoom resets back to 1.0 when you take your fingers off the screen)

app/src/main/java/com/keylesspalace/tusky/fragment/ViewImageFragment.kt
nikclayton: Yeah. I've updated the comment to make this clearer.

app/src/main/java/com/keylesspalace/tusky/fragment/ViewImageFragment.kt
nikclayton: Went with `lastDragY`.

app/src/main/java/com/keylesspalace/tusky/fragment/ViewImageFragment.kt
Lakoja: Would've introduced a variable `isZoomedAndNotEdge` :-)

Lakoja:
Code-wise ok.

This now requires another Android Studio?

(Cannot check further)

Lakoja (approved):

nikclayton:
> Code-wise ok.
> 
> This now requires another Android Studio?

No, works with Android Studio Flamingo.

PR #3903, opened 2023-07-30
Use lint version 8.1.0 to fix erroneous warning about forEach

nikclayton:
Android lint was erroneously warning that the forEach construct in Kotlin required API 24+, which is incorrect, see https://issuetracker.google.com/issues/185418482.

Work around that by forcing the Android lint version to 8.1.0.

This triggered some additional checks, which have been ignored, and a new baseline.


nikclayton:
This is necessary to unblock at least

- https://github.com/tuskyapp/Tusky/pull/3835
- https://github.com/tuskyapp/Tusky/pull/3894

charlag (approved):

PR #3904, opened 2023-07-31
Populate gradle build cache on each push to develop

nikclayton:
The build cache is used in ci.yml to speed up builds, but it needs to be populated by building code that's been merged.

### DIFF
--- a/.github/workflows/populate-gradle-build-cache.yml
+++ b/.github/workflows/populate-gradle-build-cache.yml
@@ -1,0 +1,28 @@
+# Build the app on each push to `develop`, populating the build cache to speed
+# up CI on PRs.
+
+name: Populate build cache
+
+on:
+  push:
+    branches:
+      - develop
+
+jobs:
+  build:
+    name: app:buildGreenDebug
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+
+      - uses: actions/setup-java@v3
+        with:
+          java-version: '17'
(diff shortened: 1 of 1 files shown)


.github/workflows/populate-gradle-build-cache.yml
```diff
@@ -0,0 +1,28 @@
+# Build the app on each push to `develop`, populating the build cache to speed
+# up CI on PRs.
+
+name: Populate build cache
+
+on:
+  push:
+    branches:
+      - develop
```
Lakoja, line 9: What's the difference to ci.yml?
Does that one only build on tag pushes?

nikclayton: ci.yml runs on every push. This one will run only when stuff is pushed to develop (which means, when a PR is merged to develop).

Lakoja: And if ci.yml runs on every push (also merges to develop) does this populate then do something more?

nikclayton: Sorry, should have been clearer. ci.yml runs when a pull request is opened (`pull_request:`), when a pull request has a commit pushed to it, or manually (`workflow_dispatch`).

It doesn't trigger on pushes to `develop`. That's what this one is for.

It's separate because there's no need to run the lint/test steps, because the PR will need to have passed those before it can be merged.

Lakoja (approved):

PR #3906, opened 2023-07-31
Rename "Trending" to "TrendingTags" or similar where necessary

nikclayton:
The "trending" functionality will expand to include trending links and posts. But at the moment the "Trending" references in the code are exclusively to hashtags.

Rename "Trending" to "TrendingTags" or similar everywhere necessary in order to prepare for this.

This includes a database migration, as the identifier for the "Trending tags" tab in the account preferences was changed from "Trending" to "TrendingTags". The migration updates the stored value if necessary.


nikclayton:
Note to reviewers:

1. The large size is because of the database migration JSON, the effective size of this PR is < 100 lines
2. Most of the changes are mechanical renaming in Android Studio
3. `AppDatabase.java` is the only file with new code that needs reviewing

app/src/main/java/com/keylesspalace/tusky/components/trending/TrendingTagsFragment.kt
```diff
@@ -249,6 +249,6 @@ class TrendingFragment :
     companion object {
         private const val TAG = "TrendingFragment"
```
nikclayton, line 250:
```suggestion
        private const val TAG = "TrendingTagsFragment"
```

app/src/main/java/com/keylesspalace/tusky/db/AppDatabase.java
```diff
@@ -674,4 +678,40 @@ public void migrate(@NonNull SupportSQLiteDatabase database) {
...
+    /**
+     * TabData.TRENDING was renamed to TabData.TRENDING_TAGS, and the text
+     * representation was changed from "Trending" to "TrendingTags".
+     */
+    public static final Migration MIGRATION_52_53 = new Migration(52, 53) {
+        @Override
+        public void migrate(@NonNull SupportSQLiteDatabase database) {
+            database.beginTransaction();
+            try {
+                Cursor cursor = database.query("SELECT id, tabPreferences FROM AccountEntity");
```
Lakoja, line 691: Could this be a simple `UPDATE AccountEntity SET tabPreferences = replace('..Trending..', '...Tags..', tabPreferences` (or similar)?

Lakoja:
Probably not the topic of this specific PR (but maybe that of the following ones):

What is the idea behind this "trending" (gui, usage, user desire)?

I would expect it to be a somewhat imprecise Explore page: Mixing posts, links (?) and tags?
If all of them are separate (tabs?) that would seem worse to me than the search: separation just because the api separates?

app/src/main/java/com/keylesspalace/tusky/db/AppDatabase.java
nikclayton: Yes. That would be much smarter way to do it. Done.

nikclayton:
> Probably not the topic of this specific PR (but maybe that of the following ones):
> 
> What is the idea behind this "trending" (gui, usage, user desire)?

Users ask for it, the web UI surfaces it, other apps show it.

> I would expect it to be a somewhat imprecise Explore page: Mixing posts, links (?) and tags? If all of them are separate (tabs?) that would seem worse to me than the search: separation just because the api separates?

Separation because these are separate things.

If they were a single list of things, grouped by type, there'd be no order that suited everybody. Should it be links, then posts, then hashtags? Or hashtags, links, then posts? Or something else?

Building a preference UI to allow users to reorder these is possible, but additional work that there's no proof that is needed right now, and will be obsoleted when the 5 tab limit goes away.

Lakoja (approved):

DavidEdwards (approved):
Looks good. Would be nice if the runner would run, but that isn't in the scope of this PR.

PR #3908, opened 2023-08-01
 Support trending links

nikclayton:
Expand the "trending" support to include trending links.

Main changes:

- `TrendingActivity` can now host multiple trending fragments in a viewpager, and swipe between them
- Implement a `TrendingLinks` suite (`Fragment`, `ViewModel`, `Adapter`, `Repository`, etc) to get and display the data
- Always show "Trending" on the left-nav menu
- Support adding trending links as a dedicated tab

PR #3910, opened 2023-08-01
Support trending statuses

### LINKED ISSUE #2109: Trending Hashtags page

nikclayton:
Expand the "trending" support to include trending statuses

Main changes:

- Add a new timeline kind, `TRENDING_STATUSES`. Use this everywhere necessary to fetch a timeline of trending statuses
- Update `TrendingActivity` to use this fragment in the view pager
- Support adding trending statuses as a dedicated tab

Fixes #2109


nikclayton:
This includes the changes in #3908 which you may want to review first (that, in turn, is stacked on top of #3906).



Those are interactable -- reply, favourite, bookmark, view thread, etc.

PR #3917, opened 2023-08-03
Use colorPrimary for the "edit" schedule icon

nikclayton:
The "edit" icon when showing a scheduled status' time was grey, so it's not obvious that this section is clickable.

Use colorPrimary, so it looks more like a button.

### DIFF
--- a/app/src/main/res/layout/view_compose_schedule.xml
+++ b/app/src/main/res/layout/view_compose_schedule.xml
@@ -23,7 +23,7 @@
         android:paddingBottom="16dp"
         android:textColor="?android:textColorTertiary"
         android:textSize="?attr/status_text_medium"
-        app:drawableTint="?android:attr/textColorTertiary"
+        app:drawableTint="?attr/colorPrimary"
         app:layout_constraintBottom_toTopOf="@id/invalidScheduleWarning"
         app:layout_constraintEnd_toEndOf="parent"
         app:layout_constraintHorizontal_bias="1"


app/src/main/res/layout/view_compose_schedule.xml
```diff
@@ -23,7 +23,7 @@
         android:paddingBottom="16dp"
         android:textColor="?android:textColorTertiary"
         android:textSize="?attr/status_text_medium"
-        app:drawableTint="?android:attr/textColorTertiary"
+        app:drawableTint="?attr/colorPrimary"
```
Lakoja, line 26: This is a textview that is clickable?

Lakoja:
What do I need to do to see this?

app/src/main/res/layout/view_compose_schedule.xml
nikclayton: Yes, clicking anywhere on the textview allows the user to reschedule the post.

Arguably all the text should be colorPrimary to indicate that the whole thing is clickable, but then there's quite a lot of blue in a small area and I found it distracting.

Lakoja:
>     3. Click the "clock" icon again to see the scheduled date/time

Haven't seen that, yet. Looks a bit dodgy. (Texts not aligned, for example)

But this is an improvement then. And coloring only the pen is probably correct also.

Lakoja (approved):

PR #3918, opened 2023-08-03
Disable Toot button when bottom sheets are open

nikclayton:
When the bottom sheets (pick media, set visibility, choose emoji, view/edit scheduled post time) are open the user is going to be clicking near the bottom of the screen.

This is very close to the "Toot" button, so there's a risk that the user inadvertently posts, instead of e.g., choosing an emoji, or trying to change the scheduled time.

Disable the button while the sheets are open so that the user can't post by mistake, re-enable when the sheets are closed.

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/components/compose/ComposeActivity.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/components/compose/ComposeActivity.kt
@@ -772,8 +772,10 @@ class ComposeActivity :
             addMediaBehavior.state = BottomSheetBehavior.STATE_HIDDEN
             emojiBehavior.state = BottomSheetBehavior.STATE_HIDDEN
             scheduleBehavior.setState(BottomSheetBehavior.STATE_HIDDEN)
+            binding.composeTootButton.isEnabled = false
         } else {
             composeOptionsBehavior.setState(BottomSheetBehavior.STATE_HIDDEN)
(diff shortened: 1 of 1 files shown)


Lakoja:
Hm, the cost of accidentally posting is rather limited, I guess. (click "delete and redraft")

On the other it would not be obvious why suddenly the posts button is disabled.

nikclayton:
> Hm, the cost of accidentally posting is rather limited, I guess. (click "delete and redraft")

Depends on what the user is posting. Suppose they've written something not intended to be public, have opened the privacy options, and mis-tap the line "Direct" (which is the bottom line, and therefore the one that is closest to the "Toot" button).

Now they've made something public that they wanted to be private.

And we're not in a position to judge, on a user-by-user basis, what the "cost" of that is to them.

> On the other it would not be obvious why suddenly the posts button is disabled.

It's very obvious, it changes immediately it response to the user tapping on one of the buttons at the bottom. It doesn't change in response to setting the CW, because the user's next action will be at the top of the screen, away from the Toot button.

nikclayton:
@Lakoja - ping

Lakoja:
> @Lakoja - ping

Was this requested by someone or a bug of some people?

I don't feel good to add code to solve a problem that isn't one.

nikclayton:
I saw the issue while I was working on related code.

PR #3920, opened 2023-08-04
Revert "Enable gradle build cache for Bitrise builds"

nikclayton:
Reverts tuskyapp/Tusky#3840. Turns out this needs to be enabled at Bitrise, and it's not on our plan. So every build is running this extra workflow, but it's not providing any benefit, just slowing things down.


Gryzor (approved):
I'm unfamiliar with this particular Bitrise config, but looks good.

Issue #3477, opened 2023-03-22
Links show too little contrast in dark mode


guerda:
I am using Tusky daily and I really like this app. What I see as an issue is the dark theme (night theme) in combination with links (hashtags, URLs) etc, as these have too little contrast to be readable in a night time setting.

I tried to figure it out and it seems that the link color is `tusky_blue` according to https://github.com/tuskyapp/Tusky/blob/8c519af611b27b4d99c7e5ee32c28f0111b7f413/app/src/main/res/values/colors.xml#L3, which translates to #2b90d9 
As far as I can see, the background in posts is #282c37 (`tusky_grey_20`). 

According to [WebAIM](https://webaim.org/resources/contrastchecker/), those two colors have a contrast of 4.04:1 and fails WCAG AAA in both normal and large text.

<img width="744" alt="Bildschirm­foto 2023-03-22 um 09 43 42" src="https://user-images.githubusercontent.com/230782/226848367-1a3b0ba0-8852-4622-a864-2bab73c3f0ba.png">

I propose to adapt the link color or the background color, to increase the contrast. For example, the link color #83BEF1 would pass all four checks with WebAIM.

* * * *
- Tusky Version: 21.0
- Android Version: 13
- Android Device: Pixel 7

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

Lakoja:
(Slightly related: #3178)

Lakoja:
(Just for reference: the two blue tints with 4.6 contrast each would be 217aba (for light) and 3c9add (for dark).)

guerda:
Any update on this? I assume it's a question of consideration, not feasibility.

visualstuart:
Chiming in, this is an accessibility issue for me. As I understand it, people with cataracts lose some of their ability to perceive blue.

joergi:
It's really hard to reading in the app at night in the bedroom.
It would be super helpful, if we can choose the color ourself

PR #3921, opened 2023-08-04
3477 tweak colors for contrast

Lakoja:
Fixes: #3477

This tweaks two colors: 
* for the dark theme a new slightly darker (timeline) background is used.
* for the light theme a darker shade of blue is used as highlight color (`colorPrimary` and `colorSecondary`)

Additional change:
* Use green and red more consistent: use existing colors for the diff display; move the new light variants to colors.xml like all colors

### DIFF
--- a/app/src/main/res/values/colors.xml
+++ b/app/src/main/res/values/colors.xml
@@ -4,9 +4,11 @@
     <color name="tusky_blue_light">#56a7e1</color>
     <color name="tusky_orange">#ca8f04</color>
     <color name="tusky_orange_light">#fab207</color>
-    <color name="tusky_green">#19a341</color>
+    <color name="tusky_green">#00731B</color>
     <color name="tusky_green_light">#25d069</color>
+    <color name="tusky_green_lighter">#CCFFD8</color>
     <color name="tusky_red">#DF1553</color>
+    <color name="tusky_red_lighter">#FFC0C0</color>
 
     <color name="white">#fff</color>
     <color name="black">#000</color>
--- a/app/src/main/res/values/colors.xml
+++ b/app/src/main/res/values/colors.xml
@@ -1,6 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <resources>
     <color name="tusky_blue">#2b90d9</color>
+    <color name="tusky_blue_darker">#217aba</color>
     <color name="tusky_blue_light">#56a7e1</color>
     <color name="tusky_orange">#ca8f04</color>
     <color name="tusky_orange_light">#fab207</color>
@@ -18,6 +19,7 @@
     <!-- the number roughly corresponds to the % lightness of the grey -->
     <color name="tusky_grey_05">#070b14</color>
     <color name="tusky_grey_10">#16191f</color>
+    <color name="tusky_grey_15">#21222c</color>
     <color name="tusky_grey_20">#282c37</color>
     <color name="tusky_grey_25">#313543</color>
     <color name="tusky_grey_30">#444b5d</color>
--- a/app/src/main/res/values/styles.xml
+++ b/app/src/main/res/values/styles.xml
@@ -91,6 +91,23 @@
         <item name="preferenceTheme">@style/TuskyPreferenceTheme</item>
     </style>
 
+    <style name="TuskyBlackThemeBase" parent="TuskyBaseTheme">
+        <item name="colorPrimaryDark">@color/tusky_grey_05</item>
+
+        <item name="android:colorBackground">@color/black</item>
+        <item name="windowBackgroundColor">@color/black</item>
+
+        <item name="colorSurface">@color/tusky_grey_10</item>
+
+        <item name="iconColor">@color/tusky_grey_40</item>
+        <item name="colorBackgroundHighlight">@color/tusky_grey_40</item>
+        <item name="colorBackgroundAccent">@color/tusky_grey_20</item>
+
+        <item name="dividerColor">@color/tusky_grey_20</item>
+    </style>
+
+    <style name="TuskyBlackTheme" parent="TuskyBlackThemeBase" />
+
     <style name="ViewMediaActivity.AppBarLayout" parent="ThemeOverlay.AppCompat">
         <item name="android:colorControlNormal">@color/white</item>
     </style>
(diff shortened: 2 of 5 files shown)


nikclayton (changes requested):
This changes more than the two colours that are in the PR description. Is that intended?

nikclayton:
I don't think the colour of the FAB has to change, based on https://webaim.org/resources/contrastchecker/.

If I put in `#ffffff` as the foreground colour (ie., the pencil icon) and `#2B90D9` as the background colour it reports that as a pass for "Graphical Objects and User Interface Components"

app/src/main/res/values/colors.xml
```diff
@@ -1,12 +1,15 @@
 <?xml version="1.0" encoding="utf-8"?>
 <resources>
     <color name="tusky_blue">#2b90d9</color>
+    <color name="tusky_blue_darker">#217aba</color>
     <color name="tusky_blue_light">#56a7e1</color>
     <color name="tusky_orange">#ca8f04</color>
     <color name="tusky_orange_light">#fab207</color>
-    <color name="tusky_green">#19a341</color>
+    <color name="tusky_green">#00731B</color>
```
Lakoja, line 8: The red and green changes (and mostly move them to named color values) concern the diff display.

How did you select those in the first place? @nikclayton 

For example I now used the existing red from colors.xml (came from #1952) which is df1553 not your clean red df0000.
But for the remainder I used your values.

the existing Green is only used for 3 spark buttons (are they used?). Red is used in a few more places.

nikclayton: The existing green/red colours didn't work on a dark background, see https://github.com/tuskyapp/Tusky/pull/3314#pullrequestreview-1320455937

app/src/main/res/values/styles.xml
```diff
@@ -91,6 +91,23 @@
         <item name="preferenceTheme">@style/TuskyPreferenceTheme</item>
     </style>
 
+    <style name="TuskyBlackThemeBase" parent="TuskyBaseTheme">
```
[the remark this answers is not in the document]
Lakoja: Sure, won't do. This is not unnecessarily, but related.

Lakoja:
> I don't think the colour of the FAB has to change, based on https://webaim.org/resources/contrastchecker/.

I guess that is what this PR is also partly about: Be consistent in color usage.

So if the FAB and text highlights are presented in `colorSecondary` they can and should keep it that way when that is changed.

Same goes for the green: If there is one, use it. (And change it here to improve the contrast.)

Lakoja:
> This changes more than the two colours that are in the PR description. Is that intended?

Add a part for red and green to the description.

app/src/main/res/values/styles.xml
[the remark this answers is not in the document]
nikclayton: This looks identical to the block that was deleted, lines 147-164.

nikclayton:
I think this is more complex than it needs to be for what it is trying to accomplish.

The original issue (#3477) is just about the link colour. This PR changes more than that, **and** the description is incorrect w.r.t what it currently changes.

With this PR:

**Light mode**

- Links are darker (#2b90d9 -> #217aba)
- "Widget blue" (FAB, tab icons, slider thumb, etc) is darker (#2b90d9 -> #217aba) (undocumented change)
- Diff green is the same (#ccffd8)
- Diff red is the same (#ffc0c0)

**Dark mode**

- No difference to background (#282c37), contrary to PR description
- FAB is different colour (#2b8fd7 -> #2b90d9) (undocumented change)
- Tab icons are the same colour (#2b90d9)
- Slider thumb is the same colour (#2b90d9)
- Diff green is the same (#00731b)
- Diff "red" is now purple (#df0000 -> #df1553)

The suggested new link colours from https://github.com/tuskyapp/Tusky/issues/3477#issuecomment-1489828837 are:

- #217aba -- link colour in light mode
- #3c9add -- link colour in dark mode

That can be achieved by:

1. Edit `values/theme_colors`, add an entry `<color name="colorAccent">#217aba</color>`
2. Edit `values-night/theme_colors`, add an entry `<color name="colorAccent">#3c9add</color>`
3. Edit `values/styles.xml`, add an entry `<item name="colorAccent">@color/colorAccent</item>` in the `TuskyBaseTheme` style

- In both modes that will adjust the link colour as intended.
- In dark mode the colours of the FAB, tab icons, and widgets (e.g., the slider thumb on the font size preference) are unchanged, also as intended
- In light mode the colours of the tab icons and widgets are unaffected. The colour of the FAB goes fractionally darker, #2b90d9 -> #2b8fd8

app/src/main/res/values/colors.xml
nikclayton: Just saw I missed your query about spark buttons.

They are used -- they're the buttons that provide the animations when you reply, boost, or bookmark a status.

Lakoja:
> I think this is more complex than it needs to be for what it is trying to accomplish.

I guess it's not quite as complicated as listed:

* It should unify color handling a bit - there is for example no need for very special reds and greens.
* It improves contrast between color secondary and background where needed - and that leads to some consequences (e. g. fab color).

That is sensible and not too big a change for a PR.

nikclayton:
OK, please split this in two:

1 PR -- just the accessibility changes for changing the link colours. Those are non contentious, solve a problem that users have identified and filled an issue for, and are straightforward to approve. I think the `colorAccent` changes I describe above are the smallest, simplest, way to do this, but if you have an approach that is even simpler then I'm happy to look at that too. Please do *not* change the background colour, that's a much broader change than something that just affects the link colours.

Discussion -- if you want to propose additional colour changes, or how the colours / styles / themes are currently organised (and I do see options for how they can be reorganised / updated) please do that in either a new issue or a discussion in the Matrix channels.

Thanks.

Lakoja:
> The tusky_blue is still used in some places with the light theme, e.g. The "Lock account" checkbox when editing a profile, in the elephant friend drawables, and in the spinners of SwipeToRefreshLayouts.

As far as I remember from a year ago I (only) wanted to fix the problem at hand: low contrast for text and or links.

But I guess we could do the mentioned problems in some subsequent merges.
I would feel confident to unify the SwipeToRefreshLayouts to a base layout with the proper respective color - but not in this PR.

For the others I might need to research a bit more.

I wonder however if I should include the red character warning number here - as it is really eye-watering with the dark theme...

grafik

connyduck:
Ok I'm gonna merge this for now. Please send more PRs, and yes that red text is very important.

Issue #3916, opened 2023-08-03
Scheduled posts have timezone scheduling problems


nikclayton:
Users report that opening the post scheduler late in the day can disallow them from scheduling the post that same day.

E.g., compose post at 8pm, open the scheduler, it's not possible to schedule for any later time that day, only the following day.

Happens for reporters in non-UTC timezones (e.g., UTC-4), so I suspect TZ shenanigans in the scheduling code.

* * * *
- Tusky Version: 23

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

PR #3923, opened 2023-08-04
Correct the calculations for choosing the earliest day to show in the calendar / selected day

nikclayton:
To determine the earliest day to show in the calendar, take the current date/time, add the minimum scheduled seconds buffer (which may roll the date/time over to the next day), and then clamp to the start of that day. So it's either today (if the current time + minimum scheduled seconds is less than midnight) or it's tomorrow.

When displaying the calendar work around a misfeature in Material Date Picker. It accepts UTC seconds-since-epoch, but does not convert it to the local time for display.

While I'm here, show the selected day in the time picker's title.

Fixes https://github.com/tuskyapp/Tusky/issues/3916


app/src/main/java/com/keylesspalace/tusky/components/compose/view/ComposeScheduleView.kt
```diff
@@ -98,21 +100,37 @@ class ComposeScheduleView
...
+        // millis-from-epoch, in UTC, which is good. However, it is also *displayed* in UTC
+        // instead of converting to the user's local timezone.
+        //
+        // So we have to add the TZ offset before setting it in the picker
+        val tzOffset = TimeZone.getDefault().getOffset(scheduleDateTimeUtc!!.timeInMillis)
+
         val picker = MaterialDatePicker.Builder
             .datePicker()
-            .setSelection(scheduleDateTime!!.timeInMillis)
+            .setSelection(scheduleDateTimeUtc!!.timeInMillis + tzOffset)
```
Lakoja, line 133: Does it need to be subtracted then somewhere? (when taking out?)

nikclayton: That conversion was already done, and happens in `onDateSet()`.

Lakoja: So, it always comes out as UTC but it must be set (selection?) as UTC + time zone?

nikclayton: Yes. Per the comment immediately before, it accepts the time in UTC.

What it should do is then convert that to local time before displaying it.

It doesn't. So if the UTC time is 10pm on Aug 7th, and you're in UTC+4 (so it's 2am on Aug 8th in your local timezone) the calendar showed the day as the 7th, not the 8th.

PR #3924, opened 2023-08-04
Provide specific gradle properties for builds on CI

nikclayton:
Expectation is that these will result in faster builds by disabling options (like the gradle daemon) that benefit long-lived processes but do not benefit ephemeral build workers.


nikclayton:
Note to reviewer -- I think this will be faster overall, but there's no easy way of checking except for submitting it, letting CI build with it for a week or so, and then seeing what the results are.

PR #3925, opened 2023-08-04
Add `newLintBaseline` task

nikclayton:
Add a task to regenerate the lint baseline.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -60,8 +60,7 @@ android {
 
     lint {
         lintConfig file("lint.xml")
-        // Regenerate by deleting app/lint-baseline.xml, then run:
-        //   ./gradlew lintBlueDebug
+        // Regenerate by running `./gradlew app:newLintBaseline`
         baseline = file("lint-baseline.xml")
     }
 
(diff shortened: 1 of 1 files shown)


app/build.gradle
```diff
@@ -203,3 +202,18 @@ tasks.withType(org.jetbrains.kotlin.gradle.internal.KaptWithoutKotlincTask) {
         "--add-opens", "jdk.compiler/com.sun.tools.javac.parser=ALL-UNNAMED",
         "--add-opens", "jdk.compiler/com.sun.tools.javac.code=ALL-UNNAMED"])
 }
+
+tasks.register("newLintBaseline") {
```
Lakoja, line 206: What is "trigger" to recreate that?

I never  had to do it but for the last PR I did?

Lakoja (approved):

app/build.gradle
nikclayton: The lint baseline records lint errors that we're currently ignoring. It does that by filename and line number.

If you move code around, or add new code to files that have lint errors then it can mean that the information in the baseline is obsolete.

E.g., in the errorphant PR the lint baseline had entries for the files under the old names (because the lint was complaining about the complex vector path). When you renamed the files the entries in the baseline file became obsolete, and lint started complaining about the new files.

This is a convenience task so that you can (a) run it from the command line (so it's one commend instead of two), and (b) add it as a task in Android studio so it's a single click, like so:

image

Lakoja (approved):

PR #3926, opened 2023-08-05
Make CW description selectable

### LINKED ISSUE #3826: Text in the content warning isn't selectable in the detailed view
Steps to reproduce:
- Open the detailed view for a timeline status

Expected result: Content and content warning are selectable

Actual result: Content is selectable, but content warning is not 

* * * *
- Tusky Version: nightlyish
- Android Version: 11
- Android Device: Pixel 2xl

- ✅ I searched or browsed the repo’s other issues to ensure this is not a duplicate.

tinsukE:
While the status text in the detailed layout was selectable, CW description wasn't.

Also made both CW and text selectable when viewing edits.

Fixed layouts:
- `item_status_detailed.xml`: from the bug report
- ~~`item_status_edit.xml`: I deemed it worth it to fix it there too~~

Reviewed but not made selectable:
- `item_status.xml`: as tapping on the view opens the detailed view and adding a long press to select gesture could make UX confusing
- `item_status_notification.xml`: as tapping on the notification launches the app

Fixes #3826.

### DIFF
--- a/app/src/main/res/layout/item_status_detailed.xml
+++ b/app/src/main/res/layout/item_status_detailed.xml
@@ -89,6 +89,7 @@
         android:importantForAccessibility="no"
         android:lineSpacingMultiplier="1.1"
         android:textColor="?android:textColorPrimary"
+        android:textIsSelectable="true"
         android:textSize="?attr/status_text_large"
         app:layout_constraintEnd_toEndOf="parent"
         app:layout_constraintStart_toStartOf="parent"


nikclayton (commented):
+1 to the CW changes in item_status_detailed.

Not sure about the changes to item_status_edit. Since that will include the diffs it's of limited utility.

Does the formatting come along with the copied diff? So if the user pastes the copied diff in to another app they can still clearly see what's changed?

tinsukE:
I think that for formatting to be copied along, we'd have to handle that ourselves and build a [ClipData.newHtmlTex()](https://developer.android.com/reference/kotlin/android/content/ClipData?hl=en#newhtmltext), which might be a bit much for this fix. But I think that being to copy what was removed/edited from a toot could be of use to users.

Happy to remove it if you think it's unnecessary/incorrect.

tinsukE:
I went ahead and reverted the undesired changes.

nikclayton (approved):

PR #3930, opened 2023-08-06
Fix extra clicks on media tab

nikclayton:
If:

1. You're viewing an account's media tab
2. Some of the media was marked sensitivei
3. The `alwaysShowSensitiveMedia` setting was `true`

tapping on the image (once) would do nothing visible, because it was treated as the "reveal sensitive media" tap. You had to tap on it a second time to open it.

Fix this, by passing the preference value through to the relevant code.


nikclayton:
This is https://github.com/tuskyapp/Tusky/pull/3684 (already approved) with the additional commits from me to update the PR to use the preference in additional places, and avoid an active account race condition. See https://github.com/tuskyapp/Tusky/pull/3684#issuecomment-1658243682

Lakoja (approved):

PR #3936, opened 2023-08-07
Detect Bookwyrm URL formats

bremensaki:
While helping test an issue with [Bookwyrm](https://github.com/bookwyrm-social/bookwyrm) I noticed that the URL formats used by that project aren't checked as possible profile or post links. They're quite close to a couple of others, so I just copied close examples and edited a couple of terms.

It's pretty minor, I just used a previous commit as a reference. Let me know if it needs anything more though. I've only quickly tested it on a local build with a couple of links against a live Bookwyrm and it picks them up as expected now.


nikclayton:
Looks like this is clashing with pleroma formats, given the test failure.

Tak:
The failure is for a *negative* test case  that this change obsoletes, the case should just be removed

bremensaki:
Yep, it looks like a test for a format that's invalid for Pleroma, but happens to be OK for another platform. I've updated to remove that test now.

Tak (approved):

PR #3937, opened 2023-08-07
Account: hide option to mute own domain (it is a no-op).

### LINKED ISSUE #3798: Own domain should not be muteable
Currently Tusky allows muting ones own domain. 
E.g. when I'm logged into mastodon.social, I can go to the profile of another account on mastodon.social and click "Mute mastodon.social". The Api accepts the mute just fine, but it is not doing anything, posts from mastodon.social are still being shown everywhere. 
This should be fixed by hiding the option from the drop-down menu when the account is on the same instance as the Tusky user.

tinsukE:
Fixes #3798.

I was not sure if I should put the domain comparison in the `Activity` (since it seems to keep some state and there is code there already accessing `AccountManager`) or in `ViewModel` (to me, this is the "right place" for it to be, being more encapsulated and easier to test, were this class already tested).

How do you feel about converting the cascading `if/else` to a `when` statement?

Also, should I do a case insensitive comparison, as in https://github.com/tuskyapp/Tusky/pull/3931? I opted not to, as the problematic case there was about casing for the username/handle, not the domain portion.

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/components/account/AccountViewModel.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/components/account/AccountViewModel.kt
@@ -19,6 +19,7 @@ import com.keylesspalace.tusky.util.Error
 import com.keylesspalace.tusky.util.Loading
 import com.keylesspalace.tusky.util.Resource
 import com.keylesspalace.tusky.util.Success
+import com.keylesspalace.tusky.util.getDomain
 import kotlinx.coroutines.Job
 import kotlinx.coroutines.delay
 import kotlinx.coroutines.launch
@@ -40,6 +41,7 @@ class AccountViewModel @Inject constructor(
 
     lateinit var accountId: String
     var isSelf = false
+    var isFromOwnDomain = false
 
     private var noteUpdateJob: Job? = null
 
@@ -65,6 +67,8 @@ class AccountViewModel @Inject constructor(
                             accountData.postValue(Success(account))
                             isDataLoading = false
                             isRefreshing.postValue(false)
+
+                            isFromOwnDomain = getDomain(account.url) == accountManager.activeAccount?.domain
                         },
                         { t ->
                             Log.w(TAG, "failed obtaining account", t)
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/keylesspalace/tusky/components/account/AccountViewModel.kt
```diff
@@ -40,6 +41,7 @@ class AccountViewModel @Inject constructor(
 
     lateinit var accountId: String
     var isSelf = false
+    var isFromOwnDomain = false
```
nikclayton, line 44:
```suggestion
    /** True if the viewed account has the same domain as the active account */
    var isFromOwnDomain = false
```

app/src/main/java/com/keylesspalace/tusky/components/account/AccountViewModel.kt
```diff
@@ -65,6 +67,8 @@ class AccountViewModel @Inject constructor(
                             accountData.postValue(Success(account))
                             isDataLoading = false
                             isRefreshing.postValue(false)
+
+                            isFromOwnDomain = getDomain(account.url) == accountManager.activeAccount?.domain
```
nikclayton, line 71: There are race conditions which mean that `accountManager.activeAccount` can change over the life of the activity (or fragment).

Technically it's not a problem here, but I'm trying to remove all occurrences of this pattern in the code.

You can put

```kotlin
val activeAccount = accountManager.activeAccount!!
```

at the start of the class, and then use just `activeAccount.domain` here.

nikclayton (commented):
No problem with changing the `if` to `when`.

app/src/main/java/com/keylesspalace/tusky/components/account/AccountViewModel.kt
tinsukE: I've addressed this and made sure to use the field on `setAccountInfo()` too.

nikclayton (approved):

Issue #3938, opened 2023-08-07
Notifications aren't being filtered in NotificationsFragment / ViewModel / ...


nikclayton:
1. Create a filter
2. Set the filter context to include notifications
3. Receive a notification about a status that matches the filter
4. The notification should be filtered

It's not.

This will affect Android notifications as well.

* * * *
- Tusky Version: 23.0

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

PR #3940, opened 2023-08-07
Filter notifications when appropriate

nikclayton:
Filters that have a "notifications" context should hide/warn in the notifications timeline.

This PR does that -- notifications with a status that matches a "hide" filter are completely hidden, notifications with a status that matches a "warn" filter are hidden behind a disclosure link, the same way that statuses on other timelines are.

StatusViewData / ViewDataUtils:
- Set filterAction via the constructor (with a default) to make it easier to override

NotificationsViewModel:
- Fetch the user's filters
- Determine the filter action for each notification, and include it in the status view data

NotificationsPagingAdapter:
- Add a new STATUS_FILTERED view type, use it for notifications where the status should be filtered

NotificationsFragment:
- Handle the clearWarningAction() when the user reveals a filtered status

Other:
- Move FILTER_PREF_KEYS in to FiltersViewModel, to make it more accessible
- Update test base

Fixes https://github.com/tuskyapp/Tusky/issues/3938


nikclayton:
This PR is stacked on top of https://github.com/tuskyapp/Tusky/pull/3436

PR #3959, opened 2023-08-16
Kotlin - Enable trailing comma

tinsukE:
Trailing commas on Kotlin sources [has many advantages](https://pinterest.github.io/ktlint/0.49.0/rules/standard/#trailing-comma-on-call-site):
- It makes version-control diffs cleaner – as all the focus is on the changed value.
- It makes it easy to add and reorder elements – there is no need to add or delete the comma if you manipulate elements.
- It simplifies code generation, for example, for object initializers. The last element can also have a comma.

This PR doesn't go as far as require it, but tweaks KtLint to at least allow it.

The two `.kt` files prove that the KtLint rules have been properly disabled.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -7,11 +7,13 @@ indent_style = space
 insert_final_newline = true
 trim_trailing_whitespace = true
 
+[*.{java,kt}]
 # Disable wildcard imports
-[*.{java, kt}]
 ij_kotlin_name_count_to_use_star_import = 999
 ij_kotlin_name_count_to_use_star_import_for_members = 999
 ij_java_class_count_to_use_import_on_demand = 999
+# Enable trailing comma
+ktlint_disabled_rules=trailing-comma-on-call-site,trailing-comma-on-declaration-site
 
 [*.{yml,yaml}]
 indent_size = 2
--- a/app/src/main/java/com/keylesspalace/tusky/components/report/Screen.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/components/report/Screen.kt
@@ -20,5 +20,5 @@ enum class Screen {
     Note,
     Done,
     Back,
-    Finish
+    Finish,
 }
--- a/app/src/main/java/com/keylesspalace/tusky/components/scheduled/ScheduledStatusActivity.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/components/scheduled/ScheduledStatusActivity.kt
@@ -163,8 +163,8 @@ class ScheduledStatusActivity :
                 visibility = item.params.visibility,
                 scheduledAt = item.scheduledAt,
                 sensitive = item.params.sensitive,
-                kind = ComposeActivity.ComposeKind.EDIT_SCHEDULED
-            )
+                kind = ComposeActivity.ComposeKind.EDIT_SCHEDULED,
+            ),
         )
         startActivity(intent)
     }


.editorconfig
```diff
@@ -7,11 +7,13 @@ indent_style = space
 insert_final_newline = true
 trim_trailing_whitespace = true
 
+[*.{java,kt}]
 # Disable wildcard imports
-[*.{java, kt}]
```
tinsukE, line 11: Needed to remove the whitespace between `java` and `kt`: https://pinterest.github.io/ktlint/0.50.0/rules/configuration-ktlint/

> Unfortunately [IntelliJ IDEA](https://www.jetbrains.com/idea/) has an [autoformat issue regarding .editorconfig](https://youtrack.jetbrains.com/issue/IDEA-242506). Due to this error an additional space is added between glob statements, resulting in [*{kt, kts}] instead of [*{kt,kts}]. The .editorconfig library used by ktlint [ignores sections after encountering a space in the list](https://github.com/editorconfig/editorconfig/issues/148). As a result, the rule is not applied on all files as documented in the [original ktlint issue](https://github.com/pinterest/ktlint/issues/762).

nikclayton: Good catch, thanks.

nikclayton (commented):
If I've understood this correctly (I've tried it locally to confirm) this will allow a mix of trailing and non-trailing commas in future code, with no clean way to migrate naturally as the relevant files get modified over time.

Can this be done so that existing non-trailing-comma usage is grandfathered in, but new usage is forbidden?

E.g., in `.editorconfig` have:

```
ktlint_code_style = intellij_idea
ij_kotlin_allow_trailing_comma = true
ij_kotlin_allow_trailing_comma_on_call_site = true
```

and then create a baseline that includes all the lint "violations" that this will introduce (per the discussion of the baseline file, in https://pinterest.github.io/ktlint/0.50.0/install/cli/#violation-reporting).

Over time violations would shrink, and the baseline file will get smaller (every time it's regenerated) until there are none (or only a handful) to clean up.

For this to work ktlintFormat would need to ignore violations in the baseline file too. I haven't tested that.

.editorconfig
```diff
@@ -7,11 +7,13 @@ indent_style = space
...
 ij_kotlin_name_count_to_use_star_import = 999
 ij_kotlin_name_count_to_use_star_import_for_members = 999
 ij_java_class_count_to_use_import_on_demand = 999
+# Enable trailing comma
+ktlint_disabled_rules=trailing-comma-on-call-site,trailing-comma-on-declaration-site
```
Tak, line 16: aaaaa I didn't try this method because [the docs](https://pinterest.github.io/ktlint/0.49.0/rules/configuration-ktlint/#disabled-rules) said that `ktlint_disabled_rules` is deprecated and about to be removed … but `ktlint_standard_trailing-comma-on-call-site = disabled` etc. didn't work in my attempts.
Maybe this will change by the time `ktlint_disabled_rules` is *actually* removed 🤷

Tak (approved):

Tak:
IMO it should be situation dependent - neither mandated nor forbidden.
For example, trailing commas are pretty awkward in single-line usage.

nikclayton:
> IMO it should be situation dependent - neither mandated nor forbidden. For example, trailing commas are pretty awkward in single-line usage.

These changes don't affect single-line usage, the rule / format change is only triggered on multi-line call sites and declarations.

```
// No trailing comma required, formatter does not add one
data class Foo(val x: String, val y: String)
```

```
// Trailing comma would be required, lint would fail, formatter adds one
data class Foo(
    val x: String,
    val y: String  // <-- currently allowed, would require trailing comma
)

File: .editorconfig (hunk 2 above)
Comment (tinsukE) [reply .editorconfig:16]: Yup, I was stuck when following the docs, but digging around the internet showed me this alternative.

**Comment by tinsukE** [comment]
@Tak since this is approved, shall we merge it?

Or would you prefer I look into merging/improving this other one? https://github.com/tuskyapp/Tusky/pull/3968

**Comment by Tak** [comment]
Yes, let's merge this one, thank you.

File: .editorconfig (hunk 2 above)
Comment (Goooler) [reply .editorconfig:16]: If you want to disable them, should use
```editorconfig
ij_kotlin_allow_trailing_comma = false
ij_kotlin_allow_trailing_comma_on_call_site = false
```

Seems this PR doesn't fully enable trailing commas.

PR #3968, opened 2023-08-18
Kotlin - Require trailing comma

tinsukE:
New Kotlin code will require trailing commas, but old violations are grandfathered via ktlint baseline files (`./gradlew ktlintCheck` updates the file).

This is a heavily opinionated change, it requires new and modified Kotlin code to use trailing comma. But it [has many advantages](https://pinterest.github.io/ktlint/0.49.0/rules/standard/#trailing-comma-on-call-site):
- It makes version-control diffs cleaner – as all the focus is on the changed value.
- It makes it easy to add and reorder elements – there is no need to add or delete the comma if you manipulate elements.
- It simplifies code generation, for example, for object initializers. The last element can also have a comma.

This PR is another way to tackle the same problem as https://github.com/tuskyapp/Tusky/pull/3959.

If this gets merged, already existing PRs will need to be updated:
- Rebase/merge from master
- One of:
  - Update baseline: `./gradlew ktlintCheck`; or
  - Comma all the trails

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -7,11 +7,15 @@ indent_style = space
 insert_final_newline = true
 trim_trailing_whitespace = true
 
+[*.{java,kt}]
+ktlint_code_style=intellij_idea
 # Disable wildcard imports
-[*.{java, kt}]
 ij_kotlin_name_count_to_use_star_import = 999
 ij_kotlin_name_count_to_use_star_import_for_members = 999
 ij_java_class_count_to_use_import_on_demand = 999
+# Require trailing comma
+ij_kotlin_allow_trailing_comma = true
+ij_kotlin_allow_trailing_comma_on_call_site = true
 
 [*.{yml,yaml}]
 indent_size = 2
--- a/app/config/ktlint/baseline.xml
+++ b/app/config/ktlint/baseline.xml
@@ -1,0 +1,2152 @@
+<?xml version="1.0" encoding="utf-8"?>
+<baseline version="1.0">
+    <file name="src/androidTest/java/com/keylesspalace/tusky/MigrationsTest.kt">
+        <error line="23" column="43" source="trailing-comma-on-call-site" />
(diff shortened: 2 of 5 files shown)


nikclayton:
I've tried this, and I can't get it to behave the way I'd like in Android Studio.

Ideally:

1. Creating a new line that needs a trailing comma adds it if it's not present.

   For example, if I change:

   ```kotlin
   enum class Screen {
       // ... omitted
       Back,
       Finish
   }
   ```

   to

   ```kotlin
   enum class Screen {
       // ... omitted
       Back,
       Finish,
       SomeOtherValue  // <-- no trailing comma added by me
   }
   ```

   The comma would be added on save. At the moment AS is not adding the comma.

2. Lines that are not modified do not get a trailing comma.

   I change:

   ```kotlin
   enum class Screen {
       // ... omitted
       Back,
       Finish
   }
   ```

   to

   ```kotlin
   enum class Screen {
       // ... omitted
       Back,
       SomeOtherValue,
       Finish    // <-- saving should not add a trailing comma here
   }
   ```

   This behaves as I want at the moment, and any solution should not break that.

When I was doing these tests I had Android Studio configured with "Tools > Actions on Save" set to reformat code on all file types, changed lines only.

My reasons for wanting this are:

- Android Studio should "do the expected thing" for contributors. If a contributor sends a PR that contains new code, and AS hasn't added the trailing commas then the contributor is going to see a bunch of lint issues on their PR. Hence behaviour 1.
- I would prefer that the `git blame` output still be useful. I use it probably once a week at least if I'm investigating why a particular piece of code is doing what it's doing (so I can trace back to the PR that implemented it and maybe the issue that kicked it off). Hence behaviour 2.

I also tried this with https://github.com/nbadal/ktlint-intellij-plugin installed, but that was (a) flagging errors/warnings on every site that was missing a trailing comma, and (b) adding them on save, irrespective of whether or not the line had been changed.

Tak:
Superseded by #3959

Issue #3486, opened 2023-03-24
Editing profile: No change warning on leave


Lakoja:
If you edit your profile, make some changes and then press back, then these changes are lost.

Expected: There should be a dialog with "Do you want to save?".

- [x] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

Gryzor:
Fixed.

PR #3972, opened 2023-08-19
 Editing profile: No change warning on leave.

Gryzor:
### Objective
* Prevent data loss when the user inadvertently hits back or wants to leave the profile edition with unsaved changes.

### Description
* To limit the number of changes to the existing codebase, I merely re-used the same method used by `save()` in the ViewModel to decide whether to make a network request or simply return the profile as-is. 
* ~A bit of code juggling around in the ViewModel and I was able to use the logic for all the encoding of each profile field (Which is what the ViewModel caches in memory).~ Thanks @Lakoja  for improving this in the VM. 
* A couple of internal data classes used as helpers to move all the fields around (now that they are no longer used in one single place) were introduced. 

### Potential Optimizations
* ~The profile encoding is done twice (once for checking, and then again if the user has to actually save it). I'd say this is a negligible price to pay, since the alternative would be to create a different set of comparisons and/or keeping another profile in memory for the purpose of comparison.~

### Visual Improvement
* I believe the Dialog is difficult to see, but it's being displayed with Tusky's theme. Perhaps there's a better style to apply in this case? (or maybe the edit profile activity shouldn't have the same background color as dialogs?!)

### Issue
* #3486


app/src/main/java/com/keylesspalace/tusky/viewmodel/EditProfileViewModel.kt
```diff
@@ -96,29 +106,73 @@ class EditProfileViewModel @Inject constructor(
...
-    fun save(newDisplayName: String, newNote: String, newLocked: Boolean, newFields: List<StringField>) {
+    internal fun save(newProfileData: ProfileData) {
         if (saveData.value is Loading || profileData.value !is Success) {
             return
         }
 
         saveData.value = Loading()
 
-        val displayName = if (oldProfileData?.displayName == newDisplayName) {
+        val encoded = encodeChangedProfileFields(newProfileData)
```
Lakoja, line 116: Could this be done directly with one function `hasChangedFields(newProfileData)` - instead of `encode + allFieldsAreNull``?

(As it is not really encoded but only the diff is determined?)

Gryzor: Not as-is, since the result of this "encoding" is used below; and the result of "allFieldsAreNull" is used separately as well... 

With some refactoring this could be reused, but it wasn't my intention to alter the whole EditProfile logic, as it's quite complex as it is. 

Do you have a concrete idea in mind?

Lakoja: My concrete idea (haven't tried it, will do if this comment "does not work" :-)):

- Have a data class "ProfileData" with all fields relevant for a profile.
- Use that as data type / parameter.
- For "is it changed?": the view model should compare against this data model (with the new data)
- Can also be used in the api call (or its fields)

Gryzor: Yes, that's actually what I had initially in mind, you could even compare the hash of the "old" vs. "new". 
Perhaps that could be done as either part of this PR or a new PR, because it will involve touching this VM for sure. 

I have created the ProfileData, in this PR. What initially caught me "off guard" was that the VM is "encoding" this and keeping *that* in memory, that's why I had to extract these encoding fields to compare apples to apples. 

I'd say all this encoding and what not should not be a VM problem, rather the Repo/Use case that deals with the persistence/storage/API/etc.

What do you think?

Lakoja: I see:

The api expects a) only changed data AND b) in an encoded form.

So what I now did is separate the diff-making from the encoding. Mixing them leads to confusion (e. g. an oddly named data model).
Also I renamed some stuff to honor that.

See if you like it: https://github.com/Gryzor/Tusky/pull/1

app/src/main/java/com/keylesspalace/tusky/viewmodel/EditProfileViewModel.kt
```diff
@@ -96,68 +105,49 @@ class EditProfileViewModel @Inject constructor(
...
-        val field4 = calculateFieldToUpdate(newFields.getOrNull(3), fieldsUnchanged)
-
-        if (displayName == null && note == null && locked == null && avatar == null && header == null &&
-            field1 == null && field2 == null && field3 == null && field4 == null
-        ) {
-            /** if nothing has changed, there is no need to make a network request */
+        val diff = getProfileDiff(apiProfileAccount, newProfileData)
+        if (!diff.hasChanges()) {
+            // if nothing has changed, there is no need to make an api call
             saveData.postValue(Success())
```
Lakoja, line 118: Is postValue() the same as value = ?

(Should probably be consistent.)

app/src/main/java/com/keylesspalace/tusky/EditProfileActivity.kt
```diff
@@ -306,4 +320,17 @@ class EditProfileActivity : BaseActivity(), Injectable {
...
     }
+
+    private fun showUnsavedChangesDialog() = lifecycleScope.launch {
+        when (launchAlertDialog()) {
+            AlertDialog.BUTTON_POSITIVE -> save()
+            else -> finish()
+        }
+    }
+
+    private suspend fun launchAlertDialog() = AlertDialog.Builder(this)
```
Lakoja, line 331: The `launchAlertDialog` is technical but does not say what it does: `showSaveDialog`?

(Which is the same name as the wrapping function...)

app/src/main/java/com/keylesspalace/tusky/viewmodel/EditProfileViewModel.kt
```diff
@@ -51,6 +50,16 @@ import javax.inject.Inject
 private const val HEADER_FILE_NAME = "header.png"
 private const val AVATAR_FILE_NAME = "avatar.png"
 
+/**
+ * Conveniently groups Profile Data users can modify in the UI.
```
Lakoja, line 54: `ProfileDataInUi` ?

Or simply ProfileData - because the other data class (the diff) has a more special name and can be differentiated.

Then the comment can be removed.

(Code should always speak for itself, if it can.)

app/src/main/java/com/keylesspalace/tusky/EditProfileActivity.kt
```diff
@@ -306,4 +320,17 @@ class EditProfileActivity : BaseActivity(), Injectable {
...
+        .setTitle(getString(R.string.title_edit_profile_save_changes_prompt))
```
Lakoja, line 332: These dialogs don't need a title.

app/src/main/res/values/strings.xml
```diff
@@ -819,4 +819,6 @@
     <string name="error_media_playback">Playback failed: %s</string>
     <string name="dialog_delete_filter_text">Delete filter \'%1$s\'?"</string>
     <string name="dialog_delete_filter_positive_action">Delete</string>
+    <string name="title_edit_profile_save_changes_prompt">Unsaved Changes</string>
+    <string name="message_edit_profile_save_changes_prompt">Do you want to save your profile changes?</string>
```
Lakoja, line 823: `dialog_save_profile_changes_message` ?

Is a bit shorter and conforms to some other string keys.

app/src/main/java/com/keylesspalace/tusky/EditProfileActivity.kt
```diff
@@ -306,4 +320,17 @@ class EditProfileActivity : BaseActivity(), Injectable {
         Log.w("EditProfileActivity", "failed to pick media", throwable)
         Snackbar.make(binding.avatarButton, R.string.error_media_upload_sending, Snackbar.LENGTH_LONG).show()
     }
+
+    private fun showUnsavedChangesDialog() = lifecycleScope.launch {
```
Lakoja, line 324: About these wrappings and architectural problem with the save dialog:

I see in https://github.com/tuskyapp/Tusky/pull/3965 that the AlertDialog.await() is new.

What was the idea behind that?

Other (older) locations do this rather straight forward with a `dialog.show()` and direct specification of actions on the button functions.
(See for example ComposeActivity#handleCloseButton)

app/src/main/java/com/keylesspalace/tusky/viewmodel/EditProfileViewModel.kt
Gryzor: They aren't. Post Value does so by posting the task to a Main Thread if I correctly recall. 

So if you do 

`x.postValue(1) 
x.postValue(2)
x.value = 3
`

The value will be immediately set to 3, then to 2 (as it would overwrite the 1) when the "Runnable" is dispatched and run by the main Thread message queue. (it really uses a Runnable behind the scenes if I am not mistaken).

In my experience it's often safe(r) to use postValue but I guess it depends on the use Case. I haven't used LiveData in a bit (having switched most of the time to StateFlow/SharedFLow).

app/src/main/java/com/keylesspalace/tusky/EditProfileActivity.kt
Gryzor: I'll change it to launchSaveDialog() as it's a suspend function that makes the current "showUnsavedChangesDialog()" more readable without cramping the dialog logic there as well.

app/src/main/java/com/keylesspalace/tusky/EditProfileActivity.kt
Gryzor: I'll remove it, I still think they do need a title.

app/src/main/java/com/keylesspalace/tusky/EditProfileActivity.kt
Gryzor: This was a new thing that @nikclayton asked me to introduce.

In short, it moves from a callback-based flow, to a more linear control flow, as you now "await" in a suspend function.

Is it "better"? Debatable.
Does it read "better" I'd say yes (but this is also debatable).

An Improvement I'd do (but we'd have to discuss it) is to make it more an extension of a LifecycleScope so you can do from an Activity or Fragment (for e.g.):

Pseudo-Code:
```
fun x() = launchAlertDialog(title, message, positive, negative, neutral, etc...) { result -> if (result == button positive) { ... } }
```

The reason is, it feels a bit strange to _await_ and pass the buttons in that call. At first it makes kinda sense, you construct the alert, then await on the buttons... so you pass them there. But it could be further abstracted I'd say.

Anyway, if you don't like this, your best bet is talk to Nik ;)

app/src/main/java/com/keylesspalace/tusky/viewmodel/EditProfileViewModel.kt
Lakoja: According to the documentation (of those methods): It's only a question of where you call it as `setValue` needs to be done on the main thread.

The async character of postValue (overwriting stuff) is then probably only a secondary problem.

So this can be resolved.

app/src/main/java/com/keylesspalace/tusky/EditProfileActivity.kt
Lakoja: For me this approach is much more indirect (and thus harder to read). With the old one you'll all have in one place (the call to AlertDialog.Builder).

But I guess I will discuss this if I myself have a problem (code change) like this. :-)

(Can be resolved.)

app/src/main/java/com/keylesspalace/tusky/EditProfileActivity.kt
Gryzor: is that OK?

Lakoja: Yes.

app/src/main/java/com/keylesspalace/tusky/EditProfileActivity.kt
Tak: My only question about this PR: do we still feel like this launch/await flow is better, or do we want to revert to the old style (builder)?

Lakoja: Fair point, as "my problem discussion" above will probably never happen.

I have the feeling that we should leave it here, as there is another location like this, but maybe remove it in future again. (Might research the background a bit.)

Gryzor: Yeah, I'd like to introduce another alternative, (merely as a draft) that modifies this await() into something different.

In lieu of the current situation, I agree with @Lakoja that we should decide in a different PR. (this is suspiciously very convenient for me so I don't have to push again... but I swear it's not the reason for my opinion) ;)

Gryzor: (the truth is, I am ok with AlertDialog the way it is... not everything needs to be abstracted/wrapped for the sake of ... questionable readability).

PR #3976, opened 2023-08-21
Fix exception when updating summary notifications

nikclayton:
https://github.com/tuskyapp/Tusky/commit/dc9e9f2aebfd3feb057fee4c3ac72b05cbc75a12 modifed the code that fetched the value of EXTRA_NOTIFICATION_TYPE in an intent, to use getSerializable().

However, the value was being placed in to the intent using putString().

This caused an exception when trying to update the summary notification, so it would never update.

```
java.lang.ClassCastException: java.lang.String cannot be cast to com.keylesspalace.tusky.entity.Notification$Type
    at com.keylesspalace.tusky.components.notifications.NotificationHelper.updateSummaryNotifications(NotificationHelper.java:321)
    at com.keylesspalace.tusky.components.notifications.NotificationFetcher.fetchAndShow(NotificationFetcher.kt:87)
    at com.keylesspalace.tusky.components.notifications.NotificationFetcher$fetchAndShow$1.invokeSuspend(Unknown Source:14)
```

Fix this by placing the value in to the intent using putSerializable(), to match how it will be fetched.

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/components/notifications/NotificationHelper.java
+++ b/app/src/main/java/com/keylesspalace/tusky/components/notifications/NotificationHelper.java
@@ -238,7 +238,7 @@ public class NotificationHelper {
         Bundle extras = new Bundle();
         // Add the sending account's name, so it can be used when summarising this notification
         extras.putString(EXTRA_ACCOUNT_NAME, body.getAccount().getName());
-        extras.putString(EXTRA_NOTIFICATION_TYPE, body.getType().toString());
+        extras.putSerializable(EXTRA_NOTIFICATION_TYPE, body.getType());
         builder.addExtras(extras);
 
         // Only alert for the first notification of a batch to avoid multiple alerts at once


Lakoja:
What would be a sensible way to "serialize" an enum value?

Without seeing this I would have leaned to string probably. (That's why we also have enums to have meaningful names on the values.)

Also asking because the code extracting the serializable is newer than the code placing the string. So I would have expected the new code to also extract a string (and do "fromString()" for the enum)?

nikclayton:
Android intent passing is a hot mess. Sticking arbitrary types as keys in to a map keyed by a string, and hoping that the receiver interprets it the same way is asking for problems like this.

The correct, safe way to do it is that everything that can receive an intent should declare something like

```kotlin
@Parcelize
data class IntentArgs(
    val foo: String,
    val bar: SomeEnum,
    val baz: ...,
) : Parcelable()
```

(or possibly using the `Serializable` interface instead -- I've thought about this a bit but not written any code yet to test it out)

and then insist on receiving values like that, probably via a `newInstance()` method that sets everything up. That would be nicely type-safe, while incurring what I expect would be negligible overhead due to the parcelisation (https://developer.android.com/guide/components/activities/parcelables-and-bundles#sdba).

> What would be a sensible way to "serialize" an enum value?
>
> Without seeing this I would have leaned to string probably.

You have to be careful with that. Manually serialising by string would be OK in this instance because the enum value is never persisted somewhere.

But it's a problem if the enum value is saved to disk somewhere (e.g., in a database column). Because now you can never, ever rename the enum without also ensuring that historical data is migrated. And is is not obvious that you can't do that at the place where the enum is declared, so it's a loaded foot-gun, waiting to go odd.

You can't rely on the enum ordinal value either, for similar reasons -- if you ever reorder the enums, or add a new value somewhere other than the end you've just silently broken your stored data.

**However...**

In this specific case, if you trace through the code in Android Studio (e.g., by starting with the `putSerializable` call and repeatedly hitting Ctrl-B to jump through the implementation code) you'll discover the `Bundle` stores the data as an `ArrayMap<String, Object>`, because `putSerializable` works on anything that implements the `Serializable` interface, and enums do.

So there's no "serialization" going on in the classic-sense -- it's not converting the enum to a string behind the scenes and then creating an enum using the string again, the enum object is being passed straight through.

So `putSerializable` is the right thing to do here.

Lakoja (approved):

Lakoja:
Or: using put/getSerializable frees you from worrying when all you want to do is "pass this data". :-)

Issue #3331, opened 2023-02-18
Revamping timelines, tabs, and the navigation drawer


nikclayton:
I've been looking at the different open issues around navigation and tabs. This is a single, consistent proposal to fix all of them. Feedback requested.

## Problem

The current split between what's in the navigation drawer and what's in tabs is causing problems for users.

- It's unclear what goes where
- The limit of five tabs is confusing (https://github.com/tuskyapp/Tusky/issues/2512)
- There are artificial restrictions on what can be a tab and what can be in the navigation drawer. E.g.,
  - There's no "Bookmarks" tab
  - If you remove "Notifications" from the tab list it doesn't appear in the navigation drawer. Ditto for e.g., "Direct Messages"
- Users are complaining about this and proposing fixes (e.g., https://github.com/tuskyapp/Tusky/issues/3010, https://github.com/tuskyapp/Tusky/issues/2939)

### Elevator pitch

The navigation drawer becomes the canonical location for all the timelines that the user can access.

Tabs become completely user-controlled shortcuts to those timelines. A user may choose to have zero tabs. Or many tabs. Entirely up to them.

### Details

Model it as follows:

1. The user has access to many "timelines"
2. A timeline is one of:
    - Home
    - Notifications
    - Local timeline of the user's instance
    - Local timeline of another instance (https://github.com/tuskyapp/Tusky/issues/874)
    - "Customised" timelines (e.g., posts with media only, see https://github.com/tuskyapp/Tusky/issues/1332)
    - Federated timeline
    - DMs
    - Follow requests (subset of notifications)
    - Mentions (subset of notifications, https://github.com/tuskyapp/Tusky/issues/2735, https://github.com/tuskyapp/Tusky/issues/2378)
    - Scheduled posts
    - Drafts
    - Bookmarks
    - Favourites
    - Trending (the various trending types are different timelines)
    - Announcements
    - 0-N user defined lists of accounts (currently called "Lists")
    - 0-N user defined lists of hashtags (currently the "Hashtags" tabs)
3. Every timeline that the user has appears on the navigation drawer, and is directly selectable
4. Tabs are for "quick access"
    - The user may have 0-N tabs
    - If they have 0 tabs the tab bar does not appear, and all navigation is done through the drawer
    - Otherwise, the user may add as many of their timelines to the tabs as they want
    - The user can specify the "default" tab (the tab they return to after pressing "Back", which may not be the home timeline, https://github.com/tuskyapp/Tusky/issues/1838)
[...]

PR #3982, opened 2023-08-24
Allow the user to specify an arbitrary number of tabs

### LINKED ISSUE #2512: More than five tabs
Please can we have more than five tabs - either by squashing them together a bit more or by allowing to scroll the tabs bar left/right to see the ones which are out of view?

Home, Mentions, Local, Federated, Direct Messages is already five tabs before adding ones for lists or hashtags.

nikclayton:
Fixes #2512

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/TabPreferenceActivity.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/TabPreferenceActivity.kt
@@ -164,18 +164,12 @@ class TabPreferenceActivity : BaseActivity(), Injectable, ItemInteractionListene
             toggleFab(false)
         }
 
-        binding.maxTabsInfo.text = resources.getQuantityString(R.plurals.max_tab_number_reached, MAX_TAB_COUNT, MAX_TAB_COUNT)
-
         updateAvailableTabs()
 
         onBackPressedDispatcher.addCallback(onFabDismissedCallback)
     }
 
     override fun onTabAdded(tab: TabData) {
-        if (currentTabs.size >= MAX_TAB_COUNT) {
-            return
-        }
-
         toggleFab(false)
 
         if (tab.id == HASHTAG) {
@@ -388,7 +382,6 @@ class TabPreferenceActivity : BaseActivity(), Injectable, ItemInteractionListene
 
         addTabAdapter.updateData(addableTabs)
 
-        binding.maxTabsInfo.visible(addableTabs.size == 0 || currentTabs.size >= MAX_TAB_COUNT)
         currentTabsAdapter.setRemoveButtonVisible(currentTabs.size > MIN_TAB_COUNT)
     }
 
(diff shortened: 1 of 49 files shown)


nikclayton:
+20,000 !!! -- don't panic, that's changes to the lint baseline file (this PR removes some existing lint errors as a nice side effect).

nikclayton:
This changes the initial tab presentation. Android behaviour seems to be to size the tabs appropriately and leave space, like this.

image

I think that's fine -- it's a clear indiciation to the user "Hey, you can put more stuff here".

The tabs to the right may or may not be partially visible, depending on the user's other settings.

For example, with 6 tabs configured:

image

the extra tab is peeking in from the right. But if the user has turned off the top nav it looks like this:

image

and it's not apparent there are extra tabs.

However, those extra tabs become visible when the user starts swiping through the tabs or tapping on the icons. For example, after tapping on the DM icon the icons scroll, and it's immediately obvious there is more content.

That looks like this:

image

and this (if top nav is turned off):

image

To forestall the obvious comments:

1. We can't remove the avatar icon, that's the accessible way for the user to open the navigation menu
2. We can't remove the vertical-ellipsis menu, that's the accessible way for the user to refresh and access other functionality that is normally gesture-based.

nikclayton:
Additional enhancements worth considering, but I'm not going to do as part of this PR:

- Add a fixed tab on the right that, when clicked, opens the tab preferences UI, so the user can adjust this with one click instead of three (tap avatar to open nav menu, tap "Account preferences", tap "Tabs")
- Allow the user to specify custom tab titles, and show those titles

Existing bugs that this PR does not fix, and will not fix:

- The user can add multiple tabs of the same type (e.g., multiple DM tabs). I explain how to fix that in https://github.com/tuskyapp/Tusky/issues/3331 (you can think of this PR as being part of the set of work necessary to implement the vision set out in that issue).

app/src/main/java/com/keylesspalace/tusky/TabPreferenceActivity.kt
```diff
@@ -421,6 +414,5 @@ class TabPreferenceActivity : BaseActivity(), Injectable, ItemInteractionListene
 
     companion object {
         private const val MIN_TAB_COUNT = 2
```
Lakoja, line 416: Better have 1 here?

(2 seems arbitrary, 1 is a "guard against total misconfiguration")

Lakoja:
> +20,000 !!! -- don't panic, that's changes to the lint baseline file (this PR removes some existing lint errors as a nice side effect).

Obsidian? A JS file?

nikclayton:
> > +20,000 !!! -- don't panic, that's changes to the lint baseline file (this PR removes some existing lint errors as a nice side effect).
> 
> Obsidian? A JS file?

Oh, WTF is that doing there. One sec...

PR #3985, opened 2023-08-25
Move reviewdog-suggester permissions block to before the job definition

nikclayton:
The reviewdog-suggester job was failing to leave comments, reporting:

```
reviewdog: This GitHub token doesn't have write permission of Review API [1], 
so reviewdog will report results via logging command [2] and create annotations similar to
github-pr-check reporter as a fallback.
```

Despite a `permissions` block the log showed the permissions as:

```
GITHUB_TOKEN Permissions
  Contents: read
  Issues: read
  Metadata: read
  PullRequests: read
```

Move the `permissions` block to earlier in the file, in case the position is important.

### DIFF
--- a/.github/workflows/ktlint.yml
+++ b/.github/workflows/ktlint.yml
@@ -1,5 +1,11 @@
 name: reviewdog-suggester
 on: pull_request
+
+permissions:
+  contents: read
+  issues: write
+  pull-requests: write
+
 jobs:
   ktlint:
     timeout-minutes: 5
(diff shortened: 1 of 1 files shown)


nikclayton:
Reviewers, it appears as though this does fix the problem. If you open https://github.com/tuskyapp/Tusky/actions/runs/5975036108/job/16210256090?pr=3985#step:1:20 and look at the permissions shown when it's run with this PR you'll see:

```
GITHUB_TOKEN Permissions
  Contents: read
  Issues: write
  Metadata: read
  PullRequests: write
```

And that looks correct.

PR #4003, opened 2023-09-07
Rollback aggressive linting

Tak:
- Revise linting rules
  - Make lint failures opt-in (for the project) instead of opt-out
  - Reduce noise
  - Add explicit errors for things we would ask somebody to change in a code review
  - Update baseline to only include the new errors
  - Remove baseline autoupdate task since:
    - We want this to happen very rarely
    - The autoupdater also adds warnings
- Remove reviewdog github action (that autoadds lint comments to PRs)


Tak:
What I mean by "opt in" is that we can choose which lint warnings we want to be failures by marking them in `lint.xml` - for this PR, I cherry-picked the few ones from the baseline file that seemed like actual issues and not pedantry.

Users can also see the report with *all* the warnings by running the gradle task locally and opening the report it generates (I can add this to the contributing doc).

I also plan to follow this up with another PR that fixes most of our remaining warnings as well.

app/build.gradle
```diff
@@ -202,18 +202,3 @@ tasks.withType(org.jetbrains.kotlin.gradle.internal.KaptWithoutKotlincTask) {
...
-    notCompatibleWithConfigurationCache("Is always out of date")
-    outputs.upToDateWhen { false }
-
-    doLast {
-        delete android.lint.baseline.path
-    }
-
-    // Regenerate the lint baseline
-    it.finalizedBy tasks.named("lintBlueDebug")
-}
```
DavidEdwards, line 219: Does this need to be deleted? I am not so sure. It should only generate a baseline, not cause aggressive linting.

Tak: The main issue with this is that it puts all the outstanding *warnings* into the baseline file as well - ideally, we will be adding things to the baseline so rarely that an automated task to update it should be overkill

connyduck (approved):

PR #4007, opened 2023-09-08
Add "Trending posts" (statuses) feed

tinsukE:
Add "Trending posts" (statuses) feed.

This feed is a good source of interesting accounts to follow and, personally, a sort of "Front page of the Fediverse".

Since #3908 and #3910 (which would provide a more thorough, albeit complex, access to trending things) won't get merged, I'd like to address this missing feed by simply adding another tab/feed.

~~If desired, I can move the second commit (fixing lint) to another PR.~~


Lakoja:
Thank's for your contribution.
This looks good. :-)

One (larger) question, I haven't looked this up myself, yet: In what way do the trending posts differ from tags, so that they are a timeline (and the tags are not)?

tinsukE:
> In what way do the trending posts differ from tags, so that they are a timeline (and the tags are not)?

Although both of them are "trending", they do indeed differ. To me, it makes sense to track a tag's popularity over time. But for posts, if they get popular enough, they join the "featured" feed.

At least that's how I understood their UX.

DavidEdwards:
Looks good to me too.

Tags are different in that there is no "content" related to it. The relevant information about it, is how many times have people used the tag.

Whereas for posts, the content is the single post, and how often that post has been… seen? It might be interesting to provide a metric on the post, to let the user know how relevant it is. However, not necessary I would say.

Tak:
If you merge `develop` and resolve the conflicts, I think this is ready to merge.

tinsukE:
@Tak done!
Due to #4003 I was able to drop the 2nd commit entirely.

Issue #3648, opened 2023-05-13
NotificationFetcher does not fetch all notifications


nikclayton:
To repeat the comment from https://github.com/tuskyapp/Tusky/pull/3626:

`fetchNotifications` (`fetchNewNotifications` in this PR) makes one call to get new notifications (`mastodonApi.notificationsWithAuth`).

That will return N new notifications, starting with the oldest new one, up to `limit`. `limit` is not set in this code, and is documented as being default=15 max=30 at https://docs.joinmastodon.org/methods/notifications/#get.

If we can trust these documented defaults this means that in any given notification fetch (i.e., approx every 15 minutes) Tusky will get the batch of next-newest 15 notifications and display them. If the user has more notifications than that they will have to wait another ~ 15 minutes until they are fetched.

That's probably also a source of user complaints about missing notifications.

Fix is to either:

1. Use the NotificationsRepository and fetch the notifications that way, taking advantage of its existing code for paging up (prepending) the list of notifications. Not sure how well Pager3 is suited to that though.
2. Write some additional code here that repeatedly fetches new notifications until there are no new ones.

_Originally posted by @nikclayton in https://github.com/tuskyapp/Tusky/issues/3626#issuecomment-1546621335_

Issue #3623, opened 2023-05-04
Refreshing notifications collapses already expanded content warning


connyduck:
- Have a notification with a content warning
- Expand the content warning
- pull to refresh all notifications

Expected behavior:
Local state of notifications gets preserved, e.g. collapsed state so I don't have to click to expand it multiple times

Actual behavior;
Local state is lost and already expanded warnings are collapsed again

Fairphone 4, Android 12
both Tusky 22 beta 2 and Tusky Test 22.0 beta 2-80774f35

PR #4015, opened 2023-09-11
Resets the paging3 changes of 3159 back to the (java) fragment code

Lakoja:


mcclure:
I tested this patch against regressions caused by 3159:

- "Glitchy animation" during pulldown refresh— fixed (but I have animations disabled, would be good if someone could test on a second phone)
- Pulldown refresh fetches twice— Did not test, I don't know how to test this
- "when [pulldown] refreshing while offline fullscreen error appears so one can't see read already loaded notifications anymore"— fixed (I could not reproduce)
- #3623 - Refreshing notifications collapses already expanded content warning — fixed
- Notifications scroll can get stuck in the distant past (say by days) and you can't jump up to "the present" except by using the menu / "Load More" should come back until we make an explicit decision to remove this — I don't know how to test this because it would require leaving the app open for a long period. However in all my tests, closing and reopening the app caused the notifications tab to always jump to "the present", which mostly obviates this problem. I don't know if that's "expected"/as-of-March behavior.

connyduck:
> It's missing the button functions (needs Kotlin to call the Api)

add some more "old" functions to MastodonApi?

> Most menu items are not wired (they were added after this code)

don't we want to get rid of them as well and have the old bar back?

> There are some new functions with "Old" in the name which should be removed when Kotlin

That is fine

> Some (newer) code can still be removed as it is now unused

remove it as well please

mcclure:
> don't we want to get rid of them as well and have the old bar back?
> 

There are 4 items in the menu
* Refresh
* Load newest notifications
* Filter notifications
* Delete notifications

I think when we bring the bar back we should remove "Filter notifications" and "Delete notifications". (Except when the bar is hidden, which I think you said we had a preference for.)

But it may be worth keeping the menu for "Refresh" and "Load newest notifications". You can already get these both things with gestures (pull down or tap tab header) but having a second option when that is not clear might be good.

connyduck:
> I think when we bring the bar back we should remove "Filter notifications" and "Delete notifications". (Except when the bar is hidden, which I think you said we had a preference for.)

Yes makes sense.

"Load newest notifications" - was that a workaround to jump back to the top when one is way behind on notifications? If yes it makes no sense anymore, if no what is the difference to "refresh"?

Lakoja:
Noticed something related in this code: https://github.com/tuskyapp/Tusky/issues/3648

Not sure if this is a problem (i. e. will fetch notifications endlessly as there are so many...).

Worth an investigation / followup?

Lakoja:
> > It's missing the button functions (needs Kotlin to call the Api)
> 
> add some more "old" functions to MastodonApi?
> 
> > Some (newer) code can still be removed as it is now unused
> 
> remove it as well please

Done 2x

Tak (approved):

PR #4029, opened 2023-09-24
add role badges

connyduck:
This adds support for the new Mastodon 4.2 role badges. Admins can define if a role should be visible in the interface and then we get it delivered by the Api on the `Account` object like this:
```
"roles": [
        {
              "id": "4",
              "name": "TEST",
              "color": "#ffee00"
        }
  ]
```

- keeps compatibility with older Mastodon version and non Mastodon servers
- Took me a while, but I figured out a way to use the color and have it look ok on all backgrounds (Mastodon itself ignores the color and just always uses its brand color)
- falls back to Tusky blue in case no color is configured
- I adjusted the "Follows you" and "Bot" badges so they match the new badge style
- In case the "Follows you" and "Bot" badges are visible at the same time, "Follows you" gets its own line and "Bot" goes into the same line as the role badge.
- Will work even with a lot of role badges (right now users can only have 1 role at once though)
- Will work even when the badges federate (right now they don't)

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/components/account/AccountActivity.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/components/account/AccountActivity.kt
@@ -22,9 +22,12 @@ import android.content.Context
 import android.content.Intent
 import android.content.res.ColorStateList
 import android.graphics.Color
+import android.graphics.Typeface
 import android.graphics.drawable.LayerDrawable
 import android.os.Bundle
+import android.text.SpannableStringBuilder
 import android.text.TextWatcher
+import android.text.style.StyleSpan
 import android.view.Menu
 import android.view.MenuInflater
 import android.view.MenuItem
@@ -32,10 +35,12 @@ import android.view.View
 import android.view.ViewGroup
 import androidx.activity.viewModels
 import androidx.annotation.ColorInt
+import androidx.annotation.DrawableRes
 import androidx.annotation.Px
 import androidx.appcompat.app.AlertDialog
 import androidx.appcompat.content.res.AppCompatResources
 import androidx.core.app.ActivityOptionsCompat
+import androidx.core.graphics.ColorUtils
 import androidx.core.view.MenuProvider
 import androidx.core.view.ViewCompat
 import androidx.core.view.WindowCompat
@@ -48,6 +53,7 @@ import androidx.recyclerview.widget.LinearLayoutManager
 import androidx.viewpager2.widget.MarginPageTransformer
 import com.bumptech.glide.Glide
 import com.google.android.material.appbar.AppBarLayout
+import com.google.android.material.chip.Chip
 import com.google.android.material.color.MaterialColors
 import com.google.android.material.floatingactionbutton.FloatingActionButton
 import com.google.android.material.shape.MaterialShapeDrawable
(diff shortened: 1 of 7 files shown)


app/src/main/java/com/keylesspalace/tusky/entity/Account.kt
```diff
@@ -36,8 +36,8 @@ data class Account(
     val bot: Boolean = false,
     val emojis: List<Emoji>? = emptyList(), // nullable for backward compatibility
     val fields: List<Field>? = emptyList(), // nullable for backward compatibility
-    val moved: Account? = null
-
+    val moved: Account? = null,
+    val roles: List<Role>? = emptyList()
```
Lakoja, line 40: Is this somehow cached somewhere (or is this only AccountEntity)? So could there then be accounts without roles which have some?

app/src/main/java/com/keylesspalace/tusky/components/account/AccountActivity.kt
```diff
@@ -491,6 +497,33 @@ class AccountActivity : BottomSheetActivity(), ActionButtonActivity, MenuProvide
...
+
+        val isLight = resources.getBoolean(R.bool.lightNavigationBar)
+
+        if (loadedAccount?.bot == true) {
+            val badgeView = getBadge(getColor(R.color.tusky_grey_50), R.drawable.ic_bot_24dp, getString(R.string.profile_badge_bot_text), isLight)
+            binding.accountBadgeContainer.addView(badgeView)
+        }
+
+        loadedAccount?.roles?.forEach { role ->
+            val badgeColor = if (role.color.isEmpty()) {
```
Lakoja, line 511: Would switch cases: normal one first.

app/src/main/java/com/keylesspalace/tusky/components/account/AccountActivity.kt
```diff
@@ -491,6 +497,33 @@ class AccountActivity : BottomSheetActivity(), ActionButtonActivity, MenuProvide
...
+                // sometimes the color is not set for a role, in this case fall back to our default blue
+                getColor(R.color.tusky_blue)
+            } else {
+                Color.parseColor(role.color)
+            }
+
+            val sb = SpannableStringBuilder("${role.name} ${viewModel.domain}")
```
Lakoja, line 518: Out of curiosity: Is translation an issue here? This does only depend on the language of the server?

app/src/main/java/com/keylesspalace/tusky/entity/Account.kt
connyduck: Data for AccountActivity is always freshly loaded, no cache.

app/src/main/java/com/keylesspalace/tusky/components/account/AccountActivity.kt
connyduck: The role has only one name. Admins can input whatever they want. So they probably use a name that is meaningful in the servers main language. It can't be translated.

connyduck: I think Mastodon comes with some default roles ("Owner", "Admin", "Moderator") to map the old permission system to, but admins can change them however they like.

Lakoja (approved):

digitalf0x:
[Mastodon itself renamed the `Bot` flag to `Automated`](https://github.com/mastodon/mastodon/pull/25356 ).  As you're modifying that label here, should Tusky follow?

This may include modifying the profile setting inside Tusky (Mastodon changed the label from `This is a bot account` to `This is an automated account`).

(I have no strong preference either way.)

Tak (approved):

PR #4031, opened 2023-09-24
use link icon instead of emoji when showing hidden urls

### LINKED ISSUE #4028: Incorrect formatting of link embedded in HTML toot when Blobmoji emoji style is used
This happens when Emoji style is set to Blobmoji in the Settings.

When viewing a toot with a link embedded in HTML, the link should appear formatted as `<link-text> (🔗 <url>)`

However, I'm seeing something that looks like `<link-text> (    <url>)`, where space is reserved for the "🔗", but it does not actually render.

An example toot is [this one](https://social.treehouse.systems/@rcombs/111105256891594793)

* * * *
- Tusky Version: 23.0
- Android Version: 13 (Security Update Sept. 1 2023)
- Android Device: Pixel 7 Pro
- Mastodon instance (if applicable): mendeddrum.org viewing treehouse.systems, but that doesn't seem relevant here.

connyduck:
Looks way better imho. Also closes #4028.

Screenshot_20230924_162657


connyduck:
Icon from https://pictogrammers.com/library/mdi/icon/link-variant/ 
Another one that might work: https://pictogrammers.com/library/mdi/icon/open-in-new/

app/src/main/java/com/keylesspalace/tusky/util/LinkHelper.kt
```diff
@@ -99,8 +101,22 @@ fun markupHiddenUrls(context: Context, content: CharSequence): SpannableStringBu
         val start = spannableContent.getSpanStart(span)
         val end = spannableContent.getSpanEnd(span)
         val originalText = spannableContent.subSequence(start, end)
-        val replacementText = context.getString(R.string.url_domain_notifier, originalText, getDomain(span.url))
+        val replacementText = view.context.getString(R.string.url_domain_notifier, originalText, getDomain(span.url))
         spannableContent.replace(start, end, replacementText) // this also updates the span locations
```
Lakoja, line 105: Is this then still needed?

connyduck: Because the string is no longer translateable? Yes it could be defined right in the code as well, I don't think it matters a lot. Should I change it?

Tak (approved):

app/src/main/java/com/keylesspalace/tusky/util/LinkHelper.kt
Lakoja: One of them didn't seem to be use anymore (lateron in the method). But I probably looked wrong..

Lakoja (approved):

PR #4048, opened 2023-10-06
fix boost button not updating when boosting

connyduck:


app/src/main/java/com/keylesspalace/tusky/adapter/StatusBaseViewHolder.java
```diff
@@ -725,6 +725,7 @@ private void showConfirmReblog(StatusActionListener listener,
             listener.onReblog(!buttonState, position);
             if(!buttonState) {
                 reblogButton.playAnimation();
+                reblogButton.setChecked(true);
```
connyduck, line 728: This eagerly updates the status which feels way nicer when the api is slow. We do this already when reblog confirmations are off.

Tak (approved):

PR #4051, opened 2023-10-08
2512: Allow more than 5 tabs

### LINKED ISSUE #2512: More than five tabs

Lakoja:
Fixes #2512 

grafik

Can add an arbitrary number of tabs.
Graphical behavior is unchanged for small numbers: the whole space if filled with the tabs - they are enlarged if needed.

If there are more the mode switches to "scrollable".
This does not, however, look very differently (see screenshot with the current tab scrolled out).

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/util/AdaptiveTabLayout.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/util/AdaptiveTabLayout.kt
@@ -1,0 +1,46 @@
+package com.keylesspalace.tusky.util
+
+import android.content.Context
+import android.util.AttributeSet
+import android.view.ViewGroup
+import com.google.android.material.tabs.TabLayout
+
+/**
+ * Switches the tab display mode depending on available size: start out with "scrollable" but
+ * if there is enough room switch to "fixed" (an re-measure).
+ *
+ * Idea taken from https://stackoverflow.com/a/44894143
+ */
+class AdaptiveTabLayout : TabLayout {
+    constructor(context: Context) : super(context)
+    constructor(context: Context, attrs: AttributeSet?) : super(context, attrs)
+    constructor(context: Context, attrs: AttributeSet?, defStyleAttr: Int) : super(
+        context,
+        attrs,
+        defStyleAttr
+    )
+
+    override fun onMeasure(widthMeasureSpec: Int, heightMeasureSpec: Int) {
+        tabMode = MODE_SCROLLABLE // make sure to only measure the "minimum width"
+        super.onMeasure(widthMeasureSpec, heightMeasureSpec)
+
+        if (tabCount < 2) {
+            return
+        }
+
+        try {
+            val tabLayout = getChildAt(0) as ViewGroup
+            var widthOfAllTabs = 0
+            for (i in 0 until tabLayout.childCount) {
+                widthOfAllTabs += tabLayout.getChildAt(i).measuredWidth
+            }
+            if (widthOfAllTabs <= measuredWidth) {
+                // fill all space if there is enough room
+                tabMode = MODE_FIXED
+                super.onMeasure(widthMeasureSpec, heightMeasureSpec)
+            }
+        } catch (e: Exception) {
+            e.printStackTrace()
+        }
+    }
+}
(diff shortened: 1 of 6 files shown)


connyduck:
Why not just use `app:tabMode="auto"` instead of a custom view?

Lakoja:
> Why not just use `app:tabMode="auto"` instead of a custom view?

Yes, I tried that first.
It didn't work for me. (And there are a few SO topics with that content as well.)

connyduck:
What does "didn't work" mean exactly? I just tried it, seems fine.

Lakoja:
> What does "didn't work" mean exactly? I just tried it, seems fine.

Maybe a version issue?

It behaved exactly like "scrollable": No enlargement for few items.

Lakoja:
This is how it looks with "auto" on Android 11 (emulator):

grafik

connyduck:
Ok it is not stretching the items when they are few. Seems like it is a known bug since 4 years but they don't want to fix it because it is a breaking api change. https://github.com/material-components/material-components-android/issues/650
oh well

app/src/main/java/com/keylesspalace/tusky/util/AdaptiveTabLayout.kt
```diff
@@ -0,0 +1,46 @@
...
+ * Idea taken from https://stackoverflow.com/a/44894143
+ */
+class AdaptiveTabLayout : TabLayout {
+    constructor(context: Context) : super(context)
+    constructor(context: Context, attrs: AttributeSet?) : super(context, attrs)
+    constructor(context: Context, attrs: AttributeSet?, defStyleAttr: Int) : super(
+        context,
+        attrs,
+        defStyleAttr
+    )
```
connyduck, line 21:
```suggestion
class AdaptiveTabLayout @JvmOverloads constructor(
    context: Context,
    attrs: AttributeSet? = null,
    defStyleAttr: Int = 0
) : TabLayout(context, attrs, defStyleAttr) {
```

app/src/main/java/com/keylesspalace/tusky/util/AdaptiveTabLayout.kt
```diff
@@ -0,0 +1,46 @@
+package com.keylesspalace.tusky.util
```
connyduck, line 1: should be in the `com.keylesspalace.tusky.view` package

app/src/main/java/com/keylesspalace/tusky/util/AdaptiveTabLayout.kt
```diff
@@ -0,0 +1,46 @@
...
+                widthOfAllTabs += tabLayout.getChildAt(i).measuredWidth
+            }
+            if (widthOfAllTabs <= measuredWidth) {
+                // fill all space if there is enough room
+                tabMode = MODE_FIXED
+                super.onMeasure(widthMeasureSpec, heightMeasureSpec)
+            }
+        } catch (e: Exception) {
+            e.printStackTrace()
+        }
```
connyduck, line 44: Catching all exceptions is rarely a good idea - and I don't see what would even throw here?

(also exceptions should be logged with `android.util.Log` and not `printStackTrace()`)

connyduck (changes requested):

app/src/main/java/com/keylesspalace/tusky/util/AdaptiveTabLayout.kt
Lakoja: Removed (was copy&paste)

app/src/main/java/com/keylesspalace/tusky/util/AdaptiveTabLayout.kt
Lakoja: ok

connyduck (approved):

Lakoja:
> My fault for not trying until after I merged, but it looks like this for me (pixel 2xl android 11)
> 
> Screenshot_20231011-100700

So, all the colors are wrong?

Especially also white is not correct?

You can easily disable this change: Revert `activity_main.xml`

Is it correct then?

Lakoja:
> My fault for not trying until after I merged, but it looks like this for me (pixel 2xl android 11)
> 
> Screenshot_20231011-100700

https://github.com/tuskyapp/Tusky/pull/4058

Trying to fix it there.

However, I have no idea why this happens (now) and also my fix does not fully work.

(You can set the tab icon color, but not when it is selected.)

PR #4061, opened 2023-10-12
fix theme preference defaults

### LINKED ISSUE #4060: Wrong default for theme preference
reported here: https://strangeobject.space/@bebatjof/111218192881468173

In #3813 the default theme was set to "system", but it seems like the interface default was not updated which leads to this weird behavior when installing the app on a device where light is the default and then going to Preferences:

Screenshot_20231012_211455

connyduck:
closes #4060 

Also I noticed that we had the theme defaults twice in the code so I refactored a bit so only one are still in there.


connyduck:
Wait, is an enum a good choice here if all it is used for is to get the value?

Tak (approved):
Eh, I think the enum still makes more sense than a random block of otherwise ungrouped string constants 🤷

connyduck:
Yes strings also aren't the perfect solution and this fixes a bug, so I vote for merging it.

Issue #4063, opened 2023-10-14
Text box size for audio file captions is too small


straemer:
Text box size is too small to view text properly when setting caption for audio file.

![Screenshot_20231013-195207_Tusky](https://github.com/tuskyapp/Tusky/assets/1400238/ffc12e0b-ba01-4706-8350-f79213d39e18)

To reproduce:

1. Use another app to "share" an audio file with Tusky (I used ASR)
2. Edit caption on file. 

Note that I don't see this issue when editing a post, although the text box still doesn't have a lot of vertical room.

* * * *
- Tusky Version: 23.0
- Android Version: 10
- Android Device: Samsung Galaxy S9
- Mastodon instance (if applicable):

- [X] I searched or browsed the repo’s other issues to ensure this is not a duplicate.

PR #4066, opened 2023-10-21
4063: Make dialog size more stable

Lakoja:
Fixes: #4063

Switching from an AlertDialog to only a DialogFragment.

I didn't get the AlertDialog to be sized correctly.
It also opens now directly with the right (full screen) size. When the imageView fails to load (i.e. with an audio file) it will be hidden.

This changes the button layout somewhat.

One observation: The placeholder text "... visually impaired..." is not quite right as a description for an audio file is not intended for the visually impaired. But I couldn't think of a better text just yet.

grafik

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/components/compose/dialog/CaptionDialog.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/components/compose/dialog/CaptionDialog.kt
@@ -15,7 +15,6 @@
 
 package com.keylesspalace.tusky.components.compose.dialog
 
-import android.app.Dialog
 import android.content.Context
 import android.graphics.drawable.Drawable
 import android.net.Uri
@@ -26,7 +25,7 @@ import android.view.View
 import android.view.ViewGroup
 import android.view.WindowManager
 import android.widget.EditText
-import androidx.appcompat.app.AlertDialog
+import android.widget.LinearLayout
 import androidx.core.os.BundleCompat
 import androidx.core.os.bundleOf
 import androidx.fragment.app.DialogFragment
@@ -36,6 +35,8 @@ import com.bumptech.glide.request.target.CustomTarget
 import com.bumptech.glide.request.transition.Transition
 import com.keylesspalace.tusky.R
 import com.keylesspalace.tusky.databinding.DialogImageDescriptionBinding
+import com.keylesspalace.tusky.util.hide
+import com.keylesspalace.tusky.util.viewBinding
 
 // https://github.com/tootsuite/mastodon/blob/c6904c0d3766a2ea8a81ab025c127169ecb51373/app/models/media_attachment.rb#L32
 private const val MEDIA_DESCRIPTION_CHARACTER_LIMIT = 1500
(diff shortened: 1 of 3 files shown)


app/src/main/res/layout/dialog_image_description.xml
```diff
@@ -11,28 +12,53 @@
         android:layout_width="match_parent"
         android:layout_height="0dp"
```
baghaii, line 13: Can layout height be set to match_parent here so you don't have to set it in whatever you are doing in the onStart block?

app/src/main/res/layout/dialog_image_description.xml
```diff
@@ -11,28 +12,53 @@
...
 
         <com.google.android.material.textfield.TextInputEditText
             android:id="@+id/imageDescriptionText"
-            android:inputType="textCapSentences|textMultiLine|textAutoCorrect"
             android:layout_width="match_parent"
             android:layout_height="match_parent"
-            android:hint="@string/hint_description"
             android:gravity="start"
-            android:importantForAutofill="no" />
+            android:hint="@string/hint_description"
```
connyduck, line 35: this is not necessary since it is overridden from code anyway and the string is used only here it can be deleted as well
```suggestion
            tools:hint="Description"
```

app/src/main/java/com/keylesspalace/tusky/components/compose/dialog/CaptionDialog.kt
```diff
@@ -90,27 +105,30 @@ class CaptionDialog : DialogFragment() {
                 ) {
                     imageView.setImageDrawable(resource)
                 }
+
+                override fun onLoadFailed(errorDrawable: Drawable?) {
+                    super.onLoadFailed(errorDrawable)
+                    imageView.hide()
```
connyduck, line 111: I think instead of just assuming it is an audio file when loading fails I'd rather pass an explicit flag. The flag could also be used to display a different hint that makes more sense for audio files.

connyduck (commented):
Definitely better than before.

I wonder if the dialog would be better if the image had a bit of margin around it. Looks a bit weird when it is glued to the side like that

app/src/main/java/com/keylesspalace/tusky/components/compose/dialog/CaptionDialog.kt
Lakoja: I was wondering (and didn't quickly find) how to find the type of the media thing.

Because always putting it in Glide says that the code expects an image there - which it probably shouldn't.

Lakoja:
> Definitely better than before.
> 
> I wonder if the dialog would be better if the image had a bit of margin around it. Looks a bit weird when it is glued to the side like that

That was a concious choice in the previous issue here (improving that dialog 5 months ago): Use as much space for the dialog and in it for the image as is possible/sensible.
This especially meant getting rid of one or two margins.

app/src/main/res/layout/dialog_image_description.xml
Lakoja: This is the image view (with layout_weight=1).

The things in the onStart influence the window/dialog. Normally an (alert) dialog implicitly has something like "wrap_content" that is overwritten there.

app/src/main/res/layout/dialog_image_description.xml
Lakoja: Done.

(It's a warning now: "don't use hard-coded strings" :-)

connyduck: False positive, if it starts with `tools` it is only for displaying the layout in Android Studio and won't end up in the apk.

connyduck:
> That was a concious choice in the previous issue here (improving that dialog 5 months ago): Use as much space for the dialog and in it for the image as is possible/sensible.
This especially meant getting rid of one or two margins.

Ok let's leave it then.

The dialog is now gigantic on tablets, any possibility it could have like a max size? Or at least show the image larger?

Lakoja:
> The dialog is now gigantic on tablets, any possibility it could have like a max size? Or at least show the image larger?

True.

However that is not new here and also affects other dialogs (e. g. "accounts in list").
I'd do it in a different PR/ticket.

(Generally tablets would appreciate some love. :-)

grafik

connyduck (approved):
ok

PR #4068, opened 2023-10-22
Remove garbage alt text from images attached via Gboard

untitaker:
Steps to reproduce:

1. install Gboard
   (https://play.google.com/store/apps/details?id=com.google.android.inputmethod.latin)

2. open a direct link to any image in Firefox
3. long-press the image to get a "Copy Image" dialogue (and copy the
   image)
4. compose a new post in Tusky
5. Gboard will suggest to paste the image from clipboard
6. paste image, see that when opening alt text editor, it is filled out
   with this garbage string

Why is this bad? It's not when I just fix the alt text. But it breaks
every mechanism that is supposed to remind me of adding alt text.

It's hard to argue that this is within scope of Tusky but I also don't
see it getting fixed in Gboard, so here we go.

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/components/compose/ComposeActivity.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/components/compose/ComposeActivity.kt
@@ -1064,7 +1064,10 @@ class ComposeActivity :
 
         // The Gboard android keyboard attaches this text whenever the user
         // pastes something from the keyboard's suggestion bar.
-        if (description == "Image from Gboard clipboard") {
+        // Due to different end user locales, the exact text may vary, but at
+        // least in version 13.4.08, all of the translations contained the
+        // string "Gboard".
+        if ("Gboard" in description) {
             return null
         }
 
--- a/app/src/main/java/com/keylesspalace/tusky/components/compose/ComposeActivity.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/components/compose/ComposeActivity.kt
@@ -1059,16 +1059,16 @@ class ComposeActivity :
 
     private fun sanitizePickMediaDescription(description: String?): String? {
         if (description == null) {
-            return null;
+            return null
         }
 
         // The Gboard android keyboard attaches this text whenever the user
         // pastes something from the keyboard's suggestion bar.
         if (description == "Image from Gboard clipboard") {
-            return null;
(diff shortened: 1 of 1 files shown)


connyduck:
Yes I agree this should be fixed, however your fix is only working for English. On my German device, it says "Bild aus der Gboard-Zwischenablage". No idea how we could handle all languages 😕.

untitaker:
right. some ideas:

* extract all strings from gboard apk and paste them in here
* check for case-sensitive substring "Gboard" instead of the exact string value

happy to spend time making either one of those solutions work

Lakoja:
>     5. Gboard will suggest to paste the image from clipboard
> 
>     6. paste image, see that when opening alt text editor, it is filled out
>        with this garbage string

Just as a technical question:
Why (how?) does pasting an image also set the alt text?
(Is that a sensible feature?)

untitaker:
it seems to be explicitly coded this way (also in the official Android app) so I didn't question it

On Mon, Oct 23, 2023, at 18:59, UlrichKu wrote:
> 
> 
>> `5. Gboard will suggest to paste the image from clipboard
>> 
>> 6. paste image, see that when opening alt text editor, it is filled out
>>    with this garbage string
`

connyduck:
> right. some ideas:
>
> * extract all strings from gboard apk and paste them in here
> * check for case-sensitive substring "Gboard" instead of the exact string value
>
> happy to spend time making either one of those solutions work, or some other solution

If "Gboard" occurs in all or most of the strings that is probably the best way.

untitaker:
just checked, seems this way:

```
apktool d 'Gboard - the Google Keyboard_13.4.08.559388404-beta-armeabi-v7a_Apkpure.apk'
cd .../
rg APKTOOL_DUPLICATE_string_0x7f14035d | rg -v Gboard
```

is empty

updated the PR

connyduck (approved):

Tak:
Thanks for your contribution, it will be available in the next release of Tusky Nightly

untitaker:
@Lakoja you got a real answer to your question over at https://github.com/mastodon/mastodon-android/pull/722#issuecomment-1779262851

> When you insert GIFs from Gboard, they do have sensible descriptions. Or at least they did in my testing. Also IIRC if you copy an image from a web browser, the description would be its alt text.

PR #4072, opened 2023-10-24
Support "replies policy" for lists

Tak:


connyduck:
I'm missing a text that explains what I'm selecting in that spinner. Web interface uses "Show replies to:".

connyduck (approved):

PR #4077, opened 2023-10-25
Exoplayer: Increase space between rewind, pause, ffwd buttons

mcclure:
I [posted our new video player layout] on Mastodon for comments and multiple people said the buttons were too close together. I agree. I added some space (I eyeballed it, I made it bigger until it felt too big and then I narrowed it), I think we have now increased the space from 10dp to 25dp. I added the space by wrapping the buttons in LinearLayouts, because they are <include>s and could theoretically insert more than one button.

Concerns: If the "next"/"prev" buttons ever become active, the space will not be correctly applied to those. We can fix that if it ever comes up (we don't display those buttons). If people think the buttons should be placed even further apart we can do this by just increasing the number in styles.xml.

This is what it looks like now. See previous look and comparison with 23.0 in #4071

### DIFF
--- a/app/src/main/res/layout/exo_player_control_view.xml
+++ b/app/src/main/res/layout/exo_player_control_view.xml
@@ -47,8 +47,9 @@
 
         <include layout="@layout/exo_player_control_rewind_button" />
 
+        <!-- Changes from media3 default: Different style to add more margins -->
         <ImageButton android:id="@id/exo_play_pause"
-            style="@style/ExoStyledControls.Button.Center.PlayPause"/>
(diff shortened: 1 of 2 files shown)


mcclure:
Will attempt to implement Conny's suggestion today, we should merge this only after that.

mcclure:
New patch should follow Conny's requests. Looks almost identical to above, the buttons seem to have widened by like 1px. (Moving 20dp down to 19dp made it *narrow* by 1px, so I decided to leave it at 20dp.)

45ab52df9bee3643

Tak (approved):

PR #4079, opened 2023-10-27
Fix showing filtered (warn) boosts in timelines

Tak:


app/src/main/java/com/keylesspalace/tusky/components/timeline/viewmodel/NetworkTimelineViewModel.kt
```diff
@@ -232,7 +232,7 @@ class NetworkTimelineViewModel @Inject constructor(
     }
 
     override fun clearWarning(status: StatusViewData.Concrete) {
-        updateActionableStatusById(status.actionableId) {
```
Lakoja, line 235: The actionableId was never right here?

Lakoja (approved):

connyduck (approved):

app/src/main/java/com/keylesspalace/tusky/components/timeline/viewmodel/NetworkTimelineViewModel.kt
Tak: It's because `updateActionableStatusById` looks up the status by the *outer* id, then updates the *actionable* status of that status

Issue #4005, opened 2023-09-07
Upgrade targetSdk to 34 (Android 14) and support predictive back


connyduck:
https://developer.android.com/guide/navigation/custom-back/predictive-back-gesture

https://developer.android.com/about/versions/14/behavior-changes-all

PR #4082, opened 2023-10-28
Bump compileSdk to 34

Goooler:


connyduck:
Should we also have a look how the predictive back gesture works with Tusky?

Related issue: #4005

Goooler:
This is just bumping `compileSdk`, which is irrelevant to `targetSdk`?

connyduck:
I see. I usually do them together but I guess this works as well.

connyduck:
There are some new deprecations which should probably be addressed:
`overrideActivityTransition`
`startActivityAndCollapse` - this one apparently throws `UnsupportedOperationException` on Upside Down Cake

PR #4090, opened 2023-10-31
Support code blocks

connyduck:
before

after

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/util/StatusParsingHelper.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/util/StatusParsingHelper.kt
@@ -17,15 +17,20 @@
 
 package com.keylesspalace.tusky.util
 
+import android.text.Editable
 import android.text.Html.TagHandler
+import android.text.Spannable
+import android.text.SpannableStringBuilder
 import android.text.Spanned
+import android.text.style.TypefaceSpan
 import androidx.core.text.parseAsHtml
+import org.xml.sax.XMLReader
 
 /**
  * parse a String containing html from the Mastodon api to Spanned
  */
 @JvmOverloads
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/keylesspalace/tusky/util/StatusParsingHelper.kt
```diff
@@ -17,21 +17,78 @@
...
 /**
  * parse a String containing html from the Mastodon api to Spanned
  */
 @JvmOverloads
-fun String.parseAsMastodonHtml(tagHandler: TagHandler? = null): Spanned {
+fun String.parseAsMastodonHtml(tagHandler: TagHandler? = tuskyTagHandler): Spanned {
     return this.replace("<br> ", "<br>&nbsp;")
         .replace("<br /> ", "<br />&nbsp;")
         .replace("<br/> ", "<br/>&nbsp;")
+        .replace("\n", "<br/>")
```
Tak, line 37: Do we need to inject the nonbreaking space here? Or we only get `\n` when it's a code block?

Tak (approved):

app/src/main/java/com/keylesspalace/tusky/util/StatusParsingHelper.kt
connyduck: When a browser renders HTML, it ignores newlines as a default. The Android HTML parsing code does the same.
In a browser, you can override that behavior with the [whitespace css property](https://developer.mozilla.org/en-US/docs/Web/CSS/white-space). Mastodon sets it for the whole status to `pre-wrap`. That difference did not matter, because Mastodon also replaces all newlines with `<br/>` so the Android parser would handle them correctly. But inside `<code>` blocks, newlines are not replaced by Mastodon, so that's what this replacement is for.

Tldr:
> we only get \n when it's a code block?

yes

PR #4115, opened 2023-11-20
feat: Change name of Preferences > Filters > Tabs and move them to Account Preferences(#3536)

### LINKED ISSUE #3536: Change name of Preferences > Filters > Tabs
Open Preferences and there's a section with the heading "Filters", and the single entry "Tabs".

This is confusing, and I've just figured out why.

1. These aren't filters in the normal Mastodon sense of the word. 
2. These aren't really "tab" options, they're timeline options.

Proposed concrete change:

1. Change the section title from "Filters" to "Per-timeline preferences"
2. Since the current "Tabs" entry is only for the Home timeline, change that to "Home timeline".

In the future, if other per-timeline options are needed (for example a "Show unread direct message count") then entries for those timelines can be added as additional sub-preference screens under here.

* * * *
- Tusky Version: 21

sanao1006:
# Overview
In the previous code, when you open preferences, there is a section headed "Filters" with a section called "Tabs"

This is confusing.

# Changes
- Change the section title from "Filters" to "Per-timeline preferences."
- Change the current "Tabs" section to "Home timeline" since it is only for home timelines

# Screenshots
account preference screen | detail screen
:--: | :--:
|<image src="https://github.com/tuskyapp/Tusky/assets/62137820/12694f24-b7e3-4ba3-90f5-53740e9c4269" width="250" />|<image src="https://github.com/tuskyapp/Tusky/assets/62137820/796e9ac1-76d6-43ef-a087-a1cd2d899ef8" width="250" />

# Note
- Maybe string resources should have a new property? (for translation)

# Related link
 Fixes #3536

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/settings/SettingsConstants.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/settings/SettingsConstants.kt
@@ -45,7 +45,7 @@ enum class AppTheme(val value: String) {
  *
  * - Adding a new preference that does not change the interpretation of an existing preference
  */
-const val SCHEMA_VERSION = 2023082301
+const val SCHEMA_VERSION = 2023112001
 
 /** The schema version for fresh installs */
 const val NEW_INSTALL_SCHEMA_VERSION = 0
--- a/app/src/main/java/com/keylesspalace/tusky/TuskyApplication.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/TuskyApplication.kt
@@ -146,6 +146,13 @@ class TuskyApplication : Application(), HasAndroidInjector {
                 editor.putString(APP_THEME, AppTheme.NIGHT.value)
             }
         }
+
+        if (oldVersion < 2023112001) {
+            editor.remove(PrefKeys.TAB_FILTER_HOME_REPLIES)
+            editor.remove(PrefKeys.TAB_FILTER_HOME_BOOSTS)
+            editor.remove(PrefKeys.TAB_SHOW_HOME_SELF_BOOSTS)
+        }
+
         editor.putInt(PrefKeys.SCHEMA_VERSION, newVersion)
         editor.apply()
     }
--- a/app/src/main/java/com/keylesspalace/tusky/components/preference/AccountPreferencesFragment.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/components/preference/AccountPreferencesFragment.kt
@@ -267,8 +267,7 @@ class AccountPreferencesFragment : PreferenceFragmentCompat(), Injectable {
                     preferenceDataStore = accountPreferenceDataStore
                 }
             }
-
-            preferenceCategory(R.string.pref_title_timeline_filters) {
+            preferenceCategory(R.string.pref_title_per_timeline_preferences) {
                 preference {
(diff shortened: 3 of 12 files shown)


app/src/main/java/com/keylesspalace/tusky/settings/SettingsConstants.kt
```diff
@@ -45,7 +45,7 @@ enum class AppTheme(val value: String) {
  *
  * - Adding a new preference that does not change the interpretation of an existing preference
  */
-const val SCHEMA_VERSION = 2023082301
+const val SCHEMA_VERSION = 2023112001
```
sanao1006, line 48: I don't know if this way of updating SharedPreferences is correct.

So, please point out if I am wrong!

sanao1006:
Are you possibly forgetting about this PR?
I appreciate if you can give me any response!

app/src/main/java/com/keylesspalace/tusky/settings/SettingsConstants.kt
connyduck: You did it correctly

connyduck (commented):
Very nice, thank you.

"Per-timeline preferences" is now shown twice on the account preferences which is a bit confusing. I think one of these should be renamed. Suggestion: Call the new one at the botton "local filters".

sanao1006:
Sorry, I didn't notice the "Per-timeline preferences" appearing at the top.

I thought the "Per-timeline preferences" at the top might be unnecessary rather than renaming it.

sanao1006:
Sorry, I retract my statement above. 
It is true that it would be better to rename it as you said!

sanao1006:
@connyduck 
How about setting the new one to "Per-timeline preferences" and the old one back to "Filters"?

sanao1006:
Fixed as shown in the image above.

connyduck (approved):

mcclure:
Can you explain to me the meaning of the second commit, "[update: Remove unnecessary sharedPreferences](https://github.com/tuskyapp/Tusky/pull/4115/commits/6d537dccbdc051a18047b2297b31ca0c0dc08c8e)"? Were these fields already or are they made irrelevant by this patch changing them to per-account values? Thanks.

Also, I have put up a PR #4128 , if that is accepted this PR will need to change its version number to acommodate it.

sanao1006:
> are they made irrelevant by this patch changing them to per-account values?

Yes, exactly.

> Also, I have put up a PR https://github.com/tuskyapp/Tusky/pull/4128 , if that is accepted this PR will need to change its version number to acommodate it.

In other words, this PR needs to update its DB version to 56 instead of 55, is that right?

mcclure:
Thanks for explaining.

> > Also, I have put up a PR #4128 , if that is accepted this PR will need to change its version number to acommodate it.
> 
> In other words, this PR needs to update its DB version to 56 instead of 55, is that right?

If the PR is accepted. (But there were no objections either here or on Matrix over a while so I guess it is getting accepted?)

mcclure:
Updated PR with 55->56 change as discussed. I tested this builds but did not test running it or performing an upgrade.

Sanao, this will probably cause your local copy to crash until you do an uninstall/reinstall cycle.

PR #4117, opened 2023-11-23
Regularize show/hide logic for video player scrub/play controls (fixes #4073)

### LINKED ISSUE #4073: Switching apps while video is playing leaves UI in inconsistent, unpleasant state
Perform the following steps

- Start a video playing.
- Hide the top (caption) and bottom (play controls) bars. (In the below video, I do this by waiting three seconds.)
- Switch to another app within Android.
- Switch back to Tusky.

Observed behavior: exactly ONE of the top and bottom bars will appear. If you tap, instead of dismissing as usual, it will toggle with each tap back and forth between exactly ONE of the top and bottom bars appearing, but it will switch back and forth. Once you are in this state the only way out is to either hit "back", or to wait three seconds for the bars to hide on their own.

This is clearly wrong, as well as frustrating.

This video is taken on Tusky 23.0 from the app store. However, I also see an identical problem on a self-built "develop" branch from today (with the exoplayer conversion).

Note: The video shows a screen recording of Tusky playing a video of a screen recording. This might be a little confusing.

https://github.com/tuskyapp/Tusky/assets/277318/ccf4ff6d-fcc4-4c19-83f5-310d70aa2544

mcclure:
When viewing a video in Tusky, there is a top toolbar where the description is shown and the bottom toolbar where play, forward, backward, and scrub controls are found. In both Tusky 23 and the new media3 video player code, the logic for showing these toolbars is *unrelated*; Tusky catches tap events and shows and hides the description, and the Android media library separately catches tap events and shows and hides the bottom toolbar. Meanwhile, Tusky and the Android media library each separately manage a set of logic for auto-hiding their respective toolbars after a certain number of seconds has passed. This all results in several problems:

- The top and bottom toolbars can desync, so that one is visible and the other is not, and tapping to show/hide after this will only swap which one is visible. This happens *every* time you switch to another application then back to Tusky while the video player is up.
- You can also desync the top and bottom toolbars in this way by simply tapping very rapidly.
- The autohide logic was difficult for us to control or customize, because it was partially hidden inside the Android libraries (relevant because under media3, the autohide delay increased from 3 to something like 5 or 6 seconds).

In this patch, I disabled all auto- and tap-based show/hide logic in media3 and set the Tusky-side show/hide to directly control the media3 toolbar. I then audited the code with printfs until I understood the state machine of show/hide, and removed anything irrational (some code was either unreachable, or redundant; either these lines were broken in the media3 transition, or they never worked).¹

While doing this, I made two policy changes:

- As discussed on Matrix, the autohide delay is now 4 seconds. (In discussions with users on Mastodon, some complained the previous 3 seconds was too short; but in my opinion and [I think?] charlag's, the new 5 seconds is too long).
- In the pre-existing code, if the user has hidden the controls, and they switch to another app and back, the controls display for 4 seconds then re-hide themselves, just like if the video had been presented for the first time. I think this is good and kept it— *however* I made a decision if the user intentionally taps to display the controls, *then* switches to another app and back, the controls should *not* auto-hide, because the user most recently requested those controls be shown.

Tests I performed on the final PR (successfully):

- Start video. Expect: toolbar+description hides after 4 seconds.
- Start video. Pause. Resume. Expect: t+d hides after 4 seconds.
- Start video. Wait 4 seconds until t+d hide. Switch to other app. Switch back. Expect: t+d reappears, then hides after 4 seconds.
- Start video. Wait 4 seconds until t+d hide. Tap to show t+d. Switch to other app. Switch back. Expect: t+d appear, do NOT autohide.
- Start video. Before 4 seconds up, switch to other app. Switch back. Expect: t+d reappears, then hides after 4 seconds.
- Start video. Pause. Resume. Before 4 seconds up, switch to other app. Switch back. Expect: t+d reappears, then hides after 4 seconds.
- Start video. Wait 4 seconds until t+d hide. Tap rapidly over and over for many seconds. Expect: Nothing weird
- Start *audio*. Expect: At no point does controller autohide, not even if I switch to another app and back, but I can hide it by manually tapping

These tests were performed on Android 13.  There is an entirely separate `Build.VERSION.SDK_INT <= 23` path I did not test, but Android Studio says this is dead code (I think it thinks our minimum SDK is higher than that?)

---

<small>¹ Incidentally, the underlying cause of #4073 (the show/resume part of it anyway) turned out to be that the STATE_READY event was being received not just on video load but also a second time on app resume, causing certain parts of the initialization code to run a second time although the fragment had already been fully initialized.</small>

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/fragment/ViewVideoFragment.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/fragment/ViewVideoFragment.kt
@@ -94,6 +94,15 @@ class ViewVideoFragment : ViewMediaFragment(), Injectable {
 
     private lateinit var mediaSourceFactory: DefaultMediaSourceFactory
 
+    /** Have we received at least one "READY" event? */
+    private var haveStarted = false
+
+    /** Is there a pending autohide? (We can't rely on Android's tracking because that clears on suspend.) */
+    private var pendingHideToolbar = false
+
+    /** Prevent the next play start from queueing a toolbar hide. */
+    private var suppressNextHideToolbar = false
+
     override fun onAttach(context: Context) {
         super.onAttach(context)
 
@@ -205,13 +214,27 @@ class ViewVideoFragment : ViewMediaFragment(), Injectable {
             override fun onPlaybackStateChanged(playbackState: Int) {
                 when (playbackState) {
                     Player.STATE_READY -> {
-                        // Wait until the media is loaded before accepting taps as we don't want toolbar to
-                        // be hidden until then.
-                        binding.videoView.setOnTouchListener(touchListener)
+                        if (!haveStarted) {
+                            // Wait until the media is loaded before accepting taps as we don't want toolbar to
+                            // be hidden until then.
+                            binding.videoView.setOnTouchListener(touchListener)
 
-                        binding.progressBar.hide()
-                        binding.videoView.useController = true
-                        binding.videoView.showController()
+                            binding.progressBar.hide()
+                            binding.videoView.useController = true
+                            binding.videoView.showController()
+                            haveStarted = true;
+                        } else {
+                            // This isn't a real "done loading"; this is a resume event after backgrounding.
+                            if (mediaActivity.isToolbarVisible) {
+                                // Before suspend, the toolbar/description were visible, so description is visible already.
+                                // But media3 will have automatically hidden the video controls on suspend, so we need to match the description state.
+                                binding.videoView.showController()
+                                if (!pendingHideToolbar)
+                                    suppressNextHideToolbar = true // The user most recently asked us to show the toolbar, so don't hide it when play starts.
+                            } else {
+                                mediaActivity.onPhotoTap()
+                            }
+                        }
                     }
                     else -> { /* do nothing */ }
                 }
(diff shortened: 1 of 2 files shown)


mcclure (commented):
I DIDN'T MEAN TO PRESS THIS BUTTON

app/src/main/java/com/keylesspalace/tusky/fragment/ViewVideoFragment.kt
```diff
@@ -95,6 +94,15 @@ class ViewVideoFragment : ViewMediaFragment(), Injectable {
...
     private lateinit var mediaSourceFactory: DefaultMediaSourceFactory
 
+    /** Have we received at least one "READY" event? */
+    private var haveStarted = false
+
+    /** Is there a pending autohide? (We can't rely on Android's tracking because that clears on suspend.) */
+    private var pendingHideToolbar = false
+
+    /** Prevent the next play start from queueing a toolbar hide. */
+    private var suppressNextHideToolbar = false
```
mcclure, line 104: `pendingHideToolbar` and `suppressNextHideToolbar` are flags set at a certain point in the state machine in order to get the behavior of "if you intentionally tap to display the controls, and then leave and re-enter the app, the controls autohide does not engage after four seconds". I don't like the "spooky action at a distance" nature of `suppressNextHideToolbar` but since the events media3 gives us are limited I think this is the only way to detect the relevant case was hit. If we decide the controls should *always* hide after four seconds even after a tap-to-show-then-suspend-then-resume, we can remove these two variables and their associated logic.

app/src/main/java/com/keylesspalace/tusky/fragment/ViewVideoFragment.kt
```diff
@@ -395,18 +422,24 @@ class ViewVideoFragment : ViewMediaFragment(), Injectable {
             })
             .start()
 
-        if (visible && (binding.videoView.player?.isPlaying == true) && !isAudio) {
-            hideToolbarAfterDelay(TOOLBAR_HIDE_DELAY_MS)
```
mcclure, line 399: Because there is a `hideToolbarAfterDelay `in the `onIsPlayingChanged`, we ALWAYS get an `onIsPlayingChanged`, and `onIsPlayingChanged` always happens last, we do not need any of the other `hideToolbarAfterDelay`s.

Tak (approved):
I love the level of commenting in your code!
This is something I need to be better about.

PR #4130, opened 2023-11-29
don't rehide filtered posts on interaction

### LINKED ISSUE #4125: Don't recollapse filter after every interaction
Filtered post recollapse to their filtered form after every interaction (like, boost, bookmark). This makes for a clunky experience and is genuinely just annoying as the recollapse feels unprompted.

In the vid below it happens in a hashtag feed, but it also happens in home and local, but *doesn't* happen in thread view. 

https://github.com/tuskyapp/Tusky/assets/109424320/2c82638c-d469-4c01-b35d-311ffde95751
* * * *
- Tusky Version: 24 beta 1
- Android Version: 12
- Android Device: Galaxy A12

connyduck:
closes #4125

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/db/TimelineDao.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/db/TimelineDao.kt
@@ -113,8 +113,7 @@ AND
             muted = status.muted,
             pinned = status.pinned ?: false,
             card = gson.toJson(status.card),
-            language = status.language,
-            filtered = status.filtered
+            language = status.language
         )
     }
 
@@ -139,8 +138,7 @@ AND
            muted = :muted,
            pinned = :pinned,
            card = :card,
-           language = :language,
-           filtered = :filtered
+           language = :language
            WHERE timelineUserId = :accountId AND (serverId = :statusId OR reblogServerId = :statusId)"""
     )
     @TypeConverters(Converters::class)
@@ -166,8 +164,7 @@ AND
         muted: Boolean?,
         pinned: Boolean,
         card: String?,
-        language: String?,
-        filtered: List<FilterResult>?
+        language: String?
     )
 
     @Query(


app/src/main/java/com/keylesspalace/tusky/db/TimelineDao.kt
```diff
@@ -113,8 +113,7 @@ AND
             muted = status.muted,
             pinned = status.pinned ?: false,
             card = gson.toJson(status.card),
-            language = status.language,
-            filtered = status.filtered
+            language = status.language
```
Lakoja, line 116: Hm, how does that solve the problem?

(How is "filtered" then stored?)

connyduck: Filtered contains the filters that apply to the post. When the post is revealed, we set it to null. On interaction (boost, fav, etc) it would set the filters again. Not updating the filters here means the posts stays revealed.

Lakoja: Should it also / still store the opened state once you have expanded something?

I tested this with:

- Expand a filtered post
- Fav it (ok, nothing bad happens)
- Reload timeline: the post is again filtered (collapsed)

connyduck: Yes it probably should. But looks like that would require a larger refactoring 🫣

Tak (approved):

app/src/main/java/com/keylesspalace/tusky/db/TimelineDao.kt
connyduck: I think we should merge this as it improves the situation and include it in the 24 release, and I will do the refactor later.

nehemiagurl: @Lakoja in my opinion the post *should* recollapse on timeline reload, just not on Interaction. I for example have a filter for Doctor Who to avoid spoilers, and on more than one occasion I forgot on Saturday there's a new episode up and opened the filter thinking it refers to the last one I saw. I would very much want the filter to recollapse when I reload the timeline in panic trying to avoid reading the post I've accidentally revealed.
maybe that's too extreme a position (I know recollapsing on reload doesn't happen for CWs for example) but I'd argue it justified. but then again, it's all matter of opinion.

PR #4153, opened 2023-12-08
prevent MainActivity from leaking through the DrawerImageLoader singleton

connyduck:
Another fix for a memory leak. This one is not as big as #4150, but still worth fixing for memory constrained devices imo.
The `DrawerImageLoader` implementation (a global singleton) references a member of the `MainActivity`, causing the whole activity to leak.

This weird construct was introduced in #1989 to fix a crash, but I think since we migrated to coroutines it is no longer necessary because all calls get correctly cancelled. I tried reproducing the crash but could not, so I'm pretty sure it is fine. I would appreciate it if someone else could try it as well though.

(The crash could be reproduced on slow internet, when `onFetchUserInfoSuccess` was called while the activity was being destroyed, causing Glide to crash the app because it can't use destroyed activities. `onFetchUserInfoSuccess` is now no longer called in this case because it is inside a `lifecycleScope.launch` block.)

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/MainActivity.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/MainActivity.kt
@@ -56,7 +56,6 @@ import androidx.preference.PreferenceManager
 import androidx.viewpager2.widget.MarginPageTransformer
 import at.connyduck.calladapter.networkresult.fold
 import com.bumptech.glide.Glide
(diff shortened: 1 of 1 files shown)


Tak (approved):
The changes look ok, I haven't tried to reproduce the crash yet

connyduck:
I did some more testing on another device, still can't reproduce. Merging so nightly testers get to use it.

PR #4160, opened 2023-12-10
Prevent device from dimming or sleeping screen while video/audio playing

mcclure:
We got reports on Mastodon that the media3 upgrade broke this previously functional behavior. In my testing, this restores the v23 behavior.

Note, @Gryzor suggested we do this with a Wake Lock. After looking at the docs, I conclude that (1) a wake lock is not what v23 was using, and is not appropriate for video but (2) a wake lock *is* appropriate for audio and possibly other purposes, and we should consider adding one in the v25 timeframe. I will file a followup issue explaining what I mean by this later today.

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/ViewMediaActivity.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/ViewMediaActivity.kt
@@ -158,8 +158,9 @@ class ViewMediaActivity : BaseActivity(), HasAndroidInjector, ViewImageFragment.
         })
 
         // Prevent this activity from dimming or sleeping the screen if it is playing video or audio
-        if (attachments!![binding.viewPager.currentItem].attachment.type != Attachment.Type.IMAGE)
+        if (attachments!![binding.viewPager.currentItem].attachment.type != Attachment.Type.IMAGE) {
             window.addFlags(WindowManager.LayoutParams.FLAG_KEEP_SCREEN_ON)
+        }
     }
 
     override fun onCreateOptionsMenu(menu: Menu): Boolean {
--- a/app/src/main/java/com/keylesspalace/tusky/ViewMediaActivity.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/ViewMediaActivity.kt
@@ -35,6 +35,7 @@ import android.util.Log
 import android.view.Menu
 import android.view.MenuItem
 import android.view.View
+import android.view.WindowManager
 import android.webkit.MimeTypeMap
 import android.widget.Toast
 import androidx.core.app.ShareCompat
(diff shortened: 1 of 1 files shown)


connyduck:
Shouldn't the screen stay on _only_ when a video is playing?

mcclure:
> Shouldn't the screen stay on _only_ when a video is playing?

Yes, it should. However, it turns out keeping the screen on is simple, and sleeping the screen but continuing to play audio is complicated. That is why I propose we merge this for 24.1 (restoring the 23.0 behavior) and then merge a bigger fix in 25.0. I have a full explanation in the issue [here](https://github.com/tuskyapp/Tusky/issues/4161) (it is complicated).

(If issue #4161 turns out to be easier to fix than I thought we can move that forward to 24.1 as well, of course.)

mcclure:
Update see #4162, my unsuccessful WIP for setWakeMode.

If we cannot get setWakeMode working on the 24.1 timeframe, I do believe it is better behavior to allow the audio to continue playing while keeping the screen on, than to allow the audio to cut off after 1 minute when the screen naturally sleeps.

Gryzor:
@mcclure Agree. I think the Wake Look is no longer "used" for the behavior you want here, and instead the keep screen on is "the way". This shows you how long it has been since I had to deal with one! ;)

Tak (approved):

Tak:
I agree that the pre-24 behavior was "good enough", and imo we shouldn't invest too much work in letting the screen blank while continuing to play audio (unless it's a personal interest, of course) - Tusky isn't a music player or podcast app or whatever

connyduck:
Ok so it is ok for you all that the screen will stay on when imges are shown? Then I will merge this. (I personally think it is weird and could bring us more complaining users).

mcclure:
> Ok so it is ok for you all that the screen will stay on when imges are shown? Then I will merge this. (I personally think it is weird and could bring us more complaining users).

The screen does not stay on when images are shown, see [here](https://github.com/tuskyapp/Tusky/pull/4160/files#diff-8a85b21a2327cff1933b3cf9c403daa3e87611ac32dcfcec9cf0891358a7f270R161) (and I did test this). We only request KEEP_SCREEN_ON for audio, video, gifv.

No one complained about always-on audio in v23, so I think it is ok for now.

connyduck:
But what if you swipe to an image? Shouldn't the flag be removed then? Or does it have no effect once it is set?

mcclure:
> But what if you swipe to an image? Shouldn't the flag be removed then? Or does it have no effect once it is set?

That's… a good question.
One moment, I will test.

mcclure:
Did @connyduck's test and can confirm, if you find a post with both an image and a video and swipe between them, it does the wrong thing.

If you start on an image and swipe to a video it will sleep (incorrectly), if you start on a video and swipe to an image it will not sleep (incorrectly).

I think this is worth holding off this patch for, I need to go to bed now but will attempt to fix the swipe bug tomorrow.

mcclure:
Github what the hell this was not merged

mcclure:
Sigh new PR #4168

Issue #4161, opened 2023-12-10
Feature request: Use WakeLock when playing audio


mcclure:
In v23, we used the default Android video player. While video or audio were playing with this player, the device would not dim or sleep, it just stayed on. Exoplayer, by default, does not do this; in v24 either audio or video will dim, then switch off, when they are left playing. This is a regression.

The documentation here: https://developer.android.com/training/scheduling/wakelock explainshttps://developer.android.com/training/scheduling/wakelock explains that there are several ways of keeping a Android device "awake", depending on what you want to do:

* There is a FLAG_KEEP_SCREEN_ON on Activities, which will keep the screen lit and also prevent sleeping; it does not require any special permissions.
* There is also WakeLock, which comes in several forms and *does* require a special permission. If we want to use a WakeLock, the way we should do this is using Exoplayer [setWakeMode](https://developer.android.com/reference/androidx/media3/exoplayer/ExoPlayer.Builder?hl=en#setWakeMode(int)). This comes in two flavors, each of which will require requesting an additional permission:
    * WAKE_MODE_LOCAL, which allows the screen to dim and then sleep but which keeps the CPU on. The documentation says: "It should be used together with a foreground [android.app.Service](https://developer.android.com/reference/android/app/Service.html) for use cases where playback occurs and the screen is off (e.g. background audio playback)."
    * WAKE_MODE_NETWORK, which will prevent the CPU from sleeping *and* will [prevent the wifi from sleeping](https://developer.android.com/reference/android/net/wifi/WifiManager.WifiLock.html?hl=en).

In my opinion, what we want is to use FLAG_KEEP_SCREEN_ON when playing *video* and use WAKE_MODE_LOCAL when we are playing *audio*. But there are two things I am uncertain about:

* Do we want WAKE_MODE_NETWORK? It seems like this is mainly for players that *stream* audio; we want WAKE_MODE_NETWORK when we are initially downloading the audio, but afterward we will want to relinquish it. I am not sure if this is possible and I don't know if there is something else we were already doing which will prevent device/network sleep while we are loading a video.
* What are they saying about "a foreground Service"? Does this mean that if we want to continue playing audio after screen sleep (because, I guess, at that point even if the CPU is still active, our activity suspends?) we will need to pass video playback off to a service? Can we use the service we already have (for network stuff etc) for this purpose?

[...]

PR #4162, opened 2023-12-11
Correct backgrounding behavior for both video and audio (does not work)

mcclure:
See #4161 for full explanation.

This is a followup to #4160. It causes keep-screen-alive to *only* occur for videos (and GIFVs), and for audio attachments it uses [setWakeMode](https://developer.android.com/reference/androidx/media3/exoplayer/ExoPlayer.Builder?hl=en#setWakeMode(int)) (with the required permission added to AndroidManifest.xml). This patch does not work. I can confirm we are hitting the setWakeMode branch\*, but when the screen goes to sleep the audio also stops. The documentation says that setWakeMode "should be used together with a foreground [android.app.Service](https://developer.android.com/reference/android/app/Service.html) for use cases where playback occurs and the screen is off (e.g. background audio playback)", and we have a foreground service but there is no specific provision to switch the audio playback from the activity to the foreground service. I do not know what to do from here.

If the foreground service could assume audio playback, would that mean we could allow audio attachments to keep playing even after the user switches to another app? I think that would be pretty neat.

\* I added a log.d to the setWakeMode if, and it printed. I looked in the logcat and there were no warnings around the setWakeMode indicating a permission denial or anything. One minute later I see in the log:

`ActivityTrigger activityPauseTrigger`

I then immediately see a bunch of audio-related messages I can't decipher, such as:

```
2023-12-10 20:42:26.887  6141-6315  Avrcp_ext               pid-6141                             D  AudioManager Player: AudioPlaybackConfiguration piid:55 deviceId:0 type:android.media.SoundPool u/pid:1000/1445 state:idle attr:AudioAttributes: usage=USAGE_ASSISTANCE_SONIFICATION content=CONTENT_TYPE_SONIFICATION flags=0x800 tags= bundle=null sessionId:0
2023-12-10 20:42:26.887  6141-6315  Avrcp_ext               pid-6141                             D  AudioManager Player: AudioPlaybackConfiguration piid:71 deviceId:0 type:android.media.SoundPool u/pid:1027/4928 state:idle attr:AudioAttributes: usage=USAGE_ASSISTANCE_SONIFICATION content=CONTENT_TYPE_SONIFICATION flags=0x800 tags= bundle=null sessionId:0
2023-12-10 20:42:26.887  6141-6315  Avrcp_ext               pid-6141                             D  AudioManager Player: AudioPlaybackConfiguration piid:503 deviceId:3 type:android.media.AudioTrack u/pid:10749/14450 state:paused attr:AudioAttributes: usage=USAGE_MEDIA content=CONTENT_TYPE_UNKNOWN flags=0xA00 tags= bundle=null sessionId:633
2023-12-10 20:42:26.887  6141-6315  Avrcp_ext               pid-6141                             D  AudioManager Player: AudioPlaybackConfiguration piid:63 deviceId:0 type:android.media.SoundPool u/pid:10303/2655 state:idle attr:AudioAttributes: usage=USAGE_ASSISTANCE_SONIFICATION content=CONTENT_TYPE_SONIFICATION flags=0x800 tags= bundle=null sessionId:0
2023-12-10 20:42:26.887  6141-6315  Avrcp_ext               pid-6141                             D  AudioManager isPlaying: false, mAudioPlaybackIsActive = true
2023-12-10 20:42:26.887  6141-6315  Avrcp_ext               pid-6141                             D  AudioManager Reset Active Player
2023-12-10 20:42:26.887  6141-6315  Avrcp_ext               pid-6141                             V  updateCurrentMediaState: mMediaController: null
2023-12-10 20:42:26.888  6141-6315  Avrcp_ext               pid-6141                             W  isMusicActive: true getBluetoothPlayState: 1 A2dp State: 11 mAudioPlaybackIsActive: false
2023-12-10 20:42:26.888  6141-6315  Avrcp_ext               pid-6141                             W  updateCurrentMediaState: isPlaying = true
2023-12-10 20:42:26.888  6141-6315  Avrcp_ext               pid-6141                             V  Media update: id [MediaAttributes: none] : [MediaAttributes: none]
2023-12-10 20:42:26.888  6141-6315  Avrcp_ext               pid-6141                             D  isPackageNameValid: browsedPackage = isValid = false
2023-12-10 20:42:26.888  6141-6315  Avrcp_ext               pid-6141                             V  updatePlaybackState, state: PlaybackState {state=3, position=-1, buffered position=0, speed=1.0, updated=96108045, actions=0, custom actions=[], active item id=-1, error=null} device: null
```
There were more similar messages after this but I didn't catch them all before they cleared. Various objects claiming both that the music is playing and is not playing.

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -8,6 +8,7 @@
     <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE" android:maxSdkVersion="28"/>
     <uses-permission android:name="android.permission.VIBRATE" /> <!-- For notifications -->
     <uses-permission android:name="android.permission.FOREGROUND_SERVICE" />
+    <uses-permission android:name="android.permission.WAKE_LOCK" /> <!-- For media player -->
 
     <application
         android:name=".TuskyApplication"
--- a/app/src/main/java/com/keylesspalace/tusky/ViewMediaActivity.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/ViewMediaActivity.kt
@@ -158,8 +158,9 @@ class ViewMediaActivity : BaseActivity(), HasAndroidInjector, ViewImageFragment.
         })
 
         // Prevent this activity from dimming or sleeping the screen if it is playing video or audio
-        if (attachments!![binding.viewPager.currentItem].attachment.type != Attachment.Type.IMAGE)
+        if (attachments!![binding.viewPager.currentItem].attachment.type != Attachment.Type.IMAGE) {
             window.addFlags(WindowManager.LayoutParams.FLAG_KEEP_SCREEN_ON)
+        }
     }
 
     override fun onCreateOptionsMenu(menu: Menu): Boolean {
--- a/app/src/main/java/com/keylesspalace/tusky/ViewMediaActivity.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/ViewMediaActivity.kt
@@ -159,7 +159,7 @@ class ViewMediaActivity : BaseActivity(), HasAndroidInjector, ViewImageFragment.
 
         // Prevent this activity from dimming or sleeping the screen if it is playing video
         // See also WAKE_MODE in ViewVideoFragment.kt
-        val attachmentType = attachments!![binding.viewPager.currentItem].attachment.type;
+        val attachmentType = attachments!![binding.viewPager.currentItem].attachment.type
         if (attachmentType == Attachment.Type.VIDEO || attachmentType == Attachment.Type.GIFV) {
             window.addFlags(WindowManager.LayoutParams.FLAG_KEEP_SCREEN_ON)
         }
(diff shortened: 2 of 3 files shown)


mcclure:
Feel free, I am no longer looking at this as I do not have background with the parts that need to be connected.

cbeyls:
This pull request is now outdated and should either be closed or rewritten.

Here are a few updates:

- `ViewMediaActivity` now contains a method `adjustScreenWakefulness()` which sets `FLAG_KEEP_SCREEN_ON` when the content is audio, gif or video (anything but image). So the screen is properly left on during video playback.
- ExoPlayer and media3 are able to handle the wake locks at the player level using `setWakeMode()`, but this is currently disabled because wake locks need to work in combination with a foreground service to ensure the OS will not kill the app during background playback. Since currently the audio playback only works from a regular activity in the foreground and background audio playback is not officially supported yet, then no wake locks are needed. A future pull request may be created to add proper support for background audio playback using both wake lock and a foreground service.

PR #4168, opened 2023-12-12
Prevent device from dimming or sleeping screen while video/audio playing (but really)

mcclure:
For reasons not totally clear to me, Github marked #4160 as "merged" and will not let me reopen it.

I believe this should be included for 24.1 because it fixes a 24.0 regression in the media player.

I have tested this newest commit in a number of ways, and in my testing I find (1) when viewing an image, it sleeps after about a minute (2) when viewing video, it stays awake indefinitely (3) this is true whether the image/video was opened directly, or reached by swiping from another attachment. I have not tested swiping to/from audio but I am confident it will work the same.


connyduck (approved):

charlag (commented):
will these flags be reset when we leave the activity?

Tak (approved):

connyduck:
> will these flags be reset when we leave the activity?

Yes each activity has their own window

PR #4169, opened 2023-12-13
upgrade ktlint plugin to 12.0.3

connyduck:
There are some new rules, I think they mostly make sense, except for the max line length which I had to disable because we are over it in a lot of places.


.editorconfig
```diff
@@ -12,8 +12,14 @@ trim_trailing_whitespace = true
 ij_kotlin_name_count_to_use_star_import = 999
 ij_kotlin_name_count_to_use_star_import_for_members = 999
 ij_java_class_count_to_use_import_on_demand = 999
+
+ktlint_code_style = android_studio
+
 # Enable trailing comma
```
Goooler, line 18:
```suggestion
# Disable trailing comma
```

Goooler (approved):
And we can add `ij_kotlin_imports_layout = *`.

.editorconfig
```diff
@@ -8,12 +8,20 @@ insert_final_newline = true
...
 ij_kotlin_name_count_to_use_star_import_for_members = 999
 ij_java_class_count_to_use_import_on_demand = 999
-# Enable trailing comma
-ktlint_disabled_rules=trailing-comma-on-call-site,trailing-comma-on-declaration-site
+
+ktlint_code_style = android_studio
+
+# Disable trailing comma
+ktlint_standard_trailing-comma-on-call-site = disabled
+ktlint_standard_trailing-comma-on-declaration-site = disabled
```
Goooler, line 22:
```suggestion
ij_kotlin_allow_trailing_comma = false
ij_kotlin_allow_trailing_comma_on_call_site = false
```

Goooler: We can use these instead.

https://pinterest.github.io/ktlint/latest/rules/standard/#trailing-comma-on-declaration-site
https://pinterest.github.io/ktlint/latest/rules/standard/#trailing-comma-on-call-site

.editorconfig
```diff
@@ -8,12 +8,20 @@ insert_final_newline = true
...
 ij_kotlin_name_count_to_use_star_import_for_members = 999
 ij_java_class_count_to_use_import_on_demand = 999
-# Enable trailing comma
-ktlint_disabled_rules=trailing-comma-on-call-site,trailing-comma-on-declaration-site
+
+ktlint_code_style = android_studio
+
+# Disable trailing comma
+ij_kotlin_allow_trailing_comma = false
+ij_kotlin_allow_trailing_comma_on_call_site = false
```
Tak, line 22: Again? 😢

connyduck: ?

Tak: Does this disallow *usage* of trailing commas, or disable *forced* usage of trailing commas?

Goooler: I would recommend disabling or enabling it forcibly, instead of using either state.

Tak: Yes, we want no warning either way.

I understand the urge to force it one way or the other, but it is *really* not worth it.

Tak (approved):

PR #4172, opened 2023-12-14
fix crash when opening profile image

connyduck:
```
java.lang.RuntimeException: Unable to start activity ComponentInfo{com.keylesspalace.tusky.test/com.keylesspalace.tusky.ViewMediaActivity}: java.lang.NullPointerException
	at android.app.ActivityThread.performLaunchActivity(ActivityThread.java:2778)
	at android.app.ActivityThread.handleLaunchActivity(ActivityThread.java:2856)
	at android.app.ActivityThread.-wrap11(Unknown Source:0)
	at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1589)
	at android.os.Handler.dispatchMessage(Handler.java:106)
	at android.os.Looper.loop(Looper.java:164)
	at android.app.ActivityThread.main(ActivityThread.java:6494)
	at java.lang.reflect.Method.invoke(Native Method)
	at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:438)
	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:807)
Caused by: java.lang.NullPointerException
	at com.keylesspalace.tusky.ViewMediaActivity.adjustScreenWakefulness(ViewMediaActivity.kt:351)
	at com.keylesspalace.tusky.ViewMediaActivity.onCreate(ViewMediaActivity.kt:161)
	at android.app.Activity.performCreate(Activity.java:7009)
	at android.app.Activity.performCreate(Activity.java:7000)
	at android.app.Instrumentation.callActivityOnCreate(Instrumentation.java:1214)
	at android.app.ActivityThread.performLaunchActivity(ActivityThread.java:2731)
	at android.app.ActivityThread.handleLaunchActivity(ActivityThread.java:2856) 
	at android.app.ActivityThread.-wrap11(Unknown Source:0) 
	at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1589) 
	at android.os.Handler.dispatchMessage(Handler.java:106) 
	at android.os.Looper.loop(Looper.java:164) 
	at android.app.ActivityThread.main(ActivityThread.java:6494) 
	at java.lang.reflect.Method.invoke(Native Method) 
	at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:438) 
	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:807) 
```

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/ViewMediaActivity.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/ViewMediaActivity.kt
@@ -348,10 +348,12 @@ class ViewMediaActivity : BaseActivity(), HasAndroidInjector, ViewImageFragment.
 
     // Prevent this activity from dimming or sleeping the screen if, and only if, it is playing video or audio
     private fun adjustScreenWakefulness() {
-        if (attachments!![binding.viewPager.currentItem].attachment.type == Attachment.Type.IMAGE) {
-            window.clearFlags(WindowManager.LayoutParams.FLAG_KEEP_SCREEN_ON)
-        } else {
-            window.addFlags(WindowManager.LayoutParams.FLAG_KEEP_SCREEN_ON)
+        attachments?.run {
+            if (get(binding.viewPager.currentItem).attachment.type == Attachment.Type.IMAGE) {
+                window.clearFlags(WindowManager.LayoutParams.FLAG_KEEP_SCREEN_ON)
+            } else {
+                window.addFlags(WindowManager.LayoutParams.FLAG_KEEP_SCREEN_ON)
+            }
         }
     }
 


mcclure (approved):
Sorry about that!! :(

However: I'm very concerned that this code is running outside of media tabs at all. What circumstances cause that to happen? Does this imply that the timeline feed could keep the screen awake if there is something animated on it? If so, that would be a serious regression.

Tak (approved):

connyduck:
> However: I'm very concerned that this code is running outside of media tabs at all. What circumstances cause that to happen? Does this imply that the timeline feed could keep the screen awake if there is something animated on it? If so, that would be a serious regression.

The problem here was that `ViewMediaActivity` works a bit different whn it shows profile images, as most of the other logic is not needed then. Maybe it is not ideal that it is used for regular media and profile images, but there is no need to be concerned it is used only for those things.

mcclure:
> The problem here was that `ViewMediaActivity` works a bit different whn it shows profile images, as most of the other logic is not needed then. Maybe it is not ideal that it is used for regular media and profile images, but there is no need to be concerned it is used only for those things.

Ohh, profile images as in the headers. I was confused, now I see how to trigger this. Thanks.

I am now convinced the behavior is okay, but I would like to suggest we add a comment in `adjustScreenWakefulness` explaining the reason "attachments" must be checked for nullity is in case someone has tapped on a profile image. That will make the intent clearer to the next person who edits the code.

PR #4182, opened 2023-12-18
fix ANR caused by direct message badge

connyduck:
Saw an ANR (app not responding) error being reported in the Play console and then found this. Sorry but `runBlocking` in production code is an absolute no go.


charlag (approved):
seems okay from a quick glance, I can't tell if it can lead to any races right now

connyduck:
tested this again, works fine

PR #4194, opened 2023-12-26
Add Mastodon Redirect support

zacharee:
Mastodon Redirect is my app that handles redirecting Mastodon links to supported apps, so the apps don't have to maintain their own list of supported instances.

This PR adds auto-detect support for Mastodon Redirect, so Tusky can be chosen for opening Mastodon links.

To test:
1. Download Mastodon Redirect from https://github.com/zacharee/MastodonRedirect.
2. Install Tusky from this PR's code.
3. Open and log into Tusky.
4. Find a random post and copy its link. Paste the link somewhere it'll be clickable, such as in Telegram's "Saved messages" channel.
5. Remove Tusky from Recents.
6. Open Mastodon Redirect and select Tusky in the "Discovered" section.
7. Tap the pasted link. It should launch Tusky and then resolve to the post.
8. Back out to the main feed in Tusky and go back to the link, without removing Tusky from Recents this time. The link should open Tusky and resolve to the post.


app/src/main/AndroidManifest.xml
```diff
@@ -55,6 +55,7 @@
         <activity
             android:name=".MainActivity"
             android:configChanges="orientation|screenSize|keyboardHidden|screenLayout|smallestScreenSize"
+            android:launchMode="singleTask"
```
zacharee, line 58: I needed to add this so Tusky could receive links while it's still running in the background. I don't think it breaks anything, but let me know.

app/src/main/java/com/keylesspalace/tusky/util/LinkHelper.kt
```diff
@@ -281,6 +281,10 @@ fun Context.openLink(url: String) {
  */
 private fun openLinkInBrowser(uri: Uri?, context: Context) {
     val intent = Intent(Intent.ACTION_VIEW, uri)
+
+    // Makes sure the Intent opens in the browser instead of something like Mastodon Redirect.
+    intent.selector = Intent(Intent.ACTION_VIEW, Uri.parse("https://"))
```
zacharee, line 286: A blank-URL selector guarantees the link won't accidentally open in an app that supports the domain (e.g., Mastodon Redirect, which could cause a launch loop).

connyduck:
I fail to test that.
My steps:
- use an Android Api 34 emulator
- Install Tusky from this branch
- Install Mastodon redirect 1.11.0
- make Android open links from a few domains with Mastodon redirect
- enable Tusky in Mastodon redirect
- click a link from one of the enabled domains in Chrome
- Chrome closes

```
Failed to deliver inset control state change to w=Window{bb5153c u0 com.android.chrome/com.google.android.apps.chrome.Main EXITING}
android.os.DeadObjectException
```

similar result on a real device and with Firefox

zacharee:
Try using the "Open Link" share target instead. 

Mastodon Redirect has so many domains that it can break whatever API Chrome is using to read where to open links. I've been waiting on an update to instances.social's API for better active-status filtering.

connyduck:
Same result with "open link" :( 
Seems like Android is unable to cope with an app having so many supported domains

zacharee:
Try revoking any links you enabled for auto-handling. You might also need to try opening a link from something like Keep or Telegram. I want to reduce the number of handled domains in Mastodon Redirect, but I haven't been able to find a better source yet.

zacharee:
I've reduced the number of supported domains by about 1000 in Mastodon Redirect. It's still over 11k total, but it should be less prone to crashing.

zacharee:
I also added a new share target for opening links, which uses the same path Mastodon Redirect does.

tiptoptom:
Any news here? 🙈

Lazalatin:
@zacharee This PR needs to be rebased or have a merge commit from tuskyapp:develop. If you can not afford time for solving this, I can gladly take over. (no tusky maintainer, just want to help getting this done :) )

tiptoptom:
@lazalatin can you bring that forward?

PR #4205, opened 2024-01-04
Move the statistics bar in the detailed view below the button bar

Tak:
Rationale: In the current layout, when performing multiple interactions (e.g. fav+boost, fav+reply) on a post that hasn't been interacted with before, the statistics bar appears and pushes the buttons down, so the second tap goes to the statistics bar instead


connyduck (changes requested):
Very good idea!

- Please move the `status_reblogs` and `status_favourites` down into their new correct position in xml.
- Some margin changes are needed so both the buttons and the boost/ fav count are vertically centered between the dividers

Tak:
Finally got back to this, sorry for the delay

connyduck (approved):

PR #4224, opened 2024-01-09
full sdk 34 support

### LINKED ISSUE #4005: Upgrade targetSdk to 34 (Android 14) and support predictive back

connyduck:
builds upon work from #4082 

Additionally fixes some deprecations and adds support for [predictive back](https://developer.android.com/guide/navigation/custom-back/predictive-back-gesture). I also refactored how the activity transitions work because they are closely related to predictive back. The awkward `finishWithoutSlideOutAnimation` is gone, activities that have been started with slide in will now automatically close with slide out.

closes #4082 
closes #4005


charlag (approved):
I did not test it on new SDK but on my Android 13 seems to work just fine.

I still think we need to change our transitions, vanilla masto does it well, our is *only* slide without alpha, takes 300ms and uses weird interpolation.

Tak (approved):
Seems ok here as well? (but I only have android 11)

PR #4249, opened 2024-02-13
fix compose notification action not recreating MainActivity

### LINKED ISSUE #4247: Opening Compose from Notification for Different Account Corrupts Timeline
I have two accounts setup on Tusky 24.1 and I accidentally hit "compose" instead of "quick reply" on a notification I had from the second account while the first was open in Tusky. This caused both the toot I was responding to and the toot I authred to show up in the wrong timeline.

The events happen as follows:
- Open Tusky to account A
- Open notification drawer
- Tap on "compose" for a notification from account B
- Compose and send toot
- Account A's timeline gets corrupted with content from account B

* * * *
- Tusky Version: 24.1
- Android Version: U
- Android Device: Pixel Fold

connyduck:
closes #4247

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/components/notifications/NotificationHelper.java
+++ b/app/src/main/java/com/keylesspalace/tusky/components/notifications/NotificationHelper.java
@@ -456,6 +456,7 @@ public class NotificationHelper {
 
         Intent composeIntent = MainActivity.composeIntent(context, composeOptions, account.getId(), body.getId(), (int)account.getId());
 
+        // make sure a new instance of MainActivity is started and old ones get destroyed
         composeIntent.addFlags(Intent.FLAG_ACTIVITY_NEW_TASK | Intent.FLAG_ACTIVITY_CLEAR_TASK);
 
         return PendingIntent.getActivity(context.getApplicationContext(),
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/keylesspalace/tusky/components/notifications/NotificationHelper.java
```diff
@@ -456,7 +456,7 @@ private static PendingIntent getStatusComposeIntent(Context context, Notificatio
 
         Intent composeIntent = MainActivity.composeIntent(context, composeOptions, account.getId(), body.getId(), (int)account.getId());
 
-        composeIntent.addFlags(Intent.FLAG_ACTIVITY_NEW_TASK);
+        composeIntent.addFlags(Intent.FLAG_ACTIVITY_NEW_TASK | Intent.FLAG_ACTIVITY_CLEAR_TASK);
```
charlag, line 459: I'd leave a comment in code why we do this

charlag (approved):

app/src/main/java/com/keylesspalace/tusky/components/notifications/NotificationHelper.java
Tak: ➕

Tak (approved):

PR #4250, opened 2024-02-13
fix quick replies from notifications

connyduck:
While working on #4249 I noticed that quick replies also don't work as expected. The notification just stays in the sending state forever.
There are actually 2 problems:
- Notifications are sent in `NotificationFetcher` with the id of the Mastodon notification as tag and the current account id as id. The wrong notification id was forwarded to `SendStatusBroadcastReceiver` so it never had a chance of updating the notification.
- Notifications containing an active remote input can't be cancelled (they just stop their animation when doing so). So instead I update the notification with info that the reply is being sent and have it dismiss automatically.

I also tried replacing the original notification with the "sending" notification of `SendStatusService`, but that doesn't work because `Service.startForeground` doesn't have a tag parameter, only an id.

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/receiver/SendStatusBroadcastReceiver.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/receiver/SendStatusBroadcastReceiver.kt
@@ -45,7 +45,7 @@ class SendStatusBroadcastReceiver : BroadcastReceiver() {
         AndroidInjection.inject(this, context)
 
         if (intent.action == NotificationHelper.REPLY_ACTION) {
-            val notificationId = intent.getStringExtra(NotificationHelper.KEY_NOTIFICATION_ID)
+            val serverNotificationId = intent.getStringExtra(NotificationHelper.KEY_SERVER_NOTIFICATION_ID)
             val senderId = intent.getLongExtra(NotificationHelper.KEY_SENDER_ACCOUNT_ID, -1)
             val senderIdentifier = intent.getStringExtra(
                 NotificationHelper.KEY_SENDER_ACCOUNT_IDENTIFIER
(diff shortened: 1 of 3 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -227,8 +227,8 @@
     <string name="confirmation_domain_unmuted">%s unhidden</string>
     <string name="confirmation_hashtag_unfollowed">#%s unfollowed</string>
 
-    <string name="post_sent">Sent!</string>
-    <string name="post_sent_long">Reply sent successfully.</string>
+    <string name="reply_sending">Sending...</string>
```
charlag, line 230:
```suggestion
    <string name="reply_sending">Sending…</string>
```

charlag (approved):

app/src/main/java/com/keylesspalace/tusky/receiver/SendStatusBroadcastReceiver.kt
```diff
@@ -114,25 +113,25 @@ class SendStatusBroadcastReceiver : BroadcastReceiver() {
...
-                notificationManager.cancel(notificationId)
+                    .setDefaults(0) // We don't want this to make any sound or vibration
+                    .setOnlyAlertOnce(true)
+                    .setContentTitle(context.getString(R.string.reply_sending))
+                    .setContentText(context.getString(R.string.reply_sending_long))
+                    .setSubText(senderFullName)
+                    .setVisibility(NotificationCompat.VISIBILITY_PUBLIC)
+                    .setCategory(NotificationCompat.CATEGORY_SOCIAL)
+                    .setTimeoutAfter(5000)
+                    .build()
```
Tak, line 132: Should this maybe be extracted?

Tak (approved):

app/src/main/java/com/keylesspalace/tusky/receiver/SendStatusBroadcastReceiver.kt
connyduck: I was thinking about it, it is similar to other places where we build notifications, but not enough to be easily extracted. I think its easier to read like it is.

PR #4251, opened 2024-02-13
set greenDebug as default flavor

connyduck:
I had it happen multiple times recently that I was testing green Tusky but Android Studio actually put blue Tusky on my device and I wasted a lot of time until I found out 😣
This change should tell it that greenDebug is the preferred flavor for developing.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -41,6 +41,9 @@ android {
         buildConfigField("String", "SUPPORT_ACCOUNT_URL", "\"$SUPPORT_ACCOUNT_URL\"")
     }
     buildTypes {
+        debug {
+            getIsDefault().set(true)
+        }
         release {
             minifyEnabled true
             shrinkResources true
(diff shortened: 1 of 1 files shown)


app/build.gradle
```diff
@@ -41,6 +41,9 @@ android {
         buildConfigField("String", "SUPPORT_ACCOUNT_URL", "\"$SUPPORT_ACCOUNT_URL\"")
     }
     buildTypes {
+        debug {
+            getIsDefault().set(true)
```
Goooler, line 45:
```suggestion
            isDefault true
```

app/build.gradle
```diff
@@ -55,6 +58,7 @@ android {
             resValue "string", "app_name", APP_NAME + " Test"
             applicationIdSuffix ".test"
             versionNameSuffix "-" + gitSha
+            getIsDefault().set(true)
```
Goooler, line 61:
```suggestion
            isDefault true
```

Goooler (approved):

mcclure (approved):

PR #4260, opened 2024-02-21
Improve detailed status looks

charlag:
#4205 did change how the counters for the detailed posts behave and for a good reason I believe.

However I find the changed order very confusing and not aesthetically pleasing.

I have tried a few options, including reserving space for it but it was confusing (when counters are not displayed there would be a danging separator or if we show separator together with it it would be confusing as well).

I propose we simply show the counters independent on the counts. I know we try to de-emphasize the counters but I believe this is fine to do in detailed view.

One disadvantage is that we need translators to update the translations.

Additionally I've done two spacing changes: I removed a separator between the counters and the buttons, removed padding around the counters and increased the space between the counters and the buttons instead. I believe it's better to use space than separators. This also makes the space above/below the media/counters separator balanced.

In the second commit I've also made the metadata/counters separators thinner, I think it looks better.

here's the combined version:

proposal_final


charlag:
I see a lot of indentation changes, not sure why I have different code style set

Tak (approved):
Looks fine to me, I just don't want the buttons to move 🙂

PR #4285, opened 2024-02-24
Better screen transitions

charlag:
I mostly took Android 13 transitions and removed the sliding for the "deeper"/background one because "extend" animations are not available until Android 13.

Here are the original ones: https://cs.android.com/android/platform/superproject/+/android-13.0.0_r8:frameworks/base/core/res/res/anim/;bpv=1

Initially I've made separate versions fro Android 13+ that are close to the original but I think it's not worth it to keep both.

https://github.com/tuskyapp/Tusky/assets/3099142/616fc40c-f944-45b4-bf6f-167f62d30493

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/BaseActivity.java
+++ b/app/src/main/java/com/keylesspalace/tusky/BaseActivity.java
@@ -60,6 +60,7 @@ import java.util.List;
 import javax.inject.Inject;
 
 import static com.keylesspalace.tusky.settings.PrefKeys.APP_THEME;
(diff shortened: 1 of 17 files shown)


charlag:
Noticed that "home timeline" preferences transition are slightly broken because fragment has no background color

connyduck (approved):
very nice

charlag:
should I merge it?

connyduck:
I was just waiting for Google Play to publish a nightly, because every new upload resets the position in the queue and most of the recent changes haven't been published. They have now.

PR #4026, opened 2023-09-19
Refactor notifications to Kotlin & paging

connyduck:
This refactors the NotificationsFragment and related classes to Kotlin & paging.
While trying to preserve as much of the original behavior as possible, this adds the following improvements as well:
- The "show notifications filter" preference was added again
- The "load more" button now has a background ripple effect when clicked
- The "legal" report category of Mastodon 4.2 is now supported in report notifications
- Unknown notifications now display "unknown notification type" instead of an empty line

Other code quality improvements:
- All views from xml layouts are now referenced via ViewBindings
- the classes responsible for showing system notifications were moved to a new package `systemnotifications` while the classes from this refactoring are in `notifications`
- the id of the local Tusky account is now called `tuskyAccountId` in all places I could find

closes https://github.com/tuskyapp/Tusky/issues/3429

PR #4290, opened 2024-02-27
Replace RxJava3 code with coroutines

cbeyls:
This pull request removes the remaining RxJava code and replaces it with coroutine-equivalent implementations.

- Remove all duplicate methods in `MastodonApi`:
  - Methods returning a RxJava `Single` have been replaced by suspending methods returning a `NetworkResult` in order to be consistent with the new code.
  - _sync_/_async_ method variants are replaced with the _async_ version only (suspending method), and `runBlocking{}` is used to make the async variant synchronous.
- Create a custom coroutine-based implementation of `Single` for usage in Java code where launching a coroutine is not possible. This class can be deleted after remaining Java code has been converted to Kotlin.
- `NotificationsFragment.java` can subscribe to `EventHub` events by calling the new lifecycle-aware `EventHub.subscribe()` method. This allows using the `SharedFlow` as single source of truth for all events.
- Rx Autodispose is replaced by `lifecycleScope.launch()` which will automatically cancel the coroutine when the Fragment view/Activity is destroyed.
- Background work is launched in the existing injectable `externalScope`, since using `GlobalScope` is discouraged. `externalScope` has been changed to be a `@Singleton` and to use the main dispatcher by default.
- Transform `ShareShortcutHelper` to an injectable utility class so it can use the application `Context` and `externalScope` as provided dependencies to launch a background coroutine.
- Implement a custom Glide extension method `RequestBuilder.submitAsync()` to do the same thing as `RequestBuilder.submit().get()` in a non-blocking way. This way there is no need to switch to a background dispatcher and block a background thread, and cancellation is supported out-of-the-box.
- An utility method `Fragment.updateRelativeTimePeriodically()` has been added to remove duplicate logic in `TimelineFragment` and `NotificationsFragment`, and the logic is now implemented using a simple coroutine instead of `Observable.interval()`. Note that the periodic update now happens between onStart and onStop instead of between onResume and onPause, since the Fragment is not interactive but is still visible in the started state.
- Rewrite `BottomSheetActivityTest` using coroutines tests.
- Remove all RxJava library dependencies.


app/src/main/java/com/keylesspalace/tusky/util/MediaUtils.kt
```diff
@@ -179,12 +179,10 @@ fun deleteStaleCachedMedia(mediaDirectory: File?) {
         return
     }
 
-    val twentyfourHoursAgo = Calendar.getInstance()
-    twentyfourHoursAgo.add(Calendar.HOUR, -24)
-    val unixTime = twentyfourHoursAgo.timeInMillis
+    val unixTime = System.currentTimeMillis() - 24.hours.inWholeMilliseconds
```
cbeyls, line 182: I just simplified this code a bit (no need to instantiate a `Calendar` object to substract hours).

app/src/main/java/com/keylesspalace/tusky/util/RelativeTimeUpdater.kt
```diff
@@ -0,0 +1,37 @@
...
+ * Helper method to update adapter periodically to refresh timestamp
+ * if setting absoluteTimeView is false.
+ * Start updates when the Fragment becomes visible and stop when it is hidden.
+ */
+fun Fragment.updateRelativeTimePeriodically(callback: Runnable) {
+    val preferences = PreferenceManager.getDefaultSharedPreferences(requireContext())
+    val lifecycle = viewLifecycleOwner.lifecycle
+    lifecycle.coroutineScope.launch {
+        // This child coroutine will launch each time the Fragment moves to the STARTED state
+        lifecycle.repeatOnLifecycle(Lifecycle.State.STARTED) {
```
cbeyls, line 27: Note that this coroutine is also cancelled automatically when the Fragment moves to the CREATED state (after onStop).

app/src/main/java/com/keylesspalace/tusky/util/RelativeTimeUpdater.kt
```diff
@@ -0,0 +1,37 @@
...
+            val useAbsoluteTime = preferences.getBoolean(PrefKeys.ABSOLUTE_TIME_VIEW, false)
+            if (!useAbsoluteTime) {
+                while (true) {
```
cbeyls, line 30: Note that this coroutine is also cancelled automatically when the Fragment moves to the CREATED state (after onStop). This is not a real infinite loop because it suspends in `delay()` and will be stopped on cancellation.

cbeyls (commented):
I clarified some parts of the code that could raise questions.

connyduck (approved):
This is super nice. I like the idea of the custom Single to bridge to Java.
(Will give me even more conflicts in https://github.com/tuskyapp/Tusky/pull/4026 but that is a monster anyway.)

Tak (approved):
I looked at this earlier today and didn't approve yet purely on the basis of conflict with the notification rework 🙂

cbeyls:
> This is super nice. I like the idea of the custom Single to bridge to Java. (Will give me even more conflicts in #4026 but that is a monster anyway.)

Yes, I saw that pull request right after I submitted mine so I apologize for that! I tried to keep the logic 1:1 compared to the current code (except where it made sense to improve it) to minimize impact and conflicts. Having done a similar refactoring in the past, I saw the opportunity.

cbeyls:
I noticed that the code base already includes a specialized scope for background operations that is defined in a Dagger module, so I updated my code to inject and use it.

PR #4291, opened 2024-02-28
Optimize Proguard rules

cbeyls:
Using saner defaults for R8 while reducing the app size even further.
- Add Kotlin compiler options to skip adding assertions in release builds
- Remove `optimizations`, `optimizationpasses` and `dontpreverify` rules that are ignored by R8
- Only keep runtime annotations by default. If other attributes are needed by a specific library, these will already be provided by the library rules (for example Retrofit or coroutines)
- Remove the obsolete rule allowing a View to reflectively call any arbitrary public Activity method accepting a View as argument. This has always been a bad practice and is not used in this project anyway
- Remove the rules related to enums. R8 already optimizes enums properly out-of-the-box and keeping these rules may prevent some of these optimizations
- Add support for the `@Keep` annotation. Even if it's not currently used in the code base, it can be handy in the future
- Add a missing rule to prevent generic signature of `NetworkResult` class from being removed in `MastodonApi` so Retrofit works
- Allow obfuscation and shrinking of `kotlin.coroutines.Continuation`, matching the rule defined in the next release of Retrofit
- Remove the rule forcing the removal of `String.format()`. This method is actually used in the code (and in third-party libraries) for other things than logging so forcing its removal can do more harm than good.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -48,6 +48,14 @@ android {
             minifyEnabled true
             shrinkResources true
             proguardFiles 'proguard-rules.pro'
+
+            kotlinOptions {
+                freeCompilerArgs = [
+                    "-Xno-param-assertions",
+                    "-Xno-call-assertions",
+                    "-Xno-receiver-assertions"
+                ]
+            }
         }
     }
 
--- a/app/proguard-rules.pro
+++ b/app/proguard-rules.pro
@@ -1,41 +1,52 @@
 # GENERAL OPTIONS
 
-# turn on all optimizations except those that are known to cause problems on Android
--optimizations !code/simplification/cast,!field/*,!class/merging/*
--optimizationpasses 6
 -allowaccessmodification
--dontpreverify
 
--dontusemixedcaseclassnames
--dontskipnonpubliclibraryclasses
--keepattributes *Annotation*
+# Preserve some attributes that may be required for reflection.
+-keepattributes RuntimeVisible*Annotations, AnnotationDefault
 
 # For native methods, see http://proguard.sourceforge.net/manual/examples.html#native
 -keepclasseswithmembernames class * {
     native <methods>;
 }
-# keep setters in Views so that animations can still work.
-# see http://proguard.sourceforge.net/manual/examples.html#beans
--keepclassmembers public class * extends android.view.View {
-    void set*(***);
-    *** get*();
-}
-# We want to keep methods in Activity that could be used in the XML attribute onClick
--keepclassmembers class * extends android.app.Activity {
-    public void *(android.view.View);
-}
-# For enumeration classes, see http://proguard.sourceforge.net/manual/examples.html#enumerations
--keepclassmembers enum * {
-    public static **[] values();
-    public static ** valueOf(java.lang.String);
-}
+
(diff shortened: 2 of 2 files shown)


connyduck:
Thx!

> Add support for the @Keep annotation. Even if it's not currently used in the code base, it can be handy in the future

not necessary, [that rule is included in the annotation library](https://github.com/androidx/androidx/blob/androidx-main/annotation/annotation/src/jvmMain/resources/META-INF/proguard/androidx-annotations.pro)

> Add a missing rule to prevent generic signature of NetworkResult class from being removed in MastodonApi so Retrofit works

not necessary, [that rule is included in the NetworkResult library](https://github.com/connyduck/networkresult-calladapter/blob/main/calladapter/src/main/resources/META-INF/proguard/consumer-rules.pro)

> Remove the rule forcing the removal of String.format(). This method is actually used in the code (and in third-party libraries) for other things than logging so forcing its removal can do more harm than good.

I disagree, it will only be removed if the result is not used. Take [this line](https://github.com/tuskyapp/Tusky/blob/c666a6b534d6460bf636403b9438d518bc5bcb0c/app/src/main/java/com/keylesspalace/tusky/components/login/LoginActivity.kt#L270) as an example: The log statement is removed, but without the `String.format` rule a dangling `String.format` will remain in the bytecode and we don't want that.

cbeyls:
> Thx!
> 
> > Add support for the @keep annotation. Even if it's not currently used in the code base, it can be handy in the future
> 
> not necessary, [these rules are included in the annotation library](https://github.com/androidx/androidx/blob/androidx-main/annotation/annotation/src/jvmMain/resources/META-INF/proguard/androidx-annotations.pro)

I didn't know that, thanks for the info. I'll remove this from my other projects as well.

> 
> > Add a missing rule to prevent generic signature of NetworkResult class from being removed in MastodonApi so Retrofit works
> 
> not necessary, [that rule is included in the NetworkResult library](https://github.com/connyduck/networkresult-calladapter/blob/main/calladapter/src/main/resources/META-INF/proguard/consumer-rules.pro)

I didn't think about checking the CallAdapter library rules first, indeed it makes sense to provide the rule with the library.

> > Remove the rule forcing the removal of String.format(). This method is actually used in the code (and in third-party libraries) for other things than logging so forcing its removal can do more harm than good.
> 
> I disagree, it will only be removed if the result is not used. Take [this line](https://github.com/tuskyapp/Tusky/blob/c666a6b534d6460bf636403b9438d518bc5bcb0c/app/src/main/java/com/keylesspalace/tusky/components/login/LoginActivity.kt#L270) as an example: The log statement is removed, but without the `String.format` rule a dangling `String.format` will remain in the bytecode and we don't want that.

Indeed, in the case of `String.format()` the method actually has no side effects so it's safe to keep the rule. I'll re-add it.

However, this also means that there are probably other kinds of String formatting code that end up orphaned after removing the Log calls.

Ideally the Log calls should be replaced with Kotlin inline functions taking a lamba argument (where all the formatting takes place) and checking a global boolean constant before logging, that's the only reliable way to remove all logging code and its associated formatting. But that's another topic entirely.

connyduck (approved):
This shaves about 200kB off the apk size, very nice.

PR #4292, opened 2024-02-28
change SendStatusService type to shortService

connyduck:
This way the `FOREGROUND_SERVICE_REMOTE_MESSAGING` permission is not needed and we should be able to publish on Google Play again. Drawback: The service can get killed after a while (usually 3 mins) on Android 14.
I also tried using [user initiated data transfer jobs](https://developer.android.com/about/versions/14/changes/user-initiated-data-transfers), but that is not available on all api levels, and `WorkManager`, but that is a huge refactoring and sending would probably work differently than before.


Tak (approved):

charlag (commented):
I think we don't upload media in background anyway? 3 min for just posting is more than reasonable even for very slow networks

PR #4296, opened 2024-02-29
Move ExoPlayer initialization to a Dagger module and optimize its dependencies

cbeyls:
Currently, ExoPlayer is initialized explicitly in `ViewMediaFragment` with all its dependencies, including many that are not useful for viewing Mastodon media attachments.

This pull request moves most ExoPlayer initialization and configuration to a new Dagger module, and instead a `Provider<ExoPlayer>` factory is injected in the Fragment so it can create new instances when needed.

The following ExoPlayer components will be configured:

- **Renderers**: all of them (audio, video, metadata, subtitles) except for the `CameraMotionRenderer`.
- **Extractors**: FLAC, Wav, Mp4, Ogg, Matroska/WebM and MP3 containers, to provide the same support as Firefox or Chrome browsers. Other container formats that are either image formats (already covered by Glide), not web-friendly or reserved for live streaming are skipped.
- **MediaSource**: only progressive download (through OkHttp) is provided. Live streaming support using protocols like RTSP, MPEG/Dash or HLS is skipped, because Mastodon servers don't use these protocols to download attachments.

The Mastodon documentation mentions the [supported media formats for attachments](https://docs.joinmastodon.org/user/posting/#media) and this covers them and even more. The docs also mentions that the video and audio files are transcoded to MP4 and MP3 upon upload but that was not the case in the past (for example WebM was used) and it could change again in the future.

Specifying these components manually allows reducing the APK size by about 200 KB thanks to R8 shrinking.

There are also a few extra code changes:
- Remove the code specific to API < 24 since the min SDK of the app is now 24.
- Add support for pausing a video when unplugging headphones.
- Specify the audio attributes according to content type to help the Android audio mixer.

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/di/PlayerModule.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/di/PlayerModule.kt
@@ -1,3 +1,20 @@
+/*
+ * Copyright 2024 Tusky Contributors
+ *
+ * This file is a part of Tusky.
+ *
+ * This program is free software; you can redistribute it and/or modify it under the terms of the
+ * GNU General Public License as published by the Free Software Foundation; either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * Tusky is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even
+ * the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU General
+ * Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License along with Tusky; if not,
+ * see <http://www.gnu.org/licenses>.
+ */
+
 package com.keylesspalace.tusky.di
 
 import android.content.Context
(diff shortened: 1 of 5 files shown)


app/src/main/java/com/keylesspalace/tusky/di/PlayerModule.kt
```diff
@@ -0,0 +1,123 @@
...
+    }
+
+    @Provides
+    fun provideRenderersFactory(context: Context, audioSink: AudioSink): RenderersFactory {
+        return RenderersFactory { eventHandler,
+                                  videoRendererEventListener,
+                                  audioRendererEventListener,
+                                  textRendererOutput,
+                                  metadataRendererOutput ->
+            arrayOf<Renderer>(
```
connyduck, line 52:
```suggestion
            arrayOf(
```

app/src/main/java/com/keylesspalace/tusky/di/PlayerModule.kt
```diff
@@ -0,0 +1,123 @@
+package com.keylesspalace.tusky.di
```
connyduck, line 1: missing the license header

connyduck (approved):

mcclure (approved):
This seems like a great improvement! Note, I have not had the chance to build and test locally so I'll assume you've done basic smoke tests at your end.

PR #4305, opened 2024-03-03
use Safe Calls on id.length when iterating notifications

pbielefeldt:


connyduck:
No, this is not the right fix, sorry

PR #4330, opened 2024-03-19
Retrofit 2.10.0

Goooler:
https://github.com/square/retrofit/releases/tag/2.10.0


connyduck:
Should we use `NetworkResult<Unit>` as a return type where we have `NetworkResult<ResponseBody>` now?

Goooler:
Good catch!

connyduck (approved):

PR #4335, opened 2024-03-26
fix crash when instance info fails to load

connyduck:
Steps to reproduce: Cold start the app while being logged in and offline.


charlag (approved):
oh I think I run into it today, thanks!

PR #4338, opened 2024-03-27
Add observe extensions for Flow to reduce nesting

Goooler:


Goooler:
This depends on #4337.

app/build.gradle
```diff
@@ -51,6 +51,7 @@ android {
 
             kotlinOptions {
                 freeCompilerArgs = [
+                    "-Xcontext-receivers",
```
Goooler, line 54: For using context receivers feature, see https://github.com/Kotlin/KEEP/blob/master/proposals/context-receivers.md

app/src/main/java/com/keylesspalace/tusky/util/FlowExtensions.kt
```diff
@@ -63,3 +73,45 @@ fun <T> Flow<T>.throttleFirst(timeout: Duration, timeSource: TimeSource = TimeSo
...
+
+fun <T> Flow<T>.observeLatest(
+    lifecycleOwner: LifecycleOwner,
+    action: suspend (value: T) -> Unit
+): Job = observeLatest(lifecycleOwner.lifecycleScope, action)
+
+context(LifecycleOwner)
+fun <T> Flow<T>.observe(
+    collector: FlowCollector<T>? = null
+): Job = observe(this@LifecycleOwner, collector)
```
Goooler, line 112: If we can use `Fragment`'s `viewLifecycleOwner` instead of its self for all invocations in Fragments?

connyduck:
That is nice, but also quite a big change. What do you think @charlag?

app/src/main/java/com/keylesspalace/tusky/util/FlowExtensions.kt
Goooler: Finished in 38bdb76fd8bfb4981514c0028f9fb693cab3bfc5, feel free to review.

cbeyls:
Be careful of subtle behavior changes: most of the current code doesn't use `repeatOnLifecycle()` but the new `observe()` functions do. Code that used to run once could now be repeated and in some cases that's not what you want.

cbeyls:
I think a balance needs to be found between clarity, flexibility and reducing nesting.
IMO the repeating part should be optional and mentioned explicitly. I also believe it's better to mention the lifecycle explicitly for clarity.

I think having two levels of nesting is already fine:

```
lifecycle.launch {
    flow.collect {
    }
}
```

A utility function `launchAndRepeatOnLifecycle` could be created for cases where you want the repeating behavior:

```
fun Lifecycle.launchAndRepeatOnLifecycle(
    state: Lifecycle.State = Lifecycle.State.STARTED,
    block: suspend CoroutineScope.() -> Unit
): Job = coroutineScope.launch {
    repeatOnLifecycle(state, block)
}

fun LifecycleOwner.launchAndRepeatOnLifecycle(
    state: Lifecycle.State = Lifecycle.State.STARTED,
    block: suspend CoroutineScope.() -> Unit
): Job = lifecycle.launchAndRepeatOnLifecycle(state, block)
```

Then to collect repeatedly, also two levels of nesting:

```
lifecycle.launchAndRepeatOnLifecycle {
    flow.collect {
    }
}
```

Note that you can also collect multiple things inside a lifecycle repetition by launching child coroutines. This is also more efficient than launching multiple instances of `repeatOnLifecycle()`:

```
lifecycle.launchAndRepeatOnLifecycle {
    launch {
        flow1.collect {
        }
    }
    launch {
        flow2.collect {
        }
    }
}
```

Goooler:
> Be careful of subtle behavior changes: most of the current code doesn't use repeatOnLifecycle() but the new observe() functions do.

That is what I'm considering too, remove it for now.

PR #4345, opened 2024-03-29
fix swipe-refresh spinner showing forever when refreshing AccountActivity

connyduck:
The flow must emit every update even if the values are the same, so use SharedFlow instead of StateFlow.

Regression from https://github.com/tuskyapp/Tusky/pull/4337 cc @Goooler

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/components/account/AccountViewModel.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/components/account/AccountViewModel.kt
@@ -27,6 +27,7 @@ import kotlinx.coroutines.flow.MutableSharedFlow
 import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.SharedFlow
 import kotlinx.coroutines.flow.StateFlow
+import kotlinx.coroutines.flow.asSharedFlow
 import kotlinx.coroutines.flow.asStateFlow
 import kotlinx.coroutines.launch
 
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/keylesspalace/tusky/components/account/AccountViewModel.kt
```diff
@@ -42,8 +45,8 @@ class AccountViewModel @Inject constructor(
     private val _noteSaved = MutableStateFlow(false)
     val noteSaved: StateFlow<Boolean> = _noteSaved.asStateFlow()
 
-    private val _isRefreshing = MutableStateFlow(false)
-    val isRefreshing: StateFlow<Boolean> = _isRefreshing.asStateFlow()
+    private val _isRefreshing = MutableSharedFlow<Boolean>(1, onBufferOverflow = BufferOverflow.DROP_OLDEST)
+    val isRefreshing: SharedFlow<Boolean> = _isRefreshing
```
Goooler, line 49:
```suggestion
    val isRefreshing: SharedFlow<Boolean> = _isRefreshing.asSharedFlow()
```

Goooler: Need `import kotlinx.coroutines.flow.asSharedFlow`

Goooler (approved):

Tak (approved):

cbeyls:
A better fix would be to set `isRefreshing` to true first in the coroutine that sets `isRefreshing` to false at the end.
This would also allow to properly show the refresh indicator after a configuration change if loading is still in progress.

connyduck:
Yes, I just wanted a quick fix to get rid of the regression. There are probably more improvements possible in the account view, I want to have a look soon. PRs also welcome!

Goooler:
Fixing in #4353.

PR #4347, opened 2024-03-29
Convert some sealed classes to interfaces

Goooler:
There is no non-abstract field in them, we can just fall back to interfaces.

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/util/Either.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/util/Either.kt
@@ -21,27 +21,29 @@ package com.keylesspalace.tusky.util
  * Class to represent sum type/tagged union/variant/ADT e.t.c.
  * It is either Left or Right.
  */
-sealed class Either<out L, out R> {
-    data class Left<out L, out R>(val value: L) : Either<L, R>()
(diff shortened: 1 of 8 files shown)


app/src/main/java/com/keylesspalace/tusky/util/Either.kt
```diff
@@ -21,27 +21,29 @@ package com.keylesspalace.tusky.util
...
-            Left(this.asLeft())
-        } else {
-            Right(mapper(this.asRight()))
+    companion object {
+        inline fun <L, R, N> Either<L, R>.map(crossinline mapper: (R) -> N): Either<L, N> {
+            return if (this.isLeft()) {
+                Left(this.asLeft())
+            } else {
+                Right(mapper(this.asRight()))
+            }
```
Goooler, line 46: > 'inline' modifier is not allowed on virtual members. Only private or final members can be inlined

Goooler: One more question here, seems no need to add `crossinline` for `mapper`, is there any specific reason to keep it?

Goooler: Addressing the fix to #4348.

connyduck (approved):

PR #4352, opened 2024-04-01
put back sync network methods / avoid runBlocking

connyduck:
On non-suspending methods the `NetworkResult` adapter is more efficient as it can go for `retrofit2.Call.execute` immediately and skip all the async business. And `runBlocking` in app code is generally considered a code smell. Sorry I missed that in the review of #4290 @cbeyls

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/components/compose/ComposeViewModel.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/components/compose/ComposeViewModel.kt
@@ -53,7 +53,6 @@ import kotlinx.coroutines.flow.first
 import kotlinx.coroutines.flow.shareIn
 import kotlinx.coroutines.flow.update
 import kotlinx.coroutines.launch
-import kotlinx.coroutines.runBlocking
 import kotlinx.coroutines.withContext
 
 class ComposeViewModel @Inject constructor(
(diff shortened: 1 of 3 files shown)


Tak (approved):

cbeyls:
For the record, here's the reasoning behind these removals:

- The main goal was to deduplicate calls in `MastodonApi` to make maintenance easier and reduce the risk of forgetting to update the second copy of the code. (In fact, I already found a discrepancy between another pair of blocking and non-blocking API calls while creating the pull request)
- The cost of the async business is negligible compared to the cost of completely blocking the current thread waiting for IO to complete. Both `Call.execute()` and `runBlocking()` block the thread
- The reason why `runBlocking` is considered a code smell is because it hints that a blocking caller should be replaced by a coroutine. Replacing runBlocking() with another blocking function doesn't change the smell but makes it less explicit.

In this case, the code is called in the context of the `Filter` class which creates its own background thread to filter results. Thus `runblocking` makes sense to bridge the gap between that background thread and a coroutine. Later, when that UI code is migrated to something coroutines-based, the `runBlocking()` can just be removed without changing anything else in the lower layers of the app.

connyduck:
🤔 
Ok so let's leave it as it is?

PR #4353, opened 2024-04-02
Better refreshing state indication for AccountActivity

Goooler:
Refs https://github.com/tuskyapp/Tusky/pull/4345#issuecomment-2030150905.


cbeyls:
Note: you can also remove

`binding.swipeToRefreshLayout.isRefreshing = true`

in `AccountActivity` (around line 1075). It's better to have the `ViewModel` as single source of truth for the refreshing state.

Goooler:
Good catch!

connyduck:
Now the spinner shows up on the initial load which is very irritating imo.

cbeyls:
> Now the spinner shows up on the initial load which is very irritating imo.

The state could be enriched with some additional information differentiating an initial load from a refresh.

For example, instead of a boolean, use a sealed class:

```
sealed class LoadingState {
    data object Idle : LoadingState()

    data class Loading(val isRefresh: Boolean) : LoadingState()
}

binding.swipeToRefreshLayout.isRefreshing = it is LoadingState.Loading && it.isRefresh
```

app/src/main/java/com/keylesspalace/tusky/components/account/AccountViewModel.kt
```diff
@@ -76,7 +70,7 @@ class AccountViewModel @Inject constructor(
 
     private fun obtainAccount(reload: Boolean = false) {
         if (_accountData.value == null || reload) {
-            isDataLoading = true
+            _refreshState.value = RefreshState.REFRESHING
```
connyduck, line 73:
```suggestion
            if (reload) {
                _refreshState.value = RefreshState.REFRESHING
            }
```

Goooler: I didn't notice here is the key point, can we revert 3d283bd5b97ff9a06ed6b5fa189de6e2a1662445 and use your logic here, it's a simple state.

```kt
  if (reload) {
      _refreshState.value = true
  }
```

Goooler: 6e48b6ad793ab8efd74a684d47403c6645139484

app/src/main/java/com/keylesspalace/tusky/components/account/AccountViewModel.kt
```diff
@@ -316,7 +310,7 @@ class AccountViewModel @Inject constructor(
     }
 
     private fun reload(isReload: Boolean = false) {
-        if (isDataLoading) {
+        if (_isRefreshing.value) {
```
cbeyls, line 313: Two loads may still be triggered in parallel during the initial load, because `_isRefreshing` is not set to true in that case.

Hence I suggest to keep the loading state in a separate field, or merge loading and refreshing states together using a sealed class.

Alternatively, cancel the load in progress before starting a new one.

PR #4366, opened 2024-04-09
Optimize I/O code using Okio

cbeyls:
This pull request takes advantage of the Okio library to simplify, fix or improve performance of some I/O related code in Tusky.

- Return early or throw `FileNotFoundException` early in case `contentResolver.openInputStream()` returns `null` instead of throwing `NullPointerException` later. Change the signature of `Closeable.closeQuietly()` to only accept a non-null `Closeable`.
- Reimplement `Uri.copyToFile()` using Okio. This takes advantage of the built-in high-performance buffers of the library so a buffer doesn't need to be allocated or managed manually. The new implementation also makes sure that the input and output streams are always closed, as the original code could in some cases return without properly closing a stream.
- Reimplement `ProgressRequestBody` as `Uri.asRequestBody()` (adding to the existing extension functions available in the Okio library to create a `RequestBody`). The new implementation uses Okio's `Buffer` instead of a manually managed byte array, which allows to avoid copying bytes from one buffer to the next. The max number of bytes read at once was increased from 2K to 8K to improve performance. Avoid division by zero in case `contentLength` is `0`. Finally, this implementation now takes a `Uri` as input instead of an `InputStream`, because a `RequestBody` must be replayable in case Okio retries the request, and an `InputStream` can only be used once.


app/src/main/java/com/keylesspalace/tusky/network/UriRequestBody.kt
```diff
@@ -0,0 +1,57 @@
...
+import okio.BufferedSink
+import okio.source
+
+private const val DEFAULT_CHUNK_SIZE = 8192L
+
+fun interface UploadCallback {
+    fun onProgressUpdate(percentage: Int)
+}
+
+fun Uri.asRequestBody(contentResolver: ContentResolver, contentType: MediaType? = null, contentLength: Long = -1L, uploadListener: UploadCallback? = null): RequestBody {
```
Goooler, line 32:
```suggestion
fun Uri.asRequestBody(
    contentResolver: ContentResolver, 
    contentType: MediaType? = null, 
    contentLength: Long = -1L, 
    uploadListener: UploadCallback? = null
): RequestBody {
```

Goooler (approved):
And we can replace `java.nio.ByteBuffer` and `java.nio.charset.Charset`.

https://github.com/tuskyapp/Tusky/blob/db27186b5c16137edc3fdfab0ab545933d734f0e/app/src/main/java/com/keylesspalace/tusky/util/CompositeWithOpaqueBackground.kt#L64
https://github.com/tuskyapp/Tusky/blob/db27186b5c16137edc3fdfab0ab545933d734f0e/app/src/main/java/com/keylesspalace/tusky/util/CompositeWithOpaqueBackground.kt#L114

connyduck (approved):

cbeyls:
> And we can replace `java.nio.ByteBuffer` and `java.nio.charset.Charset`.
> 
> https://github.com/tuskyapp/Tusky/blob/db27186b5c16137edc3fdfab0ab545933d734f0e/app/src/main/java/com/keylesspalace/tusky/util/CompositeWithOpaqueBackground.kt#L64
> 
> https://github.com/tuskyapp/Tusky/blob/db27186b5c16137edc3fdfab0ab545933d734f0e/app/src/main/java/com/keylesspalace/tusky/util/CompositeWithOpaqueBackground.kt#L114

Sorry I didn't have time to apply your suggested changes today.

Do you mean replacing `ByteBuffer` with Okio `Buffer` and `toByteArray()` with Okio `Buffer` or `ByteString`? I'm not sure it would be worth the change for these cases.

Goooler:
Yeah, just wanna replace all nio usages.

PR #4387, opened 2024-04-21
switch all string placeholders to positional format

### LINKED ISSUE #3297: Switch to positional substitution format in string XML
As of git 15ff6191ae and Android Studio Electric Eel, building Tusky gives you many warnings on our string XMLs saying

> Multiple substitutions specified in non-positional format of string resource

It wants us to move to [a new format for string substitutions](https://stackoverflow.com/questions/48345910/error-multiple-substitutions-specified-in-non-positional-format-did-you-mean-t), `%1$s` instead of `%s`. Doing this would actually be good for us anyway because then translations could rearrange substitutions.

connyduck:
closes #3297


app/src/main/res/values-fa/strings.xml
```diff
@@ -210,24 +210,24 @@
...
+    <string name="abbreviated_in_years">در %1$d سال</string>
+    <string name="abbreviated_in_days">در %1$d روز</string>
+    <string name="abbreviated_in_hours">در %1$d ساعت</string>
+    <string name="abbreviated_in_minutes">در %1$d دقیقه</string>
+    <string name="abbreviated_in_seconds">در %1$d ثانیه</string>
     <string name="follows_you">پیگیرتان است</string>
     <string name="pref_title_alway_show_sensitive_media">نمایش همیشگی محتوای حساس</string>
     <string name="title_media">رسانه</string>
-    <string name="replying_to">در حال پاسخ به @%s</string>
+    <string name="replying_to">در حال پاسخ به @%1$s</string>
```
connyduck, line 221: I am not sure if this is correct, but Android Studio is super weird with bidirectional files and that is what auto replace produced (also in the arabic translation)

connyduck: maybe @BoFFire knows

charlag (approved):

app/src/main/res/values-fa/strings.xml
BoFFire: Replying to @%1$s

در حال پاسخ به @%1$s

BoFFire: Cc @danialbehzadi

danialbehzadi: > I am not sure if this is correct
It seems fine to me

PR #4391, opened 2024-04-23
Refactor permissions requests to use ActivityResultContract

cbeyls:
The app currently uses a custom callback system for requesting permissions, located in `BaseActivity`.
This code doesn't work when the activity gets re-created before the permission is granted: the callback will be lost with the Activity and no action will be performed after the permission is granted.

To avoid these issues while still simplifying the code, Google recommends to use the ActivityResultContract APIs to request permissions, which allow to register persistent callbacks. This pull request removes the legacy API and replaces the calls with `registerForActivityResult(ActivityResultContracts.RequestPermission())`.

- `ActivityResultContracts.RequestPermission` will check if the permission is already granted and call the callback synchronously when possible. So this check doesn't need to be performed anymore.
- In `SearchStatusesFragment` and `SFragment`, the download action to launch after granting the permission requires an argument (a list of media URLs). This argument is temporarily stored in a Fragment field and saved and restored as part of the instance state, so the action can properly resume after an Activity re-creation.

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/BaseActivity.java
+++ b/app/src/main/java/com/keylesspalace/tusky/BaseActivity.java
@@ -19,7 +19,6 @@ import android.app.ActivityManager;
 import android.content.Context;
 import android.content.Intent;
 import android.content.SharedPreferences;
-import android.content.pm.PackageManager;
 import android.content.res.Configuration;
 import android.graphics.Bitmap;
 import android.graphics.BitmapFactory;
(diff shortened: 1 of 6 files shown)


connyduck (approved):
There is also the notifications permission in `MainActivity` but I guess we can ignore that since we don't need to react on the users choice?

cbeyls:
> There is also the notifications permission in `MainActivity` but I guess we can ignore that since we don't need to react on the users choice?

The main benefit of the new API is the callback. We could also change this one for consistency, but it wouldn't make the code simpler since like you say there is no need to react on the users choice.

PR #4411, opened 2024-05-03
Load all vector drawables with AppCompat for consistency

cbeyls:
Currently some vector drawables are loaded using the AppCompat library and others are loaded using the framework.

This pull request uniformizes this to use AppCompat to load them all.

Other changes:
- Set all compound drawables using relative positioning, since all XML layouts are also using relative positioning.
- Remove unnecessary layer list drawable used to center `R.drawable.ic_play_indicator` icon and use `ImageView.setForegroundGravity()` instead.
- Merge layers in toolbar icons `ic_arrow_back_with_background` and `ic_more_with_background` into a single vector drawable. Note that the AppCompat implementation of vector drawables is unable to load vector drawables inside layer-list drawables, so this change also makes these images compatible with older Android versions.

**Note**: technically, AppCompat will always delegate to the framework to load vector drawables on API 24+ which is the current minSDK version of the app. But at least this gives the option to lower the minSDK version in the future.


connyduck (approved):

connyduck:
please fix the merge conflict

PR #4412, opened 2024-05-03
Upgrade DrawerLayout to 1.2.0 and simplify code

cbeyls:
The main benefit of upgrading to version 1.2.0 of `DrawerLayout` is ~~to properly support **predictive back animations**: when initiating a back gesture on API 33+, the DrawerLayout will animate automatically~~ to handle back navigation automatically. The predictive back animation of the menu however depends on `NavigationView` which is not used in the project.

In addition to the upgrade, simplify DrawerLayout integration:
- Forward key events to the DrawerLayout so it can intercept them and close itself when needed.
- Don't handle the DrawerLayout closing manually using the `OnBackPressedCallback` anymore. This is not necessary since the back event will now be intercepted by the DrawerLayout when needed before reaching the `OnBackPressedDispatcher`.
- Remove legacy fix for DrawerLayout staying open after Activity recreation.


connyduck:
> properly support predictive back animations: when initiating a back gesture on API 33+, the DrawerLayout will animate automatically

That would be super cool but it doesn't do anything here (Api 34 emulator, predictive back enabled, this branch) 🙁

https://github.com/tuskyapp/Tusky/assets/10157047/46ea8f43-ead8-420d-861e-e247450b7350

cbeyls:
> > properly support predictive back animations: when initiating a back gesture on API 33+, the DrawerLayout will animate automatically
> 
> That would be super cool but it doesn't do anything here (Api 34 emulator, predictive back enabled, this branch) 🙁
> Screen_recording_20240510_161254.mp4

Indeed it turns out the `DrawerLayout` properly handles the back gesture but the animation part is handled by `NavigationView` from the Material Components library. Tusky uses the `MaterialDrawerSliderView` library instead, which doesn't support this feature (even in the newest version apparently).

It's still worth the change for the simpler code I think.

PR #4413, opened 2024-05-03
Update splash screen to properly use the Androidx SplashScreen library

cbeyls:
The Androidx SplashScreen library is added as a dependency to the project but isn't properly enabled in the current code. This pull request configures the splash screen properly.

- Remove `SplashScreenActivity` which is not needed and use `MainActivity` as main entry point to the application. `MainActivity` inherits from `BaseActivity` which already detects if no account is configured and redirects to `LoginActivity` if needed, just like `SplashScreenActivity`.
- Initialize the SplashScreen library in `MainActivity.onCreate()`.
- Instead of letting the SplashScreen library set the final theme from the `postSplashScreenTheme` attribute in SplashTheme, let `BaseActivity` set it according to the user settings.
- When no account is available in `MainActivity.onCreate()`, keep the splash screen shown until `LoginActivity` appears.
- Disable the slide-in animation when launching `LoginActivity` when no account is available because the detection happens in `onCreate()` and an Activity that finishes itself in `onCreate()` will not be drawn, so the slide-in animation will not be visible either and only `LoginActivity` will appear.
- Upgrade `core-splashscreen` to 1.2.0-alpha01 which contains a fix for corrupted app theme on API 31+.


connyduck:
I get an extra toolbar now? 
Screenshot_20240510_144144

cbeyls:
> I get an extra toolbar now? 

Looks like a bug in the SplashScreen library that was fixed recently:

> Reset enforceNavigationBarContrast to system default value from API 31, so the splash screen theme won't corrupt activity's theme.

I upgraded to the latest alpha which contains the fix.

connyduck (approved):

PR #4423, opened 2024-05-05
Replace Dagger-Android with Hilt and remove Kapt

cbeyls:
Hilt is an annotation processor built on top of Dagger which allows to remove all the Android dependency injection boilerplate code (currently around 900 lines) by writing it for us.

Hilt can use KSP instead of Kapt so Kapt can be completely removed from the project. Kapt is slow, deprecated and has a few compatibility issues. Removing Kapt will improve build times since no Java stubs have to be generated for Kotlin classes anymore (Note that KSP also processes annotations in Java classes so it can completely replace Kapt).

- Remove all modules related to manual dependency injection configuration.
- Rename `AppModule` to `StorageModule` since it now only contains configuration to retrieve the DataBase and SharedPreferences.
- Annotate all entry points (Activities, Fragments, BroadcastReceivers and Services) with `@AndroidEntryPoint`.
- Annotate all injected ViewModels with `@HiltViewModel` and replace the custom ViewModel Factory with the default one (which integrates with the one generated by Hilt).
- Add a public field to allow overriding the default ViewModelProvider.Factory in `BaseActivity` in tests.
- Annotate tested Activities with `@OptionalInject` since Activity tests currently rely on the Activities not being injected automatically.
- Annotate injected `Context` arguments with `@ApplicationContext`. Hilt provides the `Context` binding automatically but requires to specify if the Application or Activity Context is wanted.
- Add WorkManager Hilt integration so all Workers are injected by Hilt automatically using `HiltWorkerFactory`.
- Lazily initialize WorkManager in `TuskyApplication`.
- Remove Kapt and Kapt workarounds.
- ~~Remove toolchain configuration for Java 21. Toolchains force the Java bytecode to match the JDK version used to build the project, and apparently Hilt doesn't run inside the toolchain so cannot process the source code if the JDK version of the toolchain is higher than the JDK used to run Gradle. [And configuring a toolchain for an older Java version causes other issues](https://jakewharton.com/gradle-toolchains-are-rarely-a-good-idea/). **Removing toolchains configuration doesn't prevent the project from being built using JDK 21** or more recent versions but allows to build the project using older JDKs as well.~~
Added a fix to allow Hilt to properly use the JDK toolchain.
- ~~Set the Java and Kotlin bytecode target to Java 17. The standard bytecode target for Android projects is usually Java 8 or 11 (any higher version doesn't provide any benefit but may cause compatibility issues). However, since the app currently uses a library built against Java 17 bytecode (`networkresult-calladapter`), it needs to target at least Java 17 bytecode as well.~~
- Update the Dagger 2 URL in the licenses screen. Hilt is part of Dagger 2 so the label wasn't changed.


app/build.gradle
```diff
@@ -81,6 +81,16 @@ android {
...
+
+    compileOptions {
+        sourceCompatibility JavaVersion.VERSION_17
+        targetCompatibility JavaVersion.VERSION_17
+    }
+
+    kotlinOptions {
+        jvmTarget = "17"
+    }
+
```
Goooler, line 93:
```suggestion
```

This is unnecessary, it has been covered by https://github.com/tuskyapp/Tusky/blob/05c7e7b8066506830f6c0bf9112e7beab90ef00c/build.gradle#L13-L17

settings.gradle
```diff
@@ -12,7 +12,6 @@ pluginManagement {
 
 plugins {
     id 'com.gradle.develocity' version '3.17.2'
-    id 'org.gradle.toolchains.foojay-resolver-convention' version '0.8.0'
```
Goooler, line 15: This might be necessary for installing toolchain.

app/build.gradle
cbeyls: Check my comment above about why I disabled the toolchain. Hilt doesn't work with Java 21 bytecode and lowering the toolchain version forces you to build the project using an older Java version.

settings.gradle
cbeyls: And I removed it because I disabled toolchain to allow building with recent Java versions while targeting older bytecode.

app/build.gradle
cbeyls: The whole point of the toolchain is to be configured separately from the JDK used by Android Studio or the JDK running Gradle itself, so the project is always built using a specific version of the JDK no matter which JDK is installed on the machine. The Java 21 toolchain used to work on my machine, but not with Hilt.

The real question is: why do we want to use a toolchain in the first place, especially for Android projects?

Goooler: 1. Hilt is compatible with Java 21, you have to set Java 21 as the `JAVA_HOME` or use it in your AS.
2. As it's convenient for configuring `targetCompatibility`, `jvmTarget`, and so on. At least, in the past. If we have to get rid of toolchain, it would be better to address the change in a separate PR.

image

cbeyls: For the record I also tried to configure the toolchain like this which is the recommended way for Kotlin projects:

```
kotlin {
    jvmToolchain(21)
}
```

But the effect is the same and Hilt doesn't seem to run inside the toolchain and will only recognize Java 21 sources when Gradle itself is run with JDK 21.

Goooler: Kotlin's `jvmToolchain` is honoring Java's, no need to declare it repeatedly.

cbeyls:
@Goooler I agree that removing the toolchain is a different discussion and should be part of another PR, so I added it back.

The problem was that Hilt did not use JDK of the toolchain, thus defeating the purpose of using a toolchain. Fortunately I now found a fix to force Hilt to use the toolchain as well:

```
    tasks.withType(JavaCompile).configureEach {
        javaCompiler = javaToolchains.compilerFor {
            languageVersion = JavaLanguageVersion.of(21)
        }
    }
```
The project can again be built by running Gradle from any JDK version.

build.gradle
```diff
@@ -15,6 +15,13 @@ allprojects {
             toolchain.languageVersion = JavaLanguageVersion.of(21)
         }
     }
+
+    // Required for Hilt to use the toolchain
+    tasks.withType(JavaCompile).configureEach {
+        javaCompiler = javaToolchains.compilerFor {
+            languageVersion = JavaLanguageVersion.of(21)
+        }
+    }
```
Goooler, line 24:
```suggestion
```

Goooler: We don't need this config, it's honored from `JavaPluginExtension` config above, see also https://github.com/gradle/gradle/blob/95a2521ca265080e9f3c12ef0a25d10c6ff3da22/platforms/jvm/plugins-java-base/src/main/java/org/gradle/api/plugins/JavaBasePlugin.java#L229

cbeyls: Plugins inheriting from `JavaBasePlugin` configure the toolchain properly, yes, but the Hilt plugin does not.
See https://github.com/google/dagger/blob/9a67471586c3ed1bccb5d0e13a86af06e024cd1a/java/dagger/hilt/android/plugin/main/src/main/kotlin/dagger/hilt/android/plugin/HiltGradlePlugin.kt#L367

connyduck (approved):

PR #4437, opened 2024-05-10
fix notification filters behaving weirdly

### LINKED ISSUE #4433: Tusky notification-refactor branch got confused about filters, then crashed
### Detailed description

This morning I woke up and scrolled through my Tusky notifications. Because of Arguments, I had an unusually large number of notifications. I was finding it hard to see the replies among the boosts, so I tapped "filter" to hide the boosts and favs. To my surprise, "favs" and "boosts" were already disabled in the filter settings, although favs and boosts were definitely visible. Experimentally, I tapped "Apply" without making any changes. The app crashed ("this app has stopped"). Afterward I found this stack trace in my logcat:

[tusky-crash-snip-2024-05-09.txt](https://github.com/tuskyapp/Tusky/files/15262710/tusky-crash-snip-2024-05-09.txt)

On rebooting, the inconsistent state (boosts and favs visible, but unchecked in filter dropdown) persisted. "Refreshing" (IE dragging down from top) did not fix the visibility states as it sometimes does. Clicking "Apply" did not fix the visibility states (although also, it did not cause another crash). Checking "boosts" and "Favs" and hitting apply, then unchecking and hitting apply again, did fix the visibility states.

No known reproduction

### Debug information

`5ba10f2d2067` (non-final commit from notification refactor branch), self-built, greenDebug

connyduck:
This does four things

- set `enablePlaceholders = false` on `PagingConfig`s to avoid Paging Data that contains null placeholders, we don't want them (everywhere, not just in notifications)
- make sure NotificationsPagingAdapter does not crash when it encounters a null placeholder
- makes sure the notifications refresh correctly when the filters change
- the filters are now also respected when loading a gap

closes #4433

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/components/account/media/AccountMediaViewModel.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/components/account/media/AccountMediaViewModel.kt
@@ -43,7 +43,8 @@ class AccountMediaViewModel @Inject constructor(
     val media = Pager(
         config = PagingConfig(
             pageSize = LOAD_AT_ONCE,
-            prefetchDistance = LOAD_AT_ONCE * 2
+            prefetchDistance = LOAD_AT_ONCE * 2,
+            enablePlaceholders = false
         ),
         pagingSourceFactory = {
             AccountMediaPagingSource(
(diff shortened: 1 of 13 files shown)


connyduck:
meh now it sometimes makes more network requests than necessary

mcclure:
I've been using this for a couple days now and I haven't noticed the issues from #4433 recurring. Will continue testing.

Tak (approved):

PR #4441, opened 2024-05-11
inject SharedPreferences

connyduck:
(this one is for @charlag)

Calling `PreferenceManager.getDefaultSharedPreferences()` will read the preference file from disk every time. This PR makes `SharedPreferences` a singleton so they will only be created once at appstart (with a few exceptions where it is hard to inject, e.g. in the `openLink` helper) which should help getting our ANRs down.

```
StrictMode policy violation; ~duration=285 ms: android.os.strictmode.DiskReadViolation
    at android.os.StrictMode$AndroidBlockGuardPolicy.onReadFromDisk(StrictMode.java:1666)
    at libcore.io.BlockGuardOs.access(BlockGuardOs.java:74)
    at libcore.io.ForwardingOs.access(ForwardingOs.java:128)
    at android.app.ActivityThread$AndroidOs.access(ActivityThread.java:8054)
    at java.io.UnixFileSystem.checkAccess(UnixFileSystem.java:313)
    at java.io.File.exists(File.java:813)
    at android.app.ContextImpl.ensurePrivateDirExists(ContextImpl.java:790)
    at android.app.ContextImpl.ensurePrivateDirExists(ContextImpl.java:781)
    at android.app.ContextImpl.getPreferencesDir(ContextImpl.java:737)
    at android.app.ContextImpl.getSharedPreferencesPath(ContextImpl.java:962)
    at android.app.ContextImpl.getSharedPreferences(ContextImpl.java:583)
    at android.content.ContextWrapper.getSharedPreferences(ContextWrapper.java:221)
    at android.content.ContextWrapper.getSharedPreferences(ContextWrapper.java:221)
    at androidx.preference.PreferenceManager.getDefaultSharedPreferences(PreferenceManager.java:119)
    at com.keylesspalace.tusky.BaseActivity.onCreate(BaseActivity.java:96)
   ...
```


app/src/main/java/com/keylesspalace/tusky/components/login/LoginActivity.kt
```diff
@@ -99,11 +96,6 @@ class LoginActivity : BaseActivity() {
                 .into(binding.loginLogo)
         }
 
-        preferences = getSharedPreferences(
-            getString(R.string.preferences_file_key),
-            Context.MODE_PRIVATE
-        )
```
connyduck, line 105: Not sure why the SharedPreferences were created differently than everywhere else here?

app/src/main/java/com/keylesspalace/tusky/BaseActivity.java
```diff
@@ -125,7 +127,8 @@ private boolean activityTransitionWasRequested() {
 
     @Override
     protected void attachBaseContext(Context newBase) {
-        SharedPreferences preferences = PreferenceManager.getDefaultSharedPreferences(newBase);
+        // injected preferences not yet available in this point of the lifecycle
+        SharedPreferences preferences = ((TuskyApplication)newBase.getApplicationContext()).sharedPreferences;
```
connyduck, line 131: Too hacky?

cbeyls:
While I agree that it's better to inject preferences everywhere for consistency and to make testing easier, it's not true that a new instance gets created every time.

[The system keeps a static cache of SharedPreferences by name](https://android.googlesource.com/platform/frameworks/base/+/refs/heads/main/core/java/android/app/ContextImpl.java#592), and that cache includes the default SharedPreferences as well.
So it's not necessary to declare it as `@Singleton` in the dagger module because it's already a singleton.

app/src/main/java/com/keylesspalace/tusky/BaseActivity.java
```diff
@@ -125,7 +127,8 @@ private boolean activityTransitionWasRequested() {
 
     @Override
     protected void attachBaseContext(Context newBase) {
-        SharedPreferences preferences = PreferenceManager.getDefaultSharedPreferences(newBase);
+        // injected preferences not yet available at this point of the lifecycle
+        SharedPreferences preferences = ((TuskyApplication)newBase.getApplicationContext()).getPreferences();
```
cbeyls, line 131: Consider using a Dagger [EntryPoint](https://dagger.dev/hilt/entry-points.html) instead.

app/src/main/java/com/keylesspalace/tusky/components/login/LoginActivity.kt
```diff
@@ -99,11 +96,6 @@ class LoginActivity : BaseActivity() {
                 .into(binding.loginLogo)
         }
 
-        preferences = getSharedPreferences(
```
cbeyls, line 102: This separate SharedPreferences instance was used to store temporary data that should iddeally be stored to saveInstanceState instead of polluting the global SharedPreferences.

app/src/main/java/com/keylesspalace/tusky/BaseActivity.java
connyduck: oh that is nice, will do

app/src/main/java/com/keylesspalace/tusky/components/login/LoginActivity.kt
connyduck: 🧐

connyduck:
> it's not true that a new instance gets created every time.

Maybe it is not a new instance, but it definitely does the disk read every time and that is my main concern here

cbeyls:
> Maybe it is not a new instance, but according to StrictMode it does the disk read every time and that is my main concern here

I was surprised because I didn't notice that behavior in my apps before. It turns out that the cache is Context-related so the file is read once per Context. If we make sure to pass the Application Context for every retrieval like we do now, the file will only be read once and the `SharedPreferences` don't need to be declared as Singleton in the Dagger module.

Tak (approved):

PR #4443, opened 2024-05-12
Replace Either with Kotlin Result and optimize ListUtils

cbeyls:
Using `Either<Throwable, T>` is basically the same as `Result<T>` with a less friendly API. Futhermore, `Either` is currently only used in a single component.

- Replace `Either` with `Result` in `AccountsInListFragment` and `AccountsInListViewModel`.
- Add a method to convert a `NetworkResult` to a `Result` in `AccountsInListViewModel`. Alternatively, `NetworkResult` could be used everywhere in the code but other classes are already using `Result`.
- Replace `updateState()` method with `MutableStateFlow.update()` in `AccountsInListViewModel`.
- Store the current search query in a `MutableStateFlow` collected by a coroutine. This allows automatically cancelling the previous search when a new query arrives, instead of launching a new coroutine for each query which may conflict with the previous ones.
- Optimize `ListUtils`.


connyduck:
The whole account list component is due for a rewrite (move to new package structure, ViewModel & AndroidX paging), I wanted to tackle that soon, but if you want you can do it as well

cbeyls:
> The whole account list component is due for a rewrite (move to new package structure, ViewModel & AndroidX paging), I wanted to tackle that soon, but if you want you can do it as well

I can help. This branch can be used as a starting point too.

connyduck (approved):
Yes, let's merge this

PR #4445, opened 2024-05-13
Add a feature and preference to confirm follows.

fries1234:
This means a popup will appear if you have that option enabled in the preferences which will have a popup similar to the unfollow dialog asking you if you want to follow the user.


connyduck (approved):
Thank you!

PR #4461, opened 2024-05-22
Upgrade kotlin to v2

connyduck:


Tak (approved):
tbh I was expecting *much* more upheaval with this upgrade

connyduck:
Yes, I was surprised as well. Only one minor compile error? Really? :D

PR #4467, opened 2024-05-24
Optimize Glide emoji targets

cbeyls:
The current code loads emojis using Glide into basic custom `Target`s and doesn't keep a hard reference to the Target. This creates a few problems:
- Unlike images loaded into `ImageViewTarget`s, Emoji animations are not paused when the Activity/Fragment becomes invisible. GIF decoding use resources in the background.
- When `TextView`s get recycled in a RecyclerView, the loading of emojis for the previous bind are not canceled when binding the new text and starting the load of the new emojis. This is also handled automatically when using `ImageViewTarget` but not for custom targets. Also, when the obsolete emojis complete loading, the `TextView` will be unnecessarily invalidated and redrawn.
- Since Glide's `RequestManager` doesn't keep hard references to Targets after they are loaded and the emoji Target is currently not stored in any View, emojis don't get an opportunity to clean up (at least stop their animation) when the Activity/Fragment is destroyed. Depending on the Drawable implementation, animations may run forever in the background and cause memory leaks.

This pull request aims to properly track the lifecycle of emoji Targets, cancel their loading an stop animations when appropriate. It also reimplements `emojify()` to be more efficient.

- Add extension functions `View.setEmojiTargets()` and `View.clearEmojiTargets()` to store and clear lists of emoji targets in View tags, keeping a hard reference to them as long as the View is used. When clearing emoji targets, pending requests will be canceled and animations will be stopped to free memory. This is similar to what `ImageViewTarget` does, except here multiple Targets are stored for a single View instead of one.
- Add helper extension function `View.updateEmojiTargets()` to automatically clear the View emoji targets, then allowing to call `emojify()` one or more times in the `EmojiTargetScope` of that View, and finally store all the pending targets of the `EmojiTargetScope` in the View.
- Reimplement `CharSequence.emojify()` using `View.updateEmojiTargets()`. This is used in RecyclerViews as well and will automatically cancel previous emoji loadings for the same View and stop animations.
- The main logic of `emojify()` has been moved to `EmojiTargetScope`. Replace usage of slow regex `Matcher` with faster `CharSequence.indexOf()`. Use `SpannableString` (with `toSpannable()`) instead of `SpannableStringBuilder` to store the `EmojiSpan`s.
- Rename `EmojiSpan.getTarget()` to `EmojiSpan.createGlideTarget()` and improve the target to stop/resume the animation according to the parent component lifecycle, and stop the animation when clearing the target. Use a hard reference to the view instead of a weak reference, since the lifecycle of the Target now matches the one of the View and the Target will be cleared at the latest when the View is destroyed.
- Use `View.updateEmojiTargets()` in `ReportNotificationViewHolder` in order to store the targets of 2 separate sets of emojis into the same TextView.
- Fix: reimplement the code to merge the 2 emoji sets into a single `CharSequence` in `ReportNotificationViewHolder` using `TextUtils.expandTemplate()`. The current code uses `String.format()` which returns a String instead of a Spannable so the computed emojis are lost.
- Store the emoji targets in `AnnouncementAdapter` in the parent view after clearing the previous ones. It is a better location than storing one emoji target in each child `Tooltip` view because tooltips are not recycled when refreshing the data and the previous targets would not be canceled properly.
- Bonus: update `ViewVideoFragment` to use `CustomViewTarget` instead of `CustomTarget` to load the default artwork into `PlayerView`. The loading will also automatically be canceled when the fragment view is detached.


connyduck (approved):
This is very nice, thank you!

PR #4471, opened 2024-05-26
fix timeline jumping

connyduck:
Two things changed here:

The check for `positionStart`only in `onItemRangeInserted` is not always correct - we only want to jump up when something is inserted at the top, if we already are at the top.

`enablePlaceholders = false` has unintended side effects -  the recyclerview adapter sometimes receives an "onItemRangeRemoved" followed by an "onItemRangeInserted", instead of just "onItemRangeChanged".

Together they should make sure the timelines stay were they are.

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/components/trending/TrendingTagsFragment.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/components/trending/TrendingTagsFragment.kt
@@ -26,6 +26,7 @@ import androidx.fragment.app.viewModels
 import androidx.lifecycle.lifecycleScope
 import androidx.recyclerview.widget.GridLayoutManager
 import androidx.recyclerview.widget.GridLayoutManager.SpanSizeLookup
+import androidx.recyclerview.widget.LinearLayoutManager
 import androidx.recyclerview.widget.RecyclerView
 import androidx.recyclerview.widget.SimpleItemAnimator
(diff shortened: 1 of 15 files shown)


connyduck:
I also cannot reproduce the crash when clicking "load more" anymore with these changes, chances are the placeholder thing was responsible for that as well.

app/src/main/java/com/keylesspalace/tusky/components/trending/TrendingTagsFragment.kt
```diff
@@ -72,7 +73,8 @@ class TrendingTagsFragment :
 
         adapter.registerAdapterDataObserver(object : RecyclerView.AdapterDataObserver() {
             override fun onItemRangeInserted(positionStart: Int, itemCount: Int) {
-                if (positionStart == 0 && adapter.itemCount != itemCount) {
+                val firstPos = (binding.recyclerView.layoutManager as LinearLayoutManager).findFirstCompletelyVisibleItemPosition()
+                if (firstPos == 0 && positionStart == 0 && adapter.itemCount != itemCount) {
```
Tak, line 77: Maybe extract this snippet?

Tak (approved):
The code changes make sense imo, I'll start testing locally

charlag (approved):

connyduck:
I tried adding an extension function but I ended up making special cases for every fragment it is used in, I think it is better to have tis bit of duplication

PR #4472, opened 2024-05-26
Prevent parallel loading and fix duplicate ViewModel state collection in FiltersActivity

cbeyls:
This pull request fixes the following issues:

- `FiltersActivity` launches a new coroutine to collect the ViewModel state every time the Activity is resumed, without cancelling the previous coroutine.
- `FiltersActivity` reloads the filters in `onResume()`, even if loading is already in progress (without cancelling the current loading). This can lead to inconsistent state.

List of improvements:
- Implement `launchAndRepeatOnLifecycle()` to combine `coroutineScope.launch()` with `repeatOnLifecycle()` for the same `Lifecycle`. Use it in `FiltersActivity` to update the view only when the Activity is visible.
- Optimize the filters loading: load them when `FiltersViewModel` is created and when returning from `EditFilterActivity` (when receiving the Activity result). Cancel the load already in progress, if any.
- use `MutableStateFlow.update()` to update the state in a thread-safe way.
- Turn `FiltersViewModel.deleteFilter()` into a suspending function in order to perform the update in the coroutinescope of the Activity lifecycle, so the View passed as argument doesn't leak.
- Wait for an ongoing load operation to complete before performing a delete filter operation, so the state stays consistent.
- Add `Intent.withSlideInAnimation()` as a simpler and more flexible alternative to `Activity.startActivityWithSlideInAnimation(Intent)`.


app/src/main/java/com/keylesspalace/tusky/components/filters/FiltersActivity.kt
```diff
@@ -24,6 +25,12 @@ class FiltersActivity : BaseActivity(), FiltersListener {
     private val binding by viewBinding(ActivityFiltersBinding::inflate)
     private val viewModel: FiltersViewModel by viewModels()
 
+    private val editFilterLauncher =
+        registerForActivityResult(ActivityResultContracts.StartActivityForResult()) {
+            // refresh the filters upon returning from EditFilterActivity
+            loadFilters()
```
charlag: should we perhaps force load here, even if previous one is in progress?

charlag: ah I guess it's addressed in the next commit already?

charlag (approved):
didn't test it but code-wise looks fine!

app/src/main/java/com/keylesspalace/tusky/components/filters/FiltersActivity.kt
cbeyls: Yes, I changed `loadFilters()` to `reloadFilters()`.
Now filters are loaded automatically when the ViewModel is created and reloaded when `reloadFilters()` is called, cancelling the in-progress load if any. This is done using the `collectLatest()` function which cancels the processing of the previous item when a new one arrives.

cbeyls:
> didn't test it but code-wise looks fine!

I did test the screen after changes and it behaved as expected.

Issue #4459, opened 2024-05-22
Followed Hashtags UI needs an UX overhaul


igalic:
### Detailed description

Right now this UI only offers only two things:

- delete a followed hashtag (with an unintuitive looking button that looks like  this: ×👤)
- add a new one

Things i would expect from this UI:

- click on a hashtag to view the current stream
- long press on a hashtag to copy it

### Debug information

Tusky 25.2

Device:

OnePlus IV2201
Android version: 13
SDK version: 33

Account:

@meena@glitch.social
Version: 4.3.0-alpha.3

Tak:
This was completed in https://github.com/tuskyapp/Tusky/pull/4475

PR #4475, opened 2024-05-27
Followed hashtags view improvements

Tak:
Add click and longpress handler to the followed hashtags view

#4459

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/components/followedtags/FollowedTagsActivity.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/components/followedtags/FollowedTagsActivity.kt
@@ -184,7 +184,7 @@ class FollowedTagsActivity :
         return viewModel.searchAutocompleteSuggestions(token)
     }
 
-    override fun view(tagName: String) {
+    override fun viewTag(tagName: String) {
         startActivity(StatusListActivity.newHashtagIntent(this, tagName))
     }
 
--- a/app/src/main/java/com/keylesspalace/tusky/components/followedtags/FollowedTagsActivity.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/components/followedtags/FollowedTagsActivity.kt
@@ -1,14 +1,11 @@
 package com.keylesspalace.tusky.components.followedtags
 
 import android.app.Dialog
-import android.content.ClipData
-import android.content.ClipboardManager
 import android.content.DialogInterface
 import android.content.SharedPreferences
 import android.os.Bundle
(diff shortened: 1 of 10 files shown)


app/src/main/java/com/keylesspalace/tusky/components/followedtags/FollowedTagsActivity.kt
```diff
@@ -180,6 +184,15 @@ class FollowedTagsActivity :
         return viewModel.searchAutocompleteSuggestions(token)
     }
 
+    override fun view(tagName: String) {
```
charlag, line 187: `view()` is a bit of generic of a name, maybe `viewTag()`?

app/src/main/java/com/keylesspalace/tusky/components/followedtags/FollowedTagsActivity.kt
```diff
@@ -180,6 +184,15 @@ class FollowedTagsActivity :
...
+        startActivity(StatusListActivity.newHashtagIntent(this, tagName))
+    }
+
+    override fun copyTagName(tagName: String) {
+        getSystemService(ClipboardManager::class.java).setPrimaryClip(ClipData.newPlainText(null, "#$tagName"))
+        Toast.makeText(this, getString(R.string.confirmation_hashtag_copied, tagName), Toast.LENGTH_SHORT).show()
```
charlag, line 193: modern Android anyway shows popup when something is copied, I am wondering if we need anything else

charlag (approved):

app/src/main/java/com/keylesspalace/tusky/components/followedtags/FollowedTagsActivity.kt
Tak: Huh, it didn't for me (android 11), and we're manually showing a popup in other places, but I can remove it if newer android has us covered

cbeyls: From the [documentation](https://developer.android.com/develop/ui/views/touch-and-input/copy-paste#Feedback):

Users expect visual feedback when an app copies content to the clipboard. This is done automatically for users in Android 13 and higher, but it must be manually implemented in prior versions.

Tak: I moved all the slightly different "copy text to clipboard" implementations to an `Activity` extension that puts the toast behind a version check

PR #4496, opened 2024-06-06
Add option for default reply privacy set to unlisted by default

fadookie:
This PR fixes https://github.com/tuskyapp/Tusky/issues/2798 and is mostly based on and supersedes https://github.com/tuskyapp/Tusky/pull/2826 but I have fixed all merge conflicts and unit tests.

I tested the changes locally and the setting takes effect immediately for replies, and persists across killing the app.


connyduck:
It now looks as if this setting was synced with the server, which it can't be because Mastodon doesn't have this feature.



Some ideas how we could solve this:

- Move the (synced with server) out of the section title and into each setting title where it is true.
- Move the new setting into a new section "Publishing (local only)" (or something like that.

thoughs? 🤔

fadookie:
@connyduck Either is fine with me, conceptually I think it makes a bit more sense to be right where I've put it so it's right next to default post privacy. So maybe the former option you proposed is better?*

connyduck:
Ok, can you change the strings please?

fadookie:
@connyduck Okay, I've made the updates you've requested. I had to make all these settings labels multi-line so the text would fit - look good?
Screenshot_1717882942

I also cleaned up the history a little, removed some unnecessary merge commits, GPG signed and tagged the correct issue in the commit messages.

app/src/main/java/com/keylesspalace/tusky/di/StorageModule.kt
```diff
@@ -65,8 +66,11 @@ object StorageModule {
                 AppDatabase.MIGRATION_41_42, AppDatabase.MIGRATION_42_43, AppDatabase.MIGRATION_43_44,
                 AppDatabase.MIGRATION_44_45, AppDatabase.MIGRATION_45_46, AppDatabase.MIGRATION_46_47,
                 AppDatabase.MIGRATION_47_48, AppDatabase.MIGRATION_52_53, AppDatabase.MIGRATION_54_56,
-                AppDatabase.MIGRATION_58_60
+                AppDatabase.MIGRATION_58_60, AppDatabase.MIGRATION_58_60
```
connyduck, line 69:
```suggestion
                AppDatabase.MIGRATION_58_60, AppDatabase.MIGRATION_60_62
```

app/src/main/java/com/keylesspalace/tusky/di/StorageModule.kt
```diff
@@ -65,8 +66,11 @@ object StorageModule {
...
             )
             .build()
     }
+
+    @Provides
+    fun providesAccountDao(db: AppDatabase): AccountDao = db.accountDao()
```
connyduck, line 75: But `AccountDao` isn't injected anywhere? Either remove this again or actually inject it somewhere.

connyduck (changes requested):

app/src/main/java/com/keylesspalace/tusky/di/StorageModule.kt
fadookie: Oops, I had to redo this merge when I was cleaning up the history and made a mistake. Good catch.

app/src/main/java/com/keylesspalace/tusky/di/StorageModule.kt
fadookie: Good catch. This is an artifact of the merge resolution I did because `AccountDao` is now being retrieved from the DB object instead of injected. Seems simplest to remove the provider for now since we don't really need to mock it in the unit tests, I am mocking the account manager instead.

connyduck (approved):

PR #4504, opened 2024-06-10
enable dependency verification

connyduck:
https://docs.gradle.org/current/userguide/dependency_verification.html

Verification of checksums was easy to set up, but verification of signatures is more involved as gradle couldn't do it automatically for most dependencies. Checksum verification will already give us additional security though so lets try it out.


Tak (approved):
🤷

PR #4507, opened 2024-06-12
improved color scheme & theming

connyduck:
Follow up to https://github.com/tuskyapp/Tusky/pull/3921

- no more hardcoded `tusky_blue`, instead the `colorPrimary` attribute is used. This will help us when adding more themes, e.g a dynamic color one.
- The `colorPrimary` of the dark theme is now lighter for more contrast and subsequently the `colorOnPrimary` is now dark grey instead of white.
- `tusky_red_lighter` is now a bit more red than before
- Tweaked color usage in a few places for better contrast

I think this looks a bit unfamiliar but overall better and the higher contrast makes things noticeably easier to read.


app/src/main/java/com/keylesspalace/tusky/components/compose/view/ProgressImageView.kt
```diff
@@ -68,8 +71,10 @@ class ProgressImageView
     }
 
     fun setChecked(checked: Boolean) {
-        markBgPaint.color =
-            context.getColor(if (checked) R.color.tusky_blue else R.color.tusky_grey_10)
+        val backgroundColor = if (checked) com.google.android.material.R.attr.colorPrimary else android.R.attr.colorBackground
```
Lakoja, line 74: Would it make sense to import this somehow?

"com.google.android.material.R.attr.colorPrimary" looks a bit cumbersome (and lots of places, too)

app/src/main/java/com/keylesspalace/tusky/view/TuskySwipeRefreshLayout.kt
```diff
@@ -0,0 +1,39 @@
...
+import android.content.Context
+import android.util.AttributeSet
+import androidx.swiperefreshlayout.widget.SwipeRefreshLayout
+import com.google.android.material.color.MaterialColors
+
+/**
+ * SwipeRefreshLayout does not allow theming of the color scheme,
+ * so we use this class to still have a single point to change its colors.
+ */
+class TuskySwipeRefreshLayout @JvmOverloads constructor(
```
Lakoja, line 26: Should(n't) this be a style file? And all swipe to refresh have that as "style"?

Lakoja (approved):

app/src/main/java/com/keylesspalace/tusky/components/compose/view/ProgressImageView.kt
connyduck: yes definitely, thank you

app/src/main/java/com/keylesspalace/tusky/view/TuskySwipeRefreshLayout.kt
connyduck: You mean adding the missing styles here and then styling every SwipeRefreshLayout? Since the styling would be all the same and would need quite some additional lines of code I'd rather not.

Lakoja: I was thinking how all the buttons for example are style with some definition in styles.xml and the a "style=" on every usage.

But I couldn't find now how that might be done.

Tak (approved):

charlag (approved):

mcclure (approved):
Used this heavily today, light mode, could not tell the difference vs the old one, which says to me it is fine.

PR #4511, opened 2024-06-13
dependency verification: Trust sources packages used by android studio

Tak:
Sync project with gradle files fails for me in android studio without this change

See https://youtrack.jetbrains.com/issue/IDEA-258328/Dependency-verification-failed-Checksums-of-downloaded-sources-not-included-in-verification-metadata.xml


connyduck (approved):
Just run into this as well, thx!

PR #4521, opened 2024-06-18
exclude compose activity from recents

### LINKED ISSUE #4517: Sharing image to Tusky seems to leave behind residual broken post composition area?
### Detailed description

After sending a post with an image that has been shared to Tusky from another app, Tusky exits and returns you to the app that shared the image. Tusky retains some kind of existence in the recent apps view in Android though: I'm not sure if this is Android behaviour or whether Tusky isn't exiting cleanly, but I thought it was worth pointing out. There's still an entry in the recent apps for Tusky but you can't click it to open Tusky.

I seem to remember that some time ago, clicking this entry would re-upload the shared image and start a new post if you'd already sent the first post with a shared image? Perhaps this part of the code was changed since then? The current behaviour is odd / unexpected, regardless.

1. Open e.g. the photo gallery on your phone
2. Share a photo from there to Tusky. Tusky opens the post composition area and the image is uploaded
3. Send the post. For me, Tusky exits and I am returned to the gallery
4. View currently open apps in Android. The Tusky post composition area remains open in the app previews, but clicking on it opens nothing and Android says "app not available"

### Debug information

Tusky Test 25.2-125483db (but this has been happening for a while)

connyduck:
closes #4517

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -99,9 +99,13 @@
                 android:value="androidx.sharetarget.ChooserTargetServiceCompat" />
 
         </activity>
+        <!-- When content is shared to Tusky ComposeActivity can be the only active activity,
+             so it must be excluded from recents or the sharing process can be restarted from recents
+             which would be very unexpected -->
         <activity
             android:name=".components.compose.ComposeActivity"
             android:theme="@style/TuskyDialogActivityTheme"
+            android:excludeFromRecents="true"
             android:windowSoftInputMode="stateVisible|adjustResize" />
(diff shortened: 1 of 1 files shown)


connyduck:
Needs more testing, the behavior is very inconsistent between Android versions and I think there is another bug in the sharing logic.

connyduck:
Seems to work fine, I was not able to get the invalid state again where the sharing process is restarted with this change. Sometimes the app is hidden completely from recents, sometimes it is still there but MainActivity opens, which is also fine.

Tak (approved):

PR #4525, opened 2024-06-20
fix (un)muting conversations

### LINKED ISSUE #4523: Home tab: Mute then unmute a conversation -> blank timeline
### Detailed description

Open up Tusky. Scroll to the top of the Home tab. Make a post, so that the top post on your Home tab is a post by you. Tap ⋯ on this top post. Tap "Mute conversation". Tap "⋯" again. Tap "unmute conversation".

You will see this:

Screenshot_20240619-125828

The timeline will just be gone. Dragging down to refresh will make it come back.

I discovered this by accident. Note that in my testing this ONLY happens on the Home tab, it ONLY happens for the top post, and it ONLY happens if the top post is "mute conversation"able (IE by you or a reply to you).

### Debug information

I tested this with a self-built ae12ba4d84f0 (color_improvements branch)
I have not tested with a shipping Tusky.

### LINKED ISSUE #4524: Notifications tab: "Mute conversation" doesn't act "modal"
### Detailed description

Go into the notifications tab. Go to the first mention/reply. Tap ⋯. Tap "Mute conversation". Tap ⋯ again.

Expected behavior: After tapping "Mute conversation", the menu item should change to "Unmute conversation", to indicate that the conversation is now muted and the next act will unmute it. 

Observed behavior: The menu item still says "Mute conversation". If you somehow cause the tab to refresh (for example, tap the toot, then tap "back") then the menu item will correctly change to "Unmute conversation", but this will not occur automatically.

(The reverse also applies: If you tap "unmute conversation", the menu will not change to indicate the updated state.)

-

### Debug information

I tested this with a self-built ae12ba4d84f0 (color_improvements branch)
I have not tested with a shipping Tusky.

connyduck:
closes #4523 
closes #4524

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/appstore/Events.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/appstore/Events.kt
@@ -8,7 +8,6 @@ import com.keylesspalace.tusky.entity.ScheduledStatus
 import com.keylesspalace.tusky.entity.Status
 
 data class StatusChangedEvent(val status: Status) : Event
-data class MuteConversationEvent(val statusId: String, val mute: Boolean) : Event
 data class UnfollowEvent(val accountId: String) : Event
 data class BlockEvent(val accountId: String) : Event
 data class MuteEvent(val accountId: String) : Event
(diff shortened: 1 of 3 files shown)


mcclure (approved):
Makes sense to me?

Tak (approved):

PR #4542, opened 2024-07-02
prepare release v26.0 beta 1

connyduck:


charlag:
from a quick glance looks good!
why did we introduce follow preference tho instead of always doing it, uh

Tak (approved):

PR #4543, opened 2024-07-05
split out FilteredStatusViewHolder from StatusBaseViewHolder

connyduck:
This is way more efficient than before as less views need to be inflated and bound for a filtered status to be rendered. It also should fix the bug where sometimes a `StatusViewHolder` that is set up for showing a status gets bound to a status that is filtered, leading to a crash.


app/src/main/java/com/keylesspalace/tusky/components/notifications/NotificationsViewModel.kt
```diff
@@ -152,7 +152,7 @@ class NotificationsViewModel @Inject constructor(
 
     private fun shouldFilterStatus(notificationViewData: NotificationViewData): Filter.Action {
         return when ((notificationViewData as? NotificationViewData.Concrete)?.type) {
-            Notification.Type.MENTION, Notification.Type.STATUS, Notification.Type.POLL -> {
```
connyduck, line 155: Contrary to their name, `Notification.Type.STATUS` can't be filtered (they are always notifications about the users own posts)

PR #4549, opened 2024-07-07
fix updating boosts in network timeline

connyduck:
re: https://infosec.exchange/@webhat/112745609655586468

Boost were not correctly handled here, probably because they are only on profile timelines which I rarely check. This makes sure likes and boosts get correctly set to posts even when they are boosts.

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/components/timeline/viewmodel/NetworkTimelineViewModel.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/components/timeline/viewmodel/NetworkTimelineViewModel.kt
@@ -295,13 +295,7 @@ class NetworkTimelineViewModel @Inject constructor(
     }
 
     private fun handleStatusChangedEvent(status: Status) {
-        updateStatusById(status.id) { oldViewData ->
-            status.toViewData(
-                isShowingContent = oldViewData.isShowingContent,
-                isExpanded = oldViewData.isExpanded,
-                isCollapsed = oldViewData.isCollapsed
-            )
-        }
+        updateStatusByActionableId(status.id) { status }
     }
 
     override fun fullReload() {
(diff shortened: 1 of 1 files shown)


Tak (approved):

app/src/main/java/com/keylesspalace/tusky/components/timeline/viewmodel/NetworkTimelineViewModel.kt
```diff
@@ -399,23 +393,17 @@ class NetworkTimelineViewModel @Inject constructor(
...
-        val pos = statusData.indexOfFirst { it.asStatusOrNull()?.id == id }
-        if (pos == -1) return
-        updateViewDataAt(pos) { vd ->
-            if (vd.status.reblog != null) {
-                vd.copy(status = vd.status.copy(reblog = updater(vd.status.reblog)))
-            } else {
-                vd.copy(status = updater(vd.status))
+    private inline fun updateStatusByActionableId(id: String, updater: (Status) -> Status) {
+        // posts can be multiple times in the timeline, e.g. once the original and once as boost
+        statusData.forEachIndexed { index, status ->
```
charlag, line 398: hm that means scanning the whole list every time which kinda sucks but I don't have a good solution either, need some smart data structure for that index

charlag (approved):

app/src/main/java/com/keylesspalace/tusky/components/timeline/viewmodel/NetworkTimelineViewModel.kt
connyduck: yeah :/

webhat: In my head there was a hashmap which contains all toots which was updated by index. I didn't realise that you needed a full table scan 🤔

PR #4574, opened 2024-07-15
update verification metadata for tensorflow-lite-metadata

connyduck:


connyduck:
I'm not sure why this changed (I suspect it was somehow loaded from a different repo before) but the new pom file is ok

charlag (approved):

PR #4577, opened 2024-07-17
Bind account early in timeline & notification view models

charlag:
This will make it less likely that data from different accounts will get mixed up.

I checked  the codepaths and we shouldn't hit viewModel if account is not there but this is still a bit of risk to assume that so I'll investigate if we can pass in more things and use DI to make it more secure


Tak:
Looks reasonable?

connyduck (approved):

Issue #4589, opened 2024-07-28
Can't translate boosts from home timeline


jmorahan:
### Detailed description

My instance has the translation API enabled. I can translate a post when I'm viewing it individually, and I can translate a post directly from the home timeline if it's from someone I'm following, but it doesn't seem to work if it's a boost. The Translate menu item is there, but when I press it, nothing happens. However I can translate the same post if I open it individually, and I can translate it directly from the timeline on the web interface.

### Steps to reproduce the problem

1. Be on an instance with translation enabled
2. Find a post in a different language that's been boosted into your home timeline
3. Without first opening the individual post, press ... and Translate
4. Nothing happens

### Debug information

Tusky 26.0

Device:

Fairphone FP4
Android version: 13
SDK version: 33

Account:

[@jmorahan@mastodon.ie](https://mastodon.ie/@jmorahan)
Version: 4.2.10

PR #4590, opened 2024-07-28
fix translating boosts in home timeline

connyduck:
fixes https://github.com/tuskyapp/Tusky/issues/4589


charlag (commented):
I'm not sure I understand why it helps

connyduck:
We were using the id of the boost to reference its translation status. But it needs to be the id of the boosted post or nothing happens. So I changed it from `id` to `actionableId`

Tak (approved):

PR #4602, opened 2024-08-07
remove "default" from posting defaults preference names

connyduck:
This makes the interface less cluttered and matches Mastodon. The "default" is already in the preference group title, I think this is clear enough.

Before:
Screenshot_20240807_172653

After:
Screenshot_20240807_172240

Mastodon:
Screenshot 2024-08-07 at 17 21 11


Tak (approved):

mcclure (approved):
👍

Issue #4493, opened 2024-06-04
"Show more" loading the timeline jumps the timeline location to the wrong place


fin-w:
### Detailed description

It looks to me like the timeline location logic goes wrong somewhere when loading in more posts with a "load more" button.

If you have loaded in posts 0 to 100 and 150 to 200, there will be a gap 100-150 filled by the "load more" button. But when you click that button, while the correct posts are immediately displayed on screen i.e. new posts show in place of the "show more" button, any scrolling up / down loads posts as though the timeline position were somewhere up in 0-100 (I don't know how to be more exact than that sorry). Once you scroll up / down a full screen, no posts in the timeline are from your original position where the "load more" button was and the timeline only shows posts from the 0-100 region. It's like you've jumped position. You can scroll back down to your place in the timeline and everything seems fine. Loading more posts with another button causes the same issue but I think a different timeline place is jumped to.

This has been happening with the last few Tusky Test builds, probably since #4471 or possibly earlier?

### Steps to reproduce the problem

1. Sleep
2. Wake up and check the fedi- there are lots of new posts on the home timeline
3. Click a "load more" button to load / insert posts somewhere between cached posts. Currently viewed posts will not change, near the bottom of the screen new posts will be visible
4. Scroll up or down- the posts loaded in from being out of view are from a different location in the timeline (from further up / more recent, that were [I think] already cached)

### Debug information

Tusky Test 25.2-8aaca3bb

Device:
motorola moto g(7) power 
Fersiwn Android: 14 
Fersiwn SDK: 34

fin-w:
It doesn't happen every time...

fin-w:
When a jump does happen, the time of posting on visible posts switches to the time of posting on the post that is jumped to, i.e. where a post at your original position was posted 5h ago, after clicking "load more" it might display 4h ago immediately, and then after scrolling away and back will fully display the post from 4h ago, that is further up the timeline, that Tusky has replaced the original post with. So there's partial loading of data from elsewhere in the timeline, from the point posts load after clicking 'load more".

connyduck:
Are you sure this happens with build *after* #4471? Because that should be fixed there and I can't reproduce it anymore.

fin-w:
Hmm, I haven't been able to trigger it on the latest build (Tusky Test 25.2-4afce0a3) so perhaps it's gone, like you say?

fin-w:
[...]

PR #4618, opened 2024-08-18
fix "load more" progressbar sometimes not showing up

connyduck:
I don't know why exactly, but when the ViewHolder gets recycled and the ProgressBar is set to visible again, it just doesn't show anything. It seems to be a bug in the platform ProgressBar, and it doesn't happen on all devices. Replacing it with a CircularProgressIndicator fixes the bug and looks the same.
Reported here: https://github.com/tuskyapp/Tusky/issues/4493#issuecomment-2182407513


charlag (approved):

Tak (approved):
🤷

PR #4632, opened 2024-08-29
Fixes issue where alt text positioning covers video controller

### LINKED ISSUE #4165: disability to use player buttons if media alt text is long
If media alt text (for example an audio and its lyrics) is long, it can cover media player controllers, so it's not possible to use them.

Screenshot_20231212_083219_Tusky

* * * *
- Tusky Version: 24.0
- Android Version: 13

andrewhamilton0:
Fixes issue where alt text positioning covers video controller
Also adds scroll view for more responsive scrolling
closes #4165

### DIFF
--- a/app/src/main/res/layout/fragment_view_video.xml
+++ b/app/src/main/res/layout/fragment_view_video.xml
@@ -29,8 +29,7 @@
         android:layout_marginTop="?attr/actionBarSize"
         android:layout_marginBottom="160dp"
         app:layout_constraintTop_toTopOf="parent"
-        app:layout_constraintBottom_toBottomOf="parent"
-        app:layout_constraintVertical_bias="0.0">
+        app:layout_constraintBottom_toBottomOf="parent">
 
         <TextView
             android:id="@+id/mediaDescriptionTextView"
@@ -43,8 +42,6 @@
             android:textAlignment="center"
             android:textColor="#eee"
             android:textSize="?attr/status_text_medium"
-            app:layout_constraintTop_toTopOf="parent"
-            app:layout_constraintBottom_toBottomOf="parent"
             tools:text="Some media description" />
 
     </ScrollView>
--- a/app/src/main/res/layout/fragment_view_video.xml
+++ b/app/src/main/res/layout/fragment_view_video.xml
@@ -8,22 +8,6 @@
     android:clickable="true"
     android:focusable="true">
 
-    <TextView
-        android:id="@+id/mediaDescription"
-        android:layout_width="match_parent"
-        android:layout_height="wrap_content"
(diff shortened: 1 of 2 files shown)


connyduck:
Why not just limit the height in xml? E.g. to 50% of the parent.

andrewhamilton0:
> Why not just limit the height in xml? E.g. to 50% of the parent. 

I thought maybe you wanted it referenced to the height of the controls, so it could take up all available space, if needed.

I couldn't find a way to reference the controls on xml since it's built within the player and builds at runtime. So this is the solution I found.

If you prefer the 50 percent of the parent in xml, I can do that

connyduck:
Well `findViewById` on ids we don't control is very hacky and may break on every library update, so I'd really like to avoid that if possible.

Maybe we can estimate the height of the player controls and give the description a bottom margin?

(in the long run this screen should be redesigned, something where you can drag a handle to switch between video and description)

app/src/main/res/layout/fragment_view_video.xml
```diff
@@ -38,6 +22,33 @@
...
+            android:layout_height="wrap_content"
+            android:background="#60000000"
+            android:hyphenationFrequency="full"
+            android:lineSpacingMultiplier="1.1"
+            android:padding="8dp"
+            android:textAlignment="center"
+            android:textColor="#eee"
+            android:textSize="?attr/status_text_medium"
+            app:layout_constraintTop_toTopOf="parent"
+            app:layout_constraintBottom_toBottomOf="parent"
```
connyduck, line 47:
```suggestion
```

these don't do anything if the parent isn't a `ConstraintLayout`

app/src/main/res/layout/fragment_view_video.xml
```diff
@@ -38,6 +22,33 @@
...
 
+    <ScrollView
+        android:id="@+id/mediaDescriptionScrollView"
+        android:layout_width="match_parent"
+        android:layout_height="0dp"
+        android:layout_marginTop="?attr/actionBarSize"
+        android:layout_marginBottom="160dp"
+        app:layout_constraintTop_toTopOf="parent"
+        app:layout_constraintBottom_toBottomOf="parent"
+        app:layout_constraintVertical_bias="0.0">
```
connyduck, line 33:
```suggestion
        app:layout_constraintBottom_toBottomOf="parent">
```

this doesn't do anything

connyduck (changes requested):
Nice, works very well

app/src/main/res/layout/fragment_view_video.xml
andrewhamilton0: Resolved

connyduck (approved):

Issue #4573, opened 2024-07-15
On-screen keyboard pops up when navigating back from a search result


abackstrom:
### Detailed description

Performing a search dismisses the on-screen keyboard, as expected. If you navigate into any result (post, account, or hashtag) and press the Back nav button, the on-screen keyboard reappears. This feels like a bug: in most cases I want to keep browsing the results, not change my search.

### Steps to reproduce the problem

1. Search for any term. Observe that the on-screen keyboard is dismissed.
2. Tap into one of the results.
3. Tap the "Back" button in the bottom nav bar. Observe that the search box is focused and the keyboard reappears.

### Debug information

Tusky 26.0

Device:

samsung SM-F711B
Android version: 14
SDK version: 34

Account:

@annika@xoxo.zone
Version: 4.2.10

PR #4633, opened 2024-08-29
Minimizes keyboard on search screen

andrewhamilton0:
Adds feature if user clicks on toolbar, on the pages box, or on the tabs buttons that the keyboard will minimize and focus will be taken off of search bar.

Further addresses issue #4573

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/components/search/SearchActivity.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/components/search/SearchActivity.kt
@@ -106,7 +106,7 @@ class SearchActivity : BottomSheetActivity(), MenuProvider, SearchView.OnQueryTe
             searchView.clearFocus()
         }
     }
-    private fun setupClearFocusOnClickListeners(){
+    private fun setupClearFocusOnClickListeners() {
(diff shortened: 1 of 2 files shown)


connyduck:
I'm not sure this is necessary

andrewhamilton0:
> I'm not sure this is necessary

I noticed that after I performed a search, the text box becomes unfocused as expected. But then I went to edit the search, clicked the text box,  and then I decided not to edit it. The only way to get rid of the focus was to perform the search, even if I clicked into pager items and back. 

I was confused why it was doing this at first as I thought I fixed the bug but I had forgot I re-clicked on the text box without editing and completing a search.

It's not the biggest deal, but I do find it as a nice feature, but I do understand if you find it unnecessary.

connyduck:
Ok, I tried it out and I guess it makes sense. But the keyboard should alse be hidden when switching tabs, no?

andrewhamilton0:
> Ok, I tried it out and I guess it makes sense. But the keyboard should alse be hidden when switching tabs, no?

Yes you're right! I went ahead and added that functionality as well!

connyduck (approved):

PR #4636, opened 2024-08-29
fix account switching

### LINKED ISSUE #4629: Notifications attributed to incorrect account
### Detailed description

I have two accounts logged in tusky, each notification is attributed to the other account--clicking through opens the wrong account in tusky

1. Two accounts linked in tusky
2. Get a like on post of account A
3. Like is attributed to account B and clicking it opens tusky in account B
...

### Debug information

Tusky 26.1, Android

### LINKED ISSUE #4631: Can't login with multiple accounts on same server/domain as of version 26
### Detailed description

I have different accounts on the same server (personal and bots).

I can log in to the first one fine. If I log in to the second one, the first account will not be shown anymore as if I only had one account in Tusky. The timeline also doesn't change and will show the items from the first account. Sidebar will show the new one.

If I log in to the third one, the sidebar account will show that one but the timeline again does not change.

If I log out of the account, Tusky will revert to having no accounts configured.

I can't tell when it started sadly. I had the timeline/notification mixup in 26.0 and deleted and reinstalled 26.1 from F-Droid to start fresh. That version at least is affected.

Not sure if it matters, this is a gotosocial instance. It worked fine with 26.0 before reinstalling Tusky. I think I originally added all accounts with version 25 or 24.

1. Add account 1 on a Domain A
2. Add another account 2 also on Domain A
3. Only account 2 will be shown added to Tusky, account 1 will be gone as if it was never added
4. Additionally, if you log out of account 2, no account will be configured, account 1 won't come back.

### Debug information

Tusky 26.1 F-Droid
Android 14
Fairphone 5

Update with additional info: I installed 25.2 from the F-Droid archive and it works correctly with that version.

connyduck:
closes #4631 
closes #4629 

and other weirdness introduced in Tusky 26.1.
I did a lot of testing on 2 physical devices and multiple emulators. It definitely is better than before, but probably still not perfect.


connyduck:
Tested some more, I think it is good.

Tak (approved):
😅

PR #4637, opened 2024-08-31
Material Design 3

connyduck:
I tried to find a balance between going fully M3 and keeping some of the original Tusky feeling.
For example, I removed the "allCaps" setting we had on most buttons, which is recommended for M3. On the other hand, I made them less rounded than the M3 default.


app/src/main/res/values/styles.xml
```diff
@@ -53,7 +53,17 @@
...
+        <item name="colorOnPrimaryContainer">@color/colorOnPrimary</item>
+
         <item name="colorSurface">@color/colorSurface</item>
+        <item name="colorSurfaceVariant">@color/colorSurface</item>
+
+        <item name="colorSurfaceContainerLowest">@color/colorSurface</item>
+        <item name="colorSurfaceContainerLow">@color/colorSurface</item>
+        <item name="colorSurfaceContainer">@color/colorSurface</item>
+        <item name="colorSurfaceContainerHigh">@color/colorSurface</item>
+        <item name="colorSurfaceContainerHighest">@color/colorSurface</item>
```
connyduck, line 66: These are supposed to indicate containers with different importance, but setting the accordingly looks weird or doesn't provide enough contrast.

app/src/main/res/layout/activity_view_thread.xml
```diff
@@ -6,22 +6,9 @@
...
-            style="@style/Widget.AppCompat.Toolbar"
-            android:layout_width="match_parent"
-            android:layout_height="wrap_content"
-            app:contentInsetStartWithNavigation="0dp"
-            app:layout_scrollFlags="scroll|enterAlways" />
-
-    </com.google.android.material.appbar.AppBarLayout>
+    <include
+        android:id="@+id/includedToolbar"
+        layout="@layout/toolbar_basic" />
```
connyduck, line 11: For some reason, this was the only layout having its own toolbar and not including `toolbar_basic`. The toolbar scrolled with the content, unless everywhere else. I don't care either way, I just think toolbars should either scroll everywhere or nowhere. What do you prefer?

app/src/main/java/com/keylesspalace/tusky/components/compose/ComposeViewModel.kt
```diff
@@ -331,13 +331,6 @@ class ComposeViewModel @Inject constructor(
         mediaUploader.cancelUploadScope(*_media.value.map { it.localId }.toIntArray())
     }
 
-    fun shouldShowSaveDraftDialog(): Boolean {
-        // if any of the media files need to be downloaded first it could take a while, so show a loading dialog
-        return _media.value.any { mediaValue ->
-            mediaValue.uri.scheme == "https"
-        }
-    }
-
```
connyduck, line 340: This code path was impossible to hit (we don't offer a "save draft" option when editing a post).

Tak (commented):
Looks fine to me? I'll try it on device for a while

Tak:
Should the avatar in the home bar have rounded corners? (android 11, avatar is an animated gif)

Screenshot_20240902-104226

Tak:
I find the change away from all-caps on the buttons a little jarring, maybe I'm just to too used to The Old Way

connyduck:
> Should the avatar in the home bar have rounded corners? (android 11, avatar is an animated gif)
> 
> Screenshot_20240902-104226

Yes it should, but this has nothing to do with this branch (also no rounded corners on develop).

Tak (approved):
Code changes look ok to me, I've been running [022208e](https://github.com/tuskyapp/Tusky/pull/4637/commits/022208e937a89def7bda5880a982ac0dd1bf1db5) with no issues

PR #4673, opened 2024-09-12
Material3 polishing

### LINKED ISSUE #4692: The screen after reporting an account shows loading circles over block / mute buttons
### Detailed description

Report someone then see the buttons that come up in the view that follows. There are loading circles going round over the text, in the same colour as the background so they're hard to see.

_No response_

### Debug information

Tusky Test 26.2-f2a2c164

connyduck:
see commit messages for a list of changes

closes #4692


connyduck:
should be done for now

connyduck:
The biggest difference is probably how the help text looks now

Screenshot_20240915_170237

app/src/main/java/com/keylesspalace/tusky/components/report/fragments/ReportDoneFragment.kt
```diff
@@ -49,11 +47,11 @@ class ReportDoneFragment : Fragment(R.layout.fragment_report_done) {
             viewModel.muteState.collect {
                 if (it == null) return@collect
                 if (it !is Loading) {
-                    binding.buttonMute.show()
-                    binding.progressMute.show()
+                    binding.buttonMute.visibility = View.VISIBLE
+                    binding.progressMute.visibility = View.GONE
```
Tak, line 51: Is it intentional to change the behavior here? Was it wrong?

Tak (approved):

app/src/main/java/com/keylesspalace/tusky/components/report/fragments/ReportDoneFragment.kt
connyduck: Yes intentional because of two reasons:
- The `CircularProgressIndicator` has methods `show()` / `hide()` that don't do exactly what is expected, causing #4692 
- Setting the buttons to invisible instead of gone doesn't reflow the layout which is nicer

Tak (approved):

Issue #4236, opened 2024-01-21
Fully migrate Bitrise to GitHub Actions


Goooler:
Bitrise is set up in #214. Years have passed, and the ecosystem of GitHub Actions is getting stronger and faster, supporting more features. There are the same CPU cores on both Bitrise and GitHub devices, now it's time to migrate to GitHub actions completely.

Check 
- Bitrise: https://scans.gradle.com/s/mue5ayj32owgy
- GitHub Actions: https://scans.gradle.com/s/gw6ikdulzk7yu

Goooler:
Part of GHA support has been finished in 
- #3728
- #3904

Goooler:
With Gradle 8.6 and GBA v3, we can store CC data on CI, see
- https://docs.gradle.org/8.6/release-notes.html
- https://github.com/gradle/gradle-build-action/releases/tag/v3.0.0

PR #4682, opened 2024-09-14
upload to Google Play with GitHub Actions, remove Bitrise

connyduck:
closes https://github.com/tuskyapp/Tusky/issues/4236

I hope I did everything correctly, we will find out after this is merged.

I am setting up the environment variables containing the keys so they are only available to actions that need them.

This should be easier to maintain, GitHub seems to be faster then Bitrise, and we no longer have two checks on every pull request.


charlag:
Looks good so far!

.github/workflows/check-and-build.yml
```diff
@@ -1,15 +1,11 @@
-name: CI
+name: Check and build
 
 on:
-  push:
-    tags:
-      - '*'
   pull_request:
```
Goooler, line 4:
```suggestion
  push:
    branches:
      - develop
  pull_request:
```

.github/workflows/check-and-build.yml
```diff
@@ -20,14 +16,11 @@ jobs:
...
-        uses: gradle/actions/wrapper-validation@v4
-
       - name: Copy CI gradle.properties
         run: mkdir -p ~/.gradle ; cp .github/ci-gradle.properties ~/.gradle/gradle.properties
 
       - name: Gradle Build Action
-        uses: gradle/gradle-build-action@v3
+        uses: gradle/actions/setup-gradle@v4
         with:
           cache-read-only: ${{ github.ref != 'refs/heads/main' && github.ref != 'refs/heads/develop' }}
```
Goooler, line 25:
```suggestion
```

.github/workflows/deploy-release.yml
```diff
@@ -0,0 +1,45 @@
...
+
+      - uses: r0adkll/sign-android-release@f30bdd30588842ac76044ecdbd4b6d0e3e813478
+        name: Sign Tusky Blue aab
+        id: sign_aab
+        with:
+          releaseDirectory: app/build/outputs/bundle/blueRelease
+          signingKeyBase64: ${{ secrets.KEYSTORE }}
+          alias: ${{ secrets.KEY_ALIAS }}
+          keyStorePassword: ${{ secrets.KEYSTORE_PASSWORD }}
+          keyPassword: ${{ secrets.KEY_PASSWORD }}
```
Goooler, line 27: We can replace this step with:

https://github.com/LawnchairLauncher/lawnchair/blob/2d86033f7a1f0e3fd32ba395118d80e73832c2d3/.github/workflows/ci.yml#L44-L53

https://github.com/LawnchairLauncher/lawnchair/blob/2d86033f7a1f0e3fd32ba395118d80e73832c2d3/build.gradle#L194-L210

connyduck: I'd prefer it if we don't touch the build.gradle files

.github/workflows/check-and-build.yml
connyduck: Why? The action already that runs on develop already includes this action, this would run it twice?

Goooler: You mean `.github/workflows/populate-gradle-build-cache.yml`? It has been deleted in this PR.

.github/workflows/deploy-release.yml
Goooler: I'll address that in a separated PR.

.github/workflows/check-and-build.yml
connyduck: No, I meant the `deploy-test` has `uses: ./.github/workflows/check-and-build.yml` and runs on develop, but it actually doesn't 😅 
I'll fix that, thanks

Tak (approved):
Looks reasonable to me? (I'm neither an expert in github actions nor in publishing to play store)

connyduck:
Yeah, neither am I, I fully expect some kind of mistake but we will only know when we try it out.

PR #4689, opened 2024-09-17
Recreate activities on screenLayout change

### LINKED ISSUE #4255: Folding Phone Messes With Layout
When moving Tusky between different activity states with my folding phone (folding, unfolding, moving, splitting, merging) the layout can become slightly broken. Normally it takes up the expected size but for some reason even force stopping Tusky doesn't resolve this.

Screenshot_20240214-193732
Screenshot_20240214-193745
Screenshot_20240214-194205

* * * *
- Tusky Version: 24.1
- Android Version: U
- Android Device: Pixel Fold

- [X] I searched or browsed the repo’s other issues to ensure this is not a duplicate. (Specifically "is:issue folding")

JimmyJichi:
Fixes #4255 

Tested on Pixel 9 Pro Fold.

Before

https://github.com/user-attachments/assets/dd4d3a06-fb9f-4d89-94c3-401875cfe6f7

After

https://github.com/user-attachments/assets/dac01631-1add-48f4-9cb9-7365b3da2c9c

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -40,7 +40,7 @@
         <activity android:name=".components.login.LoginWebViewActivity" />
         <activity
             android:name=".MainActivity"
-            android:configChanges="orientation|screenSize|keyboardHidden|screenLayout|smallestScreenSize"
+            android:configChanges="orientation|screenSize|keyboardHidden|smallestScreenSize"
             android:exported="true"
             android:theme="@style/SplashTheme">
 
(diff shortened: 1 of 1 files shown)


connyduck:
If it only was that easy 🥺
This should only be done once MainActivity has a ViewModel, otherwise it will reload on every layout change. But this is my top priority, I'll add that soon and then merge this PR.

JimmyJichi:
> This should only be done once MainActivity has a ViewModel, otherwise it will reload on every layout change.

I can see a brief reload animation, but the timeline stays where it was. I'd argue that's acceptable. Definitely better than having a broken layout.

[output.webm](https://github.com/user-attachments/assets/8f70cf65-12ae-4615-a127-96eed6ab8ed6)

connyduck (approved):
thx

PR #4705, opened 2024-10-02
don't spam endless posts when trying to schedule on GoToSocial

### LINKED ISSUE #4703: Schedule Toot button causes repeated duplicated posts on GoToSocial
### Detailed description

When trying to schedule a post on Tusky connected to a GoToSocial instance (which doesn't support scheduling), instead of erroring, Tusky will post the message repeated indefinitely until Android application is forced stop.

More Context: https://go.rocha.social/@bruno/statuses/01J95XX6VZ3FW5K6GRR7V4H94K 

1. Using Tusky connect to a gotosocial instance (I can provide an account for testing)
2. Write a post
3. Click schedule button and set 3 hours ahead
4. Click post

Expected: Tusky would report "Error scheduling, post is saved as a draft"

What happens: Tusky will post and then indefinitelly repeat the same post, making the timeline to be spammed with multiple posts with the same content.

GTS issues:
https://github.com/superseriousbusiness/gotosocial/issues/1006 
https://github.com/superseriousbusiness/gotosocial/issues/2723
...

### Debug information

26.2 - Android 11 - From Google Play

connyduck:
closes #4703

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/appstore/Events.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/appstore/Events.kt
@@ -4,7 +4,6 @@ import com.keylesspalace.tusky.TabData
 import com.keylesspalace.tusky.entity.Account
 import com.keylesspalace.tusky.entity.Notification
 import com.keylesspalace.tusky.entity.Poll
-import com.keylesspalace.tusky.entity.ScheduledStatus
 import com.keylesspalace.tusky.entity.Status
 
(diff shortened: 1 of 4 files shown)


Tak (approved):
Should we file an issue with gts? Or is the response not specified?

connyduck:
> Should we file an issue with gts? Or is the response not specified?

If they implement one of the two things linked in #4703 it should be fine

PR #4712, opened 2024-10-08
include filtered notifications in the notification tab

connyduck:
This is a quick workaround to have Tusky behave the same on Mastodon 4.2 and 4.3 instances: All notifications are shown. This should prevent confusion for users that are using mostly Tusky.

Note, I'm also working on full support for filtered notifications, but that will still take a while.

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/components/notifications/NotificationsRemoteMediator.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/components/notifications/NotificationsRemoteMediator.kt
@@ -73,7 +73,8 @@ class NotificationsRemoteMediator(
                         // so already existing placeholders don't get accidentally overwritten
                         sinceId = topPlaceholderId,
                         limit = state.config.pageSize,
-                        excludes = excludes
+                        excludes = excludes,
+                        includeFiltered = true
                     )
 
                     val notifications = notificationResponse.body()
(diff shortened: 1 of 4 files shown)


mcclure:
Conny, I think this is possibly bad, because I think this is going to include "limited user" notifications which would not have been shown in 4.2 (e.g., users that your instance moderators believe may be bad actors).

I don't remember if we discussed this— would it be a large block of work to have *just* a preference control for which notifications are shown, with no way to see the filtered notifications within Tusky? (Possibly the "filtered notifications" preference pane could have a small noninteractive box at the bottom, explaining filtered notifications will be visible on the website but not currently Tusky.) It seems like the filtered notification display is the hard part, not viewing/setting the preference.

connyduck:
As far as I understand, notifications filtered by server admins will just never be returned, no matter what `include_filtered` is set to. @ClearlyClaire can you confirm that?

ClearlyClaire:
Notifications from limited accounts are created as filtered according to the user's notifications policy, which by default puts those in filtered notifications, and do not drop them (contrary to pre-4.3.0).

mcclure:
@ClearlyClaire , thanks!

@connyduck , if you feel confident we've reproduced 4.2.x behavior, then my objection is dropped.

But personally I am still confused, I'm sorry :(

Here is my understanding:
* For each of the five notification-filter categories, the user may set "ACCEPT", "FILTER", or "IGNORE".
* If the notification-filter category applies, and the user has set the category to FILTER, the notification gets marked "FILTERED".
* Per Claire's comment above, the default notification-filter settings are ACCEPT for the basic four categories, and FILTER for "Moderated Accounts".
* If we send `includeFiltered = true` (this patch), then FILTERED notifications are shown in Tusky.

So consider two hypothetical users, and how this patch would affect them.

* User A has not engaged with the notification filter feature. They have the default, which is ACCEPT/ACCEPT/ACCEPT/ACCEPT/FILTER.  If we do nothing, the 4.2.x behavior is reproduced (moderated accounts are hidden from the UI; everything else is shown). If we ship this patch, the 4.2.x behavior *changes*: Moderated accounts, which as Claire says above were previously "dropped", are now shown.
* User B has made full use of the notification filter feature. They have set FILTER for one of the basic four categories. If we do nothing, the 4.2.x behavior changes (posts they previously would have seen are now fully hidden). If we include this patch, it acts like 4.2.x and all these posts are shown (although *also* they will see "moderated account" posts; but this user knows how notification filters work, so if they don't like seeing moderated posts, they can set moderated accounts to HIDE and reproduce 4.2.x in Tusky).

Am I missing something?

mcclure:
Oh. I'd be very frustrated if Tusky were hiding unsolicited DMs. Then I guess I drop my objections to the patch :( :(

PR #4717, opened 2024-10-09
add checkout step to deploy pipelines

connyduck:
Re: https://github.com/tuskyapp/Tusky/pull/4682#issuecomment-2402485451
Ok that was quick.

According to the error message, that should fix the problem.
https://github.com/tuskyapp/Tusky/actions/runs/11256946411/job/31300028953


Tak (approved):

connyduck:
> According to the error message, that should fix the problem.

No it doesn't >_>

Issue #4720, opened 2024-10-10
[Bug] emotes inside of links are not rendered or duplicated


Sirs0ri:
### Detailed description

Hi! 

When a link text contains an Emote, Tusky doesn't render it correctly.

If the emote is at the end of the link, the link text will consist of _only_ the emote, no matter what other text there is before it, and the `(domain)` part after the link will show the emote instead, a second time.
If the emote is anywhere else in the link's text, it'll be displayed as its `:emote:` representation instead.

Only custom emotes seem to be affected, not emojis.

This is a test post with all different combinations of emotes and links: https://corteximplant.com/@Sirs0ri/113282965726445127

This is how the mastodon frontend renders the post (correctly): 
![grafik](https://github.com/user-attachments/assets/b84b0d15-cab7-41a4-9e70-6de6eac37713)

This is how Tusky renders the post (incorrectly) - Note "emote last" for the duplicated emote, and all other "emote" links for the text-only emotes:
![Screenshot_20241010_135535_Tusky Test](https://github.com/user-attachments/assets/68fa1b02-53e2-4fb6-9fa8-4711746f26a5)

The content of the post was as follows, it was posted from a glitch-soc instance with markdown rendering enabled:
```md
link only: [link text](https://corteximplant.com)

no space:
emote first: [:cyberdon: link text](https://corteximplant.com)
emoji first: [😃 link text](https://corteximplant.com)

emote last: [link text :cyberdon:](https://corteximplant.com)
emoji last: [link text 😃](https://corteximplant.com)

space at start/end of link:
emote first: [ :cyberdon: link text](https://corteximplant.com)
emoji first: [ 😃 link text](https://corteximplant.com)

emote middle: [link :cyberdon: text](https://corteximplant.com)
emoji middle: [link 😃 text](https://corteximplant.com)

emote last: [link text :cyberdon: ](https://corteximplant.com)
emoji last: [link text 😃 ](https://corteximplant.com)
```

### Steps to reproduce the problem

1. Create a post containing a link, e.g. from a glitch-soc instance. The link's text should:
  a. end with an emote (not a space!)
  b. have an emote somewhere in it
2. Compare the post between the Mastodon Fronted, and Tusky

### Debug information

Tusky Test 26.2-2a4e785b

Device:

samsung SM-G970F 
Android-Version: 12 
SDK-Version: 31

Account:

@Sirs0ri@corteximplant.com 
Version: 4.4.0-alpha.1+glitch.cortex

[...]

PR #4723, opened 2024-10-10
fix how emojis in link texts are displayed

connyduck:
Screenshot_20241010_204357

closes https://github.com/tuskyapp/Tusky/issues/4720


Tak (approved):
Does this change RTL behavior?

connyduck:
> Does this change RTL behavior?

Good point, I made [this post](https://akko.chir.rs/notice/AmtCzU34QtoTX4AiHY) and it looks the same before and after.

PR #4742, opened 2024-10-28
never filter own posts

connyduck:
This is to match Mastodon web behavior.

Also, make revealing filtered boosts in non-cached timelines work (can only happen on user profiles, other timelines don't have boosts).

found thanks to this: https://tech.lgbt/@darkfox/113378644538792719


app/src/main/java/com/keylesspalace/tusky/components/notifications/NotificationsViewModel.kt
```diff
@@ -153,6 +153,9 @@ class NotificationsViewModel @Inject constructor(
         return when ((notificationViewData as? NotificationViewData.Concrete)?.type) {
             Notification.Type.MENTION, Notification.Type.POLL -> {
                 notificationViewData.statusViewData?.let { statusViewData ->
+                    if (statusViewData.status.account.id == account.accountId) {
```
Tak, line 156: This doesn't need to use the actionable status too?

connyduck: No, because there are no boosts on the notifications page

Tak (approved):

PR #4743, opened 2024-10-28
fix updating filter expiration to indefinite

connyduck:
Before we would not send `expires_in` when "indefinite" was selected. But that left the expiration at the value it was before. To actually set it to indefinite we need to send `expires_in`, but leave it empty.
With a value class this was actually really nice to fix, the code now self-documents what the special values mean.

Also fixes a regression from the Material 3 redesign where the filter duration drop down would not get populated when creating a filter.

Found while working on https://github.com/tuskyapp/Tusky/pull/4742

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/components/filters/FilterExpiration.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/components/filters/FilterExpiration.kt
@@ -1,0 +1,39 @@
+/* Copyright 2024 Tusky contributors
+ *
+ * This file is a part of Tusky.
+ *
+ * This program is free software; you can redistribute it and/or modify it under the terms of the
+ * GNU General Public License as published by the Free Software Foundation; either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * Tusky is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even
+ * the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU General
+ * Public License for more details.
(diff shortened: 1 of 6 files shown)


connyduck:
Not sure if this correct yet, will need to check again with fresh eyes.

charlag (commented):
skimmed through but looks good

app/src/main/java/com/keylesspalace/tusky/components/filters/FilterExpiration.kt
```diff
@@ -0,0 +1,39 @@
...
+@JvmInline
+value class FilterExpiration private constructor(val seconds: Int) {
+
+    override fun toString(): String {
+        return if (seconds < 0) "" else seconds.toString()
+    }
+
+    companion object {
+        val unchanged: FilterExpiration?
+            get() = null
```
Tak, line 33: Does this even need to be a property? It could just be a const val, right?

Tak (approved):
Looks reasonable to me, I also didn't try yet

app/src/main/java/com/keylesspalace/tusky/components/filters/FilterExpiration.kt
connyduck: yes

connyduck:
Yeah so the `getExpirationForDurationIndex` was unnecessarily complex, that confused me a bit. Now it is shorter and everything works well on Mastodon. On GTS I found a little bug: https://github.com/superseriousbusiness/gotosocial/issues/3497

Tak (approved):

PR #4761, opened 2024-11-15
mastodon-web-like trailing hashtag bar

Tak:
Rationale: Since the mastodon web UI has started stripping "trailing" hashtags from post content and shoving it into an ellipsized section at the bottom of posts, the general hashtag : content ratio is rising.

This is an attempt at adopting a similar functionality for Tusky.

Before:

Screenshot of a hashtag-heavy post on Tusky nightly

After:
Screenshot of the same post on this branch

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/util/LinkHelper.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/util/LinkHelper.kt
@@ -107,20 +107,11 @@ fun setClickableText(
     if (showHashtagBar) {
         trailingHashtagView?.apply {
             text = SpannableStringBuilder().apply {
-                var offset = 0
                 tags?.forEachIndexed { index, tag ->
                     val text = "#${tag.name}"
-                    append(text)
-                    setSpan(
-                        getCustomSpanForTag(text, tags, URLSpan(tag.url), listener),
-                        offset,
-                        offset + tag.name.length + 1,
-                        0
-                    )
-                    offset += tag.name.length + 1
+                    append(text, getCustomSpanForTag(text, tags, URLSpan(tag.url), listener), 0)
                     if (index != tags.lastIndex) {
                         append(" ")
-                        offset += 1
                     }
                 }
             }
--- a/app/src/main/java/com/keylesspalace/tusky/util/LinkHelper.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/util/LinkHelper.kt
@@ -79,7 +79,7 @@ fun setClickableText(
     mentions: List<Mention>,
     tags: List<HashTag>?,
     listener: LinkListener,
-    trailingHashtagView: TextView?,
+    trailingHashtagView: TextView? = null,
 ) {
     val spannableContent = markupHiddenUrls(view, content)
     val (endOfContent, trailingHashtags) = when {
--- a/app/src/main/java/com/keylesspalace/tusky/util/LinkHelper.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/util/LinkHelper.kt
@@ -129,7 +129,7 @@ fun setClickableText(
 }
 
 private val trailingHashtagExpression by unsafeLazy {
-    Pattern.compile("""$wordBreakExpression(#$hashtagExpression$wordBreakFromSpaceExpression+)*""", Pattern.CASE_INSENSITIVE)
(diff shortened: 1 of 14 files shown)


connyduck:
I'm not a fan of the horizontal scrolling 🤔 
What to you think about making the `TextView` that shows the hashtags in timelines just 1 or 2 lines and then ellipsize it if there would be more hashtags? User who want to see all of them can go to the details view.

⚠️ This implementation currently messes a bit with the direct messages timeline, trailing hashtags are just not shown there. Not that many people use hashtags in DMs, but still.

Tak:
🤔 That could work, it will just mean splitting out the code for the detailed view. Do we still want them in the lower area in the detailed view, or put them back into the post content in that case?

Oops, I never use the dm timeline, I'll fix it

connyduck:
> Do we still want them in the lower area in the detailed view, or put them back into the post content in that case?

I'd be fine with both

app/src/main/java/com/keylesspalace/tusky/util/StringUtils.kt
```diff
@@ -3,10 +3,16 @@
...
 import android.text.Spanned
+import java.util.regex.Pattern
 import kotlin.random.Random
 
 private const val POSSIBLE_CHARS = "0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZ"
 
+const val WORD_BREAK_EXPRESSION = """(^|$|[^\p{L}\p{N}_])"""
+const val WORD_BREAK_FROM_SPACE_EXPRESSION = """(^|$|\s)"""
+const val HASHTAG_EXPRESSION = "([\\w_]*[\\p{Alpha}_][\\w_]*)"
+val hashtagPattern by unsafeLazy { Pattern.compile(HASHTAG_EXPRESSION, Pattern.CASE_INSENSITIVE or Pattern.MULTILINE) }
```
connyduck, line 14: I think `unsafeLazy` is unnecessary here since it is a top level property.
`unsafeLazy` is for cases where you don't want class members to be initialized at class creation time but at first access (e.g. to make sure a `Context` is available)

app/src/main/java/com/keylesspalace/tusky/util/LinkHelper.kt
```diff
@@ -70,27 +71,113 @@ fun getDomain(urlString: String?): String {
...
+ * @param trailingHashtagView a text view to fill with trailing / out-of-band hashtags
  */
 fun setClickableText(
     view: TextView,
     content: CharSequence,
     mentions: List<Mention>,
     tags: List<HashTag>?,
-    listener: LinkListener
+    listener: LinkListener,
+    trailingHashtagView: TextView?,
```
connyduck, line 82: If you give this a default value of `null` you don't need to change all the callsites

app/src/main/java/com/keylesspalace/tusky/util/LinkHelper.kt
```diff
@@ -70,27 +71,113 @@ fun getDomain(urlString: String?): String {
...
+                tags?.forEachIndexed { index, tag ->
+                    val text = "#${tag.name}"
+                    append(text)
+                    setSpan(
+                        getCustomSpanForTag(text, tags, URLSpan(tag.url), listener),
+                        offset,
+                        offset + tag.name.length + 1,
+                        0
+                    )
+                    offset += tag.name.length + 1
```
connyduck, line 120: `SpannableStringBuilder` has a method [`append` that takes the text and the span at once](https://developer.android.com/reference/android/text/SpannableStringBuilder#append(java.lang.CharSequence,%20java.lang.Object,%20int)), using it should get rid of the offset calculations here

app/src/main/java/com/keylesspalace/tusky/util/LinkHelper.kt
```diff
@@ -70,27 +71,113 @@ fun getDomain(urlString: String?): String {
...
+}
+
+private val newlineCharacters = arrayOf('\n', '\r').toCharArray()
+
+/**
+ * Get a reversed sequence of start/end line indices (not including the newline characters)
+ * for lazily searching backward by line.
+ * The idea is that we're very likely to early-out, so we want to avoid splitting lots of long messages
+ * in their entirety just to look at the last line or two.
+ * Maybe it's overkill? 🤷
```
connyduck, line 161: I think I would go with a simpler implementation and see how it works in practice

connyduck (commented):
I like how this looks now!

connyduck (approved):
Let's include this in the next release?

PR #4768, opened 2024-11-24
Notification policy

### LINKED ISSUE #4331: Mastodon.social just changed how DM notifications work, and we must add a new feature or some content may be hidden
### Detailed description

The server-side filtering Mastodon does for DMs has suddenly changed. Mastodon.social has already deployed the new version of the feature *right now*, and other Mastodon-Ruby servers will pick up the new feature when 4.3.0 ships. The new server-side feature requires Tusky to implement one or more client-side features.

Previously, Mastodon had a setting "Block direct messages from accounts you don't follow" in the Preferences tab. When this setting was checked, DMs from non-followed accounts were silently hidden (in other words, it acted like "Mute", not "Block").

As of two weeks ago, mastodon.social [changed this](https://github.com/mastodon/mastodon/pull/29366). I'm not aware of any announcement of this outside of the pull request and I discovered it by accident this morning.

There is now a large bar at the top of the notification filters pane (*not* the preferences*), listing various options that put a notification into "filtered notification" state:

image

The last of these options, "Filter out notifications from… unsolicited private mentions", is **on by default for new users**. Actually that's not true, it's on by default for everyone, but [according to Claire from Mastodon that was a bug](https://social.sitedethib.com/@Claire/112134404966364037) so apparently soon it will be ratcheted back down and only new users (and users who specifically chose to have the flag on) will have it on.

When these "filter" options are engaged, if a "filtered" post is received, it will look like this:

image

A "N notifications filtered" box appears at the top of the notification column. When clicked, a new column opens showing each *user* who has sent you filtered notifications. If you click on a username, it opens a third column showing [just the notifications from the user in question](https://github.com/tuskyapp/Tusky/assets/277318/a6a5db6c-16ff-4fa1-8991-84f7e3ab6ee7) (screenshot). To make the "N notifications filtered" box go away, you must click the "Dismiss" or "Accept" button next to each of the listed users.

There is a problem: This new feature is server-side, and [according to Claire from Mastodon](https://social.sitedethib.com/@Claire/112134341381523122)

>  both the old thing and the new thing are performed server-side, clients that don't support the new thing will just not show you the “Filtered notifications” prompt and won't let you access the filtered notifications.

In summary, **there is a new feature that is on by default for some users, and unless Tusky proactively adds support for this feature, this feature will hide content from Tusky users in such a way they cannot know content is being hidden, and cannot turn the feature off, unless they go to the website**.

## Expected behavior

The new feature is [only partially documented](https://github.com/mastodon/mastodon/pull/29366) (according to @renchap Mastodon has "an internal ticket" to add documentation) but consists of four new endpoints; a list of "notification requests" (it appears to be poll only); a RPC to accept a notification request; an RPC to reject a notification requests; and an RPC to change the server-side notification-filter "policy" (the four checkboxes above, I assume).

I think **Tusky 25 should, at minimum, contain a GUI frontend for `notifications/policy`**. As far as I know we have no checkbox for the old "Mute direct messages from people you don't follow" setting, so this is new UI. I would put this in "Account Preferences" (Mastodon puts it in the notification filters box, but I don't think this is practical on mobile.)

I think **Tusky 26 should contain full support for displaying, approving, and dismissing** (denying) **the notification requests**. (Earlier would also be nice, but it is a somewhat large feature and 26 is not far away, so maybe we should not rush it.) I think we can mostly reuse the Follow Notifications UI for this, but I think we should also add a "⋮" menu option for "deny all" (possibly this would be good in Follow Notifications as well) because [in my opinion requiring individually clicking on every request, as the Mastodon Web UI requires, would be [bad if a user is being targeted for harassment](https://files.mastodon.social/media_attachments/files/112/134/381/938/303/717/original/9580cbb4083164ec.png). 

As a stopgap before Tusky 25, I think **our official Mastodon account should make a post about the new feature**, notifying users that the feature may be on for them and showing how to disable it in the web UI.

n/a

### Debug information

n/a

### LINKED ISSUE #4550: UX proposal: Display filtered notifications "inline"
### Motivation

There is a new "filtered notifications" feature in Mastodon we must support to some degree: https://github.com/tuskyapp/Tusky/issues/4331
I think (?) Conny has work toward the preference pane for choosing your notification filters, and also the new API endpoints.
But there's a problem: When there is a filtered notification, how do we display it?
Mastodon Web shows a big box pinned to the top of your notifications. Conny has [a mockup](https://github.com/tuskyapp/Tusky/issues/4331#issuecomment-2189258973) doing similar.

But:

* [I don't like Mastodon UI for this!!](https://github.com/mastodon/mastodon/issues/29723)
* I am not the only one who seems frustrated by it, [here's Christine Lemmer-Webber complaining](https://octodon.social/@cwebber/112734674258023254) (I have not talked to her about this)

The problem in the Mastodon Web approach is filtered notifications get **too much emphasis**— more than a regular notification. This is inappropriate because filtered notifications are **less** important than a regular notification, not more.

### Pitch

I propose a solution: Show filtered notifications inline in the notification tab, so they have exactly as much emphasis of a regular notification.

A mockup:

[](https://github.com/tuskyapp/Tusky/assets/277318/1981e719-23e2-49bb-b984-d6662cfb0ec9) <big><big>➜</big></big> [](https://github.com/tuskyapp/Tusky/assets/277318/0f34328b-2aaf-4f31-8f61-93500048b3db)

A filtered notification is shown at the correct place in the timeline, but we do not show the notification, only who is messaging you (with a yellow background to indicate it is special— this is similar to Mastodon Web [coloring private mentions blue, which I still think we should adopt](https://github.com/tuskyapp/Tusky/issues/3883)) ([alternate version of image 2 without the yellow background here](https://github.com/tuskyapp/Tusky/assets/277318/0f34328b-2aaf-4f31-8f61-93500048b3db) and options to show or hide. If the user hides, the notification disappears. If the user shows, the notification is shown in full. The user is then again given inline options, whether to banish that user from their timeline forever, or to always allow their notifications in future. (This corresponds to the "checkbox" and "mute" buttons in Mastodon Web.

Doing it this way may require us to request API support from Mastodon, as currently notifications and notification requests are shown as two separate streams.

This is just one approach to the appearance. There are some things we could do to make it fancier or just different:

* We could put a "?" in the corner of the initial "Filtered notification, show/hide" display (my left image above). Tapping the ? could bring up a modal box explaining why the notification was filtered (there are four filtered notification types the user can set, plus a fifth, non-optional "your instance thinks this is a low quality notification") (I don't know if the API gives us this "why" information currently), and also link to the notification-filter-types preference pane in case the popup makes the user realize they want to adjust those.
* In the initial filtered notification display (my left image above), if the user clicks "hide", a modal box could ask "Always hide from this user?". If the user says no we zap this one notification but don't send the "always hide" to the server. (Not sure if this requires API support.)
* We could have a setting in the preferences for "never show filtered notifications". If the user clicks this, they will not even see the filtered-notification offers (left image above).

connyduck:
This was so much work wow. I think it works pretty well and is the best compromise between all the alternative we considered. Yes the pull-to-refreh on the notifications works slightly different now when the new bar is visible, but I don't think there is a way around that.

Things I plan to do later, i.e. not as part of this PR or release:
- Cache the notification policy summary for better offline behavior and less view shifting when it loads
- try to reduce some of the code duplications that are now in there
- if there is user demand, add a "legacy mode" setting where this feature is disabled even if the server would support it

closes #4331
closes #4550 as won't do
closes #4712 as won't do


app/src/main/java/com/keylesspalace/tusky/components/notifications/NotificationPolicySummaryAdapter.kt
```diff
@@ -0,0 +1,73 @@
...
+            parent,
+            false
+        )
+        binding.root.setOnClickListener {
+            onOpenDetails()
+        }
+        return BindingHolder(binding)
+    }
+
+    override fun getItemCount() = if (state.shouldShowInfo()) 0 else 0
```
Tak, line 58: ?

Tak (approved):
Looks reasonable, I'll install a build

PR #4777, opened 2024-11-27
modernize tests

connyduck:
- use `runTest` instead of `runBlocking`, where possible
- run all Robolectric tests on Api 34 (where we have most users)
- some new testcase for `TimestampUtilsTest`
- move our only instrumented Android Test, `MigrationsTest`, to unit test so it runs in CI and expand it to test all migrations
- upgrade Robolectric
- removed truth and espresso as they are no longer needed


app/src/test/java/android/text/SpannableString.kt
```diff
@@ -1,49 +0,0 @@
-package android.text
-
-// Used for stubbing Android implementation without slow & buggy Robolectric things
```
Tak, line 3: 🎉

Tak (approved):

PR #4779, opened 2024-11-30
Simplify hashtag line checker to avoid infinite loop in android matcher

Tak:
https://masto.nyc/@GetMisch/113557332197065306 causes an infinite loop in the native matcher with the original pattern, trying a different approach

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/util/LinkHelper.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/util/LinkHelper.kt
@@ -119,9 +119,8 @@ fun setClickableText(
     }
 }
 
-private val trailingHashtagExpression by unsafeLazy {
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/keylesspalace/tusky/util/LinkHelper.kt
```diff
@@ -131,7 +130,7 @@ private val trailingHashtagExpression by unsafeLazy {
 internal fun getTrailingHashtags(content: Spanned): Pair<Int, List<HashTag>> {
     // split() instead of lines() because we need to be able to account for the length of the removed delimiter
     val trailingContentLength = content.split('\r', '\n').asReversed().takeWhile { line ->
-        line.isBlank() || trailingHashtagExpression.matcher(line).matches()
+        line.split(whitespacePattern).all { it.isBlank() || hashtagWithHashPattern.matcher(it).matches() }
```
Tak, line 133: Should this maybe be `splitToSequence`?

connyduck: yes, that should be slightly more efficient

connyduck (approved):

PR #4799, opened 2024-12-08
fix caption dialog

connyduck:
The Material 3 redesign seems to have broken the dialog when very small images where used.
Also removes the weird background behind the context menu (seen in the before screenshot)

Before / after:


charlag:
400dp might be too much for some phones, not sure how they'll behave, otherwise seems good from a quick look! thank you for the fix

Tak (approved):

connyduck:
> 400dp might be too much for some phones, not sure how they'll behave, otherwise seems good from a quick look! thank you for the fix

I checked and it seems that the dialog will never grow over the available screen size, even if that is less than 400dp

PR #4813, opened 2024-12-17
make sure MainActivity doesn't hide other activities when launcher icon is clicked again

### LINKED ISSUE #4789: UI states get lost on switching to another app
### Detailed description

Since August and still present in 26.2, when drafting a toot for example, switching to another app and returning, most of the time the draft is gone and I am back in the main timeline.
Drafts lost on that way also don't show up in the saved drafts.
Same applies to opened threads (even though, there the loss usually isn't that bad.

1. Open thread, or compose toot/reply
2. Switch to another android app
3. Return to tusky, back in the timeline, opend toot or draft gone

### Debug information

Tusky 26.2

Device:

Google Pixel 6 
Android-Version: 15 
SDK-Version: 35

Account:

@mxk@hachyderm.io 
Version: 4.3.1

connyduck:
closes #4789 this time for real, hopefully, maybe

also set `alwaysRetainTaskState` for good measure

> `android:alwaysRetainTaskState`
    Whether the state of the task that the activity is in is always maintained by the system. `"true"` if it is, and `"false"` if the system can reset the task to its initial state in certain situations. The default value is `"false"`. This attribute is meaningful only for the root activity of a task. It's ignored for all other activities.

>    Normally, the system clears a task, removing all activities from the stack above the root activity, in certain situations when the user re-selects that task from the home screen. Typically, this is done if the user hasn't visited the task for a certain amount of time, such as 30 minutes.

>    However, when this attribute is `"true"`, users always return to the task in its last state, regardless of how they get there. This is useful in an application like a web browser where there is a lot of state, such as multiple open tabs, that users don't want to lose. 

https://developer.android.com/guide/topics/manifest/activity-element#always

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -41,6 +41,7 @@
         <activity
             android:name=".MainActivity"
             android:exported="true"
+            android:alwaysRetainTaskState="true"
             android:theme="@style/SplashTheme">
 
             <intent-filter>
@@ -105,6 +106,7 @@
         <activity
             android:name=".components.compose.ComposeActivity"
             android:theme="@style/TuskyDialogActivityTheme"
+            android:alwaysRetainTaskState="true"
             android:windowSoftInputMode="stateVisible|adjustResize" />
         <activity
             android:name=".components.viewthread.ViewThreadActivity"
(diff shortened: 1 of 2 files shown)


connyduck:
I'm afraid this could have unintended side effects - should we do another beta before full rollout?
On the other hand, the Tusky hashtag is full with people complaining about this bug :/

Tak (approved):

PR #4831, opened 2024-12-25
improve CaptionDialog for small screens

connyduck:
(or giant soft keyboards)

https://mastodon.social/@bewitchedmind/113703199056088283

Before / After

This makes it also look a bit weird on large screens, but I found no easy way to limit the size. Its now basically back to where it was in Tusky 26.

closes https://github.com/tuskyapp/Tusky/issues/4832


Tak (approved):
🤷

PR #4834, opened 2024-12-27
3771: Add "display replied to" functionality

Lakoja:
Earlier PR: https://github.com/tuskyapp/Tusky/pull/3778

Fixes: https://github.com/tuskyapp/Tusky/issues/3771

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/adapter/StatusBaseViewHolder.java
+++ b/app/src/main/java/com/keylesspalace/tusky/adapter/StatusBaseViewHolder.java
@@ -401,13 +401,12 @@ public abstract class StatusBaseViewHolder extends RecyclerView.ViewHolder {
         }
     }
 
-    protected void setIsReply(boolean isReply) {
+    protected void setReplyButtonImage(boolean isReply) {
         if (isReply) {
             replyButton.setImageResource(R.drawable.ic_reply_all_24dp);
         } else {
             replyButton.setImageResource(R.drawable.ic_reply_24dp);
         }
-
     }
 
     protected void setReplyCount(int repliesCount, boolean fullStats) {
@@ -768,21 +767,17 @@ public abstract class StatusBaseViewHolder extends RecyclerView.ViewHolder {
         popup.show();
     }
 
-    public void setupWithStatus(@NonNull StatusViewData.Concrete status, final @NonNull StatusActionListener listener,
-                                @NonNull StatusDisplayOptions statusDisplayOptions) {
-        this.setupWithStatus(status, listener, statusDisplayOptions, null);
-    }
-
     public void setupWithStatus(@NonNull StatusViewData.Concrete status,
                                 @NonNull final StatusActionListener listener,
                                 @NonNull StatusDisplayOptions statusDisplayOptions,
-                                @Nullable Object payloads) {
+                                @Nullable Object payloads,
+                                final boolean showStatusInfo) {
         if (payloads == null) {
             Status actionable = status.getActionable();
             setDisplayName(actionable.getAccount().getName(), actionable.getAccount().getEmojis(), statusDisplayOptions);
             setUsername(actionable.getAccount().getUsername());
             setMetaData(status, statusDisplayOptions, listener);
-            setIsReply(actionable.getInReplyToId() != null);
+            setReplyButtonImage(actionable.isReply());
             setReplyCount(actionable.getRepliesCount(), statusDisplayOptions.showStatsInline());
             setAvatar(actionable.getAccount().getAvatar(), status.getRebloggedAvatar(),
                 actionable.getAccount().getBot(), statusDisplayOptions);
--- a/app/src/main/java/com/keylesspalace/tusky/adapter/StatusViewHolder.java
+++ b/app/src/main/java/com/keylesspalace/tusky/adapter/StatusViewHolder.java
@@ -27,9 +27,9 @@ import androidx.annotation.Nullable;
 import androidx.recyclerview.widget.RecyclerView;
 
 import com.keylesspalace.tusky.R;
-import com.keylesspalace.tusky.entity.Emoji;
 import com.keylesspalace.tusky.entity.Filter;
 import com.keylesspalace.tusky.entity.Status;
+import com.keylesspalace.tusky.entity.TimelineAccount;
 import com.keylesspalace.tusky.interfaces.StatusActionListener;
 import com.keylesspalace.tusky.util.CustomEmojiHelper;
 import com.keylesspalace.tusky.util.NumberUtils;
@@ -38,7 +38,8 @@ import com.keylesspalace.tusky.util.StatusDisplayOptions;
 import com.keylesspalace.tusky.util.StringUtils;
 import com.keylesspalace.tusky.viewdata.StatusViewData;
 
-import java.util.List;
+import java.util.Collections;
+import java.util.Objects;
 
 import at.connyduck.sparkbutton.helpers.Utils;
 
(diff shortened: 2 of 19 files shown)


app/src/main/java/com/keylesspalace/tusky/entity/Status.kt
```diff
@@ -30,6 +30,7 @@ data class Status(
     val account: TimelineAccount,
     @Json(name = "in_reply_to_id") val inReplyToId: String? = null,
     @Json(name = "in_reply_to_account_id") val inReplyToAccountId: String? = null,
+    val inReplyToAccount: TimelineAccount? = null,
```
connyduck, line 33: Classes in the `entity` package are for deserializing api responses, and a Mastodon status does not have this attribute. https://docs.joinmastodon.org/entities/Status/

app/src/main/java/com/keylesspalace/tusky/adapter/StatusBaseViewHolder.java
```diff
@@ -401,15 +401,6 @@ private CharSequence getCreatedAtDescription(Date createdAt,
...
 
-    protected void setIsReply(boolean isReply) {
-        if (isReply) {
-            replyButton.setImageResource(R.drawable.ic_reply_all_24dp);
-        } else {
-            replyButton.setImageResource(R.drawable.ic_reply_24dp);
-        }
-
-    }
-
```
connyduck, line 412: Why remove this?

app/src/main/java/com/keylesspalace/tusky/adapter/StatusViewHolder.java
```diff
@@ -91,20 +96,22 @@ public void setupWithStatus(@NonNull StatusViewData.Concrete status,
...
         Context context = statusInfo.getContext();
         CharSequence wrappedName = StringUtils.unicodeWrap(name);
-        CharSequence boostedText = context.getString(R.string.post_boosted_format, wrappedName);
+        CharSequence translatedText = context.getString(isReply ? R.string.post_replied_format : R.string.post_boosted_format, wrappedName);
         CharSequence emojifiedText = CustomEmojiHelper.emojify(
-                boostedText, accountEmoji, statusInfo, statusDisplayOptions.animateEmojis()
+                translatedText, nameEmoji, statusInfo, statusDisplayOptions.animateEmojis()
         );
         statusInfo.setText(emojifiedText);
+        statusInfo.setCompoundDrawablesWithIntrinsicBounds(isReply ? R.drawable.ic_reply_all_18dp : R.drawable.ic_reblog_18dp, 0, 0, 0);
```
connyduck, line 110: The reply_all icon needs to be tinted in the same color as the text. It is currently invisible in the white theme. The easiest way to do so is probably to change the `fillColor` to the font color attribute.

connyduck (changes requested):
This seems to only work when we have the replied-to-account already in the database? But we cannot assume that, can we?

Btw, the Mastodon web interface shows "continued thread" on self-replies, I think that is nice as well.

app/src/main/java/com/keylesspalace/tusky/adapter/StatusBaseViewHolder.java
Lakoja: It is odd having this information hidden in the reply button. And that is the only usage of that function.

But also see here: https://github.com/tuskyapp/Tusky/pull/3778#issuecomment-2092897311

app/src/main/java/com/keylesspalace/tusky/adapter/StatusViewHolder.java
Lakoja: Ok.

app/src/main/java/com/keylesspalace/tusky/entity/Status.kt
Lakoja: Ok, moved to StatusViewData.Concrete

app/src/main/java/com/keylesspalace/tusky/adapter/StatusBaseViewHolder.java
connyduck: Ok, I'd also be ok with keeping the double arrow on replies because I noticed that the web interface still shows them

connyduck:
The feature only works in the home timeline (`CachedTimelineViewModel` etc), but not in others (`NetworkTimelineViewModel`), so there is no way to tell a post is a reply there.

connyduck:
> This seems to only work when we have the replied-to-account already in the database?

Yes, when the replied-to-account is not in the database, one can't tell that it is a reply. When scrolling down and the account loads because one of its posts is in the timeline as well, the info is then shown on the reply. Thats weird. 

I think to solve this there must be a check for every reply, if the replied-to-account is in the db, and if not load it.

connyduck:
https://birb.site/@charlag/113737758411679059

Lakoja:
> > This seems to only work when we have the replied-to-account already in the database?
> 
> Yes, when the replied-to-account is not in the database, one can't tell that it is a reply. When scrolling down and the account loads because one of its posts is in the timeline as well, the info is then shown on the reply. Thats weird.
> 
> I think to solve this there must be a check for every reply, if the replied-to-account is in the db, and if not load it.

Would that loading be asynchronous? How does that "then shown on the reply" work. Is this a refresh after initial display?

I think loading all necessary information should be a job of the status source (ie load it while the timeline is loaded)?

There was an easy solution in the previous PR that showed a generic "Replied" if the account was missing. Maybe that might be an improvement (to the current situation).

connyduck:
> Would that loading be asynchronous? How does that "then shown on the reply" work. Is this a refresh after initial display?

The cached timeline automatically refreshes when the database changes.
So when there is a reply without account, and the account is loaded later, the reply refreshes. 
One could check that the account is missing, load it from the server, and then insert it into the db together with the post in one transaction to solve that problem. The place to do so would probably be `CachedTimelineRemoteMediator`.

But it is an edge case that I would ignore for now.

> I think loading all necessary information should be a job of the status source (ie load it while the timeline is loaded)?

> There was an easy solution in the previous PR that showed a generic "Replied" if the account was missing. Maybe that might be an improvement (to the current situation).

Yes and yes. Here is what I think is best:

`CachedTimelineRemoteMediator`
- ignore for now
- if it turns out accounts are missing too often, we can add fetch logic here as well

`NetworkTimelineRemoteMediator` and `NetworkTimelineViewModel.loadMore`
 - when loading a post that is a reply, check all existing posts (via `viewmodel.statusData` AND the current page) for the account. Don't forget posts are nested when boosted.
 - If found, good, add it to the post
 - If not found, fetch it via api. All missing accounts of the current page can be fetched together with https://docs.joinmastodon.org/methods/accounts/#index
- only add posts to the timeline after the account has been fetched to avoid layout jumps 

When rendering posts:
- For self replies (`Status.inReplyToAccountId == Status.account.id`), show "Continued thread"
- When it is a reply, and we have an account, show "Replied to _"
- When it is a reply, and we don't have an account (yet), show "Replied in thread"

app/src/main/java/com/keylesspalace/tusky/adapter/StatusBaseViewHolder.java
connyduck: Actually, when a reply is boosted, the double arrow is still useful because the "_ boosted" would hide the "replied to_" message. Sorry for the change in mind.

Lakoja:
Oh, ok, thanks.

That sounds quite complex for something that should be a simple "display some (existing) information properly".
Or at least that is what I would plan. :-)

Maybe take a step back then:
Implement a simple "Replied" and "Continued thead".
And much later:
Once and if there is a source that delivers (or stores) all the information, then add the name to the replied to text.

In any case display and fetching should be quite separate tasks. And fetching shouldn't be a complex detective game just for this information...

connyduck:
Apparently the Mastodon web interface does it like this as well :/
It would be cool if a Status object just had an `replied_to_account` property with all the info

connyduck (approved):
Thanks, I think that is a very good trade-off now.

PR #4841, opened 2024-12-31
keep ordering when picking multiple media files at once

### LINKED ISSUE #4754: When attaching multiple images at once, they appear in random order

1. Start composing a post
2. Click paperclip, "add media"
3. In the image selector, longpress four images, tap "select"

**Expected behavior:** The four images appear in the post in the order I longpressed them

**Observed behavior:** The four images appear in what appears to be literally random order. It's not chronological.

This significantly slows down making posts with many images, because on my phone I have to do multiple clicks to get into the Photos app and then that is multiplied by four due to this bug.

Odd point for comparison: In my testing it appears the bluesky social app **does** have this problem and Discord **does not**. The fact Discord does not have the problem makes me suspect the problem is in the app and not in Android.

### Debug information

Tusky 26.2
"Version: 4.4.0-nightly.2024-11-06"
Android 13
Sony Xperia 5 III
mastodon.social

However, also @connyduck reports he saw this.

connyduck:
This restructures the code so that all picked media will be added to the upload queue in the correct order and also does some other code cleanup.

closes #4754

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/components/compose/ComposeActivity.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/components/compose/ComposeActivity.kt
@@ -73,6 +73,7 @@ import com.keylesspalace.tusky.adapter.EmojiAdapter
 import com.keylesspalace.tusky.adapter.LocaleAdapter
 import com.keylesspalace.tusky.adapter.OnEmojiSelectedListener
 import com.keylesspalace.tusky.components.compose.ComposeViewModel.ConfirmationKind
+import com.keylesspalace.tusky.components.compose.ComposeViewModel.QueuedMedia
 import com.keylesspalace.tusky.components.compose.dialog.CaptionDialog
 import com.keylesspalace.tusky.components.compose.dialog.makeFocusDialog
 import com.keylesspalace.tusky.components.compose.dialog.showAddPollDialog
@@ -102,6 +103,7 @@ import com.keylesspalace.tusky.util.getSerializableCompat
 import com.keylesspalace.tusky.util.hide
 import com.keylesspalace.tusky.util.highlightSpans
 import com.keylesspalace.tusky.util.loadAvatar
+import com.keylesspalace.tusky.util.map
 import com.keylesspalace.tusky.util.modernLanguageCode
 import com.keylesspalace.tusky.util.setDrawableTint
 import com.keylesspalace.tusky.util.show
@@ -162,7 +164,11 @@ class ComposeActivity :
     private val takePictureLauncher =
         registerForActivityResult(ActivityResultContracts.TakePicture()) { success ->
             if (success) {
-                pickMedia(photoUploadUri!!)
+                viewModel.pickMedia(
+                    listOf(
+                        ComposeViewModel.PickedMedia(photoUploadUri!!)
+                    )
+                )
             }
         }
     private val pickMediaFilePermissionLauncher =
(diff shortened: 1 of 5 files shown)


app/src/main/java/com/keylesspalace/tusky/components/compose/ComposeActivity.kt
```diff
@@ -162,7 +164,11 @@ class ComposeActivity :
     private val takePictureLauncher =
         registerForActivityResult(ActivityResultContracts.TakePicture()) { success ->
             if (success) {
-                pickMedia(photoUploadUri!!)
+                viewModel.pickMedia(
+                    listOf(
+                        ComposeViewModel.PickedMedia(photoUploadUri!!)
+                    )
+                )
             }
```
Lakoja, line 172: else - is that possible?

app/src/main/java/com/keylesspalace/tusky/components/compose/ComposeViewModel.kt
```diff
@@ -588,6 +578,36 @@ class ComposeViewModel @Inject constructor(
...
+        }
+        enum class State {
+            UPLOADING,
+            UNPROCESSED,
+            PROCESSED,
+            PUBLISHED
+        }
+    }
+
+    data class PickedMedia(
```
Lakoja, line 606: Naming: Is it important that it is "picked" here? Or more like "LocalMedia" or somesuch.

app/src/main/java/com/keylesspalace/tusky/components/compose/ComposeViewModel.kt
```diff
@@ -505,11 +497,9 @@ class ComposeViewModel @Inject constructor(
...
         if (draftAttachments != null) {
             // when coming from DraftActivity
-            viewModelScope.launch {
-                draftAttachments.forEach { attachment ->
-                    pickMedia(attachment.uri, attachment.description, attachment.focus)
-                }
-            }
+            draftAttachments.map { attachment ->
+                PickedMedia(attachment.uri, attachment.description, attachment.focus)
+            }.let { pickMedia(it) }
```
Lakoja, line 502: Is "it" a list here?

app/src/main/java/com/keylesspalace/tusky/components/compose/ComposeActivity.kt
```diff
@@ -347,14 +353,14 @@ class ComposeActivity :
                     when (intent.action) {
                         Intent.ACTION_SEND -> {
                             intent.getParcelableExtraCompat<Uri>(Intent.EXTRA_STREAM)?.let { uri ->
-                                pickMedia(uri)
+                                viewModel.pickMedia(listOf(ComposeViewModel.PickedMedia(uri)))
```
Lakoja, line 356: Overload "pickMedia"? There are three usages with only an uri here.

app/src/main/java/com/keylesspalace/tusky/components/compose/ComposeActivity.kt
connyduck: I think that is mostly when users cancel. And when the camera fails I'd expect the camera app to show an error 🤔

app/src/main/java/com/keylesspalace/tusky/components/compose/ComposeViewModel.kt
connyduck: yes

app/src/main/java/com/keylesspalace/tusky/components/compose/ComposeViewModel.kt
Lakoja: MediaData?

Lakoja (approved):

PR #4846, opened 2025-01-04
Add account preference to filter bot posts from timelines

Tak:
Making this a draft because I'm not sure what the ideal behavior is.
Like this, it hides all posts by bot accounts from all timelines, including accounts you explicitly follow and posts boosted by accounts you follow.
Is this what we want, or should it be a per-timeline preference, or more granular somehow, or…?

### DIFF
--- a/app/schemas/com.keylesspalace.tusky.db.AppDatabase/66.json
+++ b/app/schemas/com.keylesspalace.tusky.db.AppDatabase/66.json
@@ -1,0 +1,1345 @@
+{
+  "formatVersion": 1,
+  "database": {
+    "version": 66,
+    "identityHash": "379378624adb80f474a2801fc454d783",
+    "entities": [
+      {
+        "tableName": "DraftEntity",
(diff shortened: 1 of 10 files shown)


Lakoja:
Product question:
I was wondering: What is the underlying problem here?
Are all bots bad? Or what actually must be hidden? And why?

Lakoja:
There are somewhat similar filters: the three "isShow..." preferences in com/keylesspalace/tusky/db/entity/AccountEntity

They are particular only to the home timeline. Which sounds sensible to me.

Although they are server-stored and this one here would not be...

Tak:
Oops, I thought I had responded to this earlier. (Or maybe I did on matrix and not here 🤷)
Some people (including me) find bot posts annoying, between the markov chain bots (although these have really died down in the face of the general distaste for ubiquitous LLM slop), news article bots, "unattributed photo of thing every hour" bots, ad nauseam. (I had `botsin.space` limited from my instance before they shut down.)
We periodically get requests for a preference to hide posts by bots, hence this PR.

PR #4848, opened 2025-01-04
Properly summarize all notifications

Lakoja:
Do summary notifications like the Api defines it:
* Schedule and summarize without delay (in order for summerization to work)
* Always have a summary notification: simplify code with this and make more reliable
* Do not care about single notification count (the system already does that as well)
* **Bugfix: Schedule summary first: This avoids a rate limit problem that (then) not groups at all**

Testing this is probably the most difficult part.
For example I couldn't get any notification to ring with older Api versions in the debugger. (Same as for current develop)
However one hack to always get notifications: Fix "minId" in "fetchNewNotifications()" to a somewhat older value.

Next possible step: Have only one summary notification at all (for all channels/notification types). You can still configure single channels differently.
Or: For very many notifications: Only use a true summary one (something like "you have 28 favorites and 7 boosts").

Generally: The notification timeline must be improved now. Because that must be the go-to solution for any large number of notifications. It must be easy to read. E. g. with grouping per post.

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/components/systemnotifications/NotificationFetcher.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/components/systemnotifications/NotificationFetcher.kt
@@ -1,9 +1,13 @@
 package com.keylesspalace.tusky.components.systemnotifications
 
+import android.Manifest
 import android.app.NotificationManager
 import android.content.Context
+import android.content.pm.PackageManager
 import android.util.Log
 import androidx.annotation.WorkerThread
+import androidx.core.app.ActivityCompat
+import androidx.core.app.NotificationManagerCompat
 import com.keylesspalace.tusky.appstore.EventHub
 import com.keylesspalace.tusky.appstore.NewNotificationsEvent
 import com.keylesspalace.tusky.components.systemnotifications.NotificationHelper.filterNotification
@@ -16,8 +20,6 @@ import com.keylesspalace.tusky.util.HttpHeaderLink
 import com.keylesspalace.tusky.util.isLessThan
 import dagger.hilt.android.qualifiers.ApplicationContext
 import javax.inject.Inject
-import kotlin.time.Duration.Companion.milliseconds
-import kotlinx.coroutines.delay
 
 /** Models next/prev links from the "Links" header in an API response */
 data class Links(val next: String?, val prev: String?) {
@@ -52,12 +54,18 @@ class NotificationFetcher @Inject constructor(
     private val eventHub: EventHub
 ) {
     suspend fun fetchAndShow() {
+        if (ActivityCompat.checkSelfPermission(context, Manifest.permission.POST_NOTIFICATIONS) != PackageManager.PERMISSION_GRANTED) {
+            return
+        }
+
         for (account in accountManager.getAllAccountsOrderedByActive()) {
             if (account.notificationsEnabled) {
                 try {
                     val notificationManager = context.getSystemService(
                         Context.NOTIFICATION_SERVICE
                     ) as NotificationManager
+
+                    val notificationManagerCompat = NotificationManagerCompat.from(context)
 
                     // Create sorted list of new notifications
                     val notifications = fetchNewNotifications(account)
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/keylesspalace/tusky/components/systemnotifications/NotificationFetcher.kt
```diff
@@ -72,69 +72,40 @@ class NotificationFetcher @Inject constructor(
...
+                            account,
+                            notificationsGroup.key
+                        )
 
-                            // Android will rate limit / drop notifications if they're posted too
-                            // quickly. There is no indication to the user that this happened.
-                            // See https://github.com/tuskyapp/Tusky/pull/3626#discussion_r1192963664
+                        // Have a small pause before the next channel = obvious (audible, ...) notification
+                        typeIndex++
+                        if (typeIndex < notificationsByType.size)
```
Lakoja, line 105: We could get rid of this delay as well: We normally only have 1-3 summary notifications and if they appear without delay from each other the would probably only ring once.

app/src/main/java/com/keylesspalace/tusky/components/systemnotifications/NotificationHelper.java
```diff
@@ -747,29 +716,38 @@ private static void setSoundVibrationLight(AccountEntity account, NotificationCo
...
     @Nullable
-    private static String joinNames(Context context, List<StatusBarNotification> notifications) {
-        if (notifications.size() > 3) {
-            int length = notifications.size();
-            //notifications.get(0).getNotification().extras.getString(EXTRA_ACCOUNT_NAME);
+    private static String joinNames(Context context, List<String> names) {
+        if (names.size() > 3) {
+            int length = names.size();
             return String.format(context.getString(R.string.notification_summary_large),
-                    wrapItemAt(notifications.get(length - 1)),
```
connyduck, line 760: I think the `String.format` is unnecessary? Format args can be passed to `context.getString` directly

app/src/main/java/com/keylesspalace/tusky/components/systemnotifications/NotificationFetcher.kt
```diff
@@ -72,69 +71,41 @@ class NotificationFetcher @Inject constructor(
                     //   (and should therefore adhere to the notification config).
                     eventHub.dispatch(NewNotificationsEvent(account.accountId, notifications))
 
-                    // There's a maximum limit on the number of notifications an Android app
```
connyduck, line 75: `notifications` does not need to be mutable any more

(why doesn't GitHub let me comment on unchanged lines 🙄 )

app/src/main/java/com/keylesspalace/tusky/components/systemnotifications/NotificationFetcher.kt
```diff
@@ -72,69 +71,41 @@ class NotificationFetcher @Inject constructor(
...
-                        // starting with the oldest.
-                        while (notifications.size > MAX_NOTIFICATIONS) {
-                            notifications.removeAt(0)
-                        }
-                    }
 
-                    val notificationsByType = notifications.groupBy { it.type }
-
-                    // Make and send the new notifications
-                    // TODO: Use the batch notification API available in NotificationManagerCompat
```
connyduck, line 106: What about this comment? I think it makes sense.
From the docs of the batch notification api: 
> This is the preferred method for posting groups of notifications, to improve sound and animation behavior.

connyduck (commented):
Makes sense, works for me.

app/src/main/java/com/keylesspalace/tusky/components/systemnotifications/NotificationFetcher.kt
connyduck: Good idea, let's try it

app/src/main/java/com/keylesspalace/tusky/components/systemnotifications/NotificationFetcher.kt
Lakoja: Where can I find the batch notification api?

app/src/main/java/com/keylesspalace/tusky/components/systemnotifications/NotificationHelper.java
Lakoja: ok

app/src/main/java/com/keylesspalace/tusky/components/systemnotifications/NotificationFetcher.kt
Lakoja: ok

app/src/main/java/com/keylesspalace/tusky/components/systemnotifications/NotificationFetcher.kt
Lakoja: done

app/src/main/java/com/keylesspalace/tusky/components/systemnotifications/NotificationFetcher.kt
connyduck: [NotificationManagerCompat.from(context).notify(...)](https://developer.android.com/reference/androidx/core/app/NotificationManagerCompat#notify(java.util.List%3Candroidx.core.app.NotificationManagerCompat.NotificationWithIdAndTag%3E))

Lakoja: Done.

app/src/main/java/com/keylesspalace/tusky/components/systemnotifications/NotificationFetcher.kt
```diff
@@ -53,87 +54,63 @@ class NotificationFetcher @Inject constructor(
...
 
-                    NotificationHelper.updateSummaryNotifications(
-                        context,
-                        notificationManager,
-                        account
-                    )
+                        notificationManagerCompat.notify(newNotifications)
+
+                        // NOTE this can schedule multiple (summary) notifications in short succession (normally 1-3).
+                        //   They can "collapse" to only one audible one if fast enough.
```
connyduck, line 112: Could this be outside of the `notificationsByType.forEach` loop as well?

Lakoja: Done.

connyduck (approved):

PR #4849, opened 2025-01-04
increase width of tab indicator

connyduck:
There have been multiple complaints by users that the new Material 3 tab indicator is too small. 

Setting `app:tabIndicatorFullWidth="true"` alone looks weird imho, so I additionally added some padding.

cc @mcclure 

Before / After

before after


JuanjoSalvador:
New look looks far better, honestly.

Tak (approved):

anghenfil:
Looks great!

PR #4861, opened 2025-01-10
merge hashtag dialogs into one

connyduck:
A hashtag picker dialog was implemented twice, with slight differences.
Now there is only one
- with hashtag validation - no more api errors when following an invalid one
- The dialog can now be closed with the keyboard, for extra fast hashtag selection
- with autocomplete

I also added a new snackbar when following a hashtag was succesfull.

Although I'm not sure about the auto complete, it can be very annoying as the drop down covers the buttons. I found no way to make it size to its content: https://chaos.social/@ConnyDuck/113803457147888844

Should we get rid of it?


app/src/main/res/values-cy/strings.xml
```diff
@@ -690,17 +690,17 @@
     <string name="filter_description_format">%1$s: %2$s</string>
     <string name="post_media_image">Delwedd</string>
     <string name="select_list_manage">Rheoli rhestrau</string>
-    <string name="help_empty_home">Dyma\'ch <b>llinell amser cartref</b>. Mae\'n dangos negeseuon diweddar y cyfrifon rydych chi\'n eu dilyn. 
-\n 
+    <string name="help_empty_home">Dyma\'ch <b>llinell amser cartref</b>. Mae\'n dangos negeseuon diweddar y cyfrifon rydych chi\'n eu dilyn.
```
Lakoja, line 693: Hm?

Weren't those the spaces that were added before - and nobody knew why?

connyduck: I think so. Well now we know, Android Studio formats stuff it shouldn't when renaming a string.

connyduck: `sed`to the rescue 🙄

Lakoja:
Are hashtags ascii-only?
The autocomplete and especially the final hashtag after the dialog will ignore/convert umlauts for example.

The autocomplete popup not being bounded by the keyboard probably is a problem (it extends below it). If you try to close the keyboard then first the popup is removed (and only on the second try the keyboard minimized).
Feature-wise however I would expect an autocomplete - as it is in the compose screen..

A bit odd that the text field border says "without #" but the autocomplete will show the proposals with #. It probably shouldn't.
(That would however collide with the behavior in compose screen.)
Maybe don't require and don't forbid it. Only if it is omitted in manual text input then add it later. That would most conform to the compose screen.

connyduck:
>Are hashtags ascii-only?
The autocomplete and especially the final hashtag after the dialog will ignore/convert umlauts for example.

Mastodon will convert them to ascii internally. So #häshtag and #hashtag may look differently, but are actually the same.
So I think its ok if we do that as well here.


I only meant getting rid of the autocomplete in this dialog, not everywhere.

connyduck:
The thing with the dropdowns going under the keyboard affects other dialogs as well, I'm checking how it can be fixed everywhere

Tak (approved):

PR #4864, opened 2025-01-10
fix dialog background colors

connyduck:
In https://github.com/tuskyapp/Tusky/pull/4851 I changed the theme of `AccountsInListFragment`, which accidentally turned its background white for the dark theme.

Additionally this fixes the color for the preference dialogs, which I think have been incorrect since the Material3 redesign.

I also wondered if we should make dialogs darker for the black theme, but looks like there is not much interest in that https://chaos.social/deck/@ConnyDuck/113802937491059461
(Currently they are just the same as the dark theme)


Lakoja:
> I also wondered if we should make dialogs darker for the black theme, but looks like there is not much interest in that https://chaos.social/deck/@ConnyDuck/113802937491059461 (Currently they are just the same as the dark theme)

Should one mix "dark" with "dark" for a dialog? Shouldn't a dialog stand out?

connyduck:
> Should one mix "dark" with "dark" for a dialog? Shouldn't a dialog stand out?

Generally I think yes, but people using the black theme often do so to save battery and then it is better if it is as dark as possible 🤷

Tak (approved):

PR #4875, opened 2025-01-15
center icon in small send button

connyduck:
2 users complained already 😔 

Before:
before

After:
after

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/components/compose/view/TootButton.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/components/compose/view/TootButton.kt
@@ -38,9 +38,9 @@ class TootButton
     init {
         if (smallStyle) {
             setIconResource(R.drawable.ic_send_24dp)
+            iconPadding = 0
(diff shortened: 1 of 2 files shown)


Tak (approved):

Lakoja:
Isn't that icon a bit too small? (and maybe thus misplaced)

It looks smaller than for example the pencil in the fab on the main screen.

connyduck:
Yes but that is the largest it can go without overlapping other stuff. Some instances have 10000 character limit and when you upload media another icon appears. Nobody complained about the size yet. Do you have a better idea?

Lakoja:
> Yes but that is the largest it can go without overlapping other stuff. Some instances have 10000 character limit and when you upload media another icon appears. Nobody complained about the size yet. Do you have a better idea?

I meant the icon itself, not the button.

And there one could have all button images the same size. And maybe here for the "small display" reduce the normal paddings a bit.

However, I couldn't get this to work quickly just now (FAB being too different to the TootButton).

Issue #4874, opened 2025-01-15
Since v27.0 beta 1 replies are visible when selected to hide


Karmalakas:
### Detailed description

Even if I turn off the setting to show replies in the home timeline, replies are still there. Reproduceable on v27.0 beta 1, v27.0 and v27.1

It shows even replies to some posts, where I don't even participate in the discussion

This is not the case on v26.2 - on v26.2 it shows timeline as expected without any replies

### Steps to reproduce the problem

1. Turn off _Account prefernces > Home timeline > Show replies_
2. Go to home timeline
3. Observe visible replies

### Debug information

Tusky 27.1

Device:
Google Pixel 6a
Android version: 15
SDK version: 35

Version: 4.4.0-nightly.2025-01-14

connyduck:
Does restarting the app help?
Are you logged in with multiple accounts? If yes, do they all have _Show replies_ off?
Can you show me a video?

Karmalakas:
No (reinstalled different versions multiple times to test)
No (no, tested with a single account after every reinstall)
Yes (cut out parts which should not be shared)

[Youtube link here](https://www.youtube.com/shorts/mD2oIwbzrr8)
You will notice, that on both v27 and v26 before I turn off replies, it shows all of them. When I turn replies off, v27 shows mostly my own replies. I didn't bother scrolling down and searching for other peoples replies, but here's a screenshot from earlier:
![e79689f9d7a1b447](https://github.com/user-attachments/assets/d4718de5-80e0-468d-8145-ad6faafb8089)
This is v27 with replies turned off and it shows not only my own (1st), but also from someone I follow (2nd). What's even worse, I didn't even participate in that discussion and was never tagged

connyduck:
Thanks this is very helpful!
Ok, showing ones own replies is actually [expected](https://github.com/tuskyapp/Tusky/blob/develop/app/src/main/java/com/keylesspalace/tusky/components/timeline/viewmodel/TimelineViewModel.kt#L185-L186). I guess we should change that.

So changing the setting does make a difference. It filters some replies but not all 🤔 

Now, for those replies that are not your own that are not filtered- anything special about them? Something they have in common? Are they boosts? Do you see them in the web timeline? Do you follow the authors? Or the hashtags?

Karmalakas:
[...]

PR #4879, opened 2025-01-16
apply Tusky side hometimeline filters to own posts again

connyduck:
addresses part of #4874


app/src/main/java/com/keylesspalace/tusky/components/timeline/viewmodel/TimelineViewModel.kt
```diff
@@ -181,16 +181,16 @@ abstract class TimelineViewModel(
 
     protected fun shouldFilterStatus(statusViewData: StatusViewData): Filter.Action {
         val status = statusViewData.asStatusOrNull()?.status ?: return Filter.Action.NONE
-        if (status.actionableStatus.account.id == account.accountId) {
```
Lakoja, line 184: That was a sort of guess at that time "never filter own posts"?

And the actual change/fix here?

And the remainder is only reformatting?

Lakoja (approved):

app/src/main/java/com/keylesspalace/tusky/components/timeline/viewmodel/TimelineViewModel.kt
connyduck: > That was a sort of guess at that time "never filter own posts"?

I made https://github.com/tuskyapp/Tusky/pull/4742 because Mastodon filters should not apply to own posts to match Mastodon behavior. I included Tusky-only filters as well, but based on user feedback I'm reverting that part.

Tak (approved):

PR #4881, opened 2025-01-16
move renovate.json into .github

connyduck:


Tak (approved):
I don't know what the implications of this are

connyduck:
No implications, it works in both places. I just don't think it is important enough to have at the top level.

Issue #4858, opened 2025-01-08
Notifications are always disabled on first application run


Lakoja:
### Detailed description

This is at least true for non-push notifications and there for newer phones:

The ordering of steps regarding notification setup is wrong in "MainActivity.onCreate".
First "fetchUserInfo" will be done - which surprisingly by its name will try all notification setup; e. g. create channels and so on.
Only as the last block after that will the "request notification permission" dialog be triggered.

So after a very first install of Tusky you need to fully restart it (i.e. restart the phone?) in order to get notifications.

### Steps to reproduce the problem

1. Create a new virtual device with Api >= 33
2. Start Tusky there
3. Allow notifications when asked
4. There will be no notifications (for example there is the output "NotificationsDisabled" in Logcat)
5. Only after a full restart will there be some

### Debug information

Current develop. Bu probably for some time.

PR #4883, opened 2025-01-18
Refactor notifications

Lakoja:
Also fixes https://github.com/tuskyapp/Tusky/issues/4858.
But apart from that there should be no functional change.

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/MainActivity.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/MainActivity.kt
@@ -45,7 +45,6 @@ import androidx.appcompat.app.AlertDialog
 import androidx.appcompat.content.res.AppCompatResources
 import androidx.coordinatorlayout.widget.CoordinatorLayout
 import androidx.core.app.ActivityCompat
-import androidx.core.content.ContextCompat
 import androidx.core.content.pm.ShortcutManagerCompat
 import androidx.core.splashscreen.SplashScreen.Companion.installSplashScreen
 import androidx.core.view.MenuProvider
@@ -202,6 +201,16 @@ class MainActivity : BottomSheetActivity(), ActionButtonActivity, MenuProvider {
         // will be redirected to LoginActivity by BaseActivity
         activeAccount = accountManager.activeAccount ?: return
 
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.TIRAMISU &&
+            ActivityCompat.checkSelfPermission(this, Manifest.permission.POST_NOTIFICATIONS) != PackageManager.PERMISSION_GRANTED
+        ) {
+            ActivityCompat.requestPermissions(
+                this,
+                arrayOf(Manifest.permission.POST_NOTIFICATIONS),
+                ALLOW_NOTIFICATIONS_REQUEST_CODE
+            )
+        }
+
         if (explodeAnimationWasRequested()) {
             overrideActivityTransitionCompat(
                 ActivityConstants.OVERRIDE_TRANSITION_OPEN,
@@ -295,19 +304,28 @@ class MainActivity : BottomSheetActivity(), ActionButtonActivity, MenuProvider {
 
         onBackPressedDispatcher.addCallback(this@MainActivity, onBackPressedCallback)
 
-        if (
-            Build.VERSION.SDK_INT >= 33 &&
-            ContextCompat.checkSelfPermission(this@MainActivity, Manifest.permission.POST_NOTIFICATIONS) != PackageManager.PERMISSION_GRANTED
-        ) {
-            ActivityCompat.requestPermissions(
-                this@MainActivity,
-                arrayOf(Manifest.permission.POST_NOTIFICATIONS),
-                1
-            )
-        }
-
         // "Post failed" dialog should display in this activity
         draftsAlert.observeInContext(this@MainActivity, true)
+    }
+
+    override fun onRequestPermissionsResult(requestCode: Int, permissions: Array<out String>, grantResults: IntArray) {
+        super.onRequestPermissionsResult(requestCode, permissions, grantResults)
+
+        if (Build.VERSION.SDK_INT < Build.VERSION_CODES.TIRAMISU) {
+            return
+        }
+
+        if (requestCode != ALLOW_NOTIFICATIONS_REQUEST_CODE) {
+            return
+        }
+
+        val idx = permissions.indexOfFirst { it == Manifest.permission.POST_NOTIFICATIONS }
+
+        if (idx < 0 || grantResults[idx] != PackageManager.PERMISSION_GRANTED) {
+            return
+        }
+
+        viewModel.setupNotifications()
     }
 
     override fun onNewIntent(intent: Intent) {
@@ -1082,6 +1100,7 @@ class MainActivity : BottomSheetActivity(), ActionButtonActivity, MenuProvider {
     override fun getActionButton() = binding.composeButton
 
     companion object {
+        const val ALLOW_NOTIFICATIONS_REQUEST_CODE = 2389668
         const val OPEN_WITH_EXPLODE_ANIMATION = "explode"
 
         private const val TAG = "MainActivity" // logging tag
--- a/app/src/main/java/com/keylesspalace/tusky/MainActivity.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/MainActivity.kt
@@ -76,6 +76,7 @@ import com.keylesspalace.tusky.components.login.LoginActivity
 import com.keylesspalace.tusky.components.preference.PreferencesActivity
 import com.keylesspalace.tusky.components.scheduled.ScheduledStatusActivity
 import com.keylesspalace.tusky.components.search.SearchActivity
+import com.keylesspalace.tusky.components.systemnotifications.NotificationService
 import com.keylesspalace.tusky.components.trending.TrendingActivity
 import com.keylesspalace.tusky.databinding.ActivityMainBinding
 import com.keylesspalace.tusky.db.DraftsAlert
@@ -137,6 +138,9 @@ class MainActivity : BottomSheetActivity(), ActionButtonActivity, MenuProvider {
 
     @Inject
     lateinit var eventHub: EventHub
+
+    @Inject
+    lateinit var notificationService: NotificationService
 
     @Inject
     lateinit var cacheUpdater: CacheUpdater
--- a/app/src/main/java/com/keylesspalace/tusky/TuskyApplication.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/TuskyApplication.kt
@@ -16,6 +16,7 @@
 package com.keylesspalace.tusky
 
 import android.app.Application
+import android.content.Context
 import android.content.SharedPreferences
 import android.util.Log
 import androidx.hilt.work.HiltWorkerFactory
@@ -24,7 +25,6 @@ import androidx.work.Constraints
 import androidx.work.ExistingPeriodicWorkPolicy
 import androidx.work.PeriodicWorkRequestBuilder
 import androidx.work.WorkManager
-import com.keylesspalace.tusky.components.systemnotifications.NotificationHelper
 import com.keylesspalace.tusky.settings.AppTheme
 import com.keylesspalace.tusky.settings.NEW_INSTALL_SCHEMA_VERSION
 import com.keylesspalace.tusky.settings.PrefKeys
@@ -75,6 +75,7 @@ class TuskyApplication : Application(), Configuration.Provider {
             NEW_INSTALL_SCHEMA_VERSION
         )
         if (oldVersion != SCHEMA_VERSION) {
+            // TODO SCHEMA_VERSION is outdated / not updated in code
             upgradeSharedPreferences(oldVersion, SCHEMA_VERSION)
         }
 
@@ -88,8 +89,6 @@ class TuskyApplication : Application(), Configuration.Provider {
         setAppNightMode(theme)
 
         localeManager.setLocale()
-
-        NotificationHelper.createWorkerNotificationChannel(this)
 
         // Prune the database every ~ 12 hours when the device is idle.
         val pruneCacheWorker = PeriodicWorkRequestBuilder<PruneCacheWorker>(12, TimeUnit.HOURS)
--- a/app/src/main/java/com/keylesspalace/tusky/components/systemnotifications/NotificationService.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/components/systemnotifications/NotificationService.kt
@@ -1,0 +1,791 @@
+package com.keylesspalace.tusky.components.systemnotifications
+
+import android.Manifest
+import android.app.NotificationChannel
+import android.app.NotificationChannelGroup
+import android.app.NotificationManager
+import android.app.PendingIntent
+import android.content.Context
+import android.content.Intent
+import android.content.pm.PackageManager
+import android.graphics.Bitmap
+import android.graphics.BitmapFactory
+import android.os.Build
+import android.os.Bundle
+import android.provider.Settings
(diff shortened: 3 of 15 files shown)


app/src/main/java/com/keylesspalace/tusky/MainActivity.kt
```diff
@@ -76,6 +75,7 @@ import com.keylesspalace.tusky.components.login.LoginActivity
 import com.keylesspalace.tusky.components.preference.PreferencesActivity
 import com.keylesspalace.tusky.components.scheduled.ScheduledStatusActivity
 import com.keylesspalace.tusky.components.search.SearchActivity
+import com.keylesspalace.tusky.components.systemnotifications.NotificationService
```
Lakoja, line 78: I'm open for a better name.

It is no "service" in the java sense. Only in the DI sense.

app/src/main/java/com/keylesspalace/tusky/MainActivity.kt
```diff
@@ -291,21 +304,30 @@ class MainActivity : BottomSheetActivity(), ActionButtonActivity, MenuProvider {
...
-                arrayOf(Manifest.permission.POST_NOTIFICATIONS),
-                1
-            )
-        }
-
         // "Post failed" dialog should display in this activity
         draftsAlert.observeInContext(this@MainActivity, true)
     }
 
+    override fun onRequestPermissionsResult(requestCode: Int, permissions: Array<out String>, grantResults: IntArray) {
```
Lakoja, line 311: This is the bug fix.

app/src/main/java/com/keylesspalace/tusky/TuskyApplication.kt
```diff
@@ -75,6 +75,7 @@ class TuskyApplication : Application(), Configuration.Provider {
             NEW_INSTALL_SCHEMA_VERSION
         )
         if (oldVersion != SCHEMA_VERSION) {
+            // TODO SCHEMA_VERSION is outdated / not updated in code
```
Lakoja, line 78: Collateral find.

app/src/main/java/com/keylesspalace/tusky/usecase/LogoutUsecase.kt
```diff
@@ -42,12 +43,13 @@ class LogoutUsecase @Inject constructor(
         disableUnifiedPushNotificationsForAccount(context, account)
 
         // disable pull notifications
-        if (!NotificationHelper.areNotificationsEnabled(context, accountManager)) {
-            NotificationHelper.disablePullNotifications(context)
+        if (!notificationService.areNotificationsEnabled()) {
+            // TODO this is working very wrong
```
Lakoja, line 47: That would then be a next step.

Lakoja:
If this is ok I would also move the current push code to the NotificationService in this PR.

app/src/main/java/com/keylesspalace/tusky/components/systemnotifications/NotificationService.kt
```diff
@@ -0,0 +1,791 @@
...
+            if (sn.id != accountId.toInt()) continue
+
+            val channelId = sn.notification.group
+
+            if (channelId != typeChannelId) continue
+
+            val summaryTag = "$GROUP_SUMMARY_TAG.$channelId"
+            if (summaryTag == sn.tag) continue
+
+            activeNotificationsForType.add(sn)
```
connyduck, line 498:
```suggestion
        return notificationManager.activeNotifications.filter { sn ->
            val channelId = sn.notification.group
            sn.id == accountId.toInt() && channelId == typeChannelId && sn.tag != "$GROUP_SUMMARY_TAG.$channelId"
        }
```

app/src/main/java/com/keylesspalace/tusky/components/systemnotifications/NotificationService.kt
```diff
@@ -0,0 +1,791 @@
...
+            .setAutoCancel(true)
+            .setShortcutId(account.id.toString())
+            .setSubText(account.fullName)
+            .setVisibility(NotificationCompat.VISIBILITY_PRIVATE)
+            .setCategory(NotificationCompat.CATEGORY_SOCIAL)
+            .setOnlyAlertOnce(true)
+            .setGroup(channelId)
+            .setGroupAlertBehavior(NotificationCompat.GROUP_ALERT_SUMMARY) // Only ever alert for the summary notification
+
+        setSoundVibrationLight(account, builder)
```
connyduck, line 529: this could be an extension function for `NotificationCompat.Builder`, then it is just

```
return NotificationCompat.Builder(context, channelId)
    .something()
    ...
    .somethingElse()
```

app/src/main/java/com/keylesspalace/tusky/components/systemnotifications/NotificationService.kt
```diff
@@ -0,0 +1,791 @@
...
+            .setContentTitle(title)
+            .setTicker(title)
+            .setSmallIcon(R.drawable.ic_notify)
+            .setOngoing(true)
+            .build()
+    }
+
+    private fun getChannelId(account: AccountEntity, type: Notification.Type): String? {
+        return when (type) {
+            Notification.Type.MENTION -> CHANNEL_MENTION + account.identifier
```
connyduck, line 466: Why is `CHANNEL_MENTION` a constant here and the others below aren't? I don't care either way it should just be consistent 😅

app/src/main/java/com/keylesspalace/tusky/components/systemnotifications/NotificationService.kt
```diff
@@ -0,0 +1,791 @@
...
+                val channel = NotificationChannel(it.id, name, importance)
+
+                channel.description = description
+                channel.enableLights(true)
+                channel.lightColor = -0xd46f27
+                channel.enableVibration(true)
+                channel.setShowBadge(true)
+                channel.group = account.identifier
+                channels.add(channel)
+            }
```
connyduck, line 170:
```suggestion
            val channels = channelData.map {
                val name = context.getString(it.name)
                val description = context.getString(it.description)
                val importance = NotificationManager.IMPORTANCE_DEFAULT
                val channel = NotificationChannel(it.id, name, importance)

                channel.description = description
                channel.enableLights(true)
                channel.lightColor = -0xd46f27
                channel.enableVibration(true)
                channel.setShowBadge(true)
                channel.group = account.identifier
                channel
            }
```

app/src/main/java/com/keylesspalace/tusky/components/systemnotifications/NotificationService.kt
```diff
@@ -0,0 +1,791 @@
...
+    }
+
+    private fun bodyForType(notification: Notification, alwaysOpenSpoiler: Boolean): String? {
+        if (notification.status == null) {
+            return null
+        }
+
+        when (notification.type) {
+            Notification.Type.FOLLOW, Notification.Type.FOLLOW_REQUEST, Notification.Type.SIGN_UP -> return "@" + notification.account.username
+            Notification.Type.MENTION, Notification.Type.FAVOURITE, Notification.Type.REBLOG, Notification.Type.STATUS -> return if (!TextUtils.isEmpty(notification.status!!.spoilerText) && !alwaysOpenSpoiler) {
```
connyduck, line 566: the `!!` is unnecessary as you checked already for null above

app/src/main/java/com/keylesspalace/tusky/components/systemnotifications/NotificationService.kt
```diff
@@ -0,0 +1,791 @@
...
+                        buildDescription(
+                            option.title,
+                            calculatePercent(option.votesCount, poll.votersCount, poll.votesCount),
+                            poll.ownVotes.contains(i),
+                            context
+                        )
+                    )
+                    builder.append('\n')
+                    ++i
+                }
```
connyduck, line 593:
```suggestion
                poll.options.forEachIndexed { i, option ->
                    builder.append(
                        buildDescription(
                            option.title,
                            calculatePercent(option.votesCount, poll.votersCount, poll.votesCount),
                            poll.ownVotes.contains(i),
                            context
                        )
                    )
                    builder.append('\n')
                }
```

app/src/main/java/com/keylesspalace/tusky/usecase/LogoutUsecase.kt
connyduck: Could you give some more detials what the problem is?

app/src/main/java/com/keylesspalace/tusky/components/systemnotifications/NotificationService.kt
```diff
@@ -0,0 +1,791 @@
...
+            .setConstraints(Constraints.Builder().setRequiredNetworkType(NetworkType.CONNECTED).build())
+            .setInitialDelay(5, TimeUnit.MINUTES)
+            .build()
+
+        workManager.enqueue(workRequest)
+
+        Log.d(TAG, "Enabled pull checks with " + PeriodicWorkRequest.MIN_PERIODIC_INTERVAL_MILLIS + "ms interval.")
+    }
+
+    fun disablePullNotifications(context: Context) {
```
connyduck, line 210: No need for the `context` parameter, there is already a `Context` injected

app/src/main/java/com/keylesspalace/tusky/components/systemnotifications/NotificationService.kt
```diff
@@ -0,0 +1,791 @@
...
+//            newNotifications.add(summary)
+//
+//            notificationsGroup.value.forEach { notification ->
+//                val single = createNotification(notification, account)
+//
+//                if (single != null) {
+//                    newNotifications.add(single)
+//                }
+//            }
+//        }
```
connyduck, line 292: ??
Please delete

app/src/main/java/com/keylesspalace/tusky/components/systemnotifications/NotificationService.kt
```diff
@@ -0,0 +1,791 @@
...
+                .submit()
+
+            accountAvatar = target.get()
+        } catch (e: ExecutionException) {
+            Log.d(TAG, "Error loading account avatar", e)
+            accountAvatar = BitmapFactory.decodeResource(context.resources, R.drawable.avatar_default)
+        } catch (e: InterruptedException) {
+            Log.d(TAG, "Error loading account avatar", e)
+            accountAvatar = BitmapFactory.decodeResource(context.resources, R.drawable.avatar_default)
+        }
```
connyduck, line 358: In Kotlin, `try` can return a value.

```suggestion
        val accountAvatar = try {
            Glide.with(context)
                .asBitmap()
                .load(body.account.avatar)
                .transform(RoundedCorners(20))
                .submit()
                .get()
        } catch (e: ExecutionException) {
            Log.d(TAG, "Error loading account avatar", e)
            BitmapFactory.decodeResource(context.resources, R.drawable.avatar_default)
        } catch (e: InterruptedException) {
            Log.d(TAG, "Error loading account avatar", e)
            BitmapFactory.decodeResource(context.resources, R.drawable.avatar_default)
        }
```

app/src/main/java/com/keylesspalace/tusky/components/systemnotifications/NotificationService.kt
```diff
@@ -0,0 +1,791 @@
...
+        val replyVisibility = actionableStatus.visibility
+        val contentWarning = actionableStatus.spoilerText
+        val mentions = actionableStatus.mentions
+        var mentionedUsernames: MutableList<String?> = java.util.ArrayList()
+        mentionedUsernames.add(actionableStatus.account.username)
+        for (mention in mentions) {
+            mentionedUsernames.add(mention.username)
+        }
+        mentionedUsernames.removeAll(setOf(account.username))
+        mentionedUsernames = java.util.ArrayList(LinkedHashSet(mentionedUsernames))
```
connyduck, line 697: Kotlin has nice list and set builders for cases like this

```suggestion
        val mentionedUsernames = buildSet { 
            add(actionableStatus.account.username)
            for (mention in mentions) {
                add(mention.username)
            }
            remove(account.username)
        }
```

app/src/main/java/com/keylesspalace/tusky/components/systemnotifications/NotificationService.kt
```diff
@@ -0,0 +1,791 @@
...
+        val status = checkNotNull(body.status)
+
+        val citedLocalAuthor = status.account.localUsername
+        val citedText = status.content.parseAsMastodonHtml().toString()
+        val inReplyToId = status.id
+        val actionableStatus = status.actionableStatus
+        val replyVisibility = actionableStatus.visibility
+        val contentWarning = actionableStatus.spoilerText
+        val mentions = actionableStatus.mentions
+        val mentionedUsernames: MutableSet<String> = LinkedHashSet()
```
connyduck, line 728: another case for a set builder

connyduck (changes requested):
Oh this is a very good idea, thank you!

I have mostly Kotlin style complaints, and I think `NotificationHelper` can now be deleted?

app/src/main/java/com/keylesspalace/tusky/MainActivity.kt
connyduck: I think it is fine

app/src/main/java/com/keylesspalace/tusky/usecase/LogoutUsecase.kt
Lakoja: Mostly with the "thinking" behind it (or lack thereof). Earlier I had "worded very wrong" there.
It's "if notifications are disabled then disable them"?

app/src/main/java/com/keylesspalace/tusky/components/systemnotifications/NotificationService.kt
Lakoja: We want to remove that method anyway soon...

app/src/main/java/com/keylesspalace/tusky/components/systemnotifications/NotificationService.kt
Lakoja: It's the only one that is used from somewhere else.

app/src/main/java/com/keylesspalace/tusky/components/systemnotifications/NotificationService.kt
Lakoja: That is a bit of a question still: Does it matter which context is used (somewhere)?

90% of locations in the old code worked directly or indirectly with application context. But some did and do not.

connyduck: It can matter, yes, for example when taking theme attributes from a context, then you need to use the Activity Context as each Activity can have a different theme. But here it shouldn't 🤔

app/src/main/java/com/keylesspalace/tusky/MainActivity.kt
connyduck, line 311: Could use the newer `ActivityResultContracts` api instead. See `ViewMediaActivity` for an example.

app/src/main/java/com/keylesspalace/tusky/components/systemnotifications/NotificationService.kt
Lakoja: Well spotted (already removed).

app/src/main/java/com/keylesspalace/tusky/MainActivity.kt
Lakoja: Done

app/src/main/java/com/keylesspalace/tusky/components/systemnotifications/NotificationService.kt
Lakoja: All removed.

connyduck (approved):

Issue #4887, opened 2025-01-21
Problem in showing trending hashtags in RTL layouts


RZHSSNZDH:
### Detailed description

In RTL layouts (persian in my case) when you go to "Trending hashtags" part, the hashtags' position is not suitable and they are not visible as they should. Also the number of use are not suitable.

![Screenshot](https://s8.uupload.ir/files/screenshot_۲۰۲۵۰۱۱۹-۱۲۴۲۳۴_tusky_2oij.png)

### Steps to reproduce the problem

1. Change Tusky's language to an RTL language.
2. Go to Trending hashtags section.
3. See the problem :)

### Debug information

Tusky 27.1
Android 13 (LineageOS 20)
mas.to

connyduck:
Should the graphs be mirrored in RTL languages? So that "now" is on the left?

RZHSSNZDH:
Yes, "now" is on the left. Note that only hashtags' positions are not wrong, the positions of numbers of uses are wrong too.

connyduck:
Does this look right?

![Image](https://github.com/user-attachments/assets/5129ec49-c2d5-4dd0-8c8e-95ebd4060c4d)

RZHSSNZDH:
Yes. Thank you!

PR #4889, opened 2025-01-22
fix trending tags for rtl languages

connyduck:
closes https://github.com/tuskyapp/Tusky/issues/4887

And some other small improvements like better paddings and font sizes. Also the `status_info` in `item_status` now looks nicer in rtl mode.


app/src/main/java/com/keylesspalace/tusky/util/NumberUtils.kt
```diff
@@ -8,7 +8,6 @@ import kotlin.math.abs
 import kotlin.math.ln
 import kotlin.math.pow
 
-private val numberFormatter: NumberFormat = NumberFormat.getInstance()
```
connyduck, line 11: The static `NumberFormat` is bad because it doesn't update when the locale changes.

Tak (approved):

PR #4897, opened 2025-01-27
Full support for Android 15, edge-to-edge mode on Android 15

connyduck:
- Update to Api 35
- Update all dependencies that were blocked because they require Api 35
- fix some deprecation warnings
- implement the now required edge-to-edge mode

Edge-to-edge mode means that we now draw under the status bar and the navigation bar and need to make sure we don't overlap with them. Previously the system would do that for us, and we would only provide the color we want the bars in.

For the edge-to-edge mode there are two Apis that are important:
- `fitsSystemWindows` - some Widgets, mostly from the Material library, automatically handle system insets if you set this attribute on them to true
- `ViewCompat.setOnApplyWindowInsetsListener` - this allows you to manually handle the various insets. By returning new insets from the callback it is possible to tell the system which ones are handled and which ones should be taken care of by another view.

In most places edge-to-edge was straightforward to implement, except in `ComposeActivity`, `AccountActivity` and `MainActivity` which required a more custom approach, and a hacky solution to make landscape mode work in `BaseActivity`.

There is also the `ViewCompat.setWindowInsetsAnimationCallback` Api which allows animating with moving insets. I used that in `LoginActivity` and `ComposeActivity` to animate the Views together with the keyboard.

On Android Versions below 15 (Api <= 34) Tusky will look almost the same as before. I think the only exception is the main bottom bar, which is now slighty larger. We customized it to be smaller than the default, but in edge-to-edge mode the height needs to be `wrap_content` or it won't handle insets correctly.

Screenshots:


Lakoja:
Looks complicated. :-)

Some views need fine-tuning with this, I guess.
For example the Toot button looks a bit oddly placed. While the compose button seems ok.

connyduck:
> For example the Toot button looks a bit oddly placed. 

I think that is just because of the rounded display in the screenshot? Its in the same position as before.

connyduck:
There is a weird bug on this branch, when switching to a tab for the first time, it stays empty until touched. Can anybody else reproduce that?

cbeyls:
About the dependencies that were blocked because they require API 35: it's perfectly safe to upgrade the compileSdk version to be compatible with new Google libraries without changing the targetSdk version. You can always implement the new enforced behaviors later. Androidx libraries actually check both the platform SDK int and the target SDK version to ensure the correct behavior is applied.

cbeyls:
Can you post a screenshot to show padding and insets in landscape mode? I'm curious to see how the drawer menu looks like and where it's drawn.

Tak (approved):
Is all the manual padding adjustment needed just to make e2e look ok, or is it so that we can retain the previous mode on older android versions?

Tak:
Just trying it now, the bottom tab panel on the main activity seems much taller than it used to be (android 14, e2e mode not enabled)

cbeyls:
> Just trying it now, the bottom tab panel on the main activity seems much taller than it used to be (android 14, e2e mode not enabled)

I assume it's because its height is now wrap-content ?

> I think the only exception is the main bottom bar, which is now slighty larger. We customized it to be smaller than the default, but in edge-to-edge mode the height needs to be wrap_content or it won't handle insets correctly.

Tak:
Ah, makes sense - if it's intended, then no problem

connyduck:
> There is a weird bug on this branch, when switching to a tab for the first time, it stays empty until touched. Can anybody else reproduce that?

Fixed by juggling stuff around until it worked. I think the main culprit was the RecyclerView.setHasFixedSize(true) that doesn't always work in edge to edge mode 🤔 

> Can you post a screenshot to show horizontal padding and insets in landscape mode? I'm curious to see how the drawer menu looks like and where it's drawn.

Screenshot_20250130_210702

PR #4915, opened 2025-02-09
fix search pull-to-refresh

connyduck:
Before this, refreshing the status search would show the loading spinner but not actually do something (account and hashtag search worked fine).


connyduck:
No, wait this is the wrong fix, it breaks stuff like favouriting

connyduck:
Now it is correct

Tak (approved):

PR #4916, opened 2025-02-09
use updateRelativeTimePeriodically everywhere

connyduck:


connyduck:
wait this can be done much shorter

app/src/main/java/com/keylesspalace/tusky/util/RelativeTimeUpdater.kt
```diff
@@ -19,17 +21,19 @@ private val UPDATE_INTERVAL = 1.minutes
...
     lifecycle.coroutineScope.launch {
         // This child coroutine will launch each time the Fragment moves to the STARTED state
         lifecycle.repeatOnLifecycle(Lifecycle.State.STARTED) {
             val useAbsoluteTime = preferences.getBoolean(PrefKeys.ABSOLUTE_TIME_VIEW, false)
-            if (!useAbsoluteTime) {
-                while (true) {
-                    callback.run()
-                    delay(UPDATE_INTERVAL)
-                }
+            while (!useAbsoluteTime) {
```
Lakoja, line 30: useAbsoluteTime can never change value.

connyduck: That is the point. It should run until the lifecycle goes into stopped state, at which point the coroutine is cancelled.

connyduck: And I figured instead of

```
if (!useAbsoluteTime) {
    while (true) { }
}
```

I could just

```
while (!useAbsoluteTime) { }
```

Lakoja: > That is the point. It should run until the lifecycle goes into stopped state, at which point the coroutine is cancelled.

Does this "cancel hard" meaning something is forcefully interrupted?

Because I would expect the cancel condition to be there somewhere (while(coroutine alive) ...)

The "while (!useAbsoluteTime)" looks like a bug - (masking that it is a "while (true)").
A minor improvement would be an early exit?

Lakoja: Isn't the "repeatOnLifecylce" already a loop?

Some documentation hints so.

(There doesn't seem to be useful understandable documentation on this...)

connyduck: > Does this "cancel hard" meaning something is forcefully interrupted?

The `delay` will throw a `CancellationException`.

> The "while (!useAbsoluteTime)" looks like a bug - (masking that it is a "while (true)").
A minor improvement would be an early exit?

I think it is fine tbh.

> Isn't the "repeatOnLifecylce" already a loop?

Kind of. The `repeatOnLifecycle` block is repeated everytime the lifecycle gets to the started state and stopped when it gets to stopped state. In older Android style, wich you might be more familiar with, one would have started a Job in a Fragment's `onStart` and cancel it in `onStop.`  `repeatOnLifecycle` is basically the same just more convenient.

I wouldn't overthink this part of the code, its not new. And all it does is "while the Fragment is visible to the user, refresh timestamps every minute".

Tak (approved):

Issue #4835, opened 2024-12-28
Unknown Notification Type - severed relationship


edent:
### Detailed description

I received an alert saying "Unknown Notification Type" while connected to Mastodon.social

![Unknown Notification Type](https://github.com/user-attachments/assets/f719d10f-292d-4d49-9b73-66f6132e0b69)

Checking on the web, it is of type https://mastodon.social/severed_relationships

![An admin from mastodon.social has blocked momostr.pink, including 1 of your followers and 0 accounts you follow.](https://github.com/user-attachments/assets/f3796008-9d13-4466-9ed9-0d3b30081c5f)

### Steps to reproduce the problem

1. Have one of your followers on server A
2. Have an admin of your instance block server A
3. Receive notification

### Debug information

Tusky 27.0 beta 1.
Android 15 (GrapheneOS).
Pixel 8 Pro.

Lakoja:
Duplicate of/parallel to #4334

Issue #4571, opened 2024-07-13
New Follower notifications don’t use screen space well


ariasuni:
### Detailed description

Here’s an example (hope it’s clear enough even with the censoring):
![Screenshot_20240713-204438_Tusky](https://github.com/user-attachments/assets/0efed3eb-632b-48cd-87d6-79877dfe6de0)

- Display name and username should be on a separate line, so that longer display name + username can still be displayed entirely since there’s room for it
- Username shouldn’t be cut abruptly if it doesn’t fit the screen (on the screen you can see a `.s` with the s being cut off)

### Steps to reproduce the problem

1. An account with long display name and/or username follows your account

### Debug information

Tusky 26.0, Android 14, Pixel 7

ariasuni:
Nice, thanks a lot!

Issue #4334, opened 2024-03-25
Support new notification type "severed_relationships"


connyduck:
New feature of the upcoming Mastoon version, see https://github.com/mastodon/mastodon/pull/27511

MagicLike:
Are there any updates to the implementation? I have received multiple `Unknown notifications type` from severed relationships in the last few months...

connyduck:
No code has been written yet but I am planning it for v28

Issue #3883, opened 2023-07-26
Feature request: Specially format Direct Messages


mcclure:
Recently, [mastosoc](https://github.com/mastodon/mastodon) has made two changes to direct messages. First off, they have renamed them from "Direct Messages" to "Private Mentions". Second off, they have caused all "Private Mentions" to be specially highlighted, in all status views:

![image](https://github.com/tuskyapp/Tusky/assets/277318/b3115a45-c299-4418-991f-8ddd6cc9d5fc)

I have no strong opinions about the DM/PM rename. However, I think this special DM formatting feature, with the specially highlighted background is very good, and I think we should copy it.

It is fairly easy to get DMs and regular replies mixed up, but getting DMs and regular replies mixed up can be dangerous. For example a mistake I have made in the past is that I begin a DM chain with someone who I am already having a regular reply chain with. I then at some point accidentally reply to the regular reply chain thinking I am replying to the DM chain. Now a post I thought was private, and may even be sensitive, is posted publicly. The special formatting makes it clear at any one moment if I am replying to a regular or private message, and also means that if I post something private, and I do *not* see the background highlight, I realize immediately I made a mistake rather than leaving it up for hours.

Right now the only difference between a DM and regular M in Tusky is the small globe vs envelope icon. I think we should copy at minimum the background highlight and possibly also the special **✉️ Direct Message** text line at top.

nikclayton:
Drive by comment, just to note that anything that implements this should not rely solely on colour to differentiate the type of the message, there should be a textual difference (with content description) first and foremost.

PR #4929, opened 2025-02-20
various not push related notification improvements

### LINKED ISSUE #4440: Follow request entry not fully clickable
### Detailed description

When looking at the list of follow requests you can tap on one to switch to the respective account view.

This does not work when tapping the account description ("account note"). So normally more than half the area is not clickable.

Technical note: Can be fixed with simply setting another click listener on "binding.accountNote" in FollowRequestViewHolder. However the reason why this happens is probably that the account note is a "clickable span" which might contain links.

...

### Debug information

---

connyduck:
- support new notification type `severed_relationships`, closes https://github.com/tuskyapp/Tusky/issues/4835, closes https://github.com/tuskyapp/Tusky/issues/4334
- support new notification type `moderation_warning`
- the account note is now shown again for follow request and follow notifcations (was broken since https://github.com/tuskyapp/Tusky/pull/4026)
- closes https://github.com/tuskyapp/Tusky/issues/4571
- The "unknown notification type" notification now shows the unknown type and a info dialog when you click it https://chaos.social/@ConnyDuck/113601791254050485
- The notification policy banner in the notification tab is now cached for better offline behavior (and less jumping of the list on every load) and updates when interacting with the requests
- Fixes a bug where some notifications wouldn't be filtered correctly. Behavior should now match Mastodon. https://mastodon.social/@alm10965/113639206858728177
- Fixes a bug where some system notifications wouldn't have a body
- For filters and channels, report and signup notifications are now grouped as "Admin", severed relationship events and moderation warnings as "other". These lists are super long already.
- The icon for the "`<user>` just posted" notification is now a bell instead of a home
- Follow requests won't be filtered by default in the notification tab. No idea why this one got special treatment. This change will only affect new logins and not existing ones.
- closes #4440 
- Adds info about attached media or poll to StatusNotificationViewHolder. This is important context that has been missing before.
- Adds (private) reply/(private) mention text above mention notification. (Partially?) closes https://github.com/tuskyapp/Tusky/issues/3883

Some screenshots:
follow
moderation_warning
severed_relationship
unknown
unknown_info
notifications


app/src/main/res/values/strings.xml
```diff
@@ -877,4 +877,35 @@
...
+
+    <string name="moderation_warning_action_delete_statuses">Some of your posts have been removed.</string>
+    <string name="moderation_warning_action_disable">Your account has been disabled.</string>
+    <string name="moderation_warning_action_mark_statuses_as_sensitive">Some of your posts have been marked as sensitive.</string>
+    <string name="moderation_warning_action_none">Your account has received a moderation warning.</string>
+    <string name="moderation_warning_action_sensitive">Your posts will be marked as sensitive from now on.</string>
+    <string name="moderation_warning_action_silence">Your account has been limited.</string>
+    <string name="moderation_warning_action_suspend">Your account has been suspended.</string>
+
+    <string name="unknown_notification_type_explanation">Tusky does not recognize this notification type and can\'t show it to you. Your server\'s web interface may be able to do so. Please also inform the Tusky developers so they can add support for this type! </string>
```
Tak, line 899: Maybe "instance" instead of "server"? 🤷

Tak (approved):

app/src/main/java/com/keylesspalace/tusky/components/systemnotifications/NotificationService.kt
```diff
@@ -333,41 +281,49 @@ class NotificationService @Inject constructor(
...
+            getNotificationBuilder(body, account, channelId)
         } else {
             NotificationCompat.Builder(context, existingAndroidNotification)
         }
 
+        val t = titleForType(body, account)
+        val b = bodyForType(body, account)
+
+        Log.d(TAG, "titleForType: $t")
+        Log.d(TAG, "bodyForType: $b")
```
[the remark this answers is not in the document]
connyduck: nope this was debugging to see why no body was shown 🤦

app/src/main/res/values/strings.xml
connyduck: I don't really care, I used server because I think thats what Mastodon calls them these days.

mastoduy: True. Now "server" is offcial and it's better understand in other languages.

PR #4942, opened 2025-02-23
Make caption dialog non-cancellable (issue #4939)

### LINKED ISSUE #4939: Alt text lost when tapping outside of the input box
### Detailed description

in-progress alt text is lost if you tap outside the box when writing it, or click cancel

https://github.com/user-attachments/assets/e65a94ee-7f10-4d9d-b81a-2ed1a23a1892

1. start writing alt text
2. tap outside the dialogue box, or click cancel
3. in-progress alt text is lost
...

### Debug information

Tusky 27.2
android 13
pixel 7 pro

svenjacobs:
I just made the dialog non-cancellable. Actually I don't know why it was cancellable when there also is

```kotlin
dialog?.setCanceledOnTouchOutside(false)
```

Fixes #4939

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/components/compose/dialog/CaptionDialog.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/components/compose/dialog/CaptionDialog.kt
@@ -84,7 +84,7 @@ class CaptionDialog : DialogFragment() {
             binding.imageDescriptionText.setText(it)
         }
 
-        isCancelable = true
+        isCancelable = false
         dialog?.setCanceledOnTouchOutside(false) // Dialog is full screen anyway. But without this, taps in navbar while keyboard is up can dismiss the dialog.
 
         val previewUri = arguments?.getParcelableCompat<Uri>(PREVIEW_URI_ARG) ?: error("Preview Uri is null")


connyduck:
This is a good start, but I think we will still need another dialog to prevent accidental cancel with the cancel button

Ember-ruby:
> Actually I don't know why it was cancellable when there also is

this is what introduced it https://github.com/svenjacobs/Tusky/commit/9e66ccf4a6a61dc5c9841c6ac2da4f1783798333, and the only change there that looks like it had something to do with `Close dialog on back button` is setting this to true, so just setting it to false may cause an unintended regression?

svenjacobs:
So instead of making it non-cancellable, we need a warning dialog when text was entered and someone either presses Cancel or uses the back navigation.

connyduck:
> So instead of making it non-cancellable, we need a warning dialog when text was entered and someone either presses Cancel, dismissed the dialog by clicking outside or uses the back navigation.

No, the non-cancellable thing makes sense and can stay. Just add another confirmation to the cancel button.

connyduck:
> So are you fine with the back navigation not doing anything when this dialog is opened?

Well we don't want accidental cancels, so yes? Back button should either do nothing, or show the warning dialog as well.

svenjacobs:
Sorry guys but I currently struggle with this a bit as I have only coded in Jetpack Compose for the past few years and it seems I forgot how to handle `Fragments` 😂 

I tried to implement the `onCancel()` method hoping I can intercept a cancellation and show a confirmation dialog both for back navigation and clicking outside the dialog but unfortunately this method is called *after* the dialog has already been cancelled. So I guess we have to disable all cancellation and handle it manually, meaning we have to also intercept the back navigation event.

connyduck:
> we have to also intercept the back navigation event

Yes probably

Lakoja:
A few observations:
* onDismiss() is called before the dialog is closed (cancelled)
* In order to get rid of the text button "Cancel" the "setNegativeButton()" in "onCreateDialog()" needs to be removed
* The "isCancelable = false" (then) only prohibits the closing with the back button
* (Detail: That can also be placed in "onCreateDialog" - not sure which is the better location. I prefer onCreateDialog a bit.)

However: I would say that we could stay with normal UI behavior here:
* It is cancelable; in any normal way: Cancel button, back button, tap outside
* This would require however an "Are you sure?" dialog everytime the text was changed in comparisson to the initial one ("arguments?.getString(EXISTING_DESCRIPTION_ARG)")

connyduck:
@svenjacobs Do you still want to work on this? If not I can just merge it like this and do the remaining stuff myself.

svenjacobs:
Sorry, some other stuff caught up. Please go ahead and merge it. I think you will solve this in half the time than me struggling with the legacy Android view system 😉

connyduck (approved):

PR #4944, opened 2025-02-23
feat: choose boost visibility (#3095)

### LINKED ISSUE #3095: [Feature Request] Timeline visibility of Boosts
The Mastodon web interface allows you to separately set the timeline visibility of a Boost when Boosting (visible, unlisted, followers-only).

This SEEMS to be lacking in Tusky, with the confirmation dialog only displaying one button.

It would be a nice addition to bring Tusky more in line with the web features.

Thank you either way!

* * * *
- Tusky Version: 20.0
- Android Version: Android 10
- Android Device: Huawei P30 Pro

Exagone313:
Closes #3095

---

Since I had to add `visibility` on the `onReblog` method, it is possible that I have broken something. Also, I kept the old method signature (which calls the new one), but it's possible that it is not needed.

~~I'm not 100% sure that unlisted boost works, as the visibility of the boost is not shown anywehere.~~ I've confirmed that private (followers-only) boosts are not visible on private browsing on the browser.

EDIT: Confirmed visibility in devtools on web browser.

Screenshots:

image image

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/network/MastodonApi.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/network/MastodonApi.kt
@@ -252,8 +252,9 @@ interface MastodonApi {
     @DELETE("api/v1/statuses/{id}")
     suspend fun deleteStatus(@Path("id") statusId: String): NetworkResult<DeletedStatus>
 
+    @FormUrlEncoded
     @POST("api/v1/statuses/{id}/reblog")
-    suspend fun reblogStatus(@Path("id") statusId: String): NetworkResult<Status>
+    suspend fun reblogStatus(@Path("id") statusId: String, @Field("visibility") visibility: String?): NetworkResult<Status>
 
     @POST("api/v1/statuses/{id}/unreblog")
     suspend fun unreblogStatus(@Path("id") statusId: String): NetworkResult<Status>
--- a/app/src/main/java/com/keylesspalace/tusky/adapter/StatusBaseViewHolder.java
+++ b/app/src/main/java/com/keylesspalace/tusky/adapter/StatusBaseViewHolder.java
@@ -678,7 +678,7 @@ public abstract class StatusBaseViewHolder extends RecyclerView.ViewHolder {
                         showConfirmReblog(listener, buttonState, position);
                         return false;
                     } else {
-                        listener.onReblog(!buttonState, position);
+                        listener.onReblog(!buttonState, position, Status.Visibility.PUBLIC);
                         return true;
                     }
                 } else {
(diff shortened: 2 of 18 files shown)


app/src/main/java/com/keylesspalace/tusky/interfaces/StatusActionListener.java
```diff
@@ -22,9 +22,12 @@
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
 
+import com.keylesspalace.tusky.entity.Status;
+
 public interface StatusActionListener extends LinkListener {
     void onReply(int position);
     void onReblog(final boolean reblog, final int position);
+    void onReblog(final boolean reblog, final int position, final Status.Visibility visibility);
```
connyduck, line 30: Please convert this interface to Kotlin, then there can be only one `onReblog`, with `visibilty` optional.

connyduck (changes requested):
Thanks!

app/src/main/java/com/keylesspalace/tusky/interfaces/StatusActionListener.java
connyduck: You need to provide a default parameter:
```kotlin
fun onReblog(reblog: Boolean, position: Int, visibility: Status.Visibility = Status.Visibility.PUBLIC)
```

That does not work for Java callers, but there aren't many and you can just pass PUBLIC again:
```java
listener.onReblog(false, position, Status.Visibility.PUBLIC);
```

Exagone313: @connyduck I broke the problem down in one Kotlin file and I don't reproduce my issue, I'm not sure why it insists on wanting both methods. I only have Kotlin code implementing the interface.

I will push a temporary commit to illustrate the error I'm getting.

> e: file:///home/exagone/project/mobile/tusky/app/src/main/java/com/keylesspalace/tusky/components/conversation/ConversationsFragment.kt:68:1 Class 'ConversationsFragment' is not abstract and does not implement abstract base class member 'onReblog'.

connyduck: That is because `SFragment` also has a `onReblog`. The solution is to change the signature of the function in `SFragment` as well.

Exagone313: @connyduck I updated it, let me know if there are other changes to make.

app/src/main/java/com/keylesspalace/tusky/network/MastodonApi.kt
```diff
@@ -253,7 +253,7 @@ interface MastodonApi {
     suspend fun deleteStatus(@Path("id") statusId: String): NetworkResult<DeletedStatus>
 
     @POST("api/v1/statuses/{id}/reblog")
-    suspend fun reblogStatus(@Path("id") statusId: String): NetworkResult<Status>
+    suspend fun reblogStatus(@Path("id") statusId: String, @Query("visibility") visibility: String?): NetworkResult<Status>
```
connyduck, line 256: Interestingly this works, but [according to the docs](https://docs.joinmastodon.org/methods/statuses/#form-data-parameters-2), the `visibility` parameter is a form-data paramter and not a query one.

Change the `@Query` to `@Field` and add `@FormUrlEncoded` to the method please.

Exagone313: @connyduck fixed and tested

connyduck (approved):

PR #4963, opened 2025-02-28
feat: display boost visibility (#4953)

### LINKED ISSUE #4953: Show boost visibility
### Pitch

Since #4944 it is possible to choose the visibility when making a boost.

Unfortunately the interface is not showing the visibility of these boosts, only the visibility of the underlying statuses.

Note: Mastodon's UI doesn't either: https://github.com/mastodon/mastodon/issues/34013

### Motivation

Differentiating the visibility of boosts is useful if you vary the visibility depending to the subject. You might want to verify what visibility you used for your boosts.

Exagone313:
Closes #4953

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -85,6 +85,8 @@
 
     <string name="post_username_format">\@%1$s</string>
     <string name="post_boosted_format">%1$s boosted</string>
+    <string name="post_boosted_unlisted_format">%1$s boosted (Unlisted)</string>
+    <string name="post_boosted_private_format">%1$s boosted (Followers-Only)</string>
     <string name="post_replied">Replied</string>
     <string name="post_replied_format">In reply to %1$s</string>
     <string name="post_replied_self">Continued thread</string>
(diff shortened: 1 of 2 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -85,6 +85,8 @@
 
     <string name="post_username_format">\@%1$s</string>
     <string name="post_boosted_format">%1$s boosted</string>
+    <string name="post_boosted_unlisted_format">%1$s boosted (Unlisted)</string>
+    <string name="post_boosted_private_format">%1$s boosted (Followers-Only)</string>
```
Exagone313, line 89: not sure about those strings

I suggest also:

```suggestion
    <string name="post_boosted_unlisted_format">%1$s quietly boosted</string>
    <string name="post_boosted_private_format">%1$s privately boosted</string>
```

connyduck:
Yeah, so, I'm afraid that displaying it with the `StatusDetailedViewHolder` is probably a lot more complicated 
- Forward the id of the boost instead of the actual post to the `ViewThreadActivity`, everywhere it is opened
- adjust the logic to handle boosts
- Add an additional view to display the info

PR #4974, opened 2025-03-05
Show notifications as popup

Lakoja:


connyduck:
This will 100% bring us complaints. Please let us ask users first.

Lakoja:
> This will 100% bring us complaints. Please let us ask users first.

What would the problems and complaints be?
How would that question be worded?

connyduck:
People who think popup is too introduing and have to go to settings and turn it off per channel. 

Just a simple question:
Which default notification style to you prefer? Then explain the two, ideally with illustrative screenshots.

Lakoja:
> People who think popup is too introduing and have to go to settings and turn it off per channel.

Wouldn't that be equally true for audible notifications? Too intruding and must be configured for 9 channels?

app/src/main/java/com/keylesspalace/tusky/components/systemnotifications/NotificationService.kt
```diff
@@ -583,16 +581,11 @@ class NotificationService @Inject constructor(
     }
 
     private fun joinNames(notifications1: List<StatusBarNotification>, notifications2: List<Notification>): String? {
-        val names = java.util.ArrayList<String>(notifications1.size + notifications2.size)
+        val names1 = notifications1.map { it.notification.extras.getString(EXTRA_ACCOUNT_NAME) }
```
connyduck, line 584: Use `mapNotNull` here and save the `filterNotNull()` below

connyduck:
Actually we should probably just merge this and see what users think.

app/src/main/java/com/keylesspalace/tusky/components/systemnotifications/NotificationService.kt
Lakoja: Done.

Issue #4973, opened 2025-03-05
last lines of long toot get cut off in compose toot window


horsetail-fern:
### Detailed description

When I'm composing a long toot that has more lines of text than can be shown in the text box of the compose window, every additional line disappeares/gets cut off at the lower border of the window. I'm not able to scroll inside the text box.
After adding enough additional lines, the focus jumps down some lines and I can scroll a bit, but not enough to reach the lowest line
(This happens even sooner/more often if I expand the toot I'm answering since this reduces the size of the text box)

I have set a different hight for my notification bar in the developer settings. Could this be the root cause in combination with the new edge to edge view?

I have also set the display size of my android phone to a different value than the default one (wm size 2088x3800), this could also be relevant

### Steps to reproduce the problem

1. Compose new toot
2. Write much text until you reach the lower edge of the text box
3. Keep typing
4. See your text being cut off
5. Keep typing more lines
6. Focus/view jumps down some lines, but last line is still not completely visible

### Debug information

Tusky 28.0 beta 1
Google Pixel 6 Pro 
Android-Version: 15 
SDK-Version: 35

connyduck:
> I have set a different hight for my notification bar in the developer settings. Could this be the root cause in combination with the new edge to edge view?
>
> I have also set the display size of my android phone to a different value than the default one (wm size 2088x3800), this could also be relevant

How do you do that? My developer options on an Android 35 emulator don't have these settings.

connyduck:
Also, a screenshot would be helpful

horsetail-fern:
Not sure if changing the display size is even possible with the android studio emulator. I use adb to do that: `adb shell wm size 2088x3800`
If the phone is rooted you can also use `su wm size 2088x3800` in an terminal emulator

I could reset this to the default value later and check if the bug still happens

I've attached two screenshots: the first shows how the last line is cut off, the second one shows how the focus jumps down after some more lines (but the last one is still partially cut off) and instead cuts off the "Antwort an" at the top

I switched back and forth between different apps while taking those screenshots and at one point I was able to scroll the text completely normal and see all lines. But I wasn't able to reproduce that…
And I noticed that the top bar of the window changed it's colour after jumping down, not sure if that's relevant in any form

![Image](https://github.com/user-attachments/assets/e63d66ab-3749-4386-a3f9-06f10b4db2dd)

![Image](https://github.com/user-attachments/assets/232d6f37-ebb4-47b6-812d-e1851fcae50c)

[...]

PR #4978, opened 2025-03-07
fix window insets related bugs

connyduck:
- Fixes the background of the ComposeActivity in table mode which looked weird since https://github.com/tuskyapp/Tusky/pull/4897
- The ComposeActivity in tablet mode will now resize when the keyboard expands, the previously used fixed height would cause the bottom bar to hide entered text on some devices, closes https://github.com/tuskyapp/Tusky/issues/4973
- The bottom sheets in the compose view will now be in fully hidden state by default, as some weirdness on some devices caused them to peek over the bottom bar in half collapsed state. Please test @fin-w 
- The bottom sheet in the image viewer will now expand a bit more in edge-to-edge mode so it doesn't look it is deliberately obscured by the system bar. The image also moves up a bit so it won't be covered by the bottom sheet as much.
- The "Performing lookup…" bottom sheet won't be covered by the nav bar anymore


fin-w:
@connyduck I can confirm this fixes my issue, many thanks!

PR #4980, opened 2025-03-07
Add "show results" for polls

Tak:
Body of a post with the 'Show Results' button

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/db/dao/TimelineStatusDao.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/db/dao/TimelineStatusDao.kt
@@ -183,6 +183,16 @@ WHERE tuskyAccountId = :tuskyAccountId AND serverId = :statusId"""
     )
     abstract suspend fun setVoted(tuskyAccountId: Long, statusId: String, poll: String)
 
+    @Transaction
+    @OptIn(ExperimentalStdlibApi::class)
+    open suspend fun setShowResults(tuskyAccountId: Long, statusId: String, moshi: Moshi) {
+        getStatus(tuskyAccountId, statusId)?.let { status ->
+            status.poll?.let { poll ->
+                setVoted(tuskyAccountId, statusId, moshi.adapter<Poll>().toJson(poll.copy(voted = true)))
+            }
+        }
+    }
+
     @Query(
         """UPDATE TimelineStatusEntity SET expanded = :expanded
 WHERE tuskyAccountId = :tuskyAccountId AND serverId = :statusId"""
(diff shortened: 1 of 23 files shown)


app/src/main/java/com/keylesspalace/tusky/db/dao/TimelineStatusDao.kt
```diff
@@ -183,6 +183,16 @@ WHERE tuskyAccountId = :tuskyAccountId AND serverId = :statusId"""
...
 
+    @Transaction
+    @OptIn(ExperimentalStdlibApi::class)
+    open suspend fun setShowResults(tuskyAccountId: Long, statusId: String, moshi: Moshi) {
+        getStatus(tuskyAccountId, statusId)?.let { status ->
+            status.poll?.let { poll ->
+                setVoted(tuskyAccountId, statusId, moshi.adapter<Poll>().toJson(poll.copy(voted = true)))
+            }
+        }
+    }
```
Tak, line 194: I don't *love* this, but it feels better than adding another field and updating the "which poll view should I show" logic 🤷

connyduck: By *this* you mean setting the poll to voted when it isn't actually voted? I think that is fine.

But you could include the poll or the whole status in the `PollShowResultsEvent` then you don't need the `getStatus` here.

app/src/main/java/com/keylesspalace/tusky/db/dao/TimelineStatusDao.kt
```diff
@@ -183,6 +183,16 @@ WHERE tuskyAccountId = :tuskyAccountId AND serverId = :statusId"""
     )
     abstract suspend fun setVoted(tuskyAccountId: Long, statusId: String, poll: String)
```
connyduck, line 184: I just found out that one can do this:
```suggestion
    @TypeConverters(Converters::class)
    abstract suspend fun setVoted(tuskyAccountId: Long, statusId: String, poll: Poll)
```
and then there is no more need for the manual json conversion with Moshi

connyduck (commented):
nice

app/src/main/java/com/keylesspalace/tusky/db/dao/TimelineStatusDao.kt
Tak: 🤩

app/src/main/java/com/keylesspalace/tusky/db/dao/TimelineStatusDao.kt
Tak: By `this` I mean the function doing query/deserialize/update
It's not trivial to include more stuff in the `PollShowResultsEvent` - the other events are created from a network call that includes those in the response from the server
I could pass the whole status into the `StatusActionListener` callback I guess, but even then I'm not sure I always have it

connyduck: Ah I see. Its fine then.

connyduck (approved):
will merge once v28 is out

PR #4981, opened 2025-03-08
improve margin between post content and buttons

connyduck:
[A user made me aware of the fact that we have varying spacing between post content and the buttons](https://mstdn.games/@Sven/114104502703573482), so I investigated 🧐 

Indeed, there were some problems:

- The trailing hashtags bar was sometimes visible, despite not showing any content.
- The media container was visible despite showing no media
- The buttons where layed out differently when showing a conversation

This change fixes those problems and adds an additional 2dp space between buttons and contents because without these issues the buttons and the content are quite close otherwise.


app/src/main/java/com/keylesspalace/tusky/util/LinkHelper.kt
```diff
@@ -86,7 +86,7 @@ fun setClickableText(
         trailingHashtagView == null || tags.isNullOrEmpty() -> Pair(spannableContent.length, emptyList())
         else -> getTrailingHashtags(spannableContent)
     }
-    val inlineHashtags = mutableListOf<CharSequence>()
+    val inlineHashtags = mutableSetOf<CharSequence>()
```
connyduck, line 89: This is so we count hashtags only once, no matter how often they appear. Otherwise the `showHashtagBar` check below is wrong for posts that use a hashtag multiple times.

Tak (approved):

Issue #4982, opened 2025-03-08
Post failed dialogue shows in the wrong view, and shows malformed


fin-w:
### Detailed description

(Title reads "Failed to upload"):

![Image](https://github.com/user-attachments/assets/a129d0d0-4133-4e19-a8b0-fee267869202)

### Steps to reproduce the problem

1. Open your profile that is logged in to a server that is unavailable / offline. For me this was because the server was down.
2. Open a post in the timeline (there will be locally cached data even if the server is offline).
3. Respond to the post from the thread view, not the timeline view. Once the compose view disappears (message queued to send), Tusky moves back to the thread view but doesn't show the warning (it's being shown in the timeline view that you haven't moved back to yet).
4. Respond as many times as you want dialogue boxes.
5. Go back to the timeline view, and see the dialogue boxes (correctly formed and missing text versions) appear.

### Debug information

Tusky Test v28.0 beta 1-A1DA5A6B from F-Droid repo
Android 15

connyduck:
I think there are three issues, and only the first two are bugs:

1) Multiple dialogs shown on top of each other
2) One of the dialogs not having any text

These should be easily fixable.

And then
3) Where should the dialog be shown? The current idea is to just show the information somewhere where it will definitely be seen. Showing it on each  app screen is a bit more complicated since it requires syncing between all of them, I'm not sure if that additional effort is worth the minor improvement.

connyduck:
The missing text is Welsh only, I think you fixed it already in Weblate @fin-w

fin-w:
I don't mind if the warning is shown only in the timelines. I would expect it immediately but if that requires syncing a state across views then I can see it's more work than it's worth.

The missing text should default to English if the Welsh equivalent does not exist though, right? I don't see how its translation status should affect the warning dialogue.

connyduck:
It's a plural string and was missing some entries, looks like it shows as just empty then.

fin-w:
OK! So it was a translation issue I was seeing.

The only remaining issue then is that the dialogue shows multiple times. But that only happens if you (fail to) send more than one post. So I could just close this issue? Unless you think showing only once is better?

PR #4985, opened 2025-03-10
fix draft alert showing

connyduck:
- makes sure there is only ever one dialog shown at a time
- also shows the dialog in the thread view

closes https://github.com/tuskyapp/Tusky/issues/4982

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/db/DraftsAlert.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/db/DraftsAlert.kt
@@ -18,15 +18,17 @@ package com.keylesspalace.tusky.db
 import android.content.Context
 import android.content.DialogInterface
 import android.util.Log
+import androidx.appcompat.app.AlertDialog
+import androidx.lifecycle.Lifecycle
 import androidx.lifecycle.LifecycleCoroutineScope
 import androidx.lifecycle.LifecycleOwner
 import androidx.lifecycle.lifecycleScope
+import androidx.lifecycle.repeatOnLifecycle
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/keylesspalace/tusky/db/DraftsAlert.kt
```diff
@@ -37,51 +39,55 @@ import kotlinx.coroutines.launch
...
-                                    context.resources.getQuantityString(R.plurals.action_post_failed_detail, count)
-                                )
-                                .setPositiveButton(R.string.action_post_failed_show_drafts) { _: DialogInterface?, _: Int ->
-                                    clearDraftsAlert(coroutineScope, activeAccountId) // User looked at drafts
+                context.repeatOnLifecycle(Lifecycle.State.RESUMED) {
+                    val draftsNeedUserAlert = draftDao.draftsNeedUserAlert(activeAccountId)
+
+                    if (showAlert) {
+                        draftDao.draftsNeedUserAlert(activeAccountId).collect { count ->
+                            Log.d(TAG, "User id $activeAccountId changed: Notification-worthy draft count $count")
```
Lakoja, line 66: What's notification-worthy?

The meaning of this log (and thus this code) is "the user changed something and thus his draft count now is $count"?

Lakoja (approved):

app/src/main/java/com/keylesspalace/tusky/db/DraftsAlert.kt
connyduck: Its about the count of drafts with the `failedToSendNew` flag

PR #4993, opened 2025-03-16
fix loading posts with invalid `published_at` in preview card

### LINKED ISSUE #4992: Trending posts no longer load
Trending posts stopped loading in the beta.  They still load fine in browser for my instance 

App details:

Tusky 28.0 beta 1

Device:

asus ASUS_AI2401_H
Android version: 15
SDK version: 35

Account:

@vwbusguy@mastodon.online
Version: 4.4.0-nightly.2025-03-14-security

connyduck:
closes #4992

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/json/GuardedAdapter.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/json/GuardedAdapter.kt
@@ -17,6 +17,7 @@
 
 package com.keylesspalace.tusky.json
 
+import android.util.Log
 import com.squareup.moshi.JsonAdapter
 import com.squareup.moshi.JsonDataException
 import com.squareup.moshi.JsonReader
@@ -35,8 +36,9 @@ class GuardedAdapter<T> private constructor(
 
     override fun fromJson(reader: JsonReader): T? {
         return try {
-            delegate.fromJson(reader.peekJson())
-        } catch (e: JsonDataException) {
+            reader.peekJson().use { delegate.fromJson(it) }
+        } catch (e: Exception) {
+            Log.w("GuardedAdapter", "failed to read json", e)
             null
         } finally {
             reader.skipValue()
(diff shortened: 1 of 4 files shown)


cbeyls:
I don't have access to my computer today, but you should reimplement GuardedAdapter to use peekJson() and provide the peeked reader to the Delegate adapter, then call skip. That way the Delegate can crash at any point in the peeked stream, it doesn't matter because we always where we are in the original stream.
I already wrote this, I can send you the code this evening or tomorrow if you want to.

connyduck:
@cbeyls I think I figured it out, thank you very much for the help!

cbeyls:
You're welcome, that should do the trick yes. I will still check later if my code had any extra logic.

app/src/main/java/com/keylesspalace/tusky/json/GuardedAdapter.kt
```diff
@@ -35,10 +35,11 @@ class GuardedAdapter<T> private constructor(
 
     override fun fromJson(reader: JsonReader): T? {
         return try {
-            delegate.fromJson(reader)
+            delegate.fromJson(reader.peekJson())
```
cbeyls, line 38: I recommend you close the peeked reader after usage so the Okio buffer segments can move back to the pool:
`reader.peekJson().use { delegate.fromJson(it) }`

app/src/main/java/com/keylesspalace/tusky/json/GuardedAdapter.kt
```diff
@@ -35,10 +35,11 @@ class GuardedAdapter<T> private constructor(
...
         } catch (e: JsonDataException) {
```
cbeyls, line 39: You should probably catch all kinds of exceptions, not just `JsonDataException`.
`JsonDataException` is what Moshi itself and its official adapters throw but you may have other unhandled exceptions like `IllegalArgumentException` in your own custom adapters.

cbeyls:
I compared with my own implementation and the 2 comments I left in the review are the only differences I found.

PR #5006, opened 2025-03-22
fix conversations jumping down on first load

connyduck:
When opening the conversations tab with empty cache (i.e. after a frash login), it would sometimes start with 6 year old messages for me, which I find very annoying. The issue is that the loading footer is shown first and the RecyclerView tries to keep it visible when new items are added. This works around the issue. Note that it does not happen when there are already some conversations in the cache.


connyduck:
(This issue is also the reason why I didn't add loading footers to more fragments. If this workaround keeps working fine I might introduce them elsewhere as well.)

Tak (approved):

PR #5012, opened 2025-03-24
New icons

connyduck:
This changes all icons to rounded Material symbols and additionally

- removes the material typeface library
- increases the size of the subscribe icon on profiles and makes the two states easier to distinguish
- displays filled icons on selected tabs and outlined ones on unselected tabs
- adjusts some colors to better fit into the themes
- removes duplicate icons
- establishes a naming convention for icons
- when "download media previews" is off, gifs now have their own icon instead of sharing the video icon
- changes the # and @ buttons in the compose view to use icons instead of text


Tak (approved):
Nice, I like that we dropped a lot of boilerplate for setting the size/color

PR #5028, opened 2025-03-29
migrate account list to viewmodel & paging

connyduck:
This was the last of our fragments that didn't have a ViewModel and still used the old custom pagination

Additional benefits:
- Way better error handling (the old one didn't even work, in some cases it would look like success when it really failed)
- smooter scrolling


connyduck:
Funfact: https://chaos.social/@ConnyDuck/114247688577292079

cbeyls:
I don't recommend using a bufferless `SharedFlow` to send UI events because there is no guarantee that an event will reach the UI and it may be sent to nowhere and lost forever.

Either a channel should be used, where the events are enqueued until consumed.

Or even better in my opinion, you can store the current list of events (or only the latest one if it supersedes older ones) in a `StateFlow`, then call a method on the `ViewModel` to mark an event as consumed which will atomically modify the list to remove it. This way, if the handling of the event inthe  UI is interrupted for any reason, it still has a chance to be processed later.
Snackbars are a good example: if a Snackbar is shown right before a configuration change then the user will not have time to acknowledge it, so it's better to consume the event after the Snackbar is dismissed than right after creating it.

connyduck:
Like this? 🤔

Tak (approved):

PR #5034, opened 2025-04-02
fix text selection shrinking the view in ViewThreadFragment

connyduck:
https://layer8.space/@andre/114268763071943595

wtf is this bug 😳

Happens only on Android 15

I have no idea what exactly is going on here, I found this fix by trial and error. `outsideOverlay` seems to look exactly the same as `outsideInset` so i hope this is safe.


Tak (approved):
🤷

PR #5036, opened 2025-04-03
Fix unused translations lint warnings

### LINKED ISSUE #5032: Remove unused translation warnings
This is a sub-issue for https://github.com/tuskyapp/Tusky/issues/3842

The goal is to remove the unused translation warnings. 
Either by using the translation or deleting it.

jleidgens:
Fixes #5032

I added all languages that where marked as unused to the locale file.

The only language that maybe is not really worth putting in there is berber with 36 translations right now.
But in the end those 36 may be better (if they are correct).


connyduck:
I think you need to edit donottranslate.xml as well so the in-app language selector gets the languages as well

jleidgens:
Like this? 
I'm not sure how to sort the languages. I sorted all the latin languages by latin alphabet. 
How would you sort the non latin languages? I just added them to the bottoom of the list

connyduck (approved):

PR #5038, opened 2025-04-05
Add support for the blur filter action

Tak:
Add support for the [blur](https://docs.joinmastodon.org/entities/Filter/#filter_action) filter action added in mastodon 4.4.0.
Images and videos on matched posts are hidden by default, and the label reads "Filtered: ${title of applied filter}".
If a matched post has a preview card, the preview image is also blurred.

~~This is draft for now until I upgrade my instance and test, I've tested so far with spoofed data.~~

Screenshot of two posts in the timeline with blurred media. The label reads: Filtered: BLUR

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/components/timeline/viewmodel/CachedTimelineRemoteMediator.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/components/timeline/viewmodel/CachedTimelineRemoteMediator.kt
@@ -161,7 +161,7 @@ class CachedTimelineRemoteMediator(
                 if (oldStatus != null) break
             }
 
-            val filter = viewModel.kind?.let { status.getApplicableFilter(it.toFilterKind()) }
+            val filter = status.getApplicableFilter(viewModel.kind.toFilterKind())
             val expanded = oldStatus?.expanded ?: activeAccount.alwaysOpenSpoiler
             val contentShowing = oldStatus?.contentShowing ?: (activeAccount.alwaysShowSensitiveMedia || (!status.actionableStatus.sensitive && filter?.action != Filter.Action.BLUR))
             val contentCollapsed = oldStatus?.contentCollapsed != false
--- a/app/src/main/java/com/keylesspalace/tusky/components/timeline/viewmodel/CachedTimelineRemoteMediator.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/components/timeline/viewmodel/CachedTimelineRemoteMediator.kt
@@ -161,7 +161,7 @@ class CachedTimelineRemoteMediator(
                 if (oldStatus != null) break
             }
 
-            val filter = status.getApplicableFilter(viewModel.kind.toFilterKind())
+            val filter = viewModel.kind?.let { status.getApplicableFilter(it.toFilterKind()) }
             val expanded = oldStatus?.expanded ?: activeAccount.alwaysOpenSpoiler
             val contentShowing = oldStatus?.contentShowing ?: (activeAccount.alwaysShowSensitiveMedia || (!status.actionableStatus.sensitive && filter?.action != Filter.Action.BLUR))
             val contentCollapsed = oldStatus?.contentCollapsed != false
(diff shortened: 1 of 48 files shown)


app/src/main/java/com/keylesspalace/tusky/components/timeline/viewmodel/CachedTimelineRemoteMediator.kt
```diff
@@ -160,8 +161,9 @@ class CachedTimelineRemoteMediator(
                 if (oldStatus != null) break
             }
 
+            val filter = status.getApplicableFilter(viewModel.kind.toFilterKind())
             val expanded = oldStatus?.expanded ?: activeAccount.alwaysOpenSpoiler
-            val contentShowing = oldStatus?.contentShowing ?: (activeAccount.alwaysShowSensitiveMedia || !status.actionableStatus.sensitive)
+            val contentShowing = oldStatus?.contentShowing ?: (activeAccount.alwaysShowSensitiveMedia || (!status.actionableStatus.sensitive && filter?.action != Filter.Action.BLUR))
```
Tak, line 166: This logic is repeated in a bunch of places, but I didn't find a nice way to extract it 🤷

app/src/main/java/com/keylesspalace/tusky/components/timeline/viewmodel/CachedTimelineViewModel.kt
```diff
@@ -100,12 +101,15 @@ class CachedTimelineViewModel @Inject constructor(
...
+                val filter = status?.let { shouldFilterStatus(it) }
                 timelineData.toViewData(
                     isDetailed = false,
-                    translation = translation
+                    translation = translation,
+                    filter = filter,
                 )
             }.filter { statusViewData ->
-                shouldFilterStatus(statusViewData) != Filter.Action.HIDE
+                statusViewData.filter?.action != Filter.Action.HIDE
```
Tak, line 112: TODO doublecheck this 🤨

app/src/main/java/com/keylesspalace/tusky/viewdata/StatusViewData.kt
```diff
@@ -42,7 +42,7 @@ sealed interface TranslationViewData {
  */
 sealed class StatusViewData {
     abstract val id: String
-    var filterAction: Filter.Action = Filter.Action.NONE
+    var filter: Filter? = null
```
Tak, line 45: Store the whole filter, so we can get both the action and the title

Tak:
Tested with glitch.taks.garden

connyduck (commented):
Nice!

I'm wondering if we could detect this feature somehow? Or at least have a nice error message when a user tries to set a blur filter on an instance that does not support it?

app/src/main/java/com/keylesspalace/tusky/components/timeline/viewmodel/CachedTimelineRemoteMediator.kt
```diff
@@ -160,8 +161,9 @@ class CachedTimelineRemoteMediator(
                 if (oldStatus != null) break
             }
 
+            val filter = viewModel.kind?.let { status.getApplicableFilter(it.toFilterKind()) }
```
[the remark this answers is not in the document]
Tak: This one failed a bunch of tests because `kind` was null on the test viewModel (maybe I should fix the tests 🤷)

Tak:
Yeah, I had exactly that in one iteration and couldn't decide if it was better or not - if you think it's an improvement, then I'll readd it

Tak:
Yeah, I had exactly that in one iteration and couldn't decide if it was better or not - if you think it's an improvement, then I'll readd it

app/src/main/java/com/keylesspalace/tusky/components/timeline/viewmodel/CachedTimelineRemoteMediator.kt
Tak: Yeah, I had exactly that in one iteration and couldn't decide if it was better or not - if you think it's an improvement, then I'll readd it

Tak:
It doesn't *look* like the capability is advertised anywhere? It *also* doesn't look like we get a useful error when we send an unknown value.

tsmethurst:
> Is other software (e.g. gts) going to support `api_versions`?

GtS not yet, maybe in future.

connyduck:
@Tak Do you want to add feature detection? I'd prefer it but I'm also fine if we leave it out.

Tak:
I can add it, was just on vacation last week and didn't get to it

connyduck (approved):

PR #5055, opened 2025-04-11
update shrinker rules and packaging excludes

connyduck:
We wouldn't want to ship unnecessary stuff, would we

 apk size 20.788.307 ->  20.768.841 bytes 🤪


Tak (approved):
I am going to *savor* those 19KB

PR #5057, opened 2025-04-11
fix activity restart when emoji pack changed

connyduck:
When the emoji pack changes and `MainActivity` is recreated status bar and navigation bar turn dark in edge-to-edge mode (Android 15), we would need [the same workaround as here](https://github.com/tuskyapp/Tusky/blob/068d884aefcd954b72cb5f13bf569b95c9da2ebe/app/src/main/java/com/keylesspalace/tusky/components/preference/PreferencesActivity.kt#L71-L75) in `MainActivity` as well.

But why not handle the restart in `AccountActivity` as we do for all other preferences that require it.


app/src/main/java/com/keylesspalace/tusky/components/preference/PreferencesActivity.kt
```diff
@@ -159,7 +160,7 @@ class PreferencesActivity :
             }
             PrefKeys.STATUS_TEXT_SIZE, PrefKeys.ABSOLUTE_TIME_VIEW, PrefKeys.SHOW_BOT_OVERLAY, PrefKeys.ANIMATE_GIF_AVATARS, PrefKeys.USE_BLURHASH,
             PrefKeys.SHOW_SELF_USERNAME, PrefKeys.SHOW_CARDS_IN_TIMELINES, PrefKeys.CONFIRM_REBLOGS, PrefKeys.CONFIRM_FAVOURITES,
-            PrefKeys.ENABLE_SWIPE_FOR_TABS, PrefKeys.MAIN_NAV_POSITION, PrefKeys.HIDE_TOP_TOOLBAR, PrefKeys.SHOW_STATS_INLINE -> {
+            EMOJI_PREFERENCE, PrefKeys.ENABLE_SWIPE_FOR_TABS, PrefKeys.MAIN_NAV_POSITION, PrefKeys.HIDE_TOP_TOOLBAR, PrefKeys.SHOW_STATS_INLINE -> {
```
Tak, line 163: It's a little weird that `EMOJI_PREFERENCE` is the only one specifically imported here

Tak (approved):

app/src/main/java/com/keylesspalace/tusky/components/preference/PreferencesActivity.kt
connyduck: It is because its from a library and not from our code

PR #5072, opened 2025-04-22
autoformat-with-ktlint

mittwerk:
that's what happend when I used to run ktlint with default ruleset

all works fine


connyduck:
Why would you format an already formatted codebase. I don't like how long this makes some lines.

mittwerk:
@connyduck I more confused about why it was even formatting, so I thought I'd offer this

PR #5082, opened 2025-04-27
4986 - Add delete_media parameter support when deleting statuses

### LINKED ISSUE #4986: Send `delete_media` parameter when deleting status
Mastodon added a new optional `delete_media` parameter to `DELETE /api/v1/statuses/:id`. 
We should send that parameter set to true when a status is deleted without redraft. When deleting & redrafting it should be set to false. This will allow the server to delete some media earlier and not only during the next cleanup.

https://github.com/mastodon/mastodon/pull/33988

achaviara:
### Summary
This pull request updates the delete status API call to support the optional `delete_media` parameter introduced in Mastodon 4.2.

### Changes
Updated `deleteStatus` function to accept a `delete_media` query parameter

Fixes #4986

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/components/search/SearchViewModel.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/components/search/SearchViewModel.kt
@@ -121,9 +121,9 @@ class SearchViewModel @Inject constructor(
         hashtagsPagingSourceFactory.newSearch(query)
     }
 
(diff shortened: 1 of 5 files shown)


Tak:
This looks reasonable to me

connyduck (approved):
Very nice, thank you!

(the parameter is added in Mastodon 4.4)

PR #5103, opened 2025-05-04
Let OS handle text selection popup

### LINKED ISSUE #5031: add option to search for selected text
### Pitch

Currently the only to way to search the web for selected text or a word in a toot is to select the desired text, copy it to the clipboard via context menu and manually put it in a search engine. It would be a real improvement to have a "search / web search" as action (works via ACTION_PROCESS_TEXT).
Thanks in advance

### Motivation

why: by now it's a bit cumbersome to perform this
who benefits: everyone having to do the extra steps manually

tthbaltazar:
Fixes #5031.

### DIFF
--- a/app/src/main/java/com/keylesspalace/tusky/BaseActivity.kt
+++ b/app/src/main/java/com/keylesspalace/tusky/BaseActivity.kt
@@ -125,6 +125,8 @@ abstract class BaseActivity : AppCompatActivity() {
         val style = textStyle(preferences.getString(PrefKeys.STATUS_TEXT_SIZE, "medium"))
         getTheme().applyStyle(style, true)
 
+        getDelegate().setHandleNativeActionModesEnabled(false)
+        
         if (requiresLogin()) {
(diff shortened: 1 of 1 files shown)


connyduck:
What exactly is this doing? I see no difference in the text selection popup.

tthbaltazar:
Uh. Sorry, this is not the fix.

After a long time of staring into android source code I found the real bug: android hides other apps, and [this also affects the actions shows in the text selection popup](https://cs.android.com/android/platform/superproject/+/android15-qpr2-release:frameworks/base/core/java/android/widget/Editor.java;l=8133?q=initializeAccessibilityActions&ss=android%2Fplatform%2Fsuperproject).

The real solution is to declare that we want to see activities that can handle PROCESS_TEXT actions: https://developer.android.com/training/package-visibility/declaring#intent-filter-signature

This explains why the behavior is so different between apps. I'm pretty sure this is a bug and this should be visible by default to all apps. I guess androidx could fix this by putting this into the manifest of every app.

Issue #5113, opened 2025-05-08
GoToSocial 0.19.x crashes Tusky on account load


jessienab:
### Detailed description

For some reason, when attempting to open a logged in GTS 0.19.x profile with Tusky 28/28 nightly, the app immediately crashes. There's limited information even from the Android crash log, see attached logs.

This crash happens with Tusky AND Tusky Nightly 28.0-dc1f5157

Tusky 28 app crash log: [gts-crash_tusky28.log](https://github.com/user-attachments/files/20103266/gts-crash_tusky28.log)
Tusky 28 Nightly app crash log: [gts-crash_tusky28nightly.log](https://github.com/user-attachments/files/20103267/gts-crash_tusky28nightly.log)

GTS output after logging in with Tusky 28: [gts-server-login_tusky28.log](https://github.com/user-attachments/files/20103269/gts-server-login_tusky28.log)
GTS output after logging in with Tusky 28 Nightly: [gts-server-login_tusky28nightly.log](https://github.com/user-attachments/files/20103268/gts-server-login_tusky28nightly.log)

If someone running GTS 0.18.x can test that would be nice, downgrading with a backup on my live instance would be a bit cumbersome.

If I need to open an issue upstream with GoToSocial, let me know and I'll reference this for any suggestions.

Thank you for any help!

### Steps to reproduce the problem

1. Open app
2. Login to GTS instance and authorize OAUTH access
3. App will redirect to account, and app will crash

### Debug information

Tusky 28 from F-Droid / Tusky Nightly 28-dc1f5157 from https://releases.nailyk.fr/repo/
Android 14 (LineageOS)
Google Pixel 5
Instance: gts.nabein.me (GoToSocial)

connyduck:
Looks like you have a post in your home timeline that has a 0 aspect ratio, causing Tusky to divide by zero and then trying to round that NaN value. https://github.com/tuskyapp/Tusky/blob/7bb2ab49f57ba5d9aa07685685b2ad6f90bdb489/app/src/main/java/com/keylesspalace/tusky/view/MediaPreviewLayout.kt#L204


It is possible that you can login once there are enough new posts in the timeline, but scrolling down might crash again.
It would be very cool if you could somehow identify the offending post(s), both for Tusky and GtS.
Maybe by checking api responses for their aspect ratio, or maybe by checking with a webinterface if there is some attachment that looks weird.

<details>
  <summary>Deobfuscated stacktrace</summary>
  
  ```
[...]

PR #5114, opened 2025-05-08
fix crash when attachment has no size information

connyduck:
speculative fix for https://github.com/tuskyapp/Tusky/issues/5113


Tak (approved):
Looks reasonable

PR #5116, opened 2025-05-09
fix filtering boosts from the home timeline

connyduck:
Since https://github.com/tuskyapp/Tusky/pull/5038 the option to filter boosts was ignored and boosts never filtered.

The issue here was that the `HomeTimelineData.status` holds no reblog information, so `timelineData.status?.toStatus(it)` will always return a unreblogged status. `timelineData.toViewData` constructs the correct structure as it considers `HomeTimelineData.reblogAccount` as well.


Tak (approved):
Ouch, nice find

